PR #46, opened 2016-08-19
Add javadoc to pull request status

Kostronor:


Kostronor:
:shipit: @cltnschlosser @Raycoms I fixed the javadoc mess :D

Raycoms:
:shipit: nice

cltnschlosser:
So what exactly does this do? Like how does it work?

javadoc.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+echo "Javadoc-testing commit $TRAVIS_COMMIT"
```
cltnschlosser: Debug line?

javadoc.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ -n "${GITHUB_TOKEN:-}" ]; then
```
cltnschlosser: Why aren't we doing this on pull request?

javadoc.sh
```diff
@@ -0,0 +1,33 @@
...
+            --request POST \
+            --data '{"state":"pending",
+            "target_url":"https://travis-ci.org/Minecolonies/minecolonies/builds/{$TRAVIS_BUILD_ID}",
+            "description":"Javadoc build in progress...",
+            "context":"javadoc"}' \
+            "https://api.github.com/repos/Minecolonies/minecolonies/statuses/$TRAVIS_COMMIT"
+            
+      echo "Start javadoc gradle..."
+      
+      if ./gradlew javadoc; then
```
cltnschlosser: So, this allows us to create a custom PR Acceptance criteria that we change using api calls?

javadoc.sh
```diff
@@ -0,0 +1,33 @@
...
+            curl -H "Content-Type: application/json" -H "Authorization: token $GITHUB_TOKEN" \
```
cltnschlosser: Also, been curious, is this the Minecolonies organization's token?

javadoc.sh
Kostronor: Some nice output to make it look cleaner

javadoc.sh
Kostronor: Because then we would report the check two times. A pull request always has the push build.

javadoc.sh
Kostronor: Yep, exactly

javadoc.sh
Kostronor: This is the MinecoloniesSonar user's token. Someone I created so no one of us had to give the token.

PR #68, opened 2016-08-28
Translations migration

### LINKED ISSUE #67: Translation And Bugs
### Minecolonies version
MineColonies 1.8.9 0.3.472
### Expected behavior
1. Translated Achievements Menu (Turkish)
2. Some Alerts Should Translated
3. POEditor Should Have All Words in MineColonies (like WorkOrders)
4. Builder Shouldn't Be Stuck

### Actual behaviour

1. Untranslated Achievements Menu (English Achievements but Game Is Turkish)
2. Some Alerts Just Pops Out Code-Like Words (While in Turkish)
3. POEditor Hasn't Got WorkOrders word for translation.
4. Builder Stucks A While When He/She Builds His/Hers Hut.

Just Look At This Pictures: http://imgur.com/a/7vQEs

Kostronor:
Fixes #67 

- Updated all translations with the ones from POEditor
- Their format is different so we see this diff...


src/main/resources/assets/minecolonies/lang/en_US.lang
```diff
@@ -1,35 +1,32 @@
-com.minecolonies.error.supplyChestAlreadyPlaced=You have already placed a supply ship!
-
+com.minecolonies.error.supplyChestAlreadyPlaced=You have already placed a supply ship\!
```
cltnschlosser: Do we need these characters escaped? Or should I add a line to the fix-lang tool to remove all \ characters?

Raycoms: We do not need them escaped. You can add a line to it.

2016-08-28 23:45 GMT-03:00 Colton Schlosser <notifications@github.com>:

src/main/resources/assets/minecolonies/lang/en_US.lang
```diff
@@ -77,105 +69,77 @@ tile.minecolonies.blockHutMiner.name=Mine
...
-
 itemGroup.minecolonies=MineColonies
-
 entity.worker.inventoryFullChestFull=Please make space in my building chest. I am not able to dump my inventory.
-
 entity.builder.messageBuildStart=Started building %s
 entity.builder.messageBuildComplete=Finished building %s
 entity.builder.messageNeedMaterial=I need %s (Amount: %d)
 entity.builder.messageBuilderNecessary=Requiring a builder of level %d in order to build or upgrade this building.
-entity.builder.messageNoBuilder=You need a builder in order to build or upgrade this building.
```
cltnschlosser: Do we need to add this to POEditor or something? so it doesn't get removed?

src/main/resources/assets/minecolonies/lang/da_DK.lang
```diff
@@ -76,57 +68,78 @@ tile.minecolonies.blockHutFisherman.name=Fiskerhus
...
-
+entity.builder.messageBuilderNecessary=
 entity.miner.messageNeedTools=Mangler Værktøj %s!
 entity.miner.messageNeedBlockAndItem=Jeg mangler %s!
 entity.miner.messageNeedBetterTools=Jeg mangler bedre værktøj!
 entity.miner.messageMoreBlocks=Jeg skal ha flere %s!
 entity.miner.messageRequiresBetterHut=Upgradere mit hus så jeg kan komme længere ned!
-
 entity.fisherman.messageWaterTooFar=Fiskerhuset er for langt væk fra vandet!
+com.minecolonies.workOrderAdded=
```
cltnschlosser: Also do we want to comment out empty lines? So the english is used instead of blank string.

Raycoms: That would be great!

2016-08-29 9:50 GMT-03:00 Colton Schlosser <notifications@github.com>:

src/main/resources/assets/minecolonies/lang/en_US.lang
Kostronor: Seems so. We cannot import and export at the same time. @Raycoms has to add the new strings first (or tell me from which language to import them ;) )

src/main/resources/assets/minecolonies/lang/da_DK.lang
Kostronor: There is nothing really we can do (if we don't want to delete them every new push) 
I suggest we add to the languagehandler to detect empty strings and show the default in that case

cltnschlosser: @Kostronor @Raycoms I can add something to the script

src/main/resources/assets/minecolonies/lang/en_US.lang
cltnschlosser: @Raycoms Can you do this?

cltnschlosser:
:shipit:

Raycoms:
:shipit:

PR #141, opened 2016-09-19
Fixes #71 build tool sometimes places huts the wrong way

### LINKED ISSUE #71: Build tool sometimes places huts facing into the wrong direction.
### Minecolonies version
Newest
### Expected behavior
Rotate and place the hut facing the wanted direction.
### Actual behaviour
Sometimes faces opposite direction.
Place a hut using the build tool, probably citizen hut at east or west will work.

Kostronor:
Closes #71 

- Fixes the build tool placing the huts the wrong way
- Fixes the build tool achievement.

### DIFF
--- a/src/main/java/com/minecolonies/items/ItemBuildTool.java
+++ b/src/main/java/com/minecolonies/items/ItemBuildTool.java
@@ -27,13 +27,12 @@ public class ItemBuildTool extends AbstractItemMinecolonies
     @Override
     public EnumActionResult onItemUse(ItemStack stack, EntityPlayer playerIn, World worldIn, BlockPos pos, EnumHand hand, EnumFacing facing, float hitX, float hitY, float hitZ)
     {
-        if (!worldIn.isRemote)
+        if (worldIn.isRemote)
         {
-            return EnumActionResult.FAIL;
+            playerIn.addStat(ModAchievements.achievementWandOfbuilding);
+            MineColonies.proxy.openBuildToolWindow(pos.offset(facing));
+            return EnumActionResult.SUCCESS;
         }
-
-        playerIn.addStat(ModAchievements.achievementWandOfbuilding);
-        MineColonies.proxy.openBuildToolWindow(pos.offset(facing));
         return EnumActionResult.FAIL;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/items/ItemBuildTool.java
```diff
@@ -27,13 +27,12 @@ public ItemBuildTool()
     @Override
     public EnumActionResult onItemUse(ItemStack stack, EntityPlayer playerIn, World worldIn, BlockPos pos, EnumHand hand, EnumFacing facing, float hitX, float hitY, float hitZ)
     {
-        if (!worldIn.isRemote)
+        if (worldIn.isRemote)
         {
-            return EnumActionResult.FAIL;
+            playerIn.addStat(ModAchievements.achievementWandOfbuilding);
```
Kostronor: If we move the achievement gain inside a `worldIn.isRemote` then on a remote server the achievement won't be awarded.

Kostronor: Can you test that the achievement gets awarded on build tool use?

DarVillous: It did not, after looking at the code I'm not sure it ever did as the server would have exited the method before it was called.  I moved it to the top of the method to make sure it is called on use.

Kostronor:
LGTM! @Raycoms @cltnschlosser what do you think?

Raycoms:
:shipit:

src/main/java/com/minecolonies/items/ItemBuildTool.java
```diff
@@ -27,14 +27,13 @@ public ItemBuildTool()
...
+        if (worldIn.isRemote)
         {
-            return EnumActionResult.FAIL;
+            MineColonies.proxy.openBuildToolWindow(pos.offset(facing));
         }
 
-        playerIn.addStat(ModAchievements.achievementWandOfbuilding);
-        MineColonies.proxy.openBuildToolWindow(pos.offset(facing));
-        return EnumActionResult.FAIL;
+        return EnumActionResult.SUCCESS;
```
cltnschlosser: What is the difference between FAIL and SUCCESS here?

Raycoms: Previously the itemUse method returned true and false, now it's fail and success.

cltnschlosser: But what does a FAIL and a SUCCESS represent? I remember it not being always just return true if the item was used, the return value had impact

Kostronor: Success means we have worked on that interact and it is done, so forge knows that something worked. A fail is used for not allowed stuff. and pass is used for spectating events.

PR #158, opened 2016-09-22
Update PullApprove to version 2

Kostronor:

We now better represent our workflow.

Once Github releases a Pull Request Review api I will be there and integrate that :D

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,10 +1,12 @@
-approve_by_comment: true
-approve_regex: '(Approved|:shipit:|:\+1:|LGTM)'
-author_approval: ignored
-reject_regex: ^Rejected
-reset_on_push: true
-reviewers:
-  name: maintainers
-  required: 2
-  teams:
-    - maintainers
+version: 2
+groups:
+  maintainers:
+    required: 3
+    author_approval:
+      auto: true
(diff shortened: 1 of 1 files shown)


.pullapprove.yml
```diff
@@ -1,10 +1,14 @@
...
-    - maintainers
+version: 2
+groups:
+  maintainers:
+    required: 3
+    author_approval:
+      auto: true
+    approve_by_comment: 
+      enabled: true
+      approve_regex: '^(Approved|:shipit:|:\+1:|LGTM)'
```
cltnschlosser: Do we want to require the approval at the beginning on the comment ( ^ ), didn't have it before

Raycoms:
Always only accepted the approval at the beginning, as far as I know.

2016-09-23 10:00 GMT-03:00 Colton Schlosser <notifications@github.com>:

.pullapprove.yml
Kostronor: I think it will make sense, that way it it intentional and not a random :+1: somewhere in a review.

Kostronor:
@Raycoms it works as intended

Raycoms:
:shipit:

cltnschlosser:
:shipit:

PR #213, opened 2016-10-19
Should reduce some lagg

Raycoms:
Should reduce some lag from the server.

### DIFF
--- a/src/main/java/com/minecolonies/colony/Colony.java
+++ b/src/main/java/com/minecolonies/colony/Colony.java
@@ -697,9 +697,9 @@ public class Colony implements IColony
             //  Detect CitizenData whose EntityCitizen no longer exist in world, and clear the mapping
             //  Consider handing this in an ChunkUnload Event instead?
             citizens.values()
-              .stream()
-              .filter(Colony::isCitizenMissingFromWorld)
-              .forEach(CitizenData::clearCitizenEntity);
+                    .stream()
+                    .filter(Colony::isCitizenMissingFromWorld)
+                    .forEach(CitizenData::clearCitizenEntity);
 
             //  Cleanup disappeared citizens
             //  It would be really nice if we didn't have to do this... but Citizens can disappear without dying!
@@ -709,11 +709,11 @@ public class Colony implements IColony
                 //  All chunks within a good range of the colony should be loaded, so all citizens should be loaded
                 //  If we don't have any references to them, destroy the citizen
                 citizens.values().stream().filter(citizen -> citizen.getCitizenEntity() == null)
-                  .forEach(citizen ->
-                  {
-                      Log.getLogger().warn(String.format("Citizen #%d:%d has gone AWOL, respawning them!", getID(), citizen.getId()));
-                      spawnCitizen(citizen);
-                  });
+                        .forEach(citizen ->
+                        {
+                            Log.getLogger().warn(String.format("Citizen #%d:%d has gone AWOL, respawning them!", getID(), citizen.getId()));
+                            spawnCitizen(citizen);
+                        });
             }
 
             //  Cleanup Buildings whose Blocks have gone AWOL
@@ -759,7 +759,7 @@ public class Colony implements IColony
 
     private void cleanUpBuildings(@NotNull TickEvent.WorldTickEvent event)
     {
-        @Nullable List<AbstractBuilding> removedBuildings = null;
+        @Nullable List<AbstractBuilding> removedBuildings = new ArrayList<>();;
 
         //Need this list, we may enter he while we add a building in the real world.
         List<AbstractBuilding> tempBuildings = new ArrayList<>(buildings.values());
@@ -770,31 +770,27 @@ public class Colony implements IColony
             if (event.world.isBlockLoaded(loc) && !building.isMatchingBlock(event.world.getBlockState(loc).getBlock()))
             {
                 //  Sanity cleanup
-                if (removedBuildings == null)
-                {
-                    removedBuildings = new ArrayList<>();
-                }
                 removedBuildings.add(building);
             }
         }
 
-        if (removedBuildings != null)
-        {
-            removedBuildings.forEach(AbstractBuilding::destroy);
-        }
+        removedBuildings.forEach(AbstractBuilding::destroy);
 
         @NotNull final ArrayList<Field> tempFields = new ArrayList<>(fields.values());
 
         for (@NotNull final Field field : tempFields)
         {
-            @NotNull final ScarecrowTileEntity scarecrow = (ScarecrowTileEntity) world.getTileEntity(field.getID());
-            if (scarecrow == null)
+            if (event.world.isBlockLoaded(field.getLocation()))
             {
-                fields.remove(field.getID());
-            }
-            else
-            {
-                field.setInventoryField(scarecrow.getInventoryField());
+                final ScarecrowTileEntity scarecrow = (ScarecrowTileEntity) event.world.getTileEntity(field.getID());
+                if (scarecrow == null)
+                {
+                    fields.remove(field.getID());
+                }
+                else
+                {
+                    field.setInventoryField(scarecrow.getInventoryField());
+                }
             }
         }
 
--- a/src/main/java/com/minecolonies/colony/Colony.java
+++ b/src/main/java/com/minecolonies/colony/Colony.java
@@ -759,7 +759,7 @@ public class Colony implements IColony
 
     private void cleanUpBuildings(@NotNull TickEvent.WorldTickEvent event)
     {
-        @Nullable List<AbstractBuilding> removedBuildings = new ArrayList<>();;
+        @Nullable final List<AbstractBuilding> removedBuildings = new ArrayList<>();
 
         //Need this list, we may enter he while we add a building in the real world.
         List<AbstractBuilding> tempBuildings = new ArrayList<>(buildings.values());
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/colony/Colony.java
```diff
@@ -759,7 +759,7 @@ private boolean areAllColonyChunksLoaded(@NotNull TickEvent.WorldTickEvent event
 
     private void cleanUpBuildings(@NotNull TickEvent.WorldTickEvent event)
     {
-        @Nullable List<AbstractBuilding> removedBuildings = null;
+        @Nullable List<AbstractBuilding> removedBuildings = new ArrayList<>();;
```
MinecoloniesSonar: MINOR Remove this empty statement. [rule](http://home.kk-sc.de:9000/coding_rules#rule_key=squid%3AEmptyStatementUsageCheck)
MINOR Local variable 'removedBuildings' could be declared final [rule](http://home.kk-sc.de:9000/coding_rules#rule_key=pmd%3ALocalVariableCouldBeFinal)

MinecoloniesSonar:
SonarQube analysis reported 2 issues
* MINOR 2 minor

Watch the comments in this conversation to review them.

src/main/java/com/minecolonies/colony/Colony.java
```diff
@@ -759,7 +759,7 @@ private boolean areAllColonyChunksLoaded(@NotNull TickEvent.WorldTickEvent event
 
     private void cleanUpBuildings(@NotNull TickEvent.WorldTickEvent event)
     {
-        @Nullable List<AbstractBuilding> removedBuildings = null;
+        @Nullable final List<AbstractBuilding> removedBuildings = new ArrayList<>();
```
cltnschlosser: Why make this change? It simplifies the logic, but it requires us to make an ArrayList every time this method is called. Which with most calls wouldn't need to happen.

src/main/java/com/minecolonies/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -828,9 +828,9 @@ private static ItemStack handleKeepX(@NotNull Map<ItemStorage, Integer> keptX,
      */
     private static boolean keptEnough(@NotNull Map<ItemStorage, Integer> kept, @NotNull Map<ItemStorage, Integer> keep, @NotNull ItemStack stack)
```
cltnschlosser: This whole method is broken, I'm going to simply everything to an equivalent and I think you'll see some issues.

```
private static boolean keptEnough(@NotNull Map<ItemStorage, Integer> kept, @NotNull Map<ItemStorage, Integer> keep, @NotNull ItemStack stack)
      {
          for(Map.Entry<ItemStorage, Integer> tempStackEntry : keep.entrySet())
          {
              final ItemStorage tempStack = tempStackEntry.getKey();
              if(tempStack.getItem() == stack.getItem() && tempStack.getDamageValue() != stack.getItemDamage())
              {
                  keep.put(new ItemStorage(stack.getItem(), stack.getItemDamage(), 0, tempStack.ignoreDamageValue()), tempStackEntry.getValue());
                 break;
             }
         }
         return true;
     }
```

1) Entry set is better, it just wasn't being used properly.
2) Why are we even looping in the first place? I think we should be using a different data structure. Maybe Map<Item, Map<ItemStorage, Integer>>, so that we can do something like this
```
{
    Map<ItemStorage, Integer> map = keep.get(stack.getItem());
    if (map != null) {
        // This logic seems broken to me, but it's what we are doing
        // Loop through the internal map and take the first entry that doesn't have the same damage, save as tempStackEntry
        {
            map.put(new ItemStorage(stack.getItem, stack.getItemDamage(), 0, tempStackEntry.getKey().ignoreDamageValue()), tempStackEntry.getValue());
        }    
    }
    return true;
}
```

3) I doubt we really want to be always returning true.
4) Regardless of whether or not we are ignoring the damage value, we copy the same value from the map, so we potentially want 4 birch wood, well not we want 4 oak wood as well, etc.
5) We could simplify this a lot for cases where we are ignoring the damage value, and even when we aren't. Maybe two maps Map<Item, Integer> we don't care about damage, and Map<Item, Map<Integer, Integer>> for when we care about damage, where the internal map is damage, count

src/main/java/com/minecolonies/colony/Colony.java
Kostronor: The arraylist overhead is minimal. The removed null checks even that one out

src/main/java/com/minecolonies/entity/ai/basic/AbstractEntityAIBasic.java
Kostronor: @Raycoms

Raycoms: 1) How should I use it, isn't it just overhead when I could just directly access the keyset?
2,5) I wanted to avoid having to use two methods for almost the same purpose 
means we would have to override two methods etc doing in the end 90% the same thing.
3) It returns         return keep.get(tempStorage) == null || (kept.get(tempStorage) != null && kept.get(tempStorage) >= keep.get(tempStorage));
  doesn't it?

cltnschlosser: Okay, so entry set and key set is a toss up at this point, because we are only ever actually getting one value from the map. In general if you need to iterate over a map, you shouldn't be calling map.get(). Instead you should be using the entry set, but since we only do one get, I'm not sure it matters. But I really think we should just re-evaluate the data structure we are using so that we don't have to iterate over a map. Iterating over a map is going to be a poor solution every time (If this is going to be impossible, maybe we should look at LinkedHashMap, which has better iterator performance). I think the biggest problem is that we are using ItemStorage as a key and want to do comparisons on several fields of ItemStorage.

```
keep.get(tempStorage) == null || (kept.get(tempStorage) != null && kept.get(tempStorage) >= keep.get(tempStorage))

 kept.get(tempStorage) >= keep.get(tempStorage) --> true always equal
(kept.get(tempStorage) != null && true) --> kept.get(tempStorage) != null

kept.get(tempStorage) == null || kept.get(tempStorage) != null --> true
```

cltnschlosser: Also regarding 4, I guess I'm not sure what we are gaining by creating a new entry in the map, and why we are initializing its value to the same as something with the same item, but not the same damage. Maybe just explain why we are doing this?

Kostronor: We differentiate by damage value. So we can have oak wood and spruce wood at the same time in the map.

cltnschlosser: @Kostronor I understand why, just not why the way we are doing it the way we are.

@Raycoms I will have to look at it again, I definitely would recommend a variable name change, maybe have and want, or kept and want, haveKept and wantToKeep. Something more than just a single letter difference in the middle of the word :)

cltnschlosser: My TLDR at this point would be that I think we should change the data structure so that we aren't iterating through a HashMap. In a PR about improving performance, I think the changes made to this method were negligible, and I think we can improve it a lot.

src/main/java/com/minecolonies/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -828,19 +828,22 @@ private static ItemStack handleKeepX(@NotNull Map<ItemStorage, Integer> keptX,
...
     {
-        Set<ItemStorage> tempKeep = new HashSet<>(keep.keySet());
-        for(ItemStorage tempStack: tempKeep)
+        final ArrayList<Map.Entry<ItemStorage, Integer>> tempKeep = new ArrayList<>(keep.entrySet());
+        for(Map.Entry<ItemStorage, Integer> tempEntry: tempKeep)
```
cltnschlosser: Shouldn't make the list if we are using entrySet's , we want to just use the iterator, so the foreach should use keep.entrySet. But kind of besides the point now, because I think we need to remove the loop.

src/main/java/com/minecolonies/entity/ai/basic/AbstractEntityAIBasic.java
cltnschlosser: Side note, on the topic of variable names, if something is an ItemStorage, we shouldn't call it stack (tempStack). Makes it sound like an ItemStack.

Kostronor: @cltnschlosser Yeah I'll have to change things here and there anyway, so this will be an intermediary solution

src/main/java/com/minecolonies/tileentities/TileEntityColonyBuilding.java
```diff
@@ -80,11 +80,20 @@ public void update()
 
         if (!worldObj.isRemote && colonyId == 0)
         {
-            Log.getLogger().fatal(String.format("TileEntityColonyBuilding at %s:[%d,%d,%d] has no colonyId",
-              worldObj.getWorldInfo().getWorldName(),
-              pos.getX(),
-              pos.getY(),
-              pos.getZ()));
+            Colony tempColony = ColonyManager.getColony(worldObj, this.getPosition());
```
MinecoloniesSonar: MINOR Local variable 'tempColony' could be declared final [rule](http://home.kk-sc.de:9000/coding_rules#rule_key=pmd%3ALocalVariableCouldBeFinal)

MinecoloniesSonar:
SonarQube analysis reported 1 issue
* MINOR 1 minor

Watch the comments in this conversation to review them.

PR #250, opened 2016-10-31
added @Kevnar Rustic style

Kostronor:
Added Kevnar's Rustic style to the mod. Would need some testing if it works as expected...


Kostronor:
Closing this as it sadly is an old schematic and not a structure file

PR #257, opened 2016-11-02
Update PULL_REQUEST_TEMPLATE

cltnschlosser:
Closes #

-
-
-

Review please

Now that we have daily emails from pull review, we don't need to be tagged in this. It will prevent me from getting all the sonar emails :) We can tag each other or `@Minecolonies/maintainers` if needed.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -5,4 +5,4 @@ Closes #
 -
 -
 
-Review please @Minecolonies/maintainers
+Review please


Kostronor:
Why delete the ping?

Kostronor:
:shipit:

cltnschlosser:
If I'm tagged in the PR description I will get an email for every single sonar comment. Where as if someone is needed and tagged in a comment, they should only get the email for the comment (assuming a similar email notification setup as me) @Kostronor

Raycoms:
:shipit:

PR #333, opened 2016-12-10
Updated translations

marvin-bitterlich:
automatically updated


Raycoms:
Needs more work. EN_US seems to have lost lines (shouldn't happen I guess).
And some newer strings are missing + the strings are not properly parsed.
@cltnschlosser Could you take a look at this?

Raycoms:
:shipit: I guess

PR #345, opened 2016-12-12
Spawn cititens close to townhall and not city center

Raycoms:
Citizens previously always spawned at the colony center.
That place might be corrupted.

They now will spawn at the townhall directly.

Review please


marvin-bitterlich:
@Raycoms what if the townhall is not placed? Does it crash or just don't spawn new citizens?

Raycoms:
Doesn't spawn. Both methods which call the spawn city make sure that townhall != null.
Actually I can move that check even to a better place I just noticed.

marvin-bitterlich:
LGTM

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/345/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

PR #349, opened 2016-12-16
Canceling work order improvment

akupiec:
Currently there is no way to "remove" building task - or at least I didn't found a way to do so.

While removing work order in town hall builder should stop building - other way this tab in town hall do not have a much purpose.

Review please.


Raycoms:
You did try to click the X to remove the workOrders in the townhall tab?
Normally this should work. Not sure if something mixed that functionality up.

akupiec:
It's relay working :) it's seems I messed up something or checked it in old version.

Raycoms:
But I know in same cases we're getting some problems when canceling build
orders. (Sometimes he doesn't cancel them very well). You could take a look
at this.
Maybe some of your code might be reused in this case.

2016-12-15 22:11 GMT-02:00 Artur <notifications@github.com>:

PR #498, opened 2017-01-22
Update Configurations.java - More realistic, less annoying Speeds

### LINKED ISSUE #491: [updated config] Got tired of the same old names and too long build times
https://gist.github.com/424ab1a16425aeee65cf5b04e487bfa5 save the raw as minecolonies.cfg into your config folder

I've added around 1000 first names for both ladies and men (2000ish total) and around 800 surnames.

Also reduced the painfully slow build delay from 1000 to 10.

forReason:
closes #491

This is supposed to make more realistic settings since minecolonies is a realistic mod.

``` java
// raised from 30 seconds to 4 minutes since citicens are no Sexmachines
citizenRespawnInterval       = 240;

// lowered to ~4 times slower than by hand instead of 15 times slower
blockMiningDelayModifier     = 125; 

// increased delay between placing a block from 0 to .75 seconds
// since the workers are no time gods
builderBuildBlockDelay       = 15;
```

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/configuration/Configurations.java
+++ b/src/main/java/com/minecolonies/coremod/configuration/Configurations.java
@@ -8,10 +8,10 @@ public class Configurations
     public static       int     townHallPadding              = 20;
     public static       boolean supplyChests                 = true;
     public static       boolean allowInfiniteSupplyChests    = false;
-    public static       int     citizenRespawnInterval       = 30;
+    public static       int     citizenRespawnInterval       = 240;
     public static       boolean builderInfiniteResources     = false;
-    public static       int     builderBuildBlockDelay       = 0;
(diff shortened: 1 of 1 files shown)


cltnschlosser:
4 Minutes is a long time to wait for a citizen to spawn. And 0.75 seconds for a block is kind of long

forReason:
1. The first citizen will spawn after a few seconds still
2.  even 2 seconds wouldnt be much considering that the builder doesnt have to built dirtsupport to reach the roofs etc. he builds the house way faster than I could in creative by hand. also the faster mining more than compensates for it

cltnschlosser (approved):

marvin-bitterlich (approved):
LGTM @Raycoms 
I like the tweaked values

PR #521, opened 2017-01-29
Recognize previously placed entities.

Raycoms:
Builder should recognize entities which are already at the position.
Only checks if of the same type at the same position.
Comparing the entities would be quite too hard.

Review please


marvin-bitterlich:
:shipit:

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/521/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

PR #543, opened 2017-02-04
Fix upgrade button label for town hall

xavierh:
- show when the town hall is upgradeable
using com.minecolonies.coremod.gui.workerHuts.upgrade

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowBuilding.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowBuilding.java
@@ -101,7 +101,7 @@ public abstract class AbstractWindowBuilding<B extends AbstractBuildingHut.View>
         {
             final Button button = findPaneOfTypeByID(BUTTON_BUILD, Button.class);
             button.setLabel(LanguageHandler.format("com.minecolonies.coremod.gui.workerHuts.upgrade"));
-            button.disable();
+            button.enable();
         }
     }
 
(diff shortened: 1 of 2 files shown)


marvin-bitterlich:
For this one it's fine but please use branches from the repo, not from your fork 😉

xavierh:
Sorry about that.

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/543/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

cltnschlosser:
I think this should be changed in the window class instead of the xml

Raycoms:
The building still has "build" after the hut is placed or is it "upgrade" from the beginning.
If #2, we have to adjust this. We will have to set from the buildingView the "upgrade" after building is level 1.
That should be quite analogous to the other buildings.

xavierh:
Upgrade building is the default and I didn't 2 translation texts was needed.

cltnschlosser:
Okay, so this should be reverted and more logic should be added to the window class

xavierh:
@Kostronor is it better like this, or do I do something wrong still?
@cltnschlosser I kept the new string in the xml but added the logic

Raycoms:
Please change the string in the xml back and then we're good to go.
(The default version should be in the xml)

xavierh:
@Raycoms, I will but I widen the pr scope. I should update it today

marvin-bitterlich:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/543/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

PR #579, opened 2017-02-11
Feature/buildtool server side nbt

### LINKED ISSUE #486: Buildtool: custom Huts
### Minecolonies version

Future version

### Expected behavior

- Drop a new Citizen1.nbt file into minecolonnies/buildings/ to override the citizen's hut

### Actual behaviour

- To override a default building we need to edit the jar file

### Implementations

Difficulty: moderate to unknown
Change how loadStyleMap work to read from the directory first and complete with the assets.
There is no need to extract from the jar at installation, we just need to read from the assets when not available in the directory

The location of the directory should be worlds specific as you might want to use different build on different games with constantly moving folder around (The configuration file should be world specific too btw)

The same code should be used for hut and decoration. 

Need **#487 Build tool: should use server side nbt files** to be implemented first

This is covered by #145 but with a much smaller and achievable scope

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41165318-buildtool-custom-huts?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #487: Build tool: should use server side nbt files
### Minecolonies version

Future version

### Expected behavior

- Use the server nbt files, so if they have been customized on the server, the player should have the correct one without effort

### Actual behaviour

- The player need to copy the nbt file used by its server in order to work properly.
- It is event more difficult if the player use different server with different decorations

### Implementation
Difficulty: moderate

1. When the player open the build tools, a list of hut/decoration should be sent to him
2. When the player select a hut/decoration, we should:
 - Look in the player local hard drive under minecolonnies/worldname/filename
 - If the file is not found, it should request the file to the server
 - The server should look for the file on the hard drive
 - If the file is not found it should look into the asset
 - once the file/asset is found it should send it to the client (using SavescanMessage)
-  the client should store it under minecolonnies/worldname/filename (so it won't request the file every time)

world name should be unique somehow
file name would be buildings/citizen1.nbt or waypoint/bridge1.nbt

If the building style are directory bases is should work fine with it too

This would help with #486 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41166202-build-tool-should-use-server-side-nbt-files?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #488: Scan tool: Server side storage
### Minecolonies version

Future version

### Expected behavior

- Being able to scan and have it available with the build tool even when using a dedicated server
- Scan the first position
- Scan the second position
- Scan the third time open a gui where the player need to choose the directory and the name 
- enjoy your new decorations

### Actual behaviour
- The scan is done on the server and sent to the client to store on its local hard drive

### Implementation
Difficulty: easy to moderate

The scan is already done on the server, I just need to store it on it.
This option will not be on by default. It will be explicitly turn on by editing the config file
Max scan size will be defined in the configuration file too to prevent big scan.

The directory where to store the nbt will not be a free text input, the list of available directories for storage will be provided by the server (in order to avoid some security problem), the list of available directory is up to server admin for instance if we were to use buildings as a base directory, the server admin could create:
 - buildings
   - huts (for custom huts) not usable by the player but the admin could add his custom huts in it see #486
   - decorations
   - Walls
   - Monuments
Only decorations, Walls and Monuments would be available to choose from for the player
If the name already exist a number will be appended i.e. using the stonewall when stonewall.nbt and stonewall-1.nbt already  exist. the server should save it under stonewall-2.nbt

Directory structure might need a rethink with the building style feature

Need #487 Build tool: should use server side nbt files to be implemented first to be useful

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41167106-scan-tool-server-side-storage?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github).
</bountysource-plugin>

xavierh:
Closes #487 Build tool: should use server side nbt files 
Closes #486 Buildtool: custom Huts 

- Add a UUID for the server
- Load the schematic when the world is loading (so we know where to look for schematics)
- Clear separation between huts and decorations
- Allow multiple subdirectories for schematics
- Override jar  huts and decorations provided by the jar file
- Allow to create new styles without modifying the jar file,

## Schematic loading
A schematic file is loading from the  - get the schematic in this order:
   - from the schematic directory
   - from the jar

On the client (WindowBuildTool):
 - Compare the md5 hash with the one from the server
 - Request the schematic from the server if the file is not found or the md5 does not match
 - Save it in the cache directory

The location of the schematic directory depends of the type of game:
- On the logical server: **World Directory**/minecolonies/schematics
- On the logical client: ${HOME}/.minecraft/minecolonies/**Server UUID**/schematics (on linux)

## Re-organise the schematics
The schematics are clearly separated between **huts** and **decorations** using 2 different subdirectory of the schematics folder.
- huts schematic in the decoration folder will be loaded as decoration only
- Other directories  in the schematics folder are still usable by the mod but will not be available to the BuildTool (Ex: miner's schematics and SupplyChest)

 - huts/
   - stone
     - Builder1.nbt
     - ...
     - WareHouse5.nbt
   - wooden
     - Builder1.nbt
     - ...
     - WareHouse5.nbt
 - decorations/
   - walls/
     - DarkOak/
       - Gate.nbt
       - ...

**Note 1:** schematics with multiple sub directories will have their style **directory/sub-directory/sub-sub-directory**. From the example above, Gate will have **walls/DarkOak**
**Note 2:** Multiple subdirectory work in the BuilderTool but this is not optimal as a long Style name will not fully show inside the tyle button. WindowBuildTool will need some love.

### For building:
When a building style does not exist any more (because the directory have been renamed), the Building will use the md5 hash to find the new style:
Ex: 
- We have a Building *GuardTower* Level 1with style *wooden* (structure name: huts/wooden/TowerGuard1)
- The server administrator decided to rename the wooden directory to *OakWood* because he does have other wood type tower using different wood type.
- When loading the Building huts/wooden/TowerGuard1 the schematic is not found.
- We use the md5 hash to find that the schematic is now *huts/OakWood/TowerGuard1*
- Update the Building style to *OakWood*

### For a Work order:
For a work order, we do similar thing as we store the structure name and the md5. When the schematic is not found using the structure name, we use the md5 to find the new Schematic name and update the work order accordingly

**Note:** As the structure directory change prior the MD5 hash implementation some work order will fail after the upgrade because  the schematics file have changed. Those work order will be automatically cancelled (It will not be a problem for huts)

## ToDo
- Recover Building when schematic moved (done for work order)
- Clean up the code
- Add javadocs


marvin-bitterlich:
We could do an md5 checksum on the files to see if they differ

xavierh:
Yes that what I thought. I am going to do just that

xavierh:
@Kostronor I should be able to work on that again this week.

I need to add the md5 checksum and will use **java.security.MessageDigest** unless you suggest something else.

Additionally I like to be able to create new style for huts not in the jar. For that I need  reorganize the directory structure for the schematics as huts and decorations are currently together.

I am aiming for:
- huts
  - wooden
    - builder1.nbt
    - builder2.nbt
    - ...
  - stone
    - builder1.nbt
    - builder2.nbt
    - ...
  - **My Own Style**
- decorations
  - decoration (I should probably change this name)
  - walls
  - waypoints


The idea behind it is that the directories in the **huts** directory are the styles, to create a new one, we just need to add a new directory with its nbt files in it.

I will need to pay attention to the current work orders so that any workorder using stone/builder1.nbt for instance will be converted to use huts/stone/builder1.nbt

/edit for formatting bullets list

marvin-bitterlich:
@xavierh 
That is a good idea. That was what we initially planned I guess :D

I suggest you just actively change it to actually have the huts part. That would make it way easier.

xavierh:
TODO:
- recover work order when schematic have move: I will do that by adding the md5 hash to the work order. when a work order fail because the schematic does not exist anymore I will the use the md5 hash to see if the schematic location changed and update the work order accordingly.

- need to transition workorder pre pr to work order post pr. for hut it will be easy:
   if a hut structure is not found, we can look for huts/structurename if it does exit
- Transition will not be possible for decorations

Raycoms:
I would say we just cancel the decoration workOrders if we can't find them

Am Dienstag, 28. Februar 2017 schrieb xavierh :

Raycoms:
In the past they crashed I think I've changed the code to canceling but not
sure if it works for all cases
Am Dienstag, 28. Februar 2017 schrieb Ray Neiheiser :

xavierh:
@Raycoms 
After the pr, all decoration structure name should start with decorations/
For instance **walls/stone**, should be **decorations/walls/stone** now.
If I see a Work Order Decoration where the structure is not found and the structurename does not start by **decorations/** then I will try to prefix it and see if I can load it then.

But I reorganize the Walls section as it was messy (https://github.com/Minecolonies/minecolonies/pull/579/commits/ab51c8c3d284b5bea77b0fa500a36b0f8d1c9fd4)

For example 
now we had:
decorations/walls/DarkOakWallCorner.nbt
which I change to:
decorations/walls/DarkOak/Corner.nbt 

So it is easier to navigate through walls schematic by style (**walls/DarkOak**).

Those one won't transition but I don't it is a big deal anyway.

The game does not crash when the structure of a work order does not exist. I tested it extensively.

Raycoms:
Okay that's important.
Yeah I had planned that but we didn't support subfolders yet

Am Dienstag, 28. Februar 2017 schrieb xavierh :

xavierh:
Ready for review

src/main/java/com/minecolonies/coremod/colony/Structures.java
```diff
@@ -25,23 +27,32 @@
...
     private static       Map<String, List<String>> hutStyleMap           = new HashMap<>();
     //Decoration, Style
     private static       Map<String, List<String>> decorationStyleMap    = new HashMap<>();
 
+    /* md5 hash for the schematics
+     * format is:
+     * huts/stone/builder1.nbt -> hash
+     * decorations/decoration/Well.nbt -> hash
+     */
+    private static       Map<String, String>       md5Map                = new HashMap<>();
```
marchermans: This can be improved by using a BiMap. Improves readability of reverse LoopUp and has the same performance

src/main/java/com/minecolonies/coremod/colony/Structures.java
```diff
@@ -251,9 +292,81 @@ public static int getMaxLevelForHut(final String hut)
...
+        return Structures.hutLevelsMap;
+    }
+
+    /**
+     * Returns a map of styles for one specific decoration.
+     *
+     * @param decoration Decoration to get styles for.
+     * @return List of styles.
+     */
+    public static Map<String, String> getMD5s()
```
marchermans: See comment to the md5 map variable.

src/main/java/com/minecolonies/coremod/colony/Structures.java
```diff
@@ -251,9 +292,81 @@ public static int getMaxLevelForHut(final String hut)
...
+        return "";
+    }
+
+    /**
+     * get a structure name for a give md5 hash.
+     *
+     * @param md5 hash identifying the schematic
+     * @return the structure name as 'hut/wooden/Builder1' or an empty String if not found
+     */
+    public static String getStructureNameByMD5(final String md5)
```
marchermans: By using a BiMap this method can be made a lot les complicated and a lot more performant.

src/main/java/com/minecolonies/coremod/colony/Structures.java
```diff
@@ -251,9 +292,81 @@ public static int getMaxLevelForHut(final String hut)
...
+
+
+    /**
+     * For use on client side by the ColonyStylesMessage.
+     *
+     * @param md5s        new md5Map.
+     */
+    @SideOnly(Side.CLIENT)
+    public static void setMD5s(final Map<String, String> md5Map)
+    {
```
marchermans: See comment for the md5Map field up above.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -311,6 +317,26 @@ public void readFromNBT(@NotNull final NBTTagCompound compound)
 
         rotation = compound.getInteger(TAG_ROTATION);
         style = compound.getString(TAG_STYLE);
+
+        final String buildingName = compound.getString(TAG_BUILDING_TYPE);
+        final String md5 = compound.getString(TAG_SCHEMATIC_MD5);
+        final String structureName = Structures.SCHEMATICS_HUTS + '/' + style + '/' + this.getSchematicName() + buildingLevel;
+
+        if (Structures.getMD5(structureName).isEmpty())
```
marchermans: It might be a good idea to add methods to the Structures class to check if it has an MD5 or a Structure Name.
This would make these arbitrary checks against isEmpty() not needed. Improve the readability and allow for more flexibility in the future, especially when creating an API for this kind of stuff, arbitrary checks should be avoided if they can simple be wrapped inside a proper method.

src/main/java/com/minecolonies/coremod/colony/workorders/WorkOrderBuild.java
```diff
@@ -99,7 +100,23 @@ public void readFromNBT(@NotNull final NBTTagCompound compound)
             upgradeName = compound.getString(TAG_UPGRADE_NAME);
         }
         cleared = compound.getBoolean(TAG_IS_CLEARED);
+        md5 = compound.getString(TAG_SCHEMATIC_MD5);
         structureName = compound.getString(TAG_SCHEMATIC_NAME);
+        if (Structures.getMD5(structureName).isEmpty())
+        {
+            if (!Structures.getMD5(getStructurePrefix() + '/' + structureName).isEmpty())
```
marchermans: Again see comment regarding line 325 of the AbstractBuilding class:

It might be a good idea to add methods to the Structures class to check if it has an MD5 or a Structure Name.
This would make these arbitrary checks against isEmpty() not needed. Improve the readability and allow for more flexibility in the future, especially when creating an API for this kind of stuff, arbitrary checks should be avoided if they can simple be wrapped inside a proper method.

src/main/java/com/minecolonies/coremod/colony/workorders/WorkOrderBuild.java
```diff
@@ -99,7 +100,23 @@ public void readFromNBT(@NotNull final NBTTagCompound compound)
             upgradeName = compound.getString(TAG_UPGRADE_NAME);
         }
         cleared = compound.getBoolean(TAG_IS_CLEARED);
+        md5 = compound.getString(TAG_SCHEMATIC_MD5);
         structureName = compound.getString(TAG_SCHEMATIC_NAME);
+        if (Structures.getMD5(structureName).isEmpty())
```
marchermans: Again see comment regarding the AbstractBuilding class line 325:

It might be a good idea to add methods to the Structures class to check if it has an MD5 or a Structure Name.
This would make these arbitrary checks against isEmpty() not needed. Improve the readability and allow for more flexibility in the future, especially when creating an API for this kind of stuff, arbitrary checks should be avoided if they can simple be wrapped inside a proper method.

src/main/java/com/minecolonies/coremod/network/messages/ColonyStylesMessage.java
```diff
@@ -17,8 +17,10 @@
  */
 public class ColonyStylesMessage implements IMessage, IMessageHandler<ColonyStylesMessage, IMessage>
 {
+    private Map<String, Integer> hutLevelsMap;
     private Map<String, List<String>> hutStyleMap;
     private Map<String, List<String>> decorationStyleMap;
+    private Map<String, String> md5Map;
```
marchermans: If the change to a BiMap is made, it would be required here as well.

src/main/java/com/minecolonies/coremod/network/messages/ColonyStylesMessage.java
```diff
@@ -55,11 +76,39 @@ public void fromBytes(@NotNull final ByteBuf buf)
         return map;
     }
 
+    @NotNull
+    private static Map<String, String> readMD5MapFromByteBuf(@NotNull final ByteBuf buf)
```
marchermans: Again converting this to a final BiMap would improve performance and readability.

marchermans (commented):
I am leaving this review as general comment. However i suggest to @Kostronor and @Raycoms To look into these.

The BiMap change is easily made and improves the performance a lot. Especially since he is often querrying a Structure from a MD5 Hash.

On top of that with regards to the API, Voodoo-Checks like MD5.IsEmpty() should be avoided as no were in the documentation of MD5 is stated that it cannot produce an Empty String as Hash.
With regards to the API side of these things it would be a lot better if methods were added that for now just wrapped the isEmpty() checks for a given structure name, or md5 hashsum.

src/main/java/com/minecolonies/coremod/colony/Structures.java
xavierh: By BiMap do you mean com.google.common.collect.BiMap ?
Can I use this ?

In some case I can have 2 "different" schematics with the same hash.

I experienced this when I duplicate the TowerGuard schematic to do a corner version on.
Level 2,3,4,5 where different but Level 1 is identical as they is no door to place in the corner.

You are right, I will avoid mp5.isEmpty() and return null instead

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
xavierh: do you mean something like:
boolean hasMD5Hash(final String stuctureName) ?

fair point

src/main/java/com/minecolonies/coremod/colony/Structures.java
marchermans: Xavierh, if you have two schematics with the same hash, you are doing it wrong.

xavierh: Why would it be wrong ?
The 2 schematics as identical

An other case where you could have twice the same schematic is when you want a bridge standard and a bridge which act as a waypoint.
both schematic would be identical, just the style (folder location) would be different.

I use MD5 to check the the client is using the same version of a schematic not to be sure they are unique.

Also I am planning for the player to be able to scan their own schematic, each player will have their own folder but schematic might be identical between to players

marchermans: The idea behind you using these hashes is to find the differences between two files. So two different files (regardless of the fact that they belong to the same building, even if it is rotated) should never and i mean never have the same hash.

That would complete negate the principle of what you are working here.

On top of that, if you had two files with the same hash, your method to get a structure name from a hash would only return the first one. Which is maybe not the one you requested.

marchermans:
If it is the case, as stated above by @xavierh, that two files can have the same MD5 Hash i would not pull this, but find a different way of differentiating them.

src/main/java/com/minecolonies/coremod/colony/Structures.java
xavierh: > The idea behind you using these hashes is to find the differences between two files. So two different files (regardless of the fact that they belong to the same building, even if it is rotated) should never and i mean never have the same hash.

No the idea is to make sure the client use the correct schematic.

> That would complete negate the principle of what you are working here.

No

>On top of that, if you had two files with the same hash, your method to get a structure name from a hash would only return the first one. Which is maybe not the one you requested.

One is better than none.

I use the hash to fix Work Order where the schematic still exist but have been renamed or moved. Yes this is not full proof but it is better that failing a WorkOrder and cancelling it. Or have a worthless building just because the schematic moved or its style name have been changed

xavierh:
@OrionDevelopment 

As stated above, the standard use of the MD5 is to make sure client and server have the same schematic.

As a bonus, I use it to recover Work Order and Building when/if the files have moved.

I could have simply ignore this bit and let it fail (as the previous behaviour). Instead I choose to try to recover from moved/renamed schematic.

This is not a problem for decoration as it does not matter which one we use, the result will be the same.

For Building it is only a problem when two buildings are identical but their upgraded builing are different. In that case, yes, I might use the wrong style but it is still an improvement compare to the previous behaviour where you would have not be able to upgrade the building anymore.

marchermans:
I understand what you are trying to do, and i guess the case that the checksums are equal happens not so often (read so few times that is should not matter). 

Do i understand correctly that there could be two schematics with the same contents, used for different purposes?
So say one rotated and one not, if the contained building is a perfect mirror for all four horizontal axi?

If that is the case then yeah it would not matter and you cannot use a BiMap as that forces a one to one relationship.

However the problem still recides that a workorder that was rebuild from the MD5, is not capable of retrieving that second rotated (with the same MD5) structure, making it show the user the none rotated variant (granted the output of the workorder is equal, but still), so i donnot know if it is required but should be taken into account.

There are a lot of other ways of rebuilding the workorders and synchronizing the files, especially since we are storing them in NBT. However as for now, i think i can say that it should work.

src/main/java/com/minecolonies/coremod/colony/Structures.java
marchermans: I understand now what you are trying to achieve. Still i think we should prevent Hash Colisions at any cost. But that is my opinion.

xavierh:
> Do i understand correctly that there could be two schematics with the same contents, used for different purposes?

decoration with "waypoint" in the name are also use as a waypoint on top of being a decoration

> So say one rotated and one not, if the contained building is a perfect mirror for all four horizontal axi?

Rotation are not normally done via a different schematic, they are done when a work order is created, we set the position and rotation (and mirror soon) we want the builder to use.

>If that is the case then yeah it would not matter and you cannot use a BiMap as that forces a one to one relationship.

I don't understand what is the point of "forcing" the one to one relationship?

> However the problem still recides that a workorder that was rebuild from the MD5, is not capable of retrieving that second rotated (with the same MD5) structure, making it show the user the none rotated variant (granted the output of the workorder is equal, but still), so i donnot know if it is required but should be taken into account.

When the rotation is done inside the schematic, they will have different MD5 hash anyway.
When  the rotation is done within the work order, then any schematic we retrieve using the MD5 would do anyway (This is not the case for AbstactBuilding however)

marchermans:
Ah okey, i am understanding this now. Then you will not be able to fix your collision problem this way.

One last question before i shut up:
Is it possible for an other mod to override a given schematic?

xavierh:
> Ah okey, i am understanding this now. Then you will not be able to fix your collision problem this way.

I am not scared of collision. we use the structurename to identify the schematic not the md5. If we have a hash collision, that would only mean that the player could see a different build when using the BuildTool but the builder will use the server sided schematic.

> Is it possible for an other mod to override a given schematic?

Not cleanly. In theory,  a mod could dump new schematics in the WorldDir/minecolonies/schematics/ directory. but the list won't be refresh until the server restart.

marchermans:
@xavierh okey, then i am okey with this. Could you do me a favor. Directly under the class definition add a:
//TODO: [API]: Implement external API acces to Structures. Override and additional lookup.

That would be awesome.

xavierh:
I don't think this is the right place. The main branch to develop is normally **develop**.

This pr is for version/1.11 and will be back ported later to develop

marchermans:
I know, the API Stuff i am doing is for 1.11 and onwards. However tracking things like these when the person who created the PR does not take things like that into account is hard.

I made a note for it my self now.

But i might make a PR that adds twenty orso of those TODO's.

marchermans (approved):

src/main/java/com/minecolonies/coremod/colony/Structures.java
Raycoms: Why don't we enter the folder(style) and fileName into the hash? This would make it quite unique to avoid confusion.
If we want to prevent Hash colisions we shouldn't use md5 anyway. But Md5 is, for this use case, more than enough imo.

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildTool.java
```diff
@@ -397,15 +399,6 @@ private void hutDecClicked(@NotNull final Button button)
      */
     private List<String> getStyles()
     {
-        try
```
Raycoms: Does he still load the file onTheFly or do we want to relog to get the newest styles if we drop them off?

src/main/java/com/minecolonies/coremod/colony/Structures.java
```diff
@@ -123,28 +131,61 @@ public static void loadStyleMaps(final Path basePath) throws IOException
...
                     {
                         addHutStyle(hut, style);
                         incrementHutMaxLevel(hut);
                     }
-                    else
+                    else if (SCHEMATICS_HUTS.equals(schematicCategory) || SCHEMATICS_DECORATIONS.equals(schematicCategory))
                     {
+                        if (isSchematicHut(hut))
+                        {
+                            Log.getLogger().warn(path + " look like a hut but will not be considered as a hut!");
```
Raycoms: That happens if someone drops of decorations in the hut folders?

src/main/java/com/minecolonies/coremod/colony/Structures.java
```diff
@@ -123,28 +131,61 @@ public static void loadStyleMaps(final Path basePath) throws IOException
...
+                        }
                         addDecorationStyle(style, filename);
                     }
+                    else
+                    {
+                        Log.getLogger().error(path + " schematic is not a huts or decorations, ignoring it");
```
Raycoms: that can happen? How?

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildTool.java
xavierh: I removed this bit because it did work only in SP, this "scan" is only client side which mean, the player would have seen some schematic that the server is not aware of. The builder would not be able to build them anyway.

I will implement it properly with https://github.com/Minecolonies/minecolonies/issues/488 so it will work for SP with not restrictions and MP if allowed.

src/main/java/com/minecolonies/coremod/colony/Structures.java
xavierh: I (respectfully) disagree. using the filename and style in the MD5 would mean that we can not move the schematic around if needed.
Also it does not add any functionality to do it.

src/main/java/com/minecolonies/coremod/colony/Structures.java
xavierh: This happen when a decoration schematic filename match a hut but is in decorations folder.
I added this warning to the log because the user might have put a hut schematic in the decorations folder without realising that it won't be working as a hut (but he will can still be able to build it).

src/main/java/com/minecolonies/coremod/colony/Structures.java
Raycoms: the big question is. Do we need duplicated schematics?
What happens if we have duplicates on the server or on the client?
In the case of the guardTower corner, I think, the schematic which is the same for both styles should be more like a fallback than duplicated in both folders. (Check for equal md5 as required in own styles to check if we have it elsewhere then)
I see no reason to have duplicates around if we can define fallbacks or links.

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildTool.java
Raycoms: Okay that would be nice. Its very nice if you can scan and then build on the go without having to restart.

src/main/java/com/minecolonies/coremod/colony/Structures.java
xavierh: I am scanning the whole **schematics** folder which is normally:
 - schematics
   - huts
   - decorations

There is nothing preventing the user to create more directory in schematics directory for instance **schematics/mystuff**. 
So in the log I wanted to make the user aware that it does not work.

However I need to change the Line 167:
`else if (SCHEMATICS_HUTS.equals(schematicCategory) || SCHEMATICS_DECORATIONS.equals(schematicCategory))`

for 
`else if (SCHEMATICS_DECORATIONS.equals(schematicCategory))`

because at the moment workOrderBuildDecoration assume that schematics are in "schematics/decorations"

This will also change in the futur when implementing https://github.com/Minecolonies/minecolonies/issues/488

src/main/java/com/minecolonies/coremod/colony/Structures.java
xavierh: The only reason why I had duplicated schematic is because I was testing the custom huts features and see if the clients pick up the new schematics.

I decided to go for the TowerGuard because I was in need to have a corner version.

I asked my builder to build the 5 different normal TowerGuards, one for each level.

Modified them to do the corner version. then scan them.

I did not rescan the level 1 because I did not change anything, and just copy the TowerGuard1 to the **Corner** style.

Also later one, player will be be able to scan. My plan is to store those in "schematics/players/${player UUID}/"
I can not control what player will scan if two players scan a 2x2x2 Cobblestone schematic, they will have the same MD5 hash. but each schematics will be in their own  player directory, so that each player can only uses their schematic not what another player have done.

>I see no reason to have duplicates around if we can define fallbacks or links.

We have nothing in place will allow you to have different path upgrade

Raycoms:
I'm currently thinking about pretty huge structures people might download
on the same website and will both upload.
The server could recognize he has that hash already and save space?
2017-03-02 11:15 GMT-03:00 xavierh <notifications@github.com>:

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildTool.java
xavierh: Whatever method I will implement to add new schematic on the fly will trigger a refresh of this list.

xavierh:
@Raycoms I understand your point but this PR does not allow to upload stuff, that is why I did not bother about those considerations.

Changes will be made when required I guess.

Raycoms:
:shipit: for me then.

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/579/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

Raycoms:
Can you create a backport to 1.10 as well?

xavierh:
@Raycoms As discussed I changed Line 167

For 1.10, what is the best way to backport, could you briefly describe the process with git ?

Raycoms:
I usually make a squash merge in another 1.11 branch. Then cherrypick this
in a copy of develop of 1.10 and then I fix the compile errors and test it.

2017-03-02 11:44 GMT-03:00 xavierh <notifications@github.com>:

xavierh:
@Raycoms 
cool new stuff to learn. Would it not make send to squash now so it would be only one commit in version/1.11 ?

Raycoms:
We will have to wait for @Marvin or @colton to approve your pr to squash
merge already. And that might need a while =D

2017-03-02 12:08 GMT-03:00 xavierh <notifications@github.com>:

cltnschlosser:
Didn't we decide we didn't want to force users to download schematics. They can be made available through resource packs, and if you don't have the pack, you just won't see a preview in the build tool.

marvin-bitterlich:
@cltnschlosser not force maybe, but at least offer to do it

marvin-bitterlich:
@xavierh I suggest you ping is somewhere when you change the pr status (and request our review ;) )

xavierh:
@Kostronor Was I that slow ? :D

marvin-bitterlich:
@xavierh It was sitting here for quite some time. If there wasn't ray asking me to recheck it I'd just had ignored it 😛

marvin-bitterlich:
Oh btq @xavierh I guess we have to move the pr anyway because sonar etc. are not working on forks. Would that be okay for you?

xavierh:
Yes
I would like someone to have a second close look to a specific part, please

marvin-bitterlich:
It's 108 files, I won't review that before sonar has a go at it xD ping me on specific things you want answered :)

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildTool.java
xavierh: @Raycoms You are right, 
I will load the client side directory but only when the worlds is not remote.
I guess I could do the same with the scan folder while I am at it ?

Raycoms:
We can, shouldn't be a problem.

2017-03-02 19:13 GMT-03:00 xavierh <notifications@github.com>:

xavierh:
I need to work on this a bit more. as @Raycoms mention, I lost a functionality.
SP player can not drop schematic in the decoration folder and use it **without restarting** the game.

xavierh:
Could somebody comment on those lines:
https://github.com/Minecolonies/minecolonies/blob/feature/Buildtool-server-side-nbt/src/main/java/com/minecolonies/structures/helpers/Structure.java#L163

marchermans:
@xavierh Is that not always true cause you are using the directory as a prefix?

xavierh:
I think not.
The client could ask for a schematic named "decorations/../../anyfile"
The question is, does this test is enough ?

I should also do it on the client side so when the client download a file he does not write outside of its directory

xavierh:
Todo:

- add scan folder back (the location will change) for sp user
- save cached  schematic using the md5 hash instead to avoid security issue. (Also will be reused for player custom schematic on mp)

Unfortunately I will need more time too impment it correctly :(

xavierh:
Will create a clean pr again when ready

PR #863, opened 2017-04-25
Clean up the fishing loot calculus

marvin-bitterlich:
Improve the chances and clean up the calculation

@Raycoms needed?


Raycoms:
its correct its in develop already
Not needed

PR #1058, opened 2017-05-23
New crafts for build, scan and permission tools

MineShadow7:
So I think that this would be good if we can craft a scan tool from build tool, perm tool from scan tool and build tool from perm tool. 
- Craft for build tool
- Craft for scan tool
- Craft for permission tool

Review please


xavierh:
I dont think you wanted to submit a pr.

To open an issue, this way: https://github.com/Minecolonies/minecolonies/issues/new

PR #1063, opened 2017-05-24
Update build.gradle

marvin-bitterlich:
damn case zroblems.... cant checkout…………………

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,7 @@ sourceSets {
         resources {
             srcDir 'src/structures/resources'
         }
-        compileClasspath += sourceSets.api.compileClasspath
+        // compileClasspath += sourceSets.api.compileClasspath
     }
     main {
         java {
@@ -85,8 +85,11 @@ sourceSets {
 }
 
 dependencies {
-    structuresCompile sourceSets.api.output
-    blockOutCompile sourceSets.api.output
+    // structuresCompile sourceSets.api.output
+    // blockOutCompile sourceSets.api.output
+
+    testCompile sourceSets.structures.output
+    testCompile sourceSets.blockOut.output
 
     testCompile sourceSets.structures.output
     testCompile sourceSets.blockOut.output
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ sourceSets {
         resources {
             srcDir 'src/blockout/resources'
         }
-        compileClasspath += sourceSets.api.compileClasspath
+        // compileClasspath += sourceSets.api.compileClasspath
     }
     structures {
         java {
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ sourceSets {
         resources {
             srcDir 'src/blockout/resources'
         }
-        // compileClasspath += sourceSets.api.compileClasspath
+        compileClasspath += sourceSets.api.compileClasspath
     }
     structures {
         java {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -52,7 +52,7 @@ sourceSets {
         resources {
             srcDir 'src/blockout/resources'
         }
-        compileClasspath += sourceSets.api.compileClasspath
+        // compileClasspath += sourceSets.api.compileClasspath
```
marchermans: This is needed. It does not at a depencency to the api module. But to the dependencies of the API module. AKA Minecraft and such.

build.gradle
```diff
@@ -61,7 +61,7 @@ sourceSets {
         resources {
             srcDir 'src/structures/resources'
         }
-        compileClasspath += sourceSets.api.compileClasspath
+        // compileClasspath += sourceSets.api.compileClasspath
```
marchermans: Same as above.

build.gradle
```diff
@@ -85,8 +85,11 @@ sourceSets {
 }
 
 dependencies {
-    structuresCompile sourceSets.api.output
-    blockOutCompile sourceSets.api.output
+    // structuresCompile sourceSets.api.output
```
marchermans: This will remove the dependency of structures on the API, you will Need this later, as some structures classes check for hut blocks. Please Keep.

marchermans: Or Keep it at least in mind when we move the classes

build.gradle
```diff
@@ -85,8 +85,11 @@ sourceSets {
...
+    // blockOutCompile sourceSets.api.output
```
marchermans: This will remove the dependency api from blockout. This is correct.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/permissions/Permissions.java
```diff
@@ -545,8 +541,10 @@ public void savePermissions(@NotNull final NBTTagCompound compound)
      */
     public Set<UUID> getMessagePlayers()
     {
-        return players.values().stream().filter(player ->
-                                                  hasPermission(player.getRank(), RECEIVE_MESSAGES)).map(player -> player.getID()).collect(Collectors.toSet());
+        return players.values().stream()
+                 .filter(player -> hasPermission(player.getRank(), Action.RECEIVE_MESSAGES))
```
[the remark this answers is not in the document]
marchermans: Overruled, causes Problems with the ValueOf methods in both Action and Rank

src/main/java/com/minecolonies/coremod/colony/permissions/Permissions.java
```diff
@@ -651,7 +649,7 @@ public void removePermission(final Rank rank, @NotNull final Action action)
     public Rank getRank(final UUID id)
     {
         final Player player = players.get(id);
-        return player != null ? player.getRank() : NEUTRAL;
+        return player != null ? player.getRank() : Rank.NEUTRAL;
```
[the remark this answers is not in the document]
marchermans: Overruled, see above

marchermans (approved):

marchermans:
As this is done against my Feature/api-rework branch i will pull this regardless of the build not running properly through.

I need to fix the ColonyManager for the build to pass.

PR #1860, opened 2017-12-17
Resolves #1857

### LINKED ISSUE #1857: Multiple Farmers -  Assignment issue
mine craft 1.12.2
mine colonies 8.5989

Once you have a mine colony with 2 farms.   I have 5 fields.  I try to set them up to "manual".  I have one farmer hut level 4.  I have a new farmer hut lvl 1. 

Lvl 4 hut is assign f farms 1 - 4 and lvl 1 hut assigned farm 5.

If you go to the fields and right click on the scarecrow.  It has farmer 1 (lvl 4) is assigned to every scarecrow.    Even though the farmer hut shows it correctly just farms 1 - 4.    

It doesn't appear that you can get it done correctly.   If you remove all assignments from each hut  and break the scarecrow and reassign the farmers back to fields.    When it finally updates,  it shows farmer (lvl4 hut) is assigned to all the fields.

bob0bob:
This resolves issue #1857.   Assignments are not working because the farmerFields get out of sync.  This makes sure when it tried to sync the field is assigned to this farmer.

It appears it is a sync issue with farmerFields. Once it is out of sync, it can never get back into sync.
A quick look at syncWithColony(). It doesn't do what the name indicates. It doesn't sync because when it is going through the list. It assumes everything in the farmerFields is assigned to him.

I dont' know the correct solution but this worked.

I changed:
if (scarecrow == null)
to:
if (scarecrow == null || field.getOwner().compareTo(getMainWorker().getName()) != 0)

This also put a second check in if the sync of the fields are off. It check the owner. I assume there is a better version. Like a unique IDEA??? Because a rename of citizen would cause them to lose this field. but I can't locate a variable inside of Field that would use a unique ID to point back to the farmer. I don't see one. So this seems like the best solution.

BUT THIS SOLVES the assignment issue, When someone has gone through mod with bugs, or other issues. This will make sure it is in sync correctly.
Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/BuildingFarmer.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/BuildingFarmer.java
@@ -1,5 +1,6 @@
 package com.minecolonies.coremod.colony.buildings;
 
+import com.minecolonies.api.util.BlockPosUtil;
 import com.minecolonies.api.util.ItemStackUtils;
 import com.minecolonies.api.util.LanguageHandler;
 import com.minecolonies.api.util.constant.ToolType;
@@ -14,7 +15,6 @@ import com.minecolonies.coremod.colony.jobs.AbstractJob;
 import com.minecolonies.coremod.colony.jobs.JobFarmer;
 import com.minecolonies.coremod.entity.ai.citizen.farmer.Field;
 import com.minecolonies.coremod.entity.ai.citizen.farmer.FieldView;
-import com.minecolonies.coremod.entity.ai.item.handling.ItemStorage;
 import com.minecolonies.coremod.network.messages.AssignFieldMessage;
 import com.minecolonies.coremod.network.messages.AssignmentModeMessage;
 import com.minecolonies.coremod.tileentities.ScarecrowTileEntity;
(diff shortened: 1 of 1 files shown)


marchermans:
There is the citizen id. Would be cool if you would make it so it uses that instead of the Name.

marchermans (changes requested):
The formatting is wrong here i think, Please fix.

Raycoms:
Is there a way to format the file to make sure that we can see what the real changes in the pr are?

Raycoms:
YOu can also additionally store the ciitzen.id, the name is unique as well though

marchermans:
@Raycoms no it is not anymore. Since you introduced renaming.

Raycoms:
I still made sure it's unique even when renaming.

2017-12-17 16:26 GMT+00:00 Marc Hermans <notifications@github.com>:

bob0bob:
I;m not sure what happened on the upload of the new file.  I'm still new at git.   It seeing it as a whole new change instead of just the couple of things.  

Also,  I looked in the class Field,  which is the class for each field the farmer has.  There is no citizen id  being stored in the field.   Based on the coding it didn't appear to really need one except for bugs caused the sync to get out of step and there is no way to get back.     So at this point the only option is add a new field and then when assignment happens save the citizen id and then you can change the check.  but to solve current issues.  

I don't see another way.  There might be one for someone that is experienced in MC.  But I don't see it.

Would need to change assignment,  writenbt,readnbt, and remove assignment to make it work, but not sure how upgrading to a new version with new nbt works.  I leave that to experts in the knowledge of MC

Raycoms:
@bob0bob can you contact me on discord?

PR #2614, opened 2018-07-13
Town Hall - Should Save build order. #2595

### LINKED ISSUE #2595: Town Hall - Should Save build order.
When you take the time to reorder the build order from the town hall gui.  This should be saved.   On reopening the world again.  The order is lost and I have to reorder the building process again.

bob0bob:
Based on the code. when writing to the NBT file. the priority is not saved. So on load everything is set to 0 again. Class AbstractWorkOrder looks like it holds the priority but the writeNBT and createFromNBT is not using that variable. I've have code changes to fix this. if you want me to submit them for review.

Closes #2595

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
+++ b/src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
@@ -22,6 +22,7 @@ import static com.minecolonies.api.util.constant.Suppression.UNUSED_METHOD_PARAM
 public abstract class AbstractWorkOrder
 {
     private static final String                                          TAG_TYPE       = "type";
+    private static final String                                          TAG_PRIORITY   = "priority";
     private static final String                                          TAG_ID         = "id";
     private static final String                                          TAG_CLAIMED_BY = "claimedBy";
     //  Job and View Class Mapping
@@ -112,6 +113,7 @@ public abstract class AbstractWorkOrder
         }
         try
         {
+        	order.setPriority(compound.getInteger(TAG_PRIORITY));
             order.readFromNBT(compound);
         }
         catch (final RuntimeException ex)
@@ -132,6 +134,7 @@ public abstract class AbstractWorkOrder
     public void readFromNBT(@NotNull final NBTTagCompound compound)
     {
         id = compound.getInteger(TAG_ID);
+        priority = compound.getInteger(TAG_PRIORITY);
         claimedBy = compound.getInteger(TAG_CLAIMED_BY);
     }
 
@@ -278,6 +281,7 @@ public abstract class AbstractWorkOrder
             throw new IllegalStateException(this.getClass() + " is missing a mapping! This is a bug!");
         }
 
+        compound.setInteger(TAG_PRIORITY, priority);
         compound.setString(TAG_TYPE, s);
         compound.setInteger(TAG_ID, id);
         if (claimedBy != 0)
--- a/src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
+++ b/src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
@@ -134,7 +134,8 @@ public abstract class AbstractWorkOrder
     public void readFromNBT(@NotNull final NBTTagCompound compound)
     {
         id = compound.getInteger(TAG_ID);
-        priority = compound.getInteger(TAG_TH_PRIORITY);
+        if (compound.hasKey(TAG_TH_PRIORITY))
+        	priority = compound.getInteger(TAG_TH_PRIORITY);
         claimedBy = compound.getInteger(TAG_CLAIMED_BY);
     }
 
--- a/src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
+++ b/src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
@@ -122,7 +122,7 @@ public abstract class AbstractWorkOrder
         catch (final RuntimeException ex)
         {
             Log.getLogger().error(String.format("A WorkOrder %s(%s) has thrown an exception during loading, its state cannot be restored. Report this to the mod author",
-              compound.getString(TAG_TYPE), oclass.getName()), ex);
+            compound.getString(TAG_TYPE), oclass.getName()), ex);
             return null;
         }
 
@@ -162,7 +162,7 @@ public abstract class AbstractWorkOrder
         catch (final RuntimeException ex)
         {
             Log.getLogger().error(String.format("A WorkOrder.View for #%d has thrown an exception during loading, its state cannot be restored. Report this to the mod author",
-              workOrderView.getId()), ex);
+            workOrderView.getId()), ex);
             workOrderView = null;
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
```diff
@@ -132,6 +134,7 @@ public static AbstractWorkOrder createFromNBT(@NotNull final NBTTagCompound comp
     public void readFromNBT(@NotNull final NBTTagCompound compound)
     {
         id = compound.getInteger(TAG_ID);
+        priority = compound.getInteger(TAG_PRIORITY);
```
Raycoms: we have to do a compound.hasKey(TAG_PRIORITY) first, to avoid an exception.

src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
```diff
@@ -112,6 +113,7 @@ public static AbstractWorkOrder createFromNBT(@NotNull final NBTTagCompound comp
         }
         try
         {
+        	order.setPriority(compound.getInteger(TAG_PRIORITY));
```
Raycoms: Also compound.hasKey(TAG_PRIORITY), and please move on the same line as call under it.

bob0bob:
Updating code to reflect the changes.

LDTTeam-Buildsystem (dismissed):
#### Analysis Complete
##### Statistics:
|        | Count | New | Old | Diff |
| ------ |:-----:|:---:|:---:|:----:|
| Total  | 34007 |  0  |  1  |  -1  |
| Errors |  593  |  0  |  0  |  0   |
##### More Information:

src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
```diff
@@ -132,6 +134,7 @@ public static AbstractWorkOrder createFromNBT(@NotNull final NBTTagCompound comp
     public void readFromNBT(@NotNull final NBTTagCompound compound)
     {
         id = compound.getInteger(TAG_ID);
+        priority = compound.getInteger(TAG_TH_PRIORITY);
         claimedBy = compound.getInteger(TAG_CLAIMED_BY);
     }
```
Raycoms: still needs a compound.hasKey and the formatting of order.setPriority needs the same and formatting

marchermans: Yep.

LDTTeam-Buildsystem (dismissed):
#### Analysis Complete
##### Statistics:
|        | Count | New | Old | Diff |
| ------ |:-----:|:---:|:---:|:----:|
| Total  | 34007 |  0  |  1  |  -1  |
| Errors |  593  |  0  |  0  |  0   |
##### More Information:

src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
bob0bob: I'm not following!   Please explain.

Raycoms: If we load a save which does not have the priority tag set, it will crash.
So you have to check if the tag exists first.
With: compound.hasKey

Raycoms (changes requested):
We need some saveguards to avoid crashes here

bob0bob:
What do you mean to avoid crashes???
The default return for "compound.getInteger" method if tag doesn't exist is 0.   it will not throw an exception.

Raycoms:
I saw a bunch of crashes already because we accessed not existant tags,
it's always better to verify if the key exists before we access it.

2018-07-16 11:33 GMT-03:00 bob0bob <notifications@github.com>:

Raycoms:
Also because possibly, they might change the default value in an update to
-1 which can cause some issues.

2018-07-16 11:38 GMT-03:00 Ray Neiheiser <ray.neiheiser@gmail.com>:

marchermans:
@bob0bob It is always better to provide custom up and downgrade logic when you create IO based Code. Downgrade is here irrelevant since we simple will not read the NBT in older Code, but for upgrade Code it is important that you Control the upgrade process.

Mojang uses a System called Datafixers for this Kind of purpose, which would be completely Overkill here, a simple check if the tag exists suffices for us.

bob0bob:
@ Raycoms  Didn't think of the changing of default data.  I'll add that check when I have a chance.  Thanks.

LDTTeam-Buildsystem (dismissed):
#### Analysis Complete
##### Statistics:
|        | Count | New | Old | Diff |
| ------ |:-----:|:---:|:---:|:----:|
| Total  | 34008 |  1  |  0  |  +1  |
| Errors |  593  |  0  |  0  |  +0  |
##### More Information:
- [Analysis - Full](https://teamcity.minecolonies.com/viewLog.html?buildId=12258&tab=Inspection&buildTypeId=bt88)

LDTTeam-Buildsystem (dismissed):
#### Analysis Complete
##### Statistics:
|        | Count | New | Old | Diff |
| ------ |:-----:|:---:|:---:|:----:|
| Total  | 34008 |  0  |  0  |  0   |
| Errors |  593  |  0  |  0  |  0   |
##### More Information:

marchermans (dismissed):

src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
```diff
@@ -112,6 +113,7 @@ public static AbstractWorkOrder createFromNBT(@NotNull final NBTTagCompound comp
         }
         try
         {
+        	order.setPriority(compound.getInteger(TAG_TH_PRIORITY));
```
Raycoms: if (compound.hasKey(TAG_TH_PRIORITY))
Also needed here

Raycoms (changes requested):

LDTTeam-Buildsystem (dismissed):
#### Analysis Complete
##### Statistics:
|        | Count | New | Old | Diff |
| ------ |:-----:|:---:|:---:|:----:|
| Total  | 34008 |  0  |  0  |  0   |
| Errors |  593  |  0  |  0  |  0   |
##### More Information:

src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
bob0bob: You dont need a haskey  on the save.  It creates the key.  If you put that then the save would not work.  Only on the read.   The write will add it and from that point have the key.  THis would break the code from working.

Raycoms:
but it's compound.getInteger and not setInteger

2018-07-17 20:03 GMT-03:00 bob0bob <notifications@github.com>:

bob0bob:
You right,  I saw the set and over looked the get of the compound.  I checked it in.

LDTTeam-Buildsystem (dismissed):
#### Analysis Complete
##### Statistics:
|        | Count | New | Old | Diff |
| ------ |:-----:|:---:|:---:|:----:|
| Total  | 34008 |  0  |  0  |  0   |
| Errors |  593  |  0  |  0  |  0   |
##### More Information:

src/main/java/com/minecolonies/coremod/colony/workorders/AbstractWorkOrder.java
marchermans: If this is the save, why are you doing compound.getInteger()?

Issue #2610, opened 2018-07-11
Guards - standing around like crash dummies.


bob0bob:
### Minecolonies version

current MC

### Expected behavior

- Guard should patrol and attack mobs.

### Actual behaviour

- Guards basically stand around and do nothing.

Real problem is the NBT load routine does check to make sure the mob located in the save data still exists.  If the mod has been removed since last save.  The mob is deleted MC will crash (exception caught) and many things about MC doesn't work.  patrolling, some of the guis.

1. XYZ
2. ZYX
3. ...
4. Profit

bob0bob:
Fixed in pr 2609.  Wish you could have more than one pr ticket to merge back into the main branch.  Only one ticket so everything gets merge.  Wish you could select which files.

Raycoms:
What do you mean with this?

2018-07-11 1:42 GMT-03:00 bob0bob <notifications@github.com>:

PoroUsedSnax:
@bob0bob You can make separate branches to merge into version/1.12, that way you can choose which of those get merged as well as creating separate PRs for each branch

Raycoms:
You should also be able to cherrypick certain commits into these branches
to separate your existing PR into a few small ones.

2018-07-11 10:02 GMT-03:00 Mark Fisher <notifications@github.com>:

bob0bob:
Are you saying each bug fix, I have to have a new branch???   As you can see from my test branch I do not have "Create new pr".   Only option is "view" and it automatically pulls everything I check in, into that PR.   I do not see a setting that can turn that off?   I've not been able to locate anything in help on github, at least the way I'm searching for it.


![image](https://user-images.githubusercontent.com/30982580/42575999-2443126c-84e7-11e8-8333-11b908bded93.png)

Raycoms:
Yes, you have to create 1 branch per pr.

2018-07-11 10:49 GMT-03:00 bob0bob <notifications@github.com>:

PR #2620, opened 2018-07-14
Guards - standing around like crash dummies. #2610

bob0bob:
Real problem is the NBT load routine does check to make sure the mob located in the save data still exists. If the mod has been removed since last save or an update to a mob that removed a mob. The mob is deleted MC will crash (exception caught) and many things about MC doesn't work. patrolling, some of the guis.


Asherslab (dismissed):
Huh, didn't even think of it being possible for the EntityEntry to be null, good spot

bob0bob:
added the braces.

Asherslab (approved):
Looks good!

Asherslab:
once the checks are all done, i'll merge this if it's good

LDTTeam-Buildsystem (approved):
#### Analysis Complete
##### Statistics:
|        | Count | New | Old | Diff |
| ------ |:-----:|:---:|:---:|:----:|
| Total  | 34015 |  0  |  0  |  0   |
| Errors |  593  |  0  |  0  |  0   |
##### More Information:

Issue #2623, opened 2018-07-14
Name does not show up on builder huts gui.


bob0bob:
### Expected behavior

When you click on the builder hut, you should see the name on the hut

### Actual behaviour

it is blank

Builders hut gui doesn't use the windoworkerbuildingdefault.xml file like all other job huts.  It doesn't call the default code and uses it own.  I missed that.

Issue #2618, opened 2018-07-13
Client crash when clicking on builder chest


lebanni:
minecolonies-1.12.2-0.9.10-ALPHA-universal.jar

[crash-2018-07-13_16.07.14-client.txt](https://github.com/ldtteam/minecolonies/files/2193023/crash-2018-07-13_16.07.14-client.txt)

winex:
it crashes because windowhutbuilder.xml doesn't contain list, commit 01c15bc3abb6cd96fb3d491504ec5476dca2988d doesn't check for null probably at src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowWorkerBuilding.java:139
i've tried to simply edit this file (.xml) and it works now...

Raycoms:
Why don't you make a pull request for the fix you made?

winex:
um, i think that there should be null pointer check in that commit, but i don't know how to properly do it...
xml fix is the same, but crashes will happen again w/o proper checks :(

Raycoms:
Just pushed the hotfix, thanks

PR #2624, opened 2018-07-14
Name does not show up on builder huts gui. #2623

bob0bob:
When you click on the builder hut, you should see the name on the hut.
Builders hut gui doesn't use the windoworkerbuildingdefault.xml file like all other job huts. It doesn't call the default code and uses it own. I missed that.


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/2624/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

winex:
yeah, that completes the #2618 :)
btw, why builder is so special, can't it just include worker default xml as other windows does?

PR #2699, opened 2018-07-28
Barbarian - Attack Rate - not using difficulty settings in progress #…

bob0bob:
…2697

This change is to affect the barbarians so that based on the difficulty level in the settings that rate at which the barbarians will attack will be affected.
This way on a very low settings in difficulty, the barbarians will have a slow attack rate vs. higher rates. This will make it a little more different game player overall.


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/2699/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

PR #2789, opened 2018-08-15
Allow adding extra ores for miners to recognize in the config

mraof:
Allows defining a list of resource locations as strings in the config for ores for the miner to recognize in case there are ores that weren't automatically found


marchermans (approved):
Perfect small and acceptable change.

PR #2838, opened 2018-09-01
Workaround for fps

Raycoms:
- Improve FPS a bit

Review please


marchermans (approved):
Yeah the new BlockOut fixes this indefinetly, and I should have used something like fst and kryo to do serializing and object creation, but afterwards you are Always smarter.

I added the factory System to the list of RS rewrites that i Need to make anyway.

PR #2914, opened 2018-09-26
New sandstone schematics by @Sindarin

Nightenom:
Review please


Raycoms:
I think we should avoid making the jar bigger at this point, better add them to the structure mod so it's downloadable =)

PR #2952, opened 2018-10-09
[TEST - PR] Do not close.

marchermans:


marchermans:
Good that we are removing this.
Upsource produces the same results under: https://upsource.minecolonies.com/

marchermans:
Completing test review from Upsource. Testing making a comment.

PR #2955, opened 2018-10-09
Library feature - consume paper for faster levelup

someaddons:
Changes proposed in this pull request:
- Suggestion to make the library's students use up paper for faster skill levelup, to make the library a bit more interactive/faster

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLibrary.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLibrary.java
@@ -1,5 +1,7 @@
 package com.minecolonies.coremod.colony.buildings.workerbuildings;
 
+import com.minecolonies.api.configuration.Configurations;
+import com.minecolonies.api.util.Log;
 import com.minecolonies.blockout.views.Window;
 import com.minecolonies.coremod.client.gui.WindowHutWorkerPlaceholder;
 import com.minecolonies.coremod.colony.CitizenData;
@@ -8,12 +10,14 @@ import com.minecolonies.coremod.colony.ColonyView;
 import com.minecolonies.coremod.colony.buildings.AbstractBuildingWorker;
 import com.minecolonies.coremod.colony.jobs.AbstractJob;
 import com.minecolonies.coremod.colony.jobs.JobStudent;
+import com.minecolonies.coremod.entity.ai.util.StudyItem;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockBookshelf;
-import net.minecraft.init.Items;
+import net.minecraft.item.Item;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.nbt.NBTTagList;
 import net.minecraft.nbt.NBTUtil;
+import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 import net.minecraftforge.common.util.Constants;
@@ -57,6 +61,11 @@ public class BuildingLibrary extends AbstractBuildingWorker
     private final Random random = new Random();
 
     /**
+     * Study Item list, loaded from config, Typle<int,int> First is the study chance increase,second is the breakchance
+     */
+    private final ArrayList<StudyItem> studyItems;
+
+    /**
      * Instantiates the building.
      *
      * @param c the colony.
@@ -65,7 +74,40 @@ public class BuildingLibrary extends AbstractBuildingWorker
     public BuildingLibrary(final Colony c, final BlockPos l)
     {
         super(c, l);
-        keepX.put(itemStack -> itemStack.getItem() == Items.PAPER, 200);
+
+        studyItems = parseFromConfig();
+    }
+
+    /**
+     * Parses Study Items from the Config and adds them on the keepX list
+     */
+    private ArrayList<StudyItem> parseFromConfig()
+    {
+        ArrayList<StudyItem> studyItemList = new ArrayList<>();
+
+        for (String entry : Configurations.gameplay.configListStudyItems)
+        {
+            String[] entries = entry.split(";");
+            if (entries.length < 3)
+            {
+                Log.getLogger().info("Minecolonies: Parsing config for study items for Library failed for entry:" + entry);
+                continue;
+            }
+            Item item = Item.REGISTRY.getObject(new ResourceLocation(entries[0]));
+            int skillChance = Integer.parseInt(entries[1]);
+            int breakChance = Integer.parseInt(entries[2]);
+
+            if (item == null || skillChance < 100 || skillChance > 1000 || breakChance > 100 || breakChance < 0)
+            {
+                Log.getLogger().info("Minecolonies: Parsing config for study items for Library failed for entry:" + entry);
+                continue;
+            }
+
+            studyItemList.add(new StudyItem(item, skillChance, breakChance));
+            // Keep a certain part of the items in the Chest
+            keepX.put(itemStack -> itemStack.getItem() == item, breakChance < 5 ? 5 : breakChance);
+        }
+        return studyItemList;
     }
 
     @NotNull
@@ -154,6 +196,11 @@ public class BuildingLibrary extends AbstractBuildingWorker
         }
         bookCases.remove(returnPos);
         return getLocation();
+    }
+
+    public ArrayList<StudyItem> getStudyItems()
+    {
+        return studyItems;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
```diff
@@ -19,13 +24,18 @@
...
     /**
      * The current pos to study at.
      */
     private BlockPos studyPos = null;
 
+    /**
+     * The paper item to use
+     */
+    private final Item paper = GameRegistry.makeItemStack("Paper", 0, 1, null).getItem();
+
```
Raycoms: Items.Paper would do that as well, wouldn't it?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
```diff
@@ -75,7 +85,24 @@ private AIState study()
...
+        final int slot = InventoryUtils.findFirstSlotInProviderNotEmptyWith(worker,
+          itemStack -> !ItemStackUtils.isEmpty(itemStack) && itemStack.getItem() == paper);
+
+        if (slot == -1)
+        {
+            data.tryRandomLevelUp(world.rand);
+
+            if (data.getJob().getAsyncRequests().isEmpty())
+            {
+                data.createRequestAsync(new Stack(GameRegistry.makeItemStack("Paper", 0, 10, null)));
```
Raycoms: same thing here

Raycoms:
Not sure, I think in this case we should have two more things:

1) During studying they should wield the paper in their hand
2) They should do async requests to get paper delivered.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
someaddons: ooh ty, coudlnt find that item registry

someaddons:
ye that sounds good, gonna add that

src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
```diff
@@ -75,7 +80,24 @@ private AIState study()
...
+        final int slot = InventoryUtils.findFirstSlotInProviderNotEmptyWith(worker,
+          itemStack -> !ItemStackUtils.isEmpty(itemStack) && itemStack.getItem() == Items.PAPER);
+
+        if (slot == -1)
+        {
+            data.tryRandomLevelUp(world.rand);
+
+            if (data.getJob().getAsyncRequests().isEmpty())
+            {
+                data.createRequestAsync(new Stack(new ItemStack(Items.PAPER,10)));
```
Raycoms: There is also a way to check if there is a similar request already in the AbstractEntityAiBasic afaik.

someaddons: found it, thx

Raycoms:
Looks good, did you test this well, also in regards of game balance?

someaddons:
ye tested it and seems fine, at the cost of 1 paper each study try the levelup is 3times as likely, so on avg you get an additional skillpoint for 50 paper

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/2955/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

marchermans:
Maybe also make him request books and use a 3,5 times factor for paper? Maybe also add written books with a 5x factor over paper.

marchermans (changes requested):
Please make at least some components of this configurable. Their are some bigger mods out there that add some kind of paper alternative (looks at forestry for example).

We should really look into making our AIs more flexible when it comes to this kind of stuff. Not just Items.PAPER- But OreDictionary paper needs to be accepted.

Also maybe add support for books and written books (as with trading with vanilla villagers, those are worth more).

Nightenom:
as close as possible to real life it would go something like this: request writable, a thing to write (which may also include a filling) <- or just request book and quill, we could add an easter egg when you will pass them written book with lorem ipsum paragraph - this will add a degree to their name and they will be given e.g. 1.4x learning multiplier
edit: 5x multiplier looks a bit overpowered

Raycoms:
Following orions proposal we could make it completely configurable already giving defaults:
Configuration list similar to our list for the compostables:
"item multiplier" each of them with a whitespace between the other so we can parse them then.

someaddons:
Config seems like a nice idea, makes it adjustable for however ppl want to use the mod. Book and quill could be seen/used as a "tool" required to study which breaks after X uses

someaddons:
hm don't you think that requesting multiple items with various rates could be a feature overkill and leave the player confused/annoyed with the feature? Since minecolonies in gerneal is pretty simplistic

Raycoms:
I'm currently thinking about it, I think it makes sense to just move the paper to the configuration and by default make it only 1 item with its rate. If players want to go the complicated way they can add additional items with additional modifiers. Or specific modded papers.

If they want to add books, let them do it.

someaddons:
ye I feel thats the best compromise for it aswell

someaddons:
So not sure about some feature-design decisions:

1. Should all items be able to get Hold in the mainhand? if not make it configurable via another list?
2. Should all items have a chance to be consumed when studying, or just a single item at a time(maybe in a certain order? so that less durable Items are rolled first, configurable?). If a single item: Do we use the items in a certain order? or cycle through them, or randomly select one?
3. Should all items be async requests or make it configurable?

Raycoms:
I'd go with the easiest approach:
All items hold
One item at a time (the first item first, only calculate durability if the
item itself has durability and if the item has durability decrease it a
constant amount per use)
If there is one item in the list only then one item at a time, if the
player added more to the list then make a StackList request similar to the
composter.

Am Do., 11. Okt. 2018 um 17:21 Uhr schrieb someaddons <
notifications@github.com>:

someaddons:
what I meant with durability was more in terms of after how many uses is an item used up? or give it a configurable break chance upon use. When using up single items in which order would you try that, since it would change the probabilities when you do sth like if(rollItem1) return, if rollitem(2) return etc

Raycoms:
I'd use always the first item we find in the inventory and use its chance that's the easiest and most understandable for the player. In terms of durability I wouldn't implement that right now. Would make things too complex since we would have to store the durability as nbt in the item.

someaddons:
gonna go for a % break chance then, works kinda like durability but doesnt need to be saved to nbt

marchermans (dismissed):

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLibrary.java
```diff
@@ -55,14 +60,54 @@
...
+            if (entries.length < 3)
+            {
+                Log.getLogger().info("Minecolonies: Parsing config for study items for Library failed for entry:" + entry);
+                continue;
+            }
+            final Item item = Item.REGISTRY.getObject(new ResourceLocation(entries[0]));
+            final int skillChance = Integer.parseInt(entries[1]);
+            final int breakChance = Integer.parseInt(entries[2]);
+
+            if (item == null || skillChance < 100 || skillChance > 1000 || breakChance > 100 || breakChance < 0)
```
Raycoms: better create try and catch here to avoid crashes on load up

src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
```diff
@@ -75,7 +83,60 @@ private AIState study()
...
+
+        for (final StudyItem curItem : getOwnBuilding(BuildingLibrary.class).getStudyItems())
+        {
+            final int slot = InventoryUtils.findFirstSlotInProviderNotEmptyWith(worker,
+              itemStack -> !ItemStackUtils.isEmpty(itemStack) && itemStack.getItem() == curItem.getItem());
+
+            if (slot != -1)
+            {
+                curItem.setSlot(slot);
+                currentItems.add(curItem);
```
Raycoms: break?

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLibrary.java
```diff
@@ -55,14 +60,54 @@
...
      * Instantiates the building.
+     *
      * @param c the colony.
      * @param l the location.
      */
     public BuildingLibrary(final Colony c, final BlockPos l)
     {
         super(c, l);
+
+        studyItems = parseFromConfig();
```
Raycoms, line 78: we should read and write these from and to NBT as well, else we lose the break progress on world unload/load

someaddons (commented):
a

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLibrary.java
someaddons: Instead of keeping track of the break progress, the items just got a % chance to break upon usage. So nothing to save there

PR #2984, opened 2018-10-17
Feature/job registry

Raycoms:
- Extracts Job Registry in a separate class.
- Moves actions done into abstractJob to store it to nbt.

And no @OrionDevelopment I didn't move it to a forge registry on purpose yet.

Review please


marchermans (approved):
We have not done it anywhere else. So well.

PR #3004, opened 2018-10-25
Update Chinese language files

mc-kaishixiaxue:
- zh_cn: Update Simplified Chinese translation.
- zh_tw: Create Traditional Chinese translation.


Raycoms:
Could you update things to https://translate.minecolonies.com/ ?
Since if we merge this commit it will be overwritten with the next language file commit then.

PR #3026, opened 2018-10-30
Pathfinding fixes:

someaddons:

Pathfinding fixes:
-Capping deliveryman movementspeed increase
-Fixing various Movementpaths to not be recreated before the prev is completed
-Reducing Barbarian logic to 2 tps from 20, also reduces their path creation on each tick
-Introducing a cap to the amount of nodes the pathfinding can map, to reduce unnecessary calculations(e.g. prev stuck nodes:31000, new max:5000)

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
@@ -11,7 +11,9 @@ import com.minecolonies.api.util.InventoryUtils;
 import com.minecolonies.api.util.ItemStackUtils;
 import com.minecolonies.blockout.Log;
 import com.minecolonies.coremod.colony.Colony;
-import com.minecolonies.coremod.colony.buildings.*;
+import com.minecolonies.coremod.colony.buildings.AbstractBuilding;
+import com.minecolonies.coremod.colony.buildings.AbstractBuildingContainer;
+import com.minecolonies.coremod.colony.buildings.AbstractBuildingWorker;
 import com.minecolonies.coremod.colony.buildings.workerbuildings.BuildingCook;
 import com.minecolonies.coremod.colony.buildings.workerbuildings.BuildingDeliveryman;
 import com.minecolonies.coremod.colony.buildings.workerbuildings.BuildingTownHall;
@@ -55,9 +57,9 @@ public class EntityAIWorkDeliveryman extends AbstractEntityAIInteract<JobDeliver
     private static final int MIN_DISTANCE_TO_WAREHOUSE = 5;
 
     /**
-     * Walking speed double at this level.
+     * Walking speed bonus per level
      */
-    private static final double WALKING_SPEED_MULTIPLIER = 25;
+    private static final double BONUS_SPEED_PER_LEVEL = 0.003;
 
     /**
      * Delay in ticks between every inventory operation.
@@ -635,7 +637,9 @@ public class EntityAIWorkDeliveryman extends AbstractEntityAIInteract<JobDeliver
         {
             return START_WORKING;
         }
-        worker.getEntityAttribute(SharedMonsterAttributes.MOVEMENT_SPEED).setBaseValue(BASE_MOVEMENT_SPEED + BASE_MOVEMENT_SPEED * worker.getCitizenExperienceHandler().getLevel() / WALKING_SPEED_MULTIPLIER);
+        worker.getEntityAttribute(SharedMonsterAttributes.MOVEMENT_SPEED)
+          .setBaseValue(
+            BASE_MOVEMENT_SPEED + (worker.getCitizenExperienceHandler().getLevel() > 50 ? 50 : worker.getCitizenExperienceHandler().getLevel()) * BONUS_SPEED_PER_LEVEL);
 
         final AbstractBuildingWorker ownBuilding = getOwnBuilding();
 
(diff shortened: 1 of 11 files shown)


Raycoms:
Did you test that well?
Also in terms of getting people stuck on purpose to check if they recalculate the path accordingly as well.

someaddons:
ye tested it in singleplayer/smp and couldnt find any glitches/issues. For the ppl getting stuck: its the same as before, they just scan a smaller area(still pretty large) they're just acting as before

Raycoms:
Yeah in this case for getting stuck I meant where you "stopped the recalculation" while the path is in progress. Does this have any impact?

someaddons:
ah in that regard aside from performance it does affect the moveAway paths in the regard that they work better, moving away actually doesnt really work with constant refreshing, then the entitys do not really move.

Raycoms:
Ahh, sounds great.

Raycoms:
Ahh, sounds great.
Good finds, I never thought about checking these =)
:shipit: 

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/3026/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
```diff
@@ -635,7 +637,9 @@ private AIState checkWareHouse()
         {
             return START_WORKING;
         }
-        worker.getEntityAttribute(SharedMonsterAttributes.MOVEMENT_SPEED).setBaseValue(BASE_MOVEMENT_SPEED + BASE_MOVEMENT_SPEED * worker.getCitizenExperienceHandler().getLevel() / WALKING_SPEED_MULTIPLIER);
+        worker.getEntityAttribute(SharedMonsterAttributes.MOVEMENT_SPEED)
+          .setBaseValue(
+            BASE_MOVEMENT_SPEED + (worker.getCitizenExperienceHandler().getLevel() > 50 ? 50 : worker.getCitizenExperienceHandler().getLevel()) * BONUS_SPEED_PER_LEVEL);
```
marchermans, line 642: Nice this cap. But I would rather have this configurable please, in the sense that the config file specifies the max Level. Allows larger Servers to decrease this cap, to reduce load.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
```diff
@@ -165,6 +165,7 @@ private AIState follow()
             if (!isWithinPersecutionDistance(buildingGuards.getPlayerToFollow()))
             {
                 worker.getNavigator().clearPath();
+                worker.getMoveHelper().strafe(0, 0);
```
marchermans, line 168: Why is the strave inserted here?

src/main/java/com/minecolonies/coremod/entity/ai/mobs/barbarians/EntityAIAttackArcher.java
```diff
@@ -78,36 +82,53 @@ public void startExecuting()
...
+            // Stop walking when we're in attack range
+            if (entity.getDistance(target) < MAX_ATTACK_DISTANCE)
+            {
+                entity.getNavigator().clearPath();
+            }
+
+            if (lastAttack <= 0 && entity.canEntityBeSeen(target))
+            {
+
+                final EntityTippedArrow arrowEntity = new EntityTippedArrow(CompatibilityUtils.getWorld(entity), entity);
```
marchermans, line 114: Why not EntityArrow?

marchermans (changes requested):
Some small stuff.
Make the cap configurable.
Also some small Questions.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
ItsmeAga: ah good call thx!

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
ItsmeAga: because some of the guardAI's set strafing during the fight, so when the movement is reset I think it should be set to 0 aswell

src/main/java/com/minecolonies/coremod/entity/ai/mobs/barbarians/EntityAIAttackArcher.java
ItsmeAga: not sure, I just moved that code from beeing executed every tick to only execute when we're attack ready. Though I think one could use tipped arrows with some potion effects for difficulty scaling

marchermans: As far as I can tell this Code does not require EntityArrow. But @Raycoms might actually know more.

someaddons: hm we're using the same for the guard AI aswell

Raycoms: EntityArrow is abstract and can't be instantiated

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
Raycoms: And yes, resetting strafing sounds good

PR #3051, opened 2018-11-11
fix memory and CPU leak

Raycoms:
Fixes memory and CPU leaks at the herders:
- Don't request needed items every tick (set delay before every possible exit point)
- Don't fill up global lists, you have to either clear them or only build local lists..

Review please


marchermans (changes requested):
How does this prevent memory leaking and or cpu lag? Please give a better description so it can be prevented next time.

Also create a wiki article on the github wiki describing the changes. Lets start doing this for these kind of PRs.

marchermans (approved):

Issue #3035, opened 2018-11-04
Generic Raiders - Save environment where structure was placed


Raycoms:
At the moment barbarian raiders spawn in the middle of nowhere and pirates spawn on their ships.

In the future it would be nice if we could add support for barbarian camps, vikings, mammelukes etc for the different biomes there are.

For this reason we have to add a way to store the schematic of the place we paste this schematic to be able to roll the changes back after a bit.

Structures are placed with:

```
StructureWrapper.loadAndPlaceStructureWithRotation(world, Structures.SCHEMATICS_PREFIX + "/Ships/" + shipSize, targetSpawnPoint.down(3), 0, Mirror.NONE, false);
```

What we would need would be a method which is.

backUpWorld(world, Structures.SCHEMATICS_PREFIX + "/Ships/" + shipSize, targetSpawnPoint.down(3), 0, Mirror.NONE, false) 

Ahin the StructureWrapper which would do a small backup of the world matching the size of the structure before pasting the structure. This address would then be registered in the IRaidManager together with the existing structure data to roll the changes back afterwards.

PR #3080, opened 2018-11-18
Pirate ship water

someaddons:
- Closes: #3035 

- Pirate ship events now safe a backup schematic of the world where it spawns
- Pirate ships now only spawn in water pools large enough for its schematic
- Structures no longer spawn any Item entity's which have been scanned in


Raycoms:
I think it makes more sense to do this then: https://github.com/ldtteam/minecolonies/issues/3035

someaddons:
was thinking we could at least simply place water blocks when needed till its done properly

src/main/java/com/minecolonies/coremod/util/StructureWrapper.java
```diff
@@ -168,6 +180,57 @@ private void removeStructure(@NotNull final BlockPos pos)
...
+    {
+        setLocalPosition(pos);
+
+        // Find the waterlevel
+        int waterLevel = 0;
+        for (int j = structure.getHeight(); j > 0 && waterLevel == 0; j--)
+        {
+            for (int k = 0; k < structure.getLength() * 2; k++)
+            {
+                for (int i = 0; i < structure.getWidth() * 2; i++)
```
Raycoms: This looks amazingly expensive though, going by twice the length and twice the width. 
Especially because we can be sure that at structure.getLength() + 1 there should be some sort of water already.

Since the structures can be a few hundred blocks high it also makes more sense to use binary search.

someaddons: ye you're right its pretty inefficient, gonna optimize it. Doesnt get called often though

marchermans (changes requested):
I discussed this with ray. When this was implemented I told him to only merge the pirate mechanics if the next step would be a proper reversion by scanning the area first before placing the structure. Since that is not implemented here, and we are basically cheaping out on how to resolve the issue once and for all, I can not support this pr. Please change this appropriate to the relevant issue that was made.

someaddons:
ye talked with ray, working on the proper scanning now

marchermans:
Okey then it is good 😁

someaddons:
Updated to the better version, pls review :)

src/main/java/com/minecolonies/coremod/entity/ai/mobs/util/PirateEventUtils.java
```diff
@@ -20,6 +22,11 @@
  */
 public final class PirateEventUtils
 {
+    /**
+     * Folder name for the pirate ship schematics
+     */
+    private static String PIRATESHIP_FOLDER = "/Ships/";
```
marchermans: Please make this lowercase. All minecrafts directories are lowercase for better compatibility with Linux.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/entity/ai/mobs/util/PirateEventUtils.java
someaddons: its uppercase because the schematics folder containing the ship schematics is uppercase aswel, could make it lowercase too. Seems a lot of folders in the schematics are uppercase

marchermans: Yeah... that is annoying then.
Then leave it as is.

marchermans (approved):

PR #3090, opened 2018-11-22
weak reference for workerAI

someaddons:
Changing the reference to the workerAI in AbstractJobs.java to a weak reference so it gets removed when the entity which holds the strong reference in the tasklist does get removed.

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/jobs/AbstractJob.java
+++ b/src/main/java/com/minecolonies/coremod/colony/jobs/AbstractJob.java
@@ -18,6 +18,7 @@ import net.minecraft.util.SoundEvent;
 import net.minecraftforge.common.util.Constants;
 import org.jetbrains.annotations.NotNull;
 
+import java.lang.ref.WeakReference;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/Minecolonies/minecolonies/pull-request/3090/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=Minecolonies/minecolonies)

marchermans (approved):
Can we check and make sure that this is the only place we hold the reference.

someaddons:
Rechecked all and it really is only the enity's tasklist and this workerAI as reference to the AI

marchermans:
Okey. Then before we pull this do me a favor. Run a check on the entity as well. As well as on things like the entities inventory and such. Also make sure that when we do pickup and things we do not store that anywhere just to be sure. After that I think we can safely say that we did everything we can to prevent memory leaks

someaddons:
Checked and didnt find any references outside the EntityCitizen class and its AI's it stores. Except for one in AttackingPlayer.java but that should be gone once the player the guards are following isnt aviable anymore

marchermans:
You sure that it cleans itself up?

someaddons:
ye I can see it unloading when putting an output into Finalize and moving away to unload the chunk

marchermans:
Okey, then we are good to go.

Issue #500, opened 2017-01-22
[develop question]


forReason:
When I Build Beautiful things. is there a way, I can add the schematics? like More versions of huts, improved farmers barn etc?

Raycoms:
Yes you can. But you'd have it to the jar

2017-01-22 12:45 GMT-02:00 Julian B. <notifications@github.com>:

Raycoms:
you can create your own stylepack and upload it to the jar or send it to us so that we can add it to the mod =D

PR #3105, opened 2018-11-25
AI additions

someaddons:
- Using hashmap to lookup AITargets for the current states instead of looping the whole Arraylist
- Introducing 4 different priorities so that priority no longer only relies on registration order
- New Tickrate system which allows AITargets to specify a tickrate at which they get checked
- Adding Event AITargets as one time execution dynamic targets

For more details on the new stuff check these wiki entries: https://github.com/someaddons/minecoloniesDoc/wiki/workerAI-System
https://github.com/someaddons/minecoloniesDoc/wiki/AITargets

Review please


src/main/java/com/minecolonies/coremod/entity/ai/util/AITarget.java
```diff
@@ -24,78 +24,182 @@
...
+    /**
+     * The random offset for Ticks, so that AITargets get more distributed activations on server ticks
+     */
+    @NotNull
+    private final  int tickOffset;
+    /**
+     * The variant used upon creation of the AITarget to uniformly distribute the Tick offset
+     * Static variable counter that changes with each AITarget creation and affects the next one.
+     */
+    private static int tickOffsetVariant = 0;
```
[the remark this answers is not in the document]
someaddons: intended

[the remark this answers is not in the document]
someaddons: intended

marchermans (commented):
So I have been reading up on this. And i am not sure why we are limiting ourselfs to 4 AIStates.
Can i get a well enough explanation why that is just not an int with a priority? Or even a double, and use some kind of ordered map, since that would speed up iteration?

someaddons:
You're right a priority int would be possible. Inherently we have two different kinds of AITargets: Those which apply in a certain state and are Transitions in a Statemachine and those who apply globally and are called Event in a Statemachine. Now the Events should always be done before transitions, so they should be higher priority. Transitions do not need a priority, they should be unique in state and condition. 

Due to this I went with the mixed approach of giving the global Events three different states for priority, while also avoiding sorting overhead. Since there basically are only two different priorities needed I'm not sure if we really need more granular priorities than those 4.

Raycoms:
Looks good for me =)

marchermans:
I do not get the reason for the enum. Really. Why is it not just a integer?

someaddons:
eh I probably worded it badly, its main point is seperating AITargets into different types, which then also are executed in a set order. No real priorities between equal AITargets

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractAISkeleton.java
```diff
@@ -132,7 +162,20 @@ public final void resetTask()
...
-        targetList.stream().anyMatch(this::checkOnTarget);
+
+        tickCount++;
+        if (tickCount >= AITarget.MAX_AI_TICKRATE + AITarget.MAX_AI_TICKRATE_VARIANT)
+        {
+            tickCount = 1;
+        }
+
+        // Check targets in order by priority
+        if (!targetMap.get(AIState.AI_BLOCKING_PRIO).stream().anyMatch(this::checkOnTarget)
```
marvin-bitterlich: to increase readability, you could do a flatmap on the streams to concatenate them and then do anyMatch once instead of boolean voodo :)
https://www.techempower.com/blog/2016/10/19/efficient-multiple-stream-concatenation-in-java/

src/main/java/com/minecolonies/coremod/entity/ai/util/AITarget.java
```diff
@@ -24,78 +24,182 @@
...
+     * @param action    the action to apply
+     */
+    public AITarget(@NotNull final BooleanSupplier predicate, final boolean isOkayToEat, @NotNull final Supplier<AIState> action, @NotNull final int tickRate)
+    {
+        this(null, isOkayToEat, predicate, action, tickRate);
+    }
+
+    /**
+     * Construct a target.
+     * TODO: Remove once all Targets transitioned to tickRate
```
marvin-bitterlich: What is needed to transition? Can we do that already?

src/main/java/com/minecolonies/coremod/entity/ai/util/AITarget.java
```diff
@@ -24,78 +24,182 @@
...
-    public AITarget(@Nullable final AIState state, final boolean isOkayToEat, @NotNull final BooleanSupplier predicate, @NotNull final Supplier<AIState> action)
+    public AITarget(
+      @Nullable final AIState state,
+      final boolean isOkayToEat,
+      @NotNull final BooleanSupplier predicate,
+      @NotNull final Supplier<AIState> action,
+      @NotNull final int tickRate)
     {
-        this.state = state;
+        if (state == null)
```
marvin-bitterlich: why do we allow null?

src/main/java/com/minecolonies/coremod/entity/ai/util/AITarget.java
```diff
@@ -24,78 +24,182 @@
...
         this.okayToEat = isOkayToEat;
+
+        // Limit rates
+        this.tickRate = tickRate > MAX_AI_TICKRATE ? MAX_AI_TICKRATE : tickRate;
+        this.tickRate = this.tickRate < 1 ? 1 : this.tickRate;
+
+        // Calculate offSet % tickRate already to not have redundant calculations later
+        this.tickOffset = tickOffsetVariant % this.tickRate;
+        // Increase variant for next AITarget and reset variant at a certain point
+        tickOffsetVariant++;
```
marvin-bitterlich: So this is a global state that offsets actions statically in the order they are created? Is there a good reason for this design choice?

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractAISkeleton.java
```diff
@@ -27,21 +29,26 @@
...
-    protected final      EntityCitizen       worker;
-    protected final      World               world;
+    protected final      EntityCitizen                     worker;
+    protected final      World                             world;
     @NotNull
-    protected final      ChatSpamFilter      chatSpamFilter;
+    protected final      ChatSpamFilter                    chatSpamFilter;
     @NotNull
-    private final        ArrayList<AITarget> targetList;
+    private final        Map<AIState, ArrayList<AITarget>> targetMap;
```
marvin-bitterlich: I'm not yet convinced on any runtime benefits this will have with the amount of steps we currently have. Especially if we have to check each apply anyway. But I can understand that there is a reason to brewity of sorting them into different categories.

marvin-bitterlich (changes requested):
I like the tickrate addition, but am not yet sold on hashmap or global offsets. If you have something that is time queued, then it might be better to have it have a delay counter or e priority queue with some form of time.

The global offsets are dangerous if you wait 500 ticks and then have that one tick be interrupted by an event with more priority. This allows important events to starve infrequent events from how I can see it.

One thing I would like to see is a guarantee that if you specify something to be run every x ticks, that it runs the moment these ticks are done and the check tries to start the event.

I could see a priority queue. I could also see a timestamp like approach where we save the tick we currently have and check if the new tick is at least x more than last executed.

Also regarding the hashmap I noticed that you only ever iterate over them in order one after the other. What we could do is a bake step, where we compile it into a list again. This saves the runtime overhead of multiple maps and concatenating them. And we would get the sorting benefits of maps. Or otherwise we could make a custom order on a list. With this amount of elements, iterating over them as a list is blazingly fast. So the more I think of it the less sold I am on the hashmap.

@Raycoms asked for my review as I designed the initial system :) So far I see no good reasons for some of the changes in this pr. I see their intentions but I don't see them solving the problems they are set out to solve

src/main/java/com/minecolonies/coremod/entity/ai/util/AITarget.java
someaddons: After this change all current AITargets are going to be revisited and reevaluated for the tickRate they need, to not do it all at once those temporary constructors are there which provide a default tickRate for current AItargets. Once all current targets are changed those will be removed and force new AITargets to actively choose a tickRate.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractAISkeleton.java
someaddons: The runtime benefit is that instead of doing this check: "if (state != target.getState() && target.getState() != null)" for all targets in the list we just check those which have the right state

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractAISkeleton.java
someaddons: I think its actually nice to show it seperated, to make execution order clear, also isnt doing a concat each time just an unneeded overhead?

src/main/java/com/minecolonies/coremod/entity/ai/util/AITarget.java
someaddons: we've been using null for all current AITargets when they weren't limited to a certain state, aka global checks or events. Till stuff is ported over theres an default state for null.

src/main/java/com/minecolonies/coremod/entity/ai/util/AITarget.java
someaddons: the Offsets intent is to not have all AITargets with the same tickrate execute during the same tick and instead shift them up at an equal distribution. E.g. first AITarget with tickrate 10 has an offset of 0 and will be executed on ticks 0,10,20,30 and the second AITarget with tickrate 10 has an offset of 1 and will be executed on ticks 1,11,21,31

someaddons:
The starving of infrequent AITargets is a good point, if sth is set to every 500. ticks it should at least try to execute on ticknumber 500. This could be solved by having a counter for each AITarget type, so every target would see the tick #500

marchermans:
So i read this one again. And I am really not happy with the way we did this.
I think we should split this up into two / three components:

A timer component for the AI. It makes sure that the engine runs some AI code every x ticks, or after y ticks have passed.
A event component triggers for events that happen in world (inventory changes, entering leaving biomes, what ever we can come up with, properly implemented). Which can forcefully move component 3 (the statemachine) to a given state, or schedule something in the timer component.
A state machine, properly setup. Not based on AITargets or States but on, maybe properly serializable, components which we can combine together, and which get properly baked during runtime.

I have been debating and I am not sure that we actually need something priority based, we just need to seperate some logic into its own components. That way we prevent starvation, or other problems, and we can trigger some stuff externally using events.

_TLDR: I think this design is nice, but not what we need right now. It does not provide any real benefits over what we have right now (the code is just as complicated, and my preliminary performance tests between the current alpha and this branch show no real improvement, so i am not sure what this is supposed to do)._

If @someaddons would be willing to discuss this we should set an appointment and properly discuss this over Voice and figure out a way to do it properly, cause i agree that we need some rework in this department.

Orion

src/main/java/com/minecolonies/coremod/entity/ai/statemachine/basestatemachine/BasicStateMachine.java
```diff
@@ -0,0 +1,192 @@
...
+import com.minecolonies.coremod.entity.ai.statemachine.transitions.IStateMachineEvent;
+import com.minecolonies.coremod.entity.ai.statemachine.transitions.IStateMachineOneTimeEvent;
+import com.minecolonies.coremod.entity.ai.statemachine.transitions.IStateMachineTransition;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
```
Raycoms, line 15: javadoc?

PR #3107, opened 2018-11-26
Feature/store levels

Raycoms:
- Store the level and experience per job
- Reduce pirate drops
- Fix block dupe bug with racks

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/CitizenData.java
+++ b/src/main/java/com/minecolonies/coremod/colony/CitizenData.java
@@ -972,7 +972,7 @@ public class CitizenData
         {
             return 0;
         }
-        return levelExperienceMap.computeIfAbsent(job.getName(), jobKey -> new Tuple<>(0, 0.0D)).getFirst();
+        return queryLevelExperienceMap().getFirst();
     }
 
     /**
@@ -986,7 +986,7 @@ public class CitizenData
         {
             return;
         }
-        final Tuple<Integer, Double> entry = levelExperienceMap.computeIfAbsent(job.getName(), jobKey -> new Tuple<>(0, 0.0D));
+        final Tuple<Integer, Double> entry = queryLevelExperienceMap();
         this.levelExperienceMap.put(job.getName(), new Tuple<>(lvl, entry.getSecond()));
     }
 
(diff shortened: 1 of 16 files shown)


src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -279,9 +295,9 @@ public void readFromNBT(@NotNull final NBTTagCompound compound)
             isAsleep = compound.getBoolean(TAG_ASLEEP);
         }
 
-        if (level > MAX_CITIZEN_LEVEL)
+        if (job != null && levels.getOrDefault(job.getName(), 0) > MAX_CITIZEN_LEVEL)
```
someaddons: shouldnt this check be in the for loop reading the value? To check the Max of every value, not just the current jobs

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -176,15 +172,15 @@
...
+    private Map<String, Integer> levels =  new HashMap<>();
 
     /**
-     * The total amount of experience the citizen has.
-     * This also includes the amount of experience within their Experience Bar.
+     * The total amount of experiences the citizen has depending on his job.
+     * This also includes the amount of experiences within their Experience Bar.
      */
-    private double experience;
+    private Map<String, Double> experiences =  new HashMap<>();
```
someaddons: may be easier to combine it in one HashMap<String, Tuple<Integer,Double>> ? And as far as I can see the hashmaps can be final

someaddons:
Looks fine to me apart from those small things

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
Raycoms: This is actually only a hotfix to reset workers which had a too high level before we added the level cap. That's something we can delete in the future again.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
Raycoms: yeah

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
someaddons: well if we were to ever change the constant again you'd need it :D always good to validate the data upon rereading^^

PR #3113, opened 2018-11-29
Outlining of building and stop harm on fail tp

### LINKED ISSUE #3081: [Enhancement] bounding boxes before structure creation
Could it be possible to have the bounding boxes of the structures at the different levels (mostly 5) shown when presenting them before clicking the check box , so I can tell where I should put my structures so they don't mess up my roads? (simple lines like the ones on some mods to show bounding boxes for chunks, could be a config option as well, or just a simple keymapping, or even an option on the GUI), I get to know with the tape after the confirmation, but most of the time they are off by 1 or 2 blocks, so I need to destroy the initial block and try again until it is just right, would save a lot of trouble.

tx
-Winsrp

Raycoms:
Closes #3081

- Fix fall damage on fail tp
- Add an outlining when rendering previews.
-

Review please


someaddons (dismissed):
Looks good, should be a decent addition

PR #3114, opened 2018-11-29
fix pathing up half blocks and let drop down up to 3 blocks

### LINKED ISSUE #3110: Pathfinding fails when colonist is standing on half block
### Minecolonies version
Version:  1.12.2 - 0.10.27 Release

### Expected behavior
Colonist should continue on to their destination.

### Actual behaviour
Colonist remains stuck where they are indefinitely.  Sometimes shoving them until they're standing on a full block fixes the problem.  Recalling them works.

My delivery man gets stuck every time they go to the level 2 Medieval Fisherman's Hut.  The delivery point is on top of some wooden slabs on the "dock" portion of the structure.  The fisherman gets stuck in the same location.

Raycoms:
Closes #3110

- Stops workers from pathing up blocks they can't path up
- Allows workers to drop up to 3 blocks down.
-

Review please


someaddons:
Looks fine to me!

marchermans (approved):

PR #3176, opened 2018-12-18
Eat when saturation at 0 either way!

Raycoms:
Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAIEatTask.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAIEatTask.java
@@ -132,7 +132,7 @@ public class EntityAIEatTask extends EntityAIBase
         }
 
         final CitizenData citizenData = citizen.getCitizenData();
-        if (citizenData == null || citizen.getCitizenData().getSaturation() >= CitizenConstants.HIGH_SATURATION || !citizen.isOkayToEat())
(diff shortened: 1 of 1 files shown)


someaddons:
I think this might be a bad idea, since the purpose of the eating flag is to protect the current AI task. Maybe just change some of the flags instead?

Raycoms:
The idea of the eating flag for me is "in this state it's not nice to eat" but if the saturation dropped to 0 he should eat either way imo else he will get stuck there.

someaddons:
hm technically the eating isnt vital since its just a stat, but interrupting an AI for it might be

marchermans (approved):

Raycoms:
I know what you're trying to get at.
But stopping to eat something is actually not much of an issue, previously he would even change his state just to eat something every time.
For me, the states marked as "not for eating" is more because "it wouldn't be very good if he would eat right now"
But I believe, also for reasons of immersion. Always when he reaches 0 he should go after food no matter what.

Raycoms:
A lot of people were reporting their citizens being on 0 hunger and not eating anything

someaddons:
ye we can do it this way, although its more of an force overrule than integration. Afaik the issue with starving citizens is that the Waiting for a request status doesnt allow them to eat

Raycoms:
The most common one was the issue in request state (since that one is the
most common) but we also already got complains with a lot of the other
states (basically ordered by how common the state is)
People are just bothered by a worker not eating anything when their
saturation hits 0.

>

Raycoms:
It gets worse when we add one day that they take damage if they are at 0. So we should make sure that after a minimum limit they always go to eat no matter what.

someaddons (approved):

PR #3193, opened 2018-12-27
AI statemachine

someaddons:
- Decoupling state machines from the AI, the barebones statemachine now contains only:
- - Its current state
- - Its transitions
- - Its events

- Interface types for transitions,events and states
- Extracted the okay-to-eat boolean from the transition to the state

Review please


marchermans:
I will review this tomorrow.

PR #3213, opened 2019-01-02
AI Statemachine

someaddons:
- Decoupling state machines from the AI, the barebones statemachine now contains only:
- - Its current state
- - Its transitions
- - Its events

- Interface types for transitions,events and states
- Extracted the okay-to-eat boolean from the transition to the state

Review please


marchermans (dismissed):
Accept for the conflicts i am okey with this.

That said this PR could be better. Certain abstractions could be improved, as well as the documentation.
It took me a couple of times to understand it all.

PR #3236, opened 2019-01-06
Feature/post box

Raycoms:
- A post box which allows the player to request items from the request-system to his hut
- Fixes for the barbarian difficulty
- Increment structurize version

Review please


src/main/java/com/minecolonies/coremod/client/gui/WindowPostBox.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Get the list of blocks which should be added.
+     *
+     * @param filterPredicate the predicate to filter all blocks for.
+     * @return an immutable list of blocks.
+     */
+    private Collection<? extends ItemStack> getBlockList(final Predicate<ItemStack> filterPredicate)
+    {
+        return ImmutableList.copyOf(StreamSupport.stream(Spliterators.spliteratorUnknownSize(Item.REGISTRY.iterator(), Spliterator.ORDERED), false).flatMap(item -> {
```
marchermans, line 121: This can be done much more performant.
Instead of iterating over all items in one go, one could possible just collect all items once and then filter them when needed.
This would improve performance in larger modpacks.

src/main/java/com/minecolonies/coremod/client/gui/WindowPostBox.java
```diff
@@ -0,0 +1,190 @@
...
+            return stacks.stream().filter(filterPredicate);
+        }).collect(Collectors.toList()));
+    }
+
+    /**
+     * Action when a button opening an inventory is clicked.
+     */
+    private void inventoryClicked()
+    {
+        MineColonies.getNetwork().sendToServer(new OpenInventoryMessage(buildingView.getID()));
```
marchermans, line 141: When I tested this the inventory said Chest.
I would prefer it when you also fixed that: Open building inventory then render the inventory string.
This can be done by returning a custom name.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/client/gui/WindowPostBox.java
Raycoms: the problem is that either we do this on open, or we do this on scroll for each scroll action (complex), but at least the moment the player wants to filter in the list it will have to get the whole list

marchermans: Okey.... What ever

marchermans (dismissed):

src/main/java/com/minecolonies/coremod/client/gui/WindowPostBox.java
Raycoms: Done

someaddons (dismissed):
Looks fine to me

PR #3252, opened 2019-01-11
Colony border

someaddons:
Fixes the colony border display


Raycoms:
Not needed anymore with structurize update. I bumped the version in my pr.

PR #3253, opened 2019-01-11
avoid NPEs with capabilities

Raycoms:
- Fixes NPEs reported by a user
-
-

Review please


marchermans (changes requested):
Nope you are doing it wrong if you get a null here.
Do a hasCapability check before you do a getCapability.

Raycoms:
Makes our code much more complex checking everytime "hasCapability" on things which definitely should have a capability but for a stupid bug don't (Our racks, chests, etc sometimes lose it for some forge reason)

marchermans (approved):

PR #3266, opened 2019-01-13
Builder considers Grass as Dirt

### LINKED ISSUE #3265: fix Builder should consider grass as dirt
### Expected behavior

Builder threats grass as dirt

### Actual behaviour

Builder picks up one dirt block at a time and places it.
 

Build a building in which the builder has to place grass.

schafhierte204:
Closes #3265 

-Builder treats grass like dirt

Review please


Raycoms:
Not sure what you mean the problem is, that the list of resources on page 2 shows grass but he actually only needs dirt?

schafhierte204:
The Problem is that he runs back to his hut and picks up a dirt block every time he needs to place a grass block

Raycoms:
Ahhh, now I understand what you mean. Did you test this, did this patch solve this problem for you?

schafhierte204:
I have the game running at the moment and the problem is gone.

schafhierte204:
The only minor side affect is that the grass blocks now show up as dirt in the resource list

Raycoms:
Can you fix the parenthesis to go down? like:
{

}

PR #3314, opened 2019-01-31
Increment structurize, improve citizen home assignment

Raycoms:
- Improve the citizen assignment mode
- Increment structurize version
- Fix particle of baker block

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/util/constant/WindowConstants.java
+++ b/src/api/java/com/minecolonies/api/util/constant/WindowConstants.java
@@ -1091,6 +1091,41 @@ public final class WindowConstants
     public static final String BUTTON_REQUEST = "request";
 
     /**
+     * Id of the done button in the GUI.
+     */
+    public static final String BUTTON_DONE = "done";
+
+    /**
+     * Id of the citizen name in the GUI.
+     */
+    public static final String CITIZEN_LABEL = "citizen";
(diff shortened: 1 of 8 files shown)


src/api/java/com/minecolonies/api/util/constant/WindowConstants.java
```diff
@@ -1090,6 +1090,41 @@
...
+
+    /**
+     * Id of the citizen list in the GUI.
+     */
+    public static final String CITIZEN_LIST = "unassigned";
+
+    /**
+     * Assign button of the gui.
+     */
+    public static final String CITIZEN_DONE = "done";
```
Nightenom, line 1111: Why is the string "done" twice there

Nightenom (dismissed):

src/api/java/com/minecolonies/api/util/constant/WindowConstants.java
Raycoms: because it's two different buttons, so the id might change for one but not for both at th esame time

Nightenom: Ok, I was looking how many dupes are in window constants - there are fewer than I expected :D

marchermans (approved):

PR #3350, opened 2019-02-10
Fix desync

Raycoms:
- Fixes client/server desync (also improves server side performance, by writing colony only once to NBT)
- Dman ninja pickup fix (dman has to walk until the building to pickup now)
- Dman gathering crafter output prematurely (only on specific pickup)
- Fix stonemason recipe (allow stonebricks, stairs, etc, don't allow redstone, etc)

Review please


src/main/java/com/minecolonies/coremod/colony/requestsystem/management/manager/StandardRequestManager.java
```diff
@@ -175,12 +175,13 @@ private void markDirty()
     @Override
     public boolean isDirty()
     {
-        if (dirty)
```
marchermans, line 178: This is a half assed hotfix(?!??) for something without description. And I am 100% sure this breaks a lot of things. Please explain why this change is needed and how it prevents your not described actual concrete problem you are trying to fix!

marchermans (changes requested):
More explanation needed

Raycoms:
Atm we reset dirty when checking if dirty, which can break if we check
dirty two times (which is the actual problem with the client server sync)
Besides that we serialize the colony one time each person, which is also a
lot of lost performance

marchermans (approved):

PR #3407, opened 2019-02-19
Barracks housing

someaddons:
- Barracks now only provide housing for the citizens actually assigned there.
Review please


Raycoms:
Fixes our barracks problems quickly, shipit

PR #3458, opened 2019-03-03
Allow crusher to crush any block

Raycoms:
- Allow crusher to crush any blocks into each other.
-
-

Review please


marchermans (dismissed):
Approved but does this support metadata on stacks as well?

someaddons (dismissed):

someaddons (approved):

PR #3485, opened 2019-03-10
Fix handling of bigger quantities for one block placement

### LINKED ISSUE #3466: Chisel and Bits Duplication issue
### Minecolonies version
Version: 1.12.2-0.10.202-RELEASE

### Expected behavior
When using the builder to build schematics containing bitted? blocks (blocks that have been built with chisel and bits), the builder requests the amount of bits correctly and uses the bits to construct the blocks (ideally from a bits bag)

### Actual behaviour
The builder will place the blocks correctly (same pattern/ shape/ size/ material and position) BUT will do so out of thin air using NO bits. (even though the builder hut requests bits)

This can be easily abused to quickly duplicate blocks (most notably hard to obtain blocks) as the blocks can be turned back into bits by the player and use to reconstruct a full block.

1)Join a server (i have only confirmed this on a server, but haven't tested on single player)
2)Create something out of chisel and bits in survival.
3)Use the scan tool to save the schematic.
4)Use the build tool to request the schematic to be built
5)The builder will start building without anything.
6)Profit from unlimited diamonds and infinity blocks

Raycoms:
Closes #3466

- Fixes some code smells and small bugs
- Request and pickup the full quantity if necessary.

@OrionDevelopment:

I added that the contained stacks in the inventory get definitely added, I know this is not a good solution and I want your input whether we should delete the "missing" and adjust it to the deliveries.

In a bunch of cases, we gotta pick up more than 1 itemStack of an item to fulfill a request. In these cases, up to now, the builder either doesn't pick it up and dupes things.

(Chisel and bits or spawning of tileEntities with substacks > stacksize)

Review please


src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -555,6 +555,8 @@ private IAIState lookForRequests()
                     final List<ItemStack> niceToHave = itemsNiceToHave();
                     final List<ItemStack> contained = InventoryUtils.getContainedFromItemHandler(firstDeliverableRequest.getDeliveries(), worker.getItemHandlerCitizen());
                     final List<ItemStack> missing = InventoryUtils.getMissingFromItemHandler(firstDeliverableRequest.getDeliveries(), worker.getItemHandlerCitizen());
+                    missing.addAll(firstDeliverableRequest.getDeliveries());
```
someaddons: hm dont really get the purpose of this, isnt the list of missing items from the requests added in the line above this?

Raycoms: oh yeah, that's the one I only added for debug purposes. I'd like to remove the "missing" items search and actually transfer all items.
(This is the one I want to discuss with orion

marchermans: Well it depends a bit on how you see requests to be honest. In my opinion this does not make any sense. If I remember correctly that method also returns partial stack so even if you request 64 logs while having 32 that would return the required 32 as missing.

I see requests as follows: I am in need of x of item y. Bring me please. If I already have z of y in my inventory I would only need x-z to complete that requests.

Which is what this defies. 

How ever you could also see requests like this: I am missing x of item y. Bring me please. Then this behavior would be correct since you do not care about what he has.

That said. Even our old system worked like the first option, and so does the entire rs. It is why we check buildings and everything first. Before actually performing the request. If we now say we do not want that behavior anymore we would need to rewrite a hole lot of systems, subsystems and workers to make option 2 even functional.

So no I do not think that this makes sense.

Raycoms: The problem is that we can request maximum 1 stack. But if we need 128 or 512 of x and have 64 in the inventory he will never pick more up.

Raycoms: So either we extend the system to requesting bigger quantities, or we let him pick up the exact quantity he requested, since on request he should've checked already how much he needs compared to how much he has (and not on pickup)

marchermans: On the other hand if you do not care about how much he requests VS how much he already has. This would force him him to pick up every damn stack that was delivered, and the line above it would be become redundant.

marchermans: On the other hand if you do not care about how much he requests VS how much he already has. This would force him him to pick up every damn stack that was delivered, and the line above it would be become redundant.

Raycoms: There are multiple ways to fix this 
- Allow requests of more than stacksize but only deliver max stacksize at once (causing a new request on delivery)
- Allow and support requests of more than stacksize
=> And/Or as a workaround until we do that, consider that a worker knows how much he requested on purpose and let him pick up the entire delivered stack (Since I coded most of the AIs I am pretty sure that this won't break anything).

marchermans: The rs has no internal limit to how big a stack that it requests can be. Might be that the requestable however only accepts itemstacks do then you would need to expose the stack limit as well

src/main/java/com/minecolonies/coremod/tileentities/TileEntityColonyBuilding.java
```diff
@@ -399,7 +399,7 @@ public void setStyle(final String style)
     @Override
     public boolean hasCapability(@NotNull final Capability<?> capability, final EnumFacing facing)
     {
-        if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY)
+        if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY && facing == null)
```
marchermans: Why null?
This would result in interactions with other mods probably breaking.

I left the null out so things like hoppers and Refined Storage etc could connect with our buildings, including the warehouse for inventory management, i am pretty sure this would upset some people.

src/main/java/com/minecolonies/coremod/tileentities/TileEntityColonyBuilding.java
```diff
@@ -409,35 +409,40 @@ public boolean hasCapability(@NotNull final Capability<?> capability, final Enum
     @Override
     public <T> T getCapability(@NotNull final Capability<T> capability, final EnumFacing facing)
     {
-        if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY && getBuilding() != null)
+        if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY && facing == null)
```
marchermans: See above.

src/main/java/com/minecolonies/coremod/tileentities/TileEntityColonyBuilding.java
```diff
@@ -409,35 +409,40 @@ public boolean hasCapability(@NotNull final Capability<?> capability, final Enum
...
+                                                                                  .map(handler -> (IItemHandlerModifiable) handler)
+                                                                                  .distinct()
+                                                                                  .filter(handler -> handler instanceof IItemHandlerModifiable
+                                                                                                       && handler.getSlots() >= MIN_SLOTS_FOR_RECOGNITION)
+                                                                                  .toArray(IItemHandlerModifiable[]::new));
+            }
+            return super.getCapability(capability, facing);
         }
-        return super.getCapability(capability, facing);
+        return null;
```
marchermans: Never return null on the capability. Perform the building != null check in the has capability method and remove the check here.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/tileentities/TileEntityColonyBuilding.java
Raycoms: I did that to optimize for us finding the capabilities (We get like 10 handlers for each provider).
I'll undo that then and try to optimize a bit farther back, I thought we don't need this.

marchermans: This optimization is useless…. We Group them in a set later on anyway.

Raycoms: I fixed the optimization now, so we can recognize in the grouping process if two are equal.

src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -521,11 +521,18 @@ public static int getAmountOfStacksInItemHandler(@NotNull final IItemHandler ite
     @NotNull
     public static Set<IItemHandler> getItemHandlersFromProvider(@NotNull final ICapabilityProvider provider)
     {
-        final Set<IItemHandler> handlerList = Arrays.stream(EnumFacing.VALUES)
-                                                .filter(facing -> provider.hasCapability(ITEM_HANDLER_CAPABILITY, facing))
-                                                .map(facing -> provider.getCapability(ITEM_HANDLER_CAPABILITY, facing))
-                                                .filter(Objects::nonNull)
-                                                .collect(Collectors.toSet());
+        final Set<IItemHandler> handlerList = new HashSet<>();
```
marchermans: Change is irrelevant this is what a set already does. Please revert.

src/main/java/com/minecolonies/coremod/tileentities/TileEntityColonyBuilding.java
```diff
@@ -336,6 +341,12 @@ public void update()
                 colonyId = tempColony.getID();
             }
         }
+
+        /*
+         * We want a new inventory every tick.
+         * The accessed inventory in the same tick must be the same.
+         */
+        combinedInv = null;
```
marchermans, line 349: Why this? We can just keep the original IITemHandler around....

src/main/java/com/minecolonies/coremod/tileentities/TileEntityColonyBuilding.java
```diff
@@ -423,20 +434,26 @@ public boolean hasCapability(@NotNull final Capability<?> capability, final Enum
...
             if (cap instanceof IItemHandler)
             {
                 handlers.add((IItemHandler) cap);
             }
 
-            return (T) new CombinedItemHandler(building.getSchematicName(), handlers.stream()
-                                                                     .map(handler -> (IItemHandlerModifiable) handler)
-                                                                              .distinct()
-                                                                              .filter(handler -> handler instanceof IItemHandlerModifiable && handler.getSlots() >= MIN_SLOTS_FOR_RECOGNITION)
-                                                                              .toArray(IItemHandlerModifiable[]::new));
```
marchermans, line 439: Check if you already have a cap before calculating. This does not save on performance or anything else.... Just makes the code harder to understand..... Also again why recalculate the cap every tick?

marchermans (dismissed):
Better....

src/main/java/com/minecolonies/coremod/tileentities/TileEntityColonyBuilding.java
Raycoms: (This is also the answer to the question below):
In many queries we query in the same tick the capability of this container for all enum facing + null. Since it creates a new combined inventory each time, the set will add 7 inventories in it. Which costs performance and makes the results wrong.
I, therefore, added that it caches the combined inventory.
But, since the inventory of the building can change, when new containers are added or removed. I update this here regularly so its only the same during the same tick.

PR #3491, opened 2019-03-11
Fix and optimize conversion of chiseled blocks into bit stacks, and depend only on C&B's API.

### LINKED ISSUE #3466: Chisel and Bits Duplication issue

Phylogeny:
Possibly fixes #3466

1. Depend only on C&B's API, rather than referring to core classes.
2. Fix bug where the size of the stack generated for given bit in a chiseled block is equal to its x position in the blockspace, rather than 1.
3. Optimize the conversion of chiseled blocks into bit stacks by replacing the use of an `IBitVisitor` with the use of a `StateCount` list.

@Raycoms When I looked over the _[block -> bit stack]_ code and said it was fine yesterday, I overlooked  point 2 in the list above, but later noticed it.

Also, I've never used this mod before, and have not tested to verify that this change definitively fixes #3466, but it's quite likely since point 2 is certainly a bit dupe bug.

Review please

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,7 +90,7 @@ dependencies {
     contained "com.ldtteam:structurize:${config.minecraft_version}-${config.structurize_version}"
     deobfCompile "com.ldtteam:structurize:${config.minecraft_version}-${config.structurize_version}"
 
-    compile "mod.chiselsandbits:chiselsandbits:${config.c_b_version}"
+    compile "mod.chiselsandbits:chiselsandbits:${config.c_b_version}:api"
     deobfCompile "slimeknights.mantle:Mantle:1.12-${config.mantle_version}"
     deobfCompile "slimeknights:TConstruct:1.12.2-${config.tinker_version}"
     deobfCompile "com.ferreusveritas.dynamictrees:DynamicTrees:1.12.2-${config.dynamic_tree_version}"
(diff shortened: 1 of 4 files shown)


Phylogeny:
_**Fun Fact**_
The maxim factor of bit duplication (assuming a full block of bits) is 7.5 :smile:

Raycoms:
Thanks a lot for this =) Yeah I did something hackier to fix the number of bits we need for something. This will make it much easier.

Raycoms:
People are getting crashes with:
26.java.lang.NoClassDefFoundError: mod/chiselsandbits/api/APIExceptions$InvalidBitItem

PR #3511, opened 2019-03-24
Deco block

### LINKED ISSUE #332: Create wall blocks [$20]
A block used to manage the walls.
1 wall block can be used to manage a infinite (finite only to colony radius) wall segment.
Wall block always has the walls basic texture (wood or stone depending on the upgrade)

Options are: Upgrade, repair, add segment (left, right)

<bountysource-plugin>

---
Did you help close this issue? Go claim the **[$20 bounty](https://www.bountysource.com/issues/39921329-create-wall-blocks?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github)** on [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F43175463&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #3507: crash while placing a townhall inside of the padding buffer
### Minecolonies version

Version: 1.12.2-0.10.234-ALPHA

### Expected behavior

place townhall outside in the wilderness outside, and if there is an issue it tells me without crashing

- 

### Actual behaviour

I have implemented 20 chunks padding between colonies, I think that's part of the problem and the minecolonies is tripping out and doesn't know how to tell the player that they are too close to the other colony. The whereami command says they are 355 blocks away from nearest colony, which is just shy over 20 chunks.
- 

They crafted a supply camp and placed it down and then tried to place the townhall block. Before they rage quitted they reported that it would say something about being inside of another colony, which makes me think it has something to do with range and minecolonies code not handling that properly.

really interesting you don't ask for a crash report:

### Crash Report
https://paste.lucko.me/twE2wiot6f

### Minecolonies Config
http://paste.ubuntu.com/p/8dRpNGjQzr/

### LINKED ISSUE #3508: Citizen3 House causing builder to go into activity "SPAWN_STEP"
Newest version of CaveColonies

### Expected behavior
Builder actually building the rest of Citizen3

Builder should completely build citizen 3.

### Actual behaviour
Builder never completes Citizen3

Builder builds Citizen3 almost fully until the armor stand is next in queue.

1. Get a citizen2 and a builder3
2. Upgrade the citizen2 to a citizen3, and wait for the armor stand
3. do /mc citizens info colony: 1 citizen: (Citizen Name) and it should say that the current activity is SPAWN_STEP
4. ... Profit

Raycoms:
Closes #332 
Closes #3508
Closes #3507

- Adds a decoration block which allows to connect a decoration with a block to upgrade/repair it.
- Fixes rotation and mirror of tile entities (includes it)
- Fixes issues with entity placement
- Fixes issues where the blockhut would not rotate accordingly on placement

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/registry/BuildingRegistry.java
```diff
@@ -239,6 +240,7 @@ public static AbstractBuilding create(final Colony colony, @NotNull final TileEn
 
         if (building != null && parent.getWorld() != null && !(building instanceof PostBox))
         {
+            building.setRotation(BlockUtils.getRotationFromFacing(parent.getWorld().getBlockState(parent.getPosition()).getValue(AbstractBlockHut.FACING)));
```
Raycoms, line 243: This here is the default rotation, for when placed by hand, the real rotation is set later then, if existent

src/main/java/com/minecolonies/coremod/client/gui/WindowDecorationController.java
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * The world the player of the GUI is in.
+     */
+    private final World world = Minecraft.getMinecraft().world;
+
+    /**
+     * If the player opening the GUI is an admin.
+     */
+    private boolean admin = Minecraft.getMinecraft().player.isCreative();
```
Nightenom: maybe a better variable name?
since admin does not equal to creative imho

Nightenom (commented):
Looks good for me, I just have some questions and notes:
1) why does the window have no view?
2) maybe some background image for the gui
3) can you do schematics in standalone prs (as you were doing before)?

src/main/java/com/minecolonies/coremod/client/gui/WindowDecorationController.java
Raycoms: true

Raycoms:
why does the window have no view?
- it doesn't need, we can sync the tileEntity to client/server side easily.
maybe some background image for the gui - I think it fits well considering it's the same style as the other structurize guis
can you do schematics in standalone prs (as you were doing before)? - that wasn't on purpose

marchermans (dismissed):

PR #3607, opened 2019-05-12
Building inventories

someaddons:
WIP - any feedback appreciated

- Adding building index over all inventories for quick access
- Changing building tileentities from chest to rack type


Nightenom:
Just small note, move non-specific utils into the global utils folder (e.g. HashSetIndex)

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/HashSetIndex.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Predicate;
+
+/**
+ * Generic index class which holds an index of type Key and a Set of Values.
+ */
+public class HashSetIndex<Key, Value>
```
marchermans, line 15: This is complete overkill and not needed there are several classes that can do this job in the JSL or even in Guava.
As such this class can be removed and its logic replaced by one on of them.

A good replacement would be a MultiMap.

marchermans: Also this follows no JSL interface for generic collections, so that is needed as well!

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
```diff
@@ -0,0 +1,333 @@
...
+     * Index for stored Items and their slot.
+     */
+    @NotNull
+    private ItemStorageIndex<Integer> index;
+
+    /**
+     * A meta index which is to be updated. Does not have to be set.
+     */
+    @Nullable
+    private ItemStorageIndex<Value> meta;
```
marchermans, line 33: Not needed!
The implementation should fallback onto itemstacks with a custom hashing mechanism this is not fast enough and does not cover all items, since you seem to disregard NBT, which will break in 1.13 then anyway.

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
```diff
@@ -0,0 +1,333 @@
...
+        if (!isValidSlot(slot) || ItemStackUtils.isEmpty(stack))
+        {
+            return stack;
+        }
+
+        int insert = getStackLimit(slot, stack);
+
+        if (getStackInSlot(slot) != ItemStack.EMPTY)
+        {
+            if (!getStackInSlot(slot).equals(stack))
```
marchermans, line 162: You need to check for mergability not for equality!
Merge if possible or return original stack.

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
```diff
@@ -0,0 +1,333 @@
...
+        if (simulate)
+        {
+            ItemStack simulateStack = getStackInSlot(slot).copy();
+            simulateStack.setCount(toExtract);
+            return simulateStack;
+        }
+
+        ItemStack returnStack;
+        // Returning whole stack
+        if (toExtract >= getStackInSlot(slot).getCount())
```
marchermans, line 230: This should be an equal.....

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
```diff
@@ -0,0 +1,333 @@
...
+            if (meta != null)
+            {
+                meta.removeFromIndex(new ItemStorage(returnStack), metaIndexValue);
+            }
+            this.stacks.set(slot, ItemStack.EMPTY);
+            usedSlots--;
+        }
+        else
+        {
+            getStackInSlot(slot).setCount(getStackInSlot(slot).getCount() - toExtract);
```
marchermans, line 244: A grow with -toExtract is much better.

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
marchermans: This whole if branch can be done in about three lines, with a copy of the itemstack, a shrink of the original one and a set size on the copy stack who is then returned........

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.Set;
+import java.util.function.Predicate;
+
+/**
+ * ItemStackHandler which also keeps an index of its contents.
+ * It also can push any changes to a meta(upper) index, to combine multiple handlers.
+ *
+ * @param <Value> The type of the meta index, not required.
+ */
+public class ItemStackHandlerWithIndex<Value> extends ItemStackHandler
```
marchermans, line 21: Any public member should be pulled up into an interface that extends ItemStackHandler.

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
```diff
@@ -0,0 +1,333 @@
...
+            {
+                meta.addToIndex(store, metaIndexValue);
+            }
+        }
+    }
+
+    /**
+     * Clears the meta index reference and its content.
+     */
+    public void clearMetaIndex()
```
marchermans, line 78: Why again with the damn meta....?

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
marchermans: There are too many issues with this class to count.......
The implementation seems to depend on meta data leaving NBT out of side entirely, the indexing is weirdly implemented, and as far as I can tell it is not compatible with any existing IItemHandler implementation, which is what is actually needed.

In my opinion it would be much better if we created a global index off all items in a colony in every IItemHandler.
Update one inventory per tick (configurable) and use that to perform lookups....

That allows us to incorporate inventories from other mods as well, without having to screw around.

And it is the same logic that mods like AE or RefStorage apply.

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/WarehouseRequestResolver.java
```diff
@@ -56,16 +56,12 @@ public boolean canResolve(@NotNull final IRequestManager manager, final IRequest
...
 
-            try
+            if (!wareHouses.isEmpty())
             {
-                return wareHouses.stream().anyMatch(wareHouse -> wareHouse.hasMatchingItemStackInWarehouse(itemStack -> requestToCheck.getRequest().matches(itemStack), requestToCheck.getRequest().getCount()));
-            }
-            catch (Exception e)
-            {
-                Log.getLogger().error(e);
+                return true;
```
marchermans, line 64: This change breaks the RS Logic, please revert. The canResolve phase is exactly there to check if a Warehouse has a potential stack, returning true here breaks the entire logic and would result in the RS never actually checking any provider that has a lower priority, even if no warehouse has the requested item.

BLOCKING CHANGE!

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemSearchResult.java
```diff
@@ -0,0 +1,67 @@
...
+import com.minecolonies.api.crafting.ItemStorage;
+import com.minecolonies.coremod.tileentities.TileEntityRack;
+import net.minecraft.item.ItemStack;
+
+import java.util.Set;
+
+/**
+ * Wraps all needed information returned by an item search in a class.
+ */
+public class ItemSearchResult
```
marchermans, line 12: Again this is completely locked to Racks and not what is actually wanted or needed by our players.
ItemSearch should happen colony wide if possible not limited to racks or a specific inventory type.

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStorageIndex.java
```diff
@@ -0,0 +1,117 @@
...
+import net.minecraft.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+import java.util.function.Predicate;
+
+/**
+ * Index which uses Items(ItemStorage) as key.
+ */
+public class ItemStorageIndex<Value> extends HashSetIndex<ItemStorage, Value>
```
marchermans, line 15: If this is your index for ItemStacks the index works backwards and is useless in its current implementation.
What you want is a mapping from ItemStorage (preferably directly ItemStack, since ItemStorage is not strict enough) to a Map of <somekind of lookup logic for IItemHandlers> and a list of slot indexes.

Cause your current logic is the same as a Map<ItemStorage, Integer>, which i can build easily enough from an Array or a List of ItemStorages. And as such provides no speed benefits.

marchermans (changes requested):
So lets start with the good news..... I like the Idea of indexing inventories..... It is needed and should help us.....

Now, that is it for the good news i am affraid......for several reasons......
If this is your index for ItemStacks the index works backwards and is useless in its current implementation.
What you want is a mapping from ItemStorage (preferably directly ItemStack, since ItemStorage is not strict enough) to a Map of <somekind of lookup logic for IItemHandlers> and a list of slot indexes.

Cause your current logic is the same as a Map<ItemStorage, Integer>, which i can build easily enough from an Array or a List of ItemStorages. And as such provides no speed benefits, except for maybe some corner case left and right.

Also you are touching a lot of stuff here..... You broke the RS resolving for warehouses so that needs at least some fixing, and your indexing only works for racks.....
This has to change if we want to support more inventory types in the future, which is a question we get asked regularly.

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStorageIndex.java
someaddons: atm it is hierarchical and the index is split into two parts, one in the Building which maps ItemStorage to a set of racks, and an index in the racks which maps the itemstorage to the slots it is contained in

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStackHandlerWithIndex.java
someaddons: meta might've been bad wording, but it isnt actually related to itemstacks/NBT but instead is a reference to a global index which is updated too when this index gets updated. So you get local index for each rack and a global for each building which are kept in sync

marchermans: The first of all your naming is flaw as hell.
Second of all why the two tier system?

src/main/java/com/minecolonies/coremod/colony/buildings/inventory/ItemStorageIndex.java
marchermans: I would merge this all together.

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/WarehouseRequestResolver.java
someaddons: hm from what I can see this is the only place where it is used: https://github.com/ldtteam/minecolonies/blob/e9413d09d2434eeaddbe23b52989e845102fb674/src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/RequestHandler.java#L150

So it should be fine to fail on the attempt instead of the canresolve check right?

marchermans: It needs to crash on the attemptResolve() call if it returns false on the canResolve() call. It **NEEDS** to succeed if canResolve() returned true.

Indeed in the current implemenation if canResolve() is false then attemptResolve is never called.

PR #3646, opened 2019-06-01
Feature: RS Reassign on rack update

Asherslab:
Allows for the RS System to reassign a request when the needed item is added to a rack inside a warehouse.

Requires warehouses to be repaired to work, but will not be detrimental if they do not get repaired.

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
+++ b/src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
@@ -14,6 +14,7 @@ import com.minecolonies.coremod.colony.requestsystem.management.manager.wrapped.
 import org.jetbrains.annotations.Nullable;
 
 import java.util.*;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static com.minecolonies.api.util.constant.Suppression.RAWTYPES;
@@ -346,5 +347,16 @@ public final class ResolverHandler
     public static IRequestResolver<? extends IRequestable> getResolverForRequest(final IStandardRequestManager manager, final IRequest<?> request)
     {
         return getResolverForRequest(manager, (IToken<?>) request.getToken());
+    }
+
+    /**
+     * Method used to reassign requests based on a predicate
+     *
+     * @param manager The manager to update reassign requests on
+     * @param shouldTriggerReassign the predicate to determine whether a request should be reassigned
+     */
+    public static void OnColonyUpdate(final IStandardRequestManager manager, final Predicate<IRequest> shouldTriggerReassign)
+    {
+        manager.getRequestResolverIdentitiesDataStore().getIdentities().values().forEach(resolver -> resolver.onColonyUpdate(manager, shouldTriggerReassign));
     }
 }
--- a/src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
+++ b/src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
@@ -355,7 +355,7 @@ public final class ResolverHandler
      * @param manager The manager to update reassign requests on
      * @param shouldTriggerReassign the predicate to determine whether a request should be reassigned
      */
-    public static void OnColonyUpdate(final IStandardRequestManager manager, final Predicate<IRequest> shouldTriggerReassign)
+    public static void onColonyUpdate(final IStandardRequestManager manager, final Predicate<IRequest> shouldTriggerReassign)
     {
         manager.getRequestResolverIdentitiesDataStore().getIdentities().values().forEach(resolver -> resolver.onColonyUpdate(manager, shouldTriggerReassign));
     }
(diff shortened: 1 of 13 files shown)


src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
```diff
@@ -347,4 +348,15 @@ public static void removeRequestFromResolver(final IStandardRequestManager manag
...
         return getResolverForRequest(manager, (IToken<?>) request.getToken());
     }
+
+    /**
+     * Method used to reassign requests based on a predicate
+     *
+     * @param manager The manager to update reassign requests on
+     * @param shouldTriggerReassign the predicate to determine whether a request should be reassigned
+     */
+    public static void OnColonyUpdate(final IStandardRequestManager manager, final Predicate<IRequest> shouldTriggerReassign)
```
Raycoms: shouldn't it be lower case o?

src/main/java/com/minecolonies/coremod/tileentities/TileEntityRack.java
```diff
@@ -443,6 +482,8 @@ else if (compound.hasKey(TAG_RELATIVE_NEIGHBOR))
         }
         main = compound.getBoolean(TAG_MAIN);
         updateItemStorage();
+
+        this.inWarehouse = compound.getBoolean(TAG_IN_WAREHOUSE);
```
Raycoms, line 486: I like to put a "hasCompoundKey" wrapper around it, to not rely on the default value

src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
Asherslab: ugh. C# is opposite XD

src/main/java/com/minecolonies/coremod/tileentities/TileEntityRack.java
Asherslab: it returns false if there is no key. (just looked at the code)

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/StandardPlayerRequestResolver.java
```diff
@@ -237,6 +238,24 @@ public void onSystemReset()
...
+    {
+        new ArrayList<>(assignedRequests).stream()
+                .map(manager::getRequestForToken)
+                .filter(shouldTriggerReassign)
+                .filter(Objects::nonNull)
+                .forEach(request ->
+                {
+                    final IToken newResolverToken = manager.reassignRequest(request.getToken(), ImmutableList.of(token));
+
+                    if (newResolverToken != token)
```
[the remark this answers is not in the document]
Asherslab: ugh. that has the possibility of an NPE though....

src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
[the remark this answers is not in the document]
Asherslab: yeah yeah

src/main/java/com/minecolonies/coremod/tileentities/TileEntityRack.java
Asherslab: (which, if it's false, that's correct)

marchermans (approved):
Very nice work!
I have to say, this will do good for our community, it should make people more happy.

And I am finally happy that somebody else decided to take a stab at putting something in the RS, albeit it with training wheels, but hey you gotta start somewhere!

Nice work Asherslab.
One question though: Did you learn anything about how the RS operates?

src/main/java/com/minecolonies/coremod/colony/requestsystem/management/handlers/ResolverHandler.java
marchermans: It should be lower case :smile:

someaddons:
does removing the dman logic break it until you repaired the warehouse so racks are added?

marchermans:
Yes, technically.

someaddons:
I'd put sth like this in the BuildingWarehouse.java then instead, doesnt break existing colonies
```
 @Override
    public void onWorldTick(TickEvent.WorldTickEvent event)
    {
        if (init)
        {
            for (BlockPos rackPos : containerList)
            {
                TileEntity te = colony.getWorld().getTileEntity(rackPos);
                if (te instanceof TileEntityRack)
                {
                    ((TileEntityRack) te).setInWarehouse(true);
                }
            }
        }
    }
```

marchermans:
Well tbh I do not care for this, we have the repair feature for a reason......
That said we should really look into a mechanic that allows us to update colony structural data after load.......

Asherslab:
all this will mean is that IF they don't repair their warehouse (which if you look at the code i made in the warehouse class, it does it immediately on repair being pressed, so there's no need to actually have the builder do anything) they'll have to wait the normal "retry" tick time. which is defaulted at 1200 ticks, or one a minute.

Asherslab:
Also, yes orion. i certainly did learn alot about the RS system. enough that i spent like 6 hours trying to debug something in it (found that it was actually with the worker not the RS, but hey, i was able to track it through RS)

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/3646/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #3652, opened 2019-06-06
Fix RS NPE

Asherslab:
`manager.getRequestForToken` can be null, it sometimes is, i'm catching it here.
if you want me to *do* something instead of catch it (like remove it from the data store?) tell me :smile:

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/requestsystem/requests/AbstractRequest.java
+++ b/src/main/java/com/minecolonies/coremod/colony/requestsystem/requests/AbstractRequest.java
@@ -145,7 +145,7 @@ public abstract class AbstractRequest<R extends IRequestable> implements IReques
             {
                 manager.getRequestForToken(getParent()).childStateUpdated(manager, getToken());
             }
-            catch (final IllegalArgumentException ex)
+            catch (final IllegalArgumentException | NullPointerException ex)
             {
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/colony/requestsystem/requests/AbstractRequest.java
```diff
@@ -145,7 +145,7 @@ public void setState(@NotNull final IRequestManager manager, @NotNull final Requ
             {
                 manager.getRequestForToken(getParent()).childStateUpdated(manager, getToken());
             }
-            catch (final IllegalArgumentException ex)
```
marchermans, line 148: No this should crash.
I am catching it here exactly not for this reason, if this happens please, have them contact us since this is a major bug, and catching it is a Regression, I can much easier track crashes then dum log entries that nobody fucking sees.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/requestsystem/requests/AbstractRequest.java
Asherslab: @OrionDevelopment okay thanks. this is was spamming constantly on our server last night, it may still be. talk to me on discord if you need access to console / etc

PR #3661, opened 2019-06-09
Small fixes

### LINKED ISSUE #3635: REQUEST: "Child" as a Job instead of "Unemployed"
Can you guys maybe list Child Citizens as being "Child" instead of "Unemployed" in the Townhall list? This would greatly help with planning ahead so I am no longer wondering why I have 30 Unemployed but cannot hire anybody.

### LINKED ISSUE #3649: Placing the Postbox, shows the placement by hand warning.
The placing of a postbox shows the placement by hand warning.
Since this is a building that can not be placed by the wand, because no schematic exists for it, this should be disabled.

Versions: All

Reproduction: Attempt to place a new postbox in any colony.

### LINKED ISSUE #3658: Farmer stuck in loop?
### Minecolonies version
<0.10.293>

Version: 1.12.2

### Expected behavior

- 
Farmer is moving back and forth from the Farm block to the Field block and returns to Farm block repeatedly. Only exception is when it sleeps at house.

### Actual behavior
I used the commands to observe current activity. It goes as follows.
1.Preparing
2.FARMER_PLANT
3.Idle
4.Inventory_full

Number 2 doesn't function as the moment it reaches field it moves to Number 3.
As for Number 4 this shouldn't happen as it uses only 3 inventory slots for items currently.

1. Build acacia farm.
2. Build Acacia Plain Field
3. Supply farmer with tool/seed
4. ... Profit

Raycoms:
Closes #3658
Closes #3649
Closes #3635

- Fix postbox placement
- Add child count to TH statistic
- Fix field size calculation and farmer stuck 

Review please


Asherslab (approved):
From briefly looking it over. Looks good to me.

src/main/java/com/minecolonies/coremod/event/EventHandler.java
```diff
@@ -384,7 +385,7 @@ public void onBlockPlaced(@NotNull final BlockEvent.PlaceEvent event)
     {
         final EntityPlayer player = event.getPlayer();
         final World world = event.getWorld();
-        if (event.getPlacedBlock().getBlock() instanceof AbstractBlockHut)
+        if (event.getPlacedBlock().getBlock() instanceof AbstractBlockHut && event.getPlacedBlock().getBlock() != ModBlocks.blockPostBox)
```
someaddons, line 388: I think it would be better to check for the postbox below at line 397 instead, since right now it would skip the permission check upon placement.

Raycoms:
This event is only for suggestion, permission is handled in another event
handler

marchermans (approved):

PR #3668, opened 2019-06-11
Guard rework

Raycoms:
Guards are now smarter:
- Archers don't run to their targets
- Archers have farther range on high positions
- Knights run away from the target to avoid getting hit when having no shield (doesn't work well against skeletons)
- Balancing Knights and archers to be stronger early on but having less of an advantage for leveling
- Balancing armour to be available up to any level (no max cap)
- Fixing NPE with construction tape removal
- Fixing food intake for guards (eat earlier)
- Fix food intake for all workers (first try working chest then go to restaurant)

Review please


src/main/java/com/minecolonies/coremod/client/gui/WindowTownHall.java
```diff
@@ -622,8 +622,9 @@ private void createAndSetStatistics()
             }
             else
             {
-                final int length = citizen.getJob().split("\\.").length;
-                final String job = citizen.getJob().split("\\.")[length - 1].toLowerCase(Locale.ENGLISH);
+                final String[] splitString = citizen.getJob().split("\\.");
```
marchermans, line 625: This change is irrelevant, but still...… The JVM will optimize this away during runtime,

marchermans (approved):

someaddons (approved):

PR #3705, opened 2019-06-28
add way to detect unloaded chunks

Raycoms:
- Fix lag with unloaded chunks
-
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/util/FireworkUtils.java
+++ b/src/api/java/com/minecolonies/api/util/FireworkUtils.java
@@ -1,8 +1,8 @@
 package com.minecolonies.api.util;
 
 import net.minecraft.entity.item.EntityFireworkRocket;
+import net.minecraft.init.Items;
 import net.minecraft.item.ItemDye;
-import net.minecraft.item.ItemFirework;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.nbt.NBTTagList;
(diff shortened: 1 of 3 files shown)


marchermans (commented):
How does this fix the builder still iterating in the structure definition over unloaded chunks?

Raycoms:
He won't reach it

marchermans:
….. Can we just fix this for real please…… Not some half assed shit like this……
The Problem is just covered up now….......

Raycoms:
The question is what do you mean by that? We call the world from the chunk
in 3000 places, the only place where we can interrupt this centrally is
here.

Am Fr., 28. Juni 2019 um 15:11 Uhr schrieb Marc Hermans <
notifications@github.com>:

someaddons:
"Besides that, we should make sure our critique is as constructive as possible to help the developer to improve his current and future pull requests."

it would be better if you'd say sth like this instead:
I think a better solution would be X/Y because of <reason>

Raycoms:
Don'ts:
"- Use words like useless, senseless, stupid, fail, flaw ..."
I think shit and half-assed fits also in it =D

marchermans (changes requested):
Agreed I should have worded it differently, however this is the x time we do something like this. Instead of fixing the root cause, that is here fixing the structure iterator to only allow continuing interation if the next block is loaded, or else hold the iteration on the current block. Instead we put another bandaid on it and hope that in 2 months nobody touches the damn thing and it breaks again.

Raycoms:
We would have to do both though, else we will continue running into the
iterator and "progress".

Am Sa., 29. Juni 2019 um 13:41 Uhr schrieb Marc Hermans <
notifications@github.com>:

marchermans:
The builder should just do nothing, if he gets the same blockpos back from the iterator, period. Then the iterator controls everything

Raycoms:
I talked with @OrionDevelopment about this. Since the actual fix has to be a better iterator handling on the side of structurize we merge this bandaid fix and reserve the actual fix for the structurize rework.

PR #3759, opened 2019-07-30
Revert "Increase the size of the API and add an actuall API." due to crashes

Asherslab:
Reverts ldtteam/minecolonies#3756

https://pastebin.com/raw/GKxtzmbs
https://pastebin.com/raw/FJGZMweN


Asherslab:
Tested and works. am merging to fix builds

PR #3767, opened 2019-08-02
Minor fixes

Raycoms:
- Fixes issues with townhall teleport
- Bump Structurize
- Bump forge

Review please

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -3,7 +3,7 @@
 #
 # Minecraft / Forge.
 minecraft_version=1.12.2
-forge_version=14.23.5.2768
+forge_version=14.23.5.2836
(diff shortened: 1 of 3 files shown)


Asherslab (commented):
Can you test this with Server and Client, i'm not sure if this would work or not. (You're obviously crossing from server to client with the config variable, however i think that would have to have a network message. i don't think the client will change)

Raycoms:
Oh no, I am doing the opposite. We were using the client config var for a serverside setting. I changed that.

Asherslab (approved):
Looooooks good!

(I realised that those are serialise and deserialise methods, not NBT save/read methods)

PR #3770, opened 2019-08-03
setting the right home for guards.

someaddons:
Setting the right home for guards, since all assigned guards now are housed in their workbuilding they should set their home there aswell.

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
@@ -467,7 +467,7 @@ public abstract class AbstractBuildingGuards extends AbstractBuildingWorker
 
             // Set new home, since guards are housed at their workerbuilding.
             final AbstractBuilding building = citizen.getHomeBuilding();
-            if (building != null && !(building instanceof AbstractBuildingGuards))
+            if (building != null && !building.getID().equals(this.getID()))
             {
                 building.removeCitizen(citizen);
             }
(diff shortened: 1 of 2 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBarracksTower.java
```diff
@@ -85,23 +84,6 @@ public void requestUpgrade(final EntityPlayer player, final BlockPos builder)
         }
     }
 
-    @Override
```
Raycoms, line 88: Isn't this needed anymore?

Raycoms: Ah, nvm, it's now in the super class

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
```diff
@@ -464,6 +464,15 @@ public boolean assignCitizen(final CitizenData citizen)
                   .setBaseValue(SharedMonsterAttributes.ARMOR.getDefaultValue() + getDefenceBonus());
             }
             colony.getCitizenManager().calculateMaxCitizens();
+
+            // Set new home, since guards are housed at their workerbuilding.
+            final AbstractBuilding building = citizen.getHomeBuilding();
+            if (building != null && !(building instanceof AbstractBuildingGuards))
```
Raycoms: We should probably do a check if !building.getId.equals(this.getId)

someaddons: the  !(building instanceof AbstractBuildingGuards) already includes it since "this" is always a guard building

Raycoms: I mean, to compare if the building is really THIS building. (not only same type)

someaddons: ah that way, you're right, should do that to move guards from one building to another

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/3770/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #3777, opened 2019-08-06
Feature/api

marchermans:
### API
_Note this is a longer post, Orion._

Well here it is. After fixing my last attempt and completely redoing a lot of other things that came in contact with now moved abstract classes, we have a fully working API.
#### Features
- Full abstraction layer for almost all core functions of the mod:
   + Citizens
   + Jobs
   + Buildings
   + GuardTypes
   + Entities
   + Mobs (Separation between Archer / Melee / Chief, as well as Pirate / Barbarian)
- Most of the registries have been overhauled:
   + Some registries have been turned into ForgeRegistries:
       * Jobs (via JobEntries)
       * Buildings (via BuildingEntries)
       * GuardTypes (via IGuardType)
   + Others have been completely rewritten to fulfill their specific needs:
       * Mob AI
       * PathNavigate
       * ModelType
- DataManagers that take over part of the functionality the old registries performed and now mainly interact with the new registries to provide backwards compatibility.
- Complete removal of reflection.
   + Replaced with the java `Function` lambda interface call, using the same parameters, but now much more flexible.

##### Accessing the API components
Before I start listing all the changes and how they effect us, I wanted to take a look at how to access the components.
Virtually all components have a static `getInstance()`-method on their interface that can be used to access the components current implementation registered in the API.

This is done so that most people can use their existing learned names and keep on programming without a big shift. Examples are:
 - `IColonyManager.getInstance()`
 - `IBuildingRegistry.getInstance()`
These methods are of course just wrappers around their API container parts and are equivalent to:
 - `IMinecoloniesAPI.getInstance().getColonyManager()`
 - `IMinecoloniesAPI.getInstance().getBuildingRegistry()`

A quick note on some of these endpoints (in particular the API endpoints which are IForgeRegistries):
_Some of these endpoints are defined by external sources, like forge._
_As such we are not able to add a getInstance() method to the target type (like IForgeRegistry<?>)._
_To still make it as easy as possibly for coders to use the API we added marker interfaces (like IBuildingRegistry) which are not implemented anywhere, but return the relevant target type from their static getInstance() method call._
_As an example:_
```java
public interface IBuildingRegistry
{
    static IForgeRegistry<BuildingEntry> getInstance()
    {
        return IMinecoloniesAPI.getInstance().getBuildingRegistry();
    }
}
```
##### MinecoloniesAPI
This is the core access for the API, it will be populated on Pre-Init and can from then on be used by MineColonies and other mods.
Most of our internal logic has been rerouted to use the API it self (either directly during the setup of the API, or indirectly via calling into it, or ForgeRegistry events)
##### ColonyManager
The previous `ColonyManager` was a static access based core component.
Almost of all of its functionality has been kept and as such it should feel very familiar to most coders.
##### CitizenDataManger
Previously the methods from this new component have resided in the several classes, namely: `CitizenData` and `CitizenDataView`.
Their primary goal was it to create new instances of these two classes using the incoming data (NBT, Network etc).
Since these were static instances and used by the colony manager and its minion classes, they needed to be adapted.
Their static level access was removed, and they were moved to a different class to make them easier to abstract away.
This now also provides easy external access to who ever wants to create new citizens directly using the API.
##### MobAIRegistry
The primary reason for this to exist is to allow MineColonies itself to insert special `EntityAIBase` implementations into the Mobs (EG. to follow citizens etc)
However this also allows external mods to add tasks that allow our mobs to perform different tasks, based on what mod is loaded, without having to coremod or use the entire mod.
An easy and simple idea would be to make our mobs track vanilla villagers (by spawning them as part of 1.14 raids for example) and editing their tasks list.
##### PathNavigateRegistry
All of our entities use a custom high performance `PathNavigate` instance, to give the API access to this `PathNavigate` implementation this API endpoint was added and is by default only
populated by a special MineColonies instance.
However as we are planning to perform tests with different implementations, in the future, to improve performance even more.
This endpoint will come in handy, since it will allow us to replace the target `PathNavigate`, on any version of MineColonies running this API, without having to recompile MineColonies itself.
##### BuildingDataManager
This class takes over most of the instantiation duties from the the old `BuildingRegistry`.
It interacts with the new `IForgeRegistry<BuildingEntry>` based building registry and uses the their registered `BuildingEntry` instances to find and create the right buildings
based on their `ResourceLocation`-based registry name.
To make this possible `IBuildings`, `TileEntityColonyBuilding` and `AbstractBlockHut` have received methods that allow external system (including this data manager) to query which
`BuildingEntry` they represent, without having to fall back on things like class names, or even hashes from classnames for views (_Who the hell though that was a create idea?_)
##### BuildingRegisty
This is the first of three `IForgeRegistry`-based registries that have been added.
Instead of actually holding `IBuilding` it holds `BuildingEntry` instances, and functions more like a registry for the different types of buildings.
The `BuildingEntry` replaces the previous class and block registration handling that used the passed in classes and created instances via reflection on the constructors.
The new `BuildingEntry` now provides this by combining the block reference, as well as 2 references to function Lambdas.
The function Lamdas are used to create the building and its view without having to use reflection.
##### JobDataManager
As with the `BuildingDataManager` the `JobDataManagers` job, pun intended, is to create new instances of Jobs based on the network data or from NBT.
It loads the `JobEntry` based of its registry name, from the `IForgeRegistry<JobEntry>` known as the `JobRegistry`.
As again with buildings, each `IJob` holds a reference to a `JobEntry` and those are used to retrieve the correct data.
Again comparatively to the `BuildingEntry` the `JobEntry` uses Lambdas instead of reflective access which the `JobDataManager` then uses to create the new instances, before loading the data into the newly created instance.
##### JobRegistry.
This is the second of the three `IForgeRegistry`-based registries that have been added.
It holds the `JobEntry` as with the `BuildingRegistry` holding `BuildingEntry`.
##### GuardTypeDataManager
Additional utility class to perform lookups in the `GuardTypeRegistry`, while keeping nullability in check.
##### GuardTypeRegistry
The third of the initial `IForgeRegistry`-based registries that have been added.
It holds the different `IGuardType` implementations and stores the relevant information with regards to loading them.
`IGuardType` replaces the `EnumGuardType` implementations that was used before and now allows external mods to implement new guard types by simply creating a new `IJob` and `IGuardType` implementation.
One could think of guards which fight in special ways depending on a mod was installed, for example defend against taint in a tainted biome when Thaumcraft is installed.
##### ModelTypeRegistry
Allows the registration of a new `IModelType` to the API.
This makes it possible that depending on a new `IJob` a new model can be returned with custom textures.
`IModelType` replaces the previously used `EnumModelType` which was limited by MineColonies and what we provided.
It allows for the specification of texture names as well as texture counts to use.

#### Static access classes and initializers
Since we are now using `IForgeRegistry` for several registries, we are now using event based initialization of those components.
To still give us access to the relevant instances of those types I added several classes that hold the relevant entries:
 - ModBuildings
 - ModJobs
 - ModGuardTypes

Since these classes are part of the API they hold only the abstract classes or interfaces that are part of the API, not the actual implementations.
Yet to populate the registry properly and not expose the inner workings of the mod to the API.
The main implementation has accompanying classes that populate the static fields in those classes and register them to the registry:
 - ModBuildingsInitializer
 - ModJobsInitializer
 - ModGuardTypesInitializer
Each class contains static methods that are relevant it, and that initialize the registries from it.
The methods are called from event-handlers in `CommonProxy`.

#### Movement of type enumerations in classes to the parent level
A lot of our types (like scarecrow) or stages, are stored in enumerations, that are child-classes of the target class that they are relevant for.
To make abstraction possible and to make it possible for external mods to query information, most of these enumerations where pulled out of the parent class and stored in the API.
They where not turned into registries since they where heavily situational and extending them did not seem to lead to any new functionality.


someaddons (dismissed):

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=30237&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=30283&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

PR #3953, opened 2019-10-20
don't run the avoidance AI if guard

Raycoms:
Closes ##3955

- After some time of debugging I found out that the reason the guards get stuck on pathing is the avoidance AI which is not disabled for them.
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/jobs/AbstractJobGuard.java
+++ b/src/main/java/com/minecolonies/coremod/colony/jobs/AbstractJobGuard.java
@@ -83,6 +83,6 @@ public abstract class AbstractJobGuard extends AbstractJob
     @Override
     public boolean allowsAvoidance()
     {
-        return true;
+        return false;
     }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAICitizenAvoidEntity.java
```diff
@@ -241,7 +241,7 @@ public boolean shouldExecute()
     public boolean shouldContinueExecuting()
     {
         stateMachine.tick();
-        return citizen.isCurrentlyFleeing();
+        return citizen.isCurrentlyFleeing() && !(citizen.getCitizenJobHandler().getColonyJob() instanceof AbstractJobGuard);
```
marchermans: Please turn this into a method on the job instead of a instance of.
Something like: job.requiresEnemyEvasion()

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/jobs/AbstractJobGuard.java
```diff
@@ -79,4 +79,10 @@ public SoundEvent getBedTimeSound()
         }
         return null;
     }
+
+    @Override
+    public boolean allowsAvoidance()
+    {
+        return true;
```
marchermans: Should this not be false?

marchermans (changes requested):

marchermans (dismissed):

someaddons (dismissed):

PR #3965, opened 2019-10-21
ColonyTickPR#1: Chunk load feature

someaddons:
I've split the PR for the colony ticks into three, for easier reviewing.

- Stores the loaded chunks of the colony in the colony

Feel free to review already :)


Asherslab:
I realise it's just a draft, however what's the purpose of storing these chunks?

someaddons:
they're used to see if a colony is loaded and it then ticks the right stuff, depending on it. Probably should've mentioned that all these three drafts belong together and are just split for easier reviewing

marchermans (approved):

PR #3967, opened 2019-10-21
ColonyTickPR#3: Colony tick handling feature

someaddons:
I've split the PR for the colony ticks into three, for easier reviewing.

- Introduces states to the colony
- Simplifies ticking logic and makes it about 3x faster
- Reworked some older ticking logic stuff

Details on how it works:

The basic idea is that states for a colony lets us freely control what the colony updates in each state.
For now there are three states:
 - Active: All managers /other things updated/tick as usual
 - Unloaded: Only logic which should tick even when the colony chunks are not loaded, requires a player(global subscriber) online for the colony with certain permissions.
 - Inactive: Colony does nothing except occasionally checking if it needs to go into another state.

Using the statemachine for this allows us to define transitions for logic which needs to be executed on ticks, simply by choosing in which state it should tick, giving it a function to execute and specifying at which rate we need the updates.

Regarding performance this makes our colonies require about three times less time for its ticks/updates. This gain is from adjusting the update rates of on-tick logic to a fitting rate instead of doing it all the time. Also some stuff is done event based now, instead of constant updates like adding/removing subscribers which makes it more scaleable. Basically the gain is from executing updates/logic only when actually required. 

Few examples of when stuff is beeing ticked:
(Active state)
Day/nighttime check is done every 20 ticks
Waypoints are still checked every 100 ticks
the RS system is still checked every 1 tick
Raids and most managers are updated every 500 ticks

Feel free to review already :)


src/api/java/com/minecolonies/api/entity/mobs/util/MobEventsUtils.java
```diff
@@ -90,6 +92,23 @@ public static void raiderEvent(final World world, final IColony colony)
...
+        if (colony.getWorld().getDifficulty() != EnumDifficulty.PEACEFUL
+              && Configurations.gameplay.doBarbariansSpawn
+              && colony.getRaiderManager().canHaveRaiderEvents()
+              && !colony.getPackageManager().getOfficerSubscribers().isEmpty()
+              && MobEventsUtils.isItTimeToRaid(colony.getWorld(), colony))
+        {
+            MobEventsUtils.raiderEvent(colony.getWorld(), colony);
+        }
+
+        return false;
```
Asherslab: Currently this always returns false, might want to change that =P

Asherslab:
Interesting idea

src/api/java/com/minecolonies/api/entity/mobs/util/MobEventsUtils.java
someaddons: oh the false is there because it previously was used as a transition in the colonys statemachine, but it isn't anymore, so ye would make sense to change it to reflect whether a raid is triggered or not :)

Raycoms:
Looks good now as well :shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/3967/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/api/java/com/minecolonies/api/colony/workorders/IWorkManager.java
```diff
@@ -109,9 +109,9 @@
      * Process updates on the World Tick.
      * Currently, does periodic Work Order cleanup.
      *
-     * @param event {@link net.minecraftforge.fml.common.gameevent.TickEvent.WorldTickEvent}.
+     * @param colony {@link TickEvent.WorldTickEvent}.
      */
-    void onWorldTick(@NotNull TickEvent.WorldTickEvent event);
+    void onWorldTick(@NotNull IColony colony);
```
marchermans: Again rename to onColonyTick and fix the Docu

src/api/java/com/minecolonies/api/colony/managers/interfaces/ICitizenManager.java
```diff
@@ -158,9 +158,9 @@ default ICitizenData spawnOrCreateCitizen(final ICitizenData data, @NotNull fina
 
     /**
      * Actions to execute on a specific world tick event.
-     * @param event the event.
+     * @param colony the event.
      */
-    void onWorldTick(final TickEvent.WorldTickEvent event);
+    void onWorldTick(final IColony colony);
```
marchermans: Again rename to onColonyTick and fix the Docu

src/api/java/com/minecolonies/api/colony/managers/interfaces/IBuildingManager.java
```diff
@@ -55,15 +56,15 @@
 
     /**
      * Tick the buildings on world tick.
-     * @param event the event.
+     * @param colony the event.
      */
-    void onWorldTick(TickEvent.WorldTickEvent event);
+    void onWorldTick(IColony colony);
```
marchermans: Again rename to onColonyTick and fix the docu,

src/api/java/com/minecolonies/api/colony/buildings/ICitizenAssignable.java
```diff
@@ -35,9 +35,9 @@
     /**
      * On tick of the world.
      *
-     * @param event {@link TickEvent.WorldTickEvent}
+     * @param colony {@link TickEvent.WorldTickEvent}
      */
-    void onWorldTick(TickEvent.WorldTickEvent event);
+    void onWorldTick(IColony colony);
```
marchermans: Rename to onColonyTick and fix the Docu.
This will come up a couple of times in this PR, as such I would suggest pulling it up into a super Interface: 'IColonyUpdatable(IColony colony)' and then using that in its place.

src/api/java/com/minecolonies/api/entity/mobs/util/MobEventsUtils.java
```diff
@@ -90,6 +91,24 @@ public static void raiderEvent(final World world, final IColony colony)
...
+     */
+    public static boolean checkForRaid(final IColony colony)
+    {
+        if (colony.getWorld().getDifficulty() != EnumDifficulty.PEACEFUL
+              && Configurations.gameplay.doBarbariansSpawn
+              && colony.getRaiderManager().canHaveRaiderEvents()
+              && !colony.getPackageManager().getGlobalSubscribers().isEmpty()
+              && MobEventsUtils.isItTimeToRaid(colony.getWorld(), colony))
+        {
+            MobEventsUtils.raiderEvent(colony.getWorld(), colony);
```
marchermans, line 115: I do not agree with the fact that 'checkForRaid' also trigger said raid.
Split this into two, or name the method and update the documentation.

Personally I would split it into two methods, since it is clearly a different peace of functionality to check if a raid can be triggered, and to trigger the actual raid.
Possibly have a third method which first checks for if a Raid is possible then triggers the raid, finally returning the result from the check, so it is semantically the same as this proposed changed, however in that case it needs to be abundantly clear what that method does!

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -305,6 +314,173 @@ protected Colony(final int id, @Nullable final World world)
...
+        colonyStateMachine.addTransition(new TickingTransition(UNLOADED, () -> true, this::updateState, UPDATE_STATE_INTERVAL));
+        colonyStateMachine.addTransition(new TickingTransition(ACTIVE, () -> true, this::updateState, UPDATE_STATE_INTERVAL));
+
+        colonyStateMachine.addTransition(new TickingTransition(ACTIVE, this::updateSubscribers, () -> ACTIVE, UPDATE_SUBSCRIBERS_INTERVAL));
+        colonyStateMachine.addTransition(new TickingTransition(ACTIVE, this::tickRequests, () -> ACTIVE, UPDATE_RS_INTERVAL));
+        colonyStateMachine.addTransition(new TickingTransition(ACTIVE, this::checkDayTime, () -> ACTIVE, UPDATE_DAYTIME_INTERVAL));
+        colonyStateMachine.addTransition(new TickingTransition(ACTIVE, this::updateWayPoints, () -> ACTIVE, CHECK_WAYPOINT_EVERY));
+        colonyStateMachine.addTransition(new TickingTransition(ACTIVE, this::worldTickSlow , () -> ACTIVE, MAX_TICKRATE));
+        colonyStateMachine.addTransition(new TickingTransition(UNLOADED,this::worldTickUnloaded, () -> UNLOADED, MAX_TICKRATE));
+    }
```
marchermans, line 331: More documentation on how this system works is needed, why is this better then previously?
Does it slow down ticking or not?
How does it interact with other systems which rely on accurate timings?

src/main/java/com/minecolonies/coremod/colony/ColonyView.java
```diff
@@ -1092,6 +1092,12 @@ public void onWorldTick(@NotNull final TickEvent.WorldTickEvent event)
 
     }
 
+    @Override
+    public boolean areAllColonyChunksLoaded()
+    {
+        return false;
```
marchermans, line 1098: Please check if this is used anywhere on the client side.
Especially in entity AI's or in methods who call entity AI's this is dangerous and can cause desyncs!

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -449,19 +448,12 @@ public void removeRecipe(final IToken token)
...
-        super.onWorldTick(event);
-
-        //
-        // Code below this check won't lag each tick anymore
-        //
-        if (!Colony.shallUpdate(event.world, ONWORLD_TICK_AVERAGE))
-        {
-            return;
-        }
+        super.onWorldTick(colony);
```
marchermans: Again rename and update docu.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractCitizenAssignable.java
```diff
@@ -88,10 +88,10 @@ public void onServerTick(final TickEvent.ServerTickEvent event)
     /**
      * On tick of the world.
      *
-     * @param event {@link net.minecraftforge.fml.common.gameevent.TickEvent.WorldTickEvent}
+     * @param colony {@link TickEvent.WorldTickEvent}
      */
     @Override
-    public void onWorldTick(final TickEvent.WorldTickEvent event)
+    public void onWorldTick(final IColony colony)
```
marchermans: Again rename and update docu.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBaker.java
```diff
@@ -262,17 +261,9 @@ public String getJobName()
...
-        super.onWorldTick(event);
-
-        //
-        // Code below this check won't lag each tick anymore
-        //
-        if (!Colony.shallUpdate(event.world, WAIT_TICKS))
-        {
-            return;
-        }
+        super.onWorldTick(colony);
```
marchermans: Again rename and update docu.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingHome.java
```diff
@@ -240,26 +241,14 @@ public void removeCitizen(@NotNull final ICitizenData citizen)
...
-    public void onWorldTick(@NotNull final TickEvent.WorldTickEvent event)
-    {
-        //
-        // Code below this check won't lag each tick anymore
-        //
-        if (!Colony.shallUpdate(event.world, ONWORLD_TICK_AVERAGE))
-        {
-            return;
-        }
+        childCreationTimer+= TWENTYFIVESEC;
```
marchermans, line 253: Again rename and update docu.

src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
```diff
@@ -169,27 +165,16 @@ public void sendPackets(final Set<EntityPlayerMP> closeSubscribers, final Set<En
...
+                building.onWorldTick(colony);
             }
         }
-
-        if (tickCounter == 20)
-        {
-            tickCounter = 0;
-        }
-        tickCounter++;
     }
```
marchermans: Again rename and update docu.

src/main/java/com/minecolonies/coremod/colony/managers/CitizenManager.java
```diff
@@ -440,26 +445,26 @@ else if (relation < 1)
...
+        if (this.colony.hasTownHall() && getCitizens().size() < Configurations.gameplay.initialCitizenAmount)
         {
-            int respawnInterval = Configurations.gameplay.citizenRespawnInterval * TICKS_SECOND;
-            respawnInterval -= (SECONDS_A_MINUTE * colony.getBuildingManager().getTownHall().getBuildingLevel());
+            respawnInterval -= 500 + (SECONDS_A_MINUTE * colony.getBuildingManager().getTownHall().getBuildingLevel());
 
-            if ((event.world.getTotalWorldTime() + 1) % (respawnInterval + 1) == 0)
+            if (respawnInterval <= 0)
             {
+                respawnInterval = Configurations.gameplay.citizenRespawnInterval * TICKS_SECOND;
```
marchermans: Again rename and update docu.

src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
```diff
@@ -223,37 +224,34 @@ else if(horde.size() <= SMALL_HORDE_SIZE)
     @Override
     public void onWorldTick(@NotNull final World world)
     {
-        if (Colony.shallUpdate(world, TICKS_SECOND * SECONDS_A_MINUTE))
+        for (final Map.Entry<BlockPos, Tuple<String, Long>> entry : new HashMap<>(schematicMap).entrySet())
```
marchermans, line 233: Why was the need for an update tick check here removed?
Also since this is a RaidManager, should this not be colony dependent and tick just as the other managers does with the Colony, not the world?

src/main/java/com/minecolonies/coremod/colony/workorders/WorkManager.java
```diff
@@ -262,47 +259,43 @@ public void addWorkOrder(@NotNull final IWorkOrder order, final boolean readingF
...
+                iter.remove();
+                dirty = true;
+            }
+            else if (o.hasChanged())
+            {
+                dirty = true;
+                o.resetChange();
             }
         }
     }
```
marchermans: Again rename and update docu.

marchermans (changes requested):
Again lots of documentation and explaining todo.
In particular but not in this order are needed:
  * A list of points and idea why this is needed.
  * How does this improve performance, without causing problems for players?
  * How does this new sync system with subscribers interact with this?
  * How much performance gain is created by this system?
  * Where is this performance gain coming from?
  * How does it work? Which states are checked every tick? Which aren't? Which states are checked randomly? And which states are ticked every X/Y/Z Ticks? If so how were these decisions to tick a given state randomly, or only in X/Y/Z ticks, made? Was there any research done on how external systems and the API react to these changes?

In particular: All of these three are not emergency PR's, we can do a week or two without them just fine. However for the amount of code and logic that all of them change, the Initial PR comment is way, way, way to short.

src/main/java/com/minecolonies/coremod/colony/ColonyView.java
someaddons: isnt called on client side, its a new function I added

src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
someaddons: ye I changed it to the same way the other managers are

someaddons:
added a more detailed description to the initial post

marchermans (approved):

Issue #1470, opened 2017-09-23
[Feature] Empty huts counter


MineShadow7:
In Townhall there could be a counter to huts without workers/citizens. 

They should not include warehouse and barracks.

PR #4005, opened 2019-10-30
Added empty huts counter

MrGrim48:
Feature request #1470 

- Added empty huts counter

EmptyHutsCounter

This is my first contribution to another repo. 
Correct me if I may have used incorrect branches, formatting, etc.

Review please


Raycoms:
Nice work, I noticed that the formatting is a bit off, you can find in the documention folder in the repo (which you downloaded) a formatter you should be able to apply to intellij or eclipse to fix that easily.

For the language file you can add the string in the en_us file to finish the todo

MrGrim48:
I've modified the language file to accommodate the max job value and removed both todo.
This code is going on the assumption that the building name is the same as the job name as I am not sure how I can get the job names.

In any case the max jobs value should show as long as the language supports it.
Thinking about it now, a mismatch with the job/building name may show a null value.

Raycoms:
:shipit: 

@OrionDevelopment 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4005/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/client/gui/WindowTownHall.java
```diff
@@ -635,6 +635,23 @@ private void createAndSetStatistics()
...
             }
+		}
+
+        final Map<String, Integer> jobMaxCountMap = new HashMap<>();
+        for (@NotNull final IBuildingView building : townHall.getColony().getBuildings())
+        {
+            if (!building.isBuilding())
+            {
+                String buildingName = building.getSchematicName().toLowerCase();
+                if (!buildingName.equals("warehouse") && !buildingName.equals("barracks") && !buildingName.equals("citizen") && !buildingName.equals("townhall"))
```
Asherslab: I'd much rather here you do something akin to `building instanceOf BuildingWarehouse` or etc, i think string references in this case are a lesser alternative

Asherslab (changes requested):
Looks good, not sure about these string comparisons though. Instanceof usage will always work whereas a name string comparison may break if a name or etc is changed in the future

src/main/java/com/minecolonies/coremod/client/gui/WindowTownHall.java
marchermans: Yep!

Raycoms: Would be BuildWarehouse.View probably in this case though (as a note)

someaddons: instanceof AbstractBuildingWorker.View might make this even simpler

MrGrim48:
A result of job names differing from building names. 
I've noticed a mixed use of "Baker" and "Bakery".

doubleBaker

Raycoms:
We could check if the job is contained in the building for this @MrGrim48

Raycoms:
@MrGrim48 I thought about it.
In the Building classes we already jave a getJobName(), if we simply add this method to the view as well. We avoid any issues and can use it consistently for all buildings.

    @NotNull
    @Override
    public String getJobName()

MrGrim48:
Initially I wanted to use the method `getJobName()` but didn't directly have access to it.
From what I can see with the baker, that method would return the same string as the schematic name.

The bakers job returns the string constant `COM_MINECOLONIES_COREMOD_JOB_BAKER `from `TranslationConstants.java` which returns bakery instead of baker.

From my understanding, the buildings create the jobs and pass it to the citizen to hold onto and if that is the case, I don't know how I could obtain the jobs name in the event that no citizens have the job.

Even if we have `getJobName()` return the job constant instead, I feel that adding the method to the view is beyond my current understanding.
The general use of abstraction and interfaces alone is currently foreign to me with which I'm still trying to get used to.

Raycoms:
Okay, sounds good, this should work for most of the cases for now.
:shipit:

Asherslab:
My review still stands

Raycoms:
Yeah, he adapted it now to instanceof

Asherslab (approved):
If it’s tested and working, looks good to go

PR #4013, opened 2019-10-31
Fix chunk data/claim issue

someaddons:
- Fixes an issue with chunk claims, where the claim happens and later is overwritten by data from our ChunkLoadStorage cache. Now we apply the cache before modifying the final chunk data. Likely a cause of randomly unclaimed/orphaned chunks
- Added detection to double colonies which could happen because of this claim bug and cause accessibility issues with citizens/buildings and other glitches when two are present in the very same area/with the same owner. Those are logged as warning now, so that users can manually choose the right colony to keep/delete the other.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/IColonyManagerCapability.java
+++ b/src/main/java/com/minecolonies/coremod/colony/IColonyManagerCapability.java
@@ -1,6 +1,9 @@
 package com.minecolonies.coremod.colony;
 
+import com.google.common.collect.ArrayListMultimap;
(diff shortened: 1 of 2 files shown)


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4013/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

Raycoms:
@OrionDevelopment

marchermans (approved):

PR #4024, opened 2019-11-02
Enchanter

Raycoms:
Closes #115

# Changes proposed in this pull request:
- the enchanter has a GUI where you can set from who he will drain experience
- he will walk to these workers daily and drain some experience (you can set how many levels per day)
- there is a chance of him enchanting randomly something while draining (side effect)
- if he has a high enough experience level he will request enchanted tomes 
- he will enchant this tome with a random enchantment, depending on his level, there is a chance to get high level enchantments
A few notes
a) It is probably one of the nicest workers to watch working
b) He takes only 50% of the xp he gains (because he is very skilled) but this percentage is configurable in the config file
c) you can configure which books he can produce in the config file as well
So, also modded enchantments work

- Also adds potential max, to allow to recruit workers if there is a free space for the guard tower.

Review please

PR #4035, opened 2019-11-03
Enchanter 1.14

Raycoms:
Closes #4024 

- the enchanter has a GUI where you can set from who he will drain experience
- he will walk to these workers daily and drain some experience (you can set how many levels per day)
- there is a chance of him enchanting randomly something while draining (side effect)
if he has a high enough experience level he will request enchanted tomes

he will enchant this tome with a random enchantment, depending on his level, there is a chance to get high level enchantments
A few notes
a) It is probably one of the nicest workers to watch working
b) He takes only 50% of the xp he gains (because he is very skilled) but this percentage is configurable in the config file
c) you can configure which books he can produce in the config file as well
So, also modded enchantments work

Also adds potential max, to allow to recruit workers if there is a free space for the guard tower.

Review please


marchermans:
:shipit:

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4035/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #4036, opened 2019-11-03
Improve/fix some chunk data issues

someaddons:
- made sure the chunk is mark dirty whenever the capability data changes
- removed unecessary chunk load counter which goes out of sync and could cause orphaned chunks
- changed backup loading to not re-claim all colonies when auto-reloading
- Fixed three more chunk cache merging issues, where we didnt end up with the correct data.
- Chunk-cache claims are now done in-order to avoid information loss and end up with the correct data.

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4036/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

Asherslab (dismissed):
Looks good to me, only question is if there’s an IChunk interface or something similar we can use instead, just for greater compatibility with mods that may change that, if not then sure.

someaddons:
they added the IChunk stuff with 1.14^^

Asherslab:
ahh, well then in the 1.14 port if you could use that instead that'd be great

someaddons:
ye I'll try to use it where possible, I remember the interface has a few missing functions yet :D

someaddons:
seems I've found another bug, don't merge it yet xD

someaddons:
Ready to be reviewed again :)

Raycoms:
:shipit: amazing work on this, great job on finding these small bugs by debugging it over and over and over... =D

@OrionDevelopment @Asherslab

marchermans (approved):

PR #4038, opened 2019-11-05
Feature/empty hut counter

MrGrim48:
- getJobName method added to View.
- getMaxInhabitants method added to View.
- Additions to logic in getting the total max jobs.

Review please


MrGrim48:
Maybe I should have created a new branch.

Raycoms:
since you changed it to baker and not bakery, you will have to fix that in the language file too

MrGrim48:
The following constants that was used in the TranslationConstants file does not exist in any of the translation files, at least according to the search function. 
```
"com.minecolonies.coremod.gui.workerHuts.bakery"
"com.minecolonies.coremod.job.bakery"
"com.minecolonies.coremod.bakery.noFurnace"
"com.minecolonies.coremod.bakery.noRecipe"
```
I realised in typing this, I missed WindowHutBaker attempting to reference a non existent bakery string.
I merely changed these constants to reference existing strings in the language files.

The BakerSounds file still uses bakery since I don't know enough about the sound system to make any changes. My assumption at first glance is that it's a path to the sound file and unless I'm looking at the wrong place, the folder is named baker instead of bakery.

Edit:
There were duplicate entries for bakers in practically all language files. 100 matches each.

```
com.minecolonies.coremod.gui.townHall.population.bakerys=Bakeries: %d
com.minecolonies.coremod.gui.townHall.population.bakers=Bakers: %d/%d
```

Raycoms:
I think the combat and archer training names might have to be lower case.

@Asherslab also review please

Asherslab:
Other than some code style stuff (things that could be final and etc), looks fine

MrGrim48:
The getJobName method is used logging, player messages and sounds.

In the case of the player massages, it may look better to keep the capitalization. Perhaps the method should instead return a localization textContent using the same constants as the jobs classes.

The sound class compares the string from getWorkerString in job sounds with getJobName. I could instead add a method to the worker buildings called getSoundName or something to that effect.

I may work on that in a new branch if that sounds alright.

Edit:
I don't want to make too many changes but is there a better way for worker buildings to return a jobs name? In the case of the bakers, the jobs class returned a slightly different string from the worker buildings. Should the job have a method to return a non-localized text that worker buildings can refer to?

Raycoms:
What we could do is that the worker building gets the localization for the job directly from the job building, do we use this anywhere important in a registry or can we simply change the "getJobName" to guarantee it is equal to the one used in the job?

MrGrim48:
SoundUtils compares the job name from the building and the getWorkerString from the sound class. 
Other instances it would _seem_ fine to use a localized name, except for CitizenInfoCommand, not sure how that would be affected.

Raycoms:
We should probably unify them all to guarantee they use the same one. Probably via a static method in the job class JobBuilder.getJobName() which we then use in the SoundUtils and in the building. The result of this we can then use to build the localized one "com.minecolonies."jobName".desc"

MrGrim48:
Would this static getJobName() exist along side the current getName() which would return the job description and localized string respectively?
I am in favour of the static getJobName() though.

Raycoms:
I am currently thinking about it, maybe we actually have the job get the JobName from the building. So that we have a guarantee on this and we can always get it from the building easily and it is guaranteed to be the same.

MrGrim48:
So, BuildingBuilder.getJobName() would return "Builder" the the job would use this to form the localized string for it's getName()?

Raycoms:
No, in the building it would then not be static

MrGrim48:
I now have have less of an understanding. How would the job and soundutils get the job name from the building.
On another note, would each sound class be tied to one job? At the moment the ranger guard tower and archer training buildings use the same sounds.

Raycoms:
Ranger and archer is the same thing, from the job we should be able to get an instance of the building somehow.

MrGrim48:
I believe I found it, more simpler than I thought.
```
@NotNull
@Override
public String getName()
{
    final String jobName = getCitizen().getWorkBuilding().getJobName();
    return "com.minecolonies.coremod.job." + jobName;
}
```

Raycoms:
Wait, I got a better thing, we got a registry, where you should be able to get the building from the job, etc.

MrGrim48:
I can't figure out how else to get the instance of the building from the job.
As for this pull request, I don't know what changes I could make.

marvin-bitterlich:


```diff

Complexity increasing per file
==============================
- src/main/java/com/minecolonies/coremod/client/gui/WindowTownHall.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/Minecolonies/minecolonies/pullRequest?prid=4438728&bid=15015008)

MrGrim48:
I've removed some changes.
Review please.

Raycoms:
:shipit: now for me @OrionDevelopment @Asherslab @someaddons 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4038/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

marchermans (approved):

Asherslab (approved):

PR #4042, opened 2019-11-05
feature/cactus milling

### LINKED ISSUE #4000: All of the cactus blocks are missing from the sawmill
### Minecolonies version

Version: minecolonies-0.10.465-ALPHA-universal

### Expected behavior

- Being able to add recipes for: Cactus planks and all cactus related blocks (Stairs, slabs, fences, doors, etc)

### Actual behaviour

- Met with error message: _Unable to store the blueprint. Either the sawmill has no space for any more blueprints, or the blueprint doesn't correspond with its profession._

1. Teach recipe on sawmill hut block
2. Add recipe for any cactus block
3. Failed result

---

Also missing is the the iron ingot (from Iron nuggets) recipe from the blacksmith 
^ would be very useful with the sifter

RyTechro:
Version/1.14

Closes #4000 

[x] Add cactus to sawmill
[] add iron nugget to blacksmith

The cactus issue is a problem that cactus is now seen as a log, so should have the ItemTags.Log same for cactus planks with ItemTags.Planks. This is a structurize issue but for now this works too.

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSawmill.java
```diff
@@ -107,7 +107,7 @@ public boolean canRecipeBeAdded(final IToken token)
         {
             if(!ItemStackUtils.isEmpty(stack))
             {
-                if (stack.getItem().isIn(ItemTags.PLANKS) || stack.getItem().isIn(ItemTags.LOGS))
+                if (stack.getItem().isIn(ItemTags.PLANKS) || stack.getItem().isIn(ItemTags.LOGS) || stack.getItem().getRegistryName().getPath().contains("cactus"))
```
Asherslab: It'd be better to add the cactus blocks to the item tags for planks and logs. 
or in the least make custom tags for them, because string comparisons are going to break easily

Raycoms:
Agreed

RyTechro:
I agree, but that should be done inside structurize in my opinion. Thats because that mod adds the block

Asherslab:
well then we'll need a PR into structurize won't we =D

RyTechro:
And do we have a central place for these kind of fixes, if we decide to add the ItemTags inside minecolonies? The constructor for the sawmill seems like a solution because thats where the issue arrises but that feels like the entirely wrong place to do such a thing

RyTechro:
Dang it, forgot to select the cla signed github user again...

Raycoms:
@RyTechro when you add this to structurize https://github.com/ldtteam/structurize/pulls
You would add this to the data folder. In 1.14 itemtags are handled in the data folder (there are already a few items there to copy)

PR #4055, opened 2019-11-08
Chunk data fixes v1.14

someaddons:
- made sure the chunk is mark dirty whenever the capability data changes
- removed unecessary chunk load counter which goes out of sync and could cause orphaned chunks
- changed backup loading to not re-claim all colonies when auto-reloading
- Fixed three more chunk cache merging issues, where we didnt end up with the correct data.
- Chunk-cache claims are now done in-order to avoid information loss and end up with the correct data.

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4055/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #4056, opened 2019-11-08
improve state typing/add dynamic tick rates

someaddons:
Added dynamic state types for the machines, so that they no longer rely on the AI-defined state interface, which forces you to implement AI specific stuff even when not beeing used in an AI environment. The generic type allows you to use custom state with matching return types nicely.

Dynamic ticking added to the statemachine gives them the ability to skip/throttle down the ticks they receive, which is currently used for an AI config which allows lower update rates.(was previously done outside the statemachine). Having this within the statemachine lets logic which uses tick as time measurement to adapt the timers accordingly so that despite the machine ticking less the actions work as intended, e.g. setDelay does -= getTickRate();

Changed the internal streams to loops as the lists the streams run on are very short and calls are quite frequent, so loops are better suited performance wise. (Also showed when doing a profiler test with the big server world(450+) colonies, resulting in about half the tick time needed). Also this should make debugging AI a bit better, as I think streams are harder to debug on.

Review please


src/api/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickRateStateMachine.java
```diff
@@ -40,26 +52,59 @@ public TickRateStateMachine(@NotNull final IAIState initialState, @NotNull final
...
+        tickRateCounter = 0;
+
+        // Update the tick counter for transitions
         tickCounter++;
         if (tickCounter > MAX_TICKRATE)
         {
             tickCounter = 1;
         }
 
+        for (final ITickingTransition<S> transition:eventTransitionMap.get(AIBlockingEventType.AI_BLOCKING))
```
Raycoms, line 70: For @OrionDevelopment we do this with for loops because the lists are very small and the streaming overhead is considerably (@someaddons did some tests and it  improved the tps considerably)

marchermans (changes requested):
I do not understand the need for this pr in the slightest. I would love to see much more documentation on all of this. Not just on the code, but on the pr in particular. Two bullet points are not going to get a pr that touches a core mechanic merged.

I need to see at least proper documentation on the change to generic state machines as well as the dynamic ticking, how does it work. Why was it introduced? How do what classes interact with this? How can this be co trolled via the api? Etc all of the entire enchilada.

I would accept bullet points for Bugfixes, in particular hotpatches, but this is a feature pr and as such more documentation is needed. It takes the OP 5 minutes to write that docu but spares reviewers and people who need to work with that change hours trying to understand what is going on.

someaddons:
@OrionDevelopment updated the description

marchermans (approved):

marvin-bitterlich:


```diff

Complexity increasing per file
==============================
- src/api/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickRateStateMachine.java  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/minecolonies/minecolonies/pullRequest?prid=4464736&bid=15080508)

PR #4065, opened 2019-11-10
Fix/msg clientside

Raycoms:
- Fixes an exception on the server because the wrong class is used.
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/network/messages/RecallCitizenMessage.java
+++ b/src/main/java/com/minecolonies/coremod/network/messages/RecallCitizenMessage.java
@@ -18,6 +18,7 @@ import net.minecraftforge.fml.common.network.simpleimpl.IMessage;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/network/messages/RecallCitizenMessage.java
```diff
@@ -111,20 +112,25 @@ public void messageOnServerThread(final RecallCitizenMessage message, final Enti
...
-                            citizenData.updateCitizenEntityIfNecessary();
-                        }
-                        else
-                        {
-                            citizenData.setCitizenEntity(list.get(0));
+
+                            final List<AbstractEntityCitizen> list = new ArrayList<>(player.getServerWorld()
+                                                                                       .getEntities(AbstractEntityCitizen.class,
+                                                                                         entityCitizen -> entityCitizen.getCitizenColonyHandler().getColonyId() == colony.getID()
+                                                                                                            && entityCitizen.getCitizenData().getId() == oldCitizenData.getId()));
```
Nightenom, line 123: formatting

Nightenom (approved):

src/main/java/com/minecolonies/coremod/network/messages/RecallCitizenMessage.java
Raycoms: wdym?

someaddons (approved):

src/main/java/com/minecolonies/coremod/network/messages/RecallCitizenMessage.java
Raycoms: That's the best way to format it imo

Nightenom: like the amount of spaces is disgusting, I usually use one indent layer for multi-line expressions, someone does two indent layers to distinguish between blocks and multi-line
edit: or same indentation for every line (just looks weird to me when each line is indented differently)

Raycoms: I usually go with 1 extra indent for the lambda, and then a new one for each logical subexpression.

Raycoms: The important thing is that it is readable. The lambda is the central point of the expression, so it has to be understandable where it starts and what the conditions are

Nightenom: ah I see, you just use two spaces instead of four
what I was writing about is the number of spaces when the line is shifted to the new indent layer according to the previous line

Raycoms: Oh, that's what Intellij does for me on Linux =D

PR #4084, opened 2019-11-18
Schematics 1.14

lunadragnstar:
mineshafts


Raycoms:
Fix miner shaft schematics

PR #4091, opened 2019-11-20
improve state typing/add dynamic tick rates v1.14

someaddons:
Added dynamic state types for the machines, so that they no longer rely on the AI-defined state interface, which forces you to implement AI specific stuff even when not beeing used in an AI environment. The generic type allows you to use custom state with matching return types nicely.

Dynamic ticking added to the statemachine gives them the ability to skip/throttle down the ticks they receive, which is currently used for an AI config which allows lower update rates.(was previously done outside the statemachine). Having this within the statemachine lets logic which uses tick as time measurement to adapt the timers accordingly so that despite the machine ticking less the actions work as intended, e.g. setDelay does -= getTickRate();

Changed the internal streams to loops as the lists the streams run on are very short and calls are quite frequent, so loops are better suited performance wise. (Also showed when doing a profiler test with the big server world(450+) colonies, resulting in about half the tick time needed). Also this should make debugging AI a bit better, as I think streams are harder to debug on.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/ai/statemachine/AIEventTarget.java
+++ b/src/api/java/com/minecolonies/api/entity/ai/statemachine/AIEventTarget.java
@@ -12,7 +12,7 @@ import java.util.function.Supplier;
  * Special AI Targets which are used for preState cecks and limits.
  * They are checked before normal AITargets always
  */
-public class AIEventTarget extends TickingEvent
+public class AIEventTarget extends TickingEvent<IAIState>
 {
     /**
      * Construct a special target.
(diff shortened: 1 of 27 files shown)


marvin-bitterlich:


```diff

Complexity increasing per file
==============================
- src/api/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickRateStateMachine.java  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/minecolonies/minecolonies/pullRequest?prid=4538066&bid=15274356)

PR #4093, opened 2019-11-21
Add pathfinding shutdown

### LINKED ISSUE #4087: Pathfinding ThreadPoolExecutor doesn't get shutdown
### Minecolonies version

Version: minecolonies-0.10.552-BETA-universal
Minecraft: 1.14.4
Forge 28.1.90

### Expected behavior

- The dedicated server shuts down normally, when closed with the `stop` command

### Actual behaviour

- When shutting down the dedicated server, the jvm doesn't close.
I think the problem is, that the ThreadPoolExecutor for pathfinding https://github.com/ldtteam/minecolonies/blob/3c1e5aede7389a31c635d18f1c033c9a308cf442/src/main/java/com/minecolonies/coremod/entity/pathfinding/Pathfinding.java#L29
never gets shut down when closing the server, because the created threads are non-daemon threads.

1. Start a dedicated server
2. Play a bit, so villagers use pathfinding
3. Stop the server with the `stop` command
4. The jvm never closes

### Verification
1. Add a simple ThreadFactory to the ThreadPoolExecutor:
```java
private static class DebugThreadFactory implements ThreadFactory {
        private ThreadFactory defaultFactory;
        
        public DebugThreadFactory() {
            this.defaultFactory = Executors.defaultThreadFactory();
        }
        
        @Override
        public Thread newThread(Runnable r) {
            Thread t = this.defaultFactory.newThread(r);
            t.setName("minecolonies-pathfinding-" + t.getName());
            return t;
        }
        
    }
```
2. Get a thread dump after stoping the server. See that there is only one non-deamon thread running, which is the minecolonies-pathfinding  thread.

Bricktricker:
Closes #4087

- Add a call to shutdown the pathfindig ThreadPoolExecutor on the FMLServerStoppedEvent

Review please


src/main/java/com/minecolonies/coremod/entity/pathfinding/Pathfinding.java
```diff
@@ -46,6 +46,11 @@ private Pathfinding()
     {
         return executor.submit(job);
     }
+    
+    public static void shutdown()
```
Raycoms: Can you add some javadoc here please?

Raycoms:
Besides the javadoc it looks good, as soon as that is added :shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4093/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

Asherslab (commented):
Do we start a new Pathfinding executor when another server starts? Where do we do that (not entirely certain). 

I say this because in single player this might cause issues if someone exits the world and enters it again. That being if we don’t already start another one

Raycoms:
@Asherslab it's not on world stop but on server stop, so it should be fine.

marchermans:
FML Server Stop and World stop is identical for SinglePlayer.
Please validate that this does not break it.

Asherslab:
@Raycoms  Exactly as Orion said, server stop is world stop in single player. I know because I use it in ChatChain

Raycoms:
Oh, I didn't know that, then we'll probably a onServerStart to initiate the executor as well.

Asherslab:
Something like that, i don't know where we initiate/start it, so i hence my question

Bricktricker:
Sorry, I haven‘t testet my PR correctly.

I will fix it later and move the ThreadPoolExecutor construction to onServerStarting, like Raycoms suggested.

src/main/java/com/minecolonies/coremod/entity/pathfinding/Pathfinding.java
Bricktricker: Added javadoc to start and shutdown

Bricktricker:
I have updated the PR to start the ThreadPoolExecutor during the FMLServerStartingEvent.
I have also successfully tested it in singleplayer and on a dedicated server.

marchermans (approved):

marchermans:
Approved but it seems like the deployment test is failing.

Raycoms:
Because in his own fork.

someaddons:
fails always for 1.14

someaddons (approved):

marchermans:
Then somebody needs to add the missing files...... Cause that was not merged properly then....

marchermans:
I will see what I can do tomorrow

Raycoms:
Seems to not have solved the problem completely:

https://github.com/LordILarrion/Minecolonies/blob/master/pathing_crash.txt

at com.minecolonies.coremod.entity.pathfinding.Pathfinding.enqueue(Pathfinding.java:62) ~[?:0.10.564-ALPHA] {re:classloading}
	at com.minecolonies.coremod.entity.pathfinding.MinecoloniesAdvancedPathNavigate.setPathJob(MinecoloniesAdvancedPathNavigate.java:99) ~[?:0.10.564-ALPHA] {re:classloading}
	at com.minecolonies.coremod.entity.pathfinding.MinecoloniesAdvancedPathNavigate.moveAwayFromXYZ(MinecoloniesAdvancedPathNavigate.java:82) ~[?:0.10.564-ALPHA] {re:classloading}
	
I guess we can start the pathfinding queue if it is null, instead on world start

Bricktricker:
I didn't know there is pathfinding on the client too.
A null check in `enqueue` should work.

Raycoms:
there shouldn't be on the client side, but in singleplayer it should happen in general.

Raycoms:
But that crash was on th eserver, the one above, wasn't it?

Bricktricker:
No, the crash happens on the client, while playing on a dedicated server.
Look at the stacktrace from the linked crash report, its going through client classes
(e.g. `net.minecraft.client.world.ClientWorld.func_217418_a(ClientWorld.java:165)`)

Raycoms:
Ah, hadn't seen that, but weird, I didn't know pathfinding was called on the client.

Raycoms:
I think the easiest way to fix this
is just by "if pool is null -> start pool" and then remove it from serverStart

PR #4151, opened 2019-12-15
for non tooltype blocks, use current hand

Raycoms:
- For blocks with invalid tooltype, use hand.
-
-

Review please


someaddons (approved):
tested, works.

PR #4215, opened 2020-01-08
Add Permission Check to Visitor Notification

cjm721:
This is useful for server admins so they can checkup on players without the players noticing on a per admin configurable basis.

- Adds a Forge Permission check to see if a player's entering or leaving a village should trigger a message to the village owners.


src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -1476,7 +1477,9 @@ public void addVisitingPlayer(final EntityPlayer player)
         {
             visitingPlayers.add(player);
             LanguageHandler.sendPlayerMessage(player, ENTERING_COLONY_MESSAGE, this.getPermissions().getOwnerName());
-            LanguageHandler.sendPlayersMessage(getImportantMessageEntityPlayers(), ENTERING_COLONY_MESSAGE_NOTIFY, player.getName(), this.getName());
+            if(!PermissionAPI.hasPermission(player, "minecolonies.colony.visitor.silent")) {
```
Asherslab: make this an or `if the player is OP`

Raycoms:
Where does this one here come from? "minecolonies.colony.visitor.silent"

Asherslab:
> Where does this one here come from? "minecolonies.colony.visitor.silent"

(this was explained in DMs on Discord, ignore)

src/main/java/com/minecolonies/coremod/colony/Colony.java
cjm721: Added the permission by default to operator in MineColonies.java. Seemed cleaner then directly checking for operator status here (only way I know to check it is to grab the operator list and see if UUID is in it).

This does have the slight disadvantage though for users of default Forge perm handler in single player that Forge considers everyone operator if they are allowed to run all commands (See DefaultPermissionsHandler.java in Forge).

Asherslab (changes requested):
Looks good, last thing i'd like is a config option to disable that. Some servers may wish OPs to also be announced

Asherslab (approved):

Raycoms:
Merge it

Raycoms:
Thanks for your contribution

Raycoms:
Can you please port this to 1.14 too?

PR #4254, opened 2020-01-21
Added Inventory Tweak support

Goten87:
- added inventory tweaks support 
- 
-

Review please

(my first pull request hope i did it correct)


build.properties
```diff
@@ -3,8 +3,9 @@
 #
 # Minecraft / Forge.
 minecraft_version=1.12.2
-forge_version=14.23.5.2836
+forge_version=14.23.5.2768
```
Raycoms: any particular reason to rollback the forge version?

Goten87: My setup dint want to build so I went to stable and it worked can try to change it back to the version you used now

Raycoms: Yeah, we used a higher version on purpose because there is a forge fix in there we need.

src/main/java/com/minecolonies/coremod/inventory/ContainerRack.java
```diff
@@ -17,6 +18,7 @@
 /**
  * The container class for the rack.
  */
+@ChestContainer
 public class ContainerRack extends net.minecraft.inventory.Container
```
Raycoms, line 22: Did you produce a jar and tested it with and without invtweaks installed?

Goten87: I did and no errors as far as I know

Raycoms: okay, sounds good

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4254/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/inventory/ContainerRack.java
marchermans: This will not work if InventoryTweaks is not available as far as i know.
@ChestContainer still needs to be loadable in this case...... Please verify this again!.

marchermans (changes requested):

marchermans:
Granted i am not sure this will work
But okey

marchermans:
approved

src/main/java/com/minecolonies/coremod/inventory/ContainerRack.java
Asherslab: I agree with @OrionDevelopment here, don't think this would work

Asherslab (changes requested):

Raycoms:
He did test it and said it worked with and without (with the jar)

PR #4279, opened 2020-01-28
Feature/uni 1.15

Raycoms:
This pull-request introduces the initial design of the research-system 
Including the initial part of the combat research tree.

Research is centered around the university building which has, on Page 2 of the GUI the list of main branches.

image

There will be three main branches in the future.
- Combat
- Technology
- Civilian

Each research has 
- Unique id
- Description
- Requirement of resources
- Requirement of colony buildings
- An Effect

Research can either apply bonuses to existing workers, or unlock certain feature or buildings.
(Effect).

There is one global research tree that contains the entire research tree that exists.
Colonies will copy over some parts of this tree to their individual trees while they progress.
However, most information stays at the global tree to allow the developers to change stats there and keep things forward compatible. 

With this, the colony, on load, will query the tree for the effects which apply to the research the colony has already unlocked.

Research might be in progress:

image

Finished:

image

Not started and locked (research requirements are not met)

image

or not started and locked because requirements like building level are not met.

image

or, simply not started (with the button enabled).

Researching takes the resources from the player inventory.

There are "and" relationships allowing to research both childs.

image

And "or" relationships where one branch will be locked forever after one of them is researched.

image

The university allows to research up to "number of employees" researches to be processed in parallel.

However, in this initial design (because of the limited number of research options) this is limited to 1.

Review please


marchermans:
Get we get some more docu in the pr message?

Raycoms:
Sorry yes.

Am Fr., 31. Jan. 2020 um 13:32 Uhr schrieb Marc Hermans <
notifications@github.com>:

src/api/java/com/minecolonies/api/research/GlobalResearch.java
```diff
@@ -0,0 +1,267 @@
...
+import net.minecraftforge.common.ForgeConfigSpec;
+import net.minecraftforge.items.IItemHandler;
+import net.minecraftforge.items.wrapper.InvWrapper;
+import net.minecraftforge.registries.ForgeRegistries;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+
+/**
+ * The implementation of the IResearch interface which represents one type of research.
```
someaddons: which type of research?^^ could use some more details

src/api/java/com/minecolonies/api/research/GlobalResearch.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    @Override
+    public void startResearch(@NotNull final PlayerEntity player, @NotNull final LocalResearchTree localResearchTree)
+    {
+        if (localResearchTree.getResearch(this.branch, this.id) == null && hasEnoughResources(new InvWrapper(player.inventory)))
+        {
+            final ILocalResearch research = new LocalResearch(this.id, this.branch, this.depth);
+            research.setState(ResearchState.IN_PROGRESS);
+            localResearchTree.addResearch(branch, research);
```
someaddons: should we mark the colony dirty here to save the resource consume?

src/api/java/com/minecolonies/api/research/GlobalResearch.java
Raycoms: good point, fixed

src/api/java/com/minecolonies/api/research/GlobalResearch.java
Raycoms: We mark the colony dirty after we call this in the message.

marchermans:
I would like to see a command here that can be used to hand a colony a research level.... Also have you considered making this some how data driven possibly powered of the advancement system that Mc has?

Raycoms:
As gamemode 1, you can instant research things if you have the requirement.
I may add a command in the next PR for it. But I wanted to restrict this
pull-request to the necessary basics.

I have considered making it json based, but the complexity would be
essentially higher and there is no way to nicely define effects in json. So
I would have to make effects in the code anyway and then make jsons
reference it.
So I decided to go with full in code anyway.

Maybe in a future version we can extend this to read from json as well and
use the available effects.



Am Fr., 31. Jan. 2020 um 14:48 Uhr schrieb Marc Hermans <
notifications@github.com>:

someaddons (commented):
looks good overall to me, looking forward to having researches :D

src/api/java/com/minecolonies/api/colony/IColony.java
```diff
@@ -259,6 +261,18 @@
      */
     boolean isRemote();
 
+    /**
+     * Getter for the research tree.
+     * @return the colony own research tree.
+     */
+    LocalResearchTree getResearchTree();
```
marchermans: This should be an interface and possibly in its own manager.

src/api/java/com/minecolonies/api/colony/IColony.java
```diff
@@ -259,6 +261,18 @@
...
+
+    /**
+     * Get the colonies own research effects.
+     * @return the ResearchEffects object.
+     */
+    ResearchEffects getResearchEffects();
```
marchermans: Again this should be an interface and in its own manager.

src/api/java/com/minecolonies/api/crafting/ItemStorageFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import com.minecolonies.api.colony.requestsystem.factory.IFactoryController;
+import com.minecolonies.api.util.constant.TypeConstants;
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.CompoundNBT;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Factory implementation taking care of creating new instances, serializing and deserializing RecipeStorages.
+ */
+public class ItemStorageFactory implements IItemStorageFactory
```
marchermans: This class is probably not supposed to be part of the api.

src/api/java/com/minecolonies/api/research/BuildingResearchRequirement.java
```diff
@@ -0,0 +1,44 @@
+package com.minecolonies.api.research;
+
+import com.minecolonies.api.colony.IColony;
+import com.minecolonies.api.research.interfaces.IResearchRequirement;
+import net.minecraft.util.text.TranslationTextComponent;
+
+/**
+ * Certain building research requirements.
+ */
+public class BuildingResearchRequirement implements IResearchRequirement
```
marchermans: I am pretty sure that this class should not be in the api.

src/api/java/com/minecolonies/api/research/GlobalResearch.java
```diff
@@ -0,0 +1,267 @@
...
+import net.minecraftforge.items.wrapper.InvWrapper;
+import net.minecraftforge.registries.ForgeRegistries;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+
+/**
+ * The implementation of the IGlobalResearch interface which represents the research on the global level.
+ */
+public class GlobalResearch implements IGlobalResearch
```
marchermans: Again should not be part of the api.

src/api/java/com/minecolonies/api/research/LocalResearch.java
```diff
@@ -0,0 +1,108 @@
...
+import com.minecolonies.api.research.effects.ResearchEffects;
+import com.minecolonies.api.research.interfaces.ILocalResearch;
+import com.minecolonies.api.research.util.ResearchState;
+
+import static com.minecolonies.api.research.util.ResearchConstants.BASE_RESEARCH_TIME;
+
+/**
+ * The implementation of the ILocalResearch interface which represents one type of research, stored in each colony.
+ */
+public class LocalResearch implements ILocalResearch
```
marchermans: Again this should not be part of the api.

src/api/java/com/minecolonies/api/research/LocalResearchTree.java
```diff
@@ -0,0 +1,136 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+
+import static com.minecolonies.api.research.util.ResearchConstants.*;
+
+/**
+ * The class which contains all research.
+ */
+public class LocalResearchTree
```
marchermans: Again this should be abstracted away into an interface if need be and then this implementation should not be in the api.

src/api/java/com/minecolonies/api/research/ResearchTree.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static com.minecolonies.api.research.util.ResearchConstants.TAG_RESEARCH_TREE;
+
+/**
+ * The class which contains all research.
+ */
+public class ResearchTree
```
marchermans: Create an interface.....

marchermans (changes requested):
I have stopped reviewing after 16 files.......
Please refactor all files that you added to the api package.

There is no real reason why all these implementation need to be in the api package, and the way the global research is handled is less then optimal. What is the conceptual idea behind it?

src/api/java/com/minecolonies/api/research/effects/AbstractResearchEffect.java
```diff
@@ -0,0 +1,60 @@
+package com.minecolonies.api.research.effects;
+
+import com.minecolonies.api.research.interfaces.IResearchEffect;
+import net.minecraft.util.text.TranslationTextComponent;
+
+/**
+ * The modifier research effect, it returns a double modifier.
```
marchermans: How does this return a double modifier?

src/api/java/com/minecolonies/api/research/effects/IResearchEffects.java
```diff
@@ -0,0 +1,38 @@
+package com.minecolonies.api.research.effects;
+
+import com.minecolonies.api.research.interfaces.IResearchEffect;
+import net.minecraft.nbt.CompoundNBT;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The map of unlocked research effects of a given colony.
+ */
+public interface IResearchEffects
```
marchermans: Rename to IResearchEffectsManager.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
```diff
@@ -0,0 +1,137 @@
+package com.minecolonies.api.research.interfaces;
```
marchermans: Since it is part of the api anyway -> Scap the interfaces package name and just dump them in the parent.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Add a child to a research.
+     * @param child the child to add.
+     */
+    void addChild(IGlobalResearch child);
+
+    /**
+     * Get the list of childs of the research.
+     */
+    List<String> getChilds();
```
marchermans: ImmutableList....

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
```diff
@@ -0,0 +1,137 @@
...
+     * Getter for the research requirement.
+     * @return the requirement.
+     */
+    IResearchRequirement getResearchRequirement();
+
+    /**
+     * Get the effect of the research.
+     * @return the effect.
+     */
+    IResearchEffect getEffect();
```
marchermans: In my opinion a single research might have more then one effect in the future. This should already be accommodated in the api from the get go.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Set the research requirement.
+     * @param requirement the requirement.
+     */
+    void setRequirement(final IResearchRequirement requirement);
+    /**
+     * Getter for the research requirement.
+     * @return the requirement.
+     */
+    IResearchRequirement getResearchRequirement();
```
marchermans: In my opinion, it might be a nice idea to have requests which have multiple requirements.
This could be adapted by making this a collection.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
```diff
@@ -0,0 +1,137 @@
...
+     * Set the parent of a research.
+     * @param id the id of the parent.
+     */
+    void setParent(String id);
+
+    /**
+     * Set the research requirement.
+     * @param requirement the requirement.
+     */
+    void setRequirement(final IResearchRequirement requirement);
```
marchermans: Again it might be a good idea to make it possible for a single research to have multiple requirements.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearchTree.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * The class which contains all research.
+ */
+public interface IGlobalResearchTree
+{
+    /**
+     * Get an instance of this Tree.
+     * @return the instance.
+     */
+    static IGlobalResearchTree getInstance()
```
marchermans: Nice much better.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearchTree.java
```diff
@@ -0,0 +1,66 @@
+package com.minecolonies.api.research.interfaces;
+
+import com.minecolonies.api.MinecoloniesAPIProxy;
+import net.minecraft.nbt.CompoundNBT;
+
+import java.util.*;
+
+/**
+ * The class which contains all research.
```
marchermans: This is not a class.
And this documentation can be made better.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearchTree.java
```diff
@@ -0,0 +1,66 @@
...
+        return MinecoloniesAPIProxy.getInstance().getGlobalResearchTree();
+    }
+
+    /**
+     * Get a research by id.
+     * @param id the id of the research.
+     * @param branch the branch of the research.
+     * @return the IResearch object.
+     */
+    IGlobalResearch getResearch(final String branch, final String id);
```
marchermans: In my opinion a branch should be denoted as a single root research element which gets unlocked by default.
As such all ids should be unique and the notion of branches should not be needed.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearchTree.java
```diff
@@ -0,0 +1,66 @@
...
+     * Add a research to the tree.
+     * @param research the research to add.
+     */
+    void addResearch(final String branch, final IGlobalResearch research);
+
+    /**
+     * Get the list of all branches.
+     * @return the list of branches.
+     */
+    List<String> getBranches();
```
marchermans: Yeah not sure about the design decision to treat branches differently, but i guess it could work.

One question though why is there not an object that represents a branch with at least the following properties:
+ id
+ display translation key
+ display icon
??

Would allow for much better UX experience and tracking -> Possibly register the objects via a forge registry?

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearchTree.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param branch the branch it belongs to.
+     * @return the list of research without parent.
+     */
+    List<String> getPrimaryResearch(final String branch);
+
+    /**
+     * Write the research tree to NBT.
+     * @param compound the compound.
+     */
+    void writeToNBT(final CompoundNBT compound);
```
marchermans: I am not sure why we are writing global objects into NBT.
This tree should be constant and as such no reading and writing from NBT should be required.

If this tree is not constant, it has no reason to exist, since then it can not 100% reliably be used as a reference for the local colony trees.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearchTree.java
```diff
@@ -0,0 +1,66 @@
...
+     * Read the research tree from NBT.
+     * @param compound the compound to read it from.
+    +
+     */
+    void readFromNBT(final CompoundNBT compound);
+
+    /**
+     * Load cost for all research.
+     */
+    void loadCost();
```
marchermans: Cost? Why is this handled separately? Are you possibly initializing this object too early?
If so this should be handled differently.

src/api/java/com/minecolonies/api/research/interfaces/ILocalResearchTree.java
```diff
@@ -0,0 +1,57 @@
+package com.minecolonies.api.research.interfaces;
+
+import com.minecolonies.api.research.interfaces.ILocalResearch;
+import net.minecraft.nbt.CompoundNBT;
+import java.util.List;
+
+/**
+ * The class which contains all research.
```
marchermans: Nope

marchermans: It does not contain all research and it is not a class.

src/api/java/com/minecolonies/api/research/interfaces/ILocalResearchTree.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public interface ILocalResearchTree
```
marchermans: TBH The way global and local trees are implemented right now leaves a lot to be desired.

In my eyes the global research tree should be build once at startup, and then referenced by the colony.
The local tree just is a kind of lookup structure that says: This colony has this research done. It does not even need to be a tree, but a map with some state suffices. Effects of completed researches can then be easily looked up in the global and if we ever change the global it will automatically update all the locals.

src/api/java/com/minecolonies/api/research/interfaces/IResearchEffect.java
```diff
@@ -0,0 +1,53 @@
+package com.minecolonies.api.research.interfaces;
+
+import net.minecraft.util.text.TranslationTextComponent;
+
+/**
+ * The effect of a research.
+ * @param <T> the type of the value.
+ */
+public interface IResearchEffect<T>
```
marchermans: This is named wrongly, or implemented wrongly.

As it states right now this is implemented as research effect type. However an effect type should have no notion of its branch, description etc.

Especially should an effect have no notion of its parent..... An effect should be active or not, based on if a research marks it as active or not....

This should be redesigned, cause right now it is a bit of a mess.

src/api/java/com/minecolonies/api/research/interfaces/IResearchRequirement.java
```diff
@@ -0,0 +1,23 @@
+package com.minecolonies.api.research.interfaces;
+
+import com.minecolonies.api.colony.IColony;
+import net.minecraft.util.text.TranslationTextComponent;
+
+/**
+ * Interface of research requirements.
```
marchermans: Nope. Document better this tells me nothing about what this interface does.

src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -2084,6 +2084,40 @@ public static boolean removeStackFromItemHandler(final IItemHandler handler, fin
...
+        int tries = 0;
+        while (tries < count)
+        {
+            final int slot = findFirstSlotInItemHandlerNotEmptyWith(handler, stack -> ItemStackUtils.compareItemStacksIgnoreStackSize(workingStack, stack));
+            if (slot == -1)
+            {
+                return;
+            }
+
+            final int removedSize = ItemStackUtils.getSize(handler.extractItem(slot, localCount, false));
```
marchermans, line 2107: This does not work. You also need to check if removedSize is bigger then the localCount, so you will need to simulate first. You have no idea how the IItemHandler is implemented, if it even allows partial extraction. Just because you now only use it for extraction from the Player or Citizen does not mean this is a valid utility implementation, and we might possibly run into this problem in the future.

src/main/java/com/minecolonies/coremod/MineColonies.java
```diff
@@ -67,6 +70,11 @@
      */
     private static Configuration config;
 
+    /**
+     * Create the global research tree.
+     */
+    private static IGlobalResearchTree globalResearchTree;
```
marchermans: Nope no reason for this object to be here!

marchermans: The api needs to manage this, Minecolonies should not hold an instance to it.

src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowSkeleton.java
```diff
@@ -107,10 +107,6 @@ public void onButtonClicked(@NotNull final Button button)
             buttons.get(button.getID()).accept(button);
             Network.getNetwork().sendToServer(new ClickGuiButtonTriggerMessage(button.getID(), this.resource));
         }
-        else
```
marchermans, line 110: Why was this removed?!?!?!??!!??!?!?!?!

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -814,6 +832,18 @@ public boolean isRemote()
         return false;
     }
 
+    @Override
```
marchermans, line 832: Okey i do not understand your definition logic here:
Hey i do have a research manager cause i was asked to implement it -> But I still refuse to rewrite the surface properly and just leave the methods here?

These two methods have nothing directly to do with the colony -> The colony needs to give access to the manager (as with all other managers) and the manager gives access to these two methods. Not the colony gives access to the managers values....

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBarracks.java
```diff
@@ -96,6 +98,18 @@ public void onDestroyed()
         super.onDestroyed();
     }
 
+    @Override
+    public void requestUpgrade(final PlayerEntity player, final BlockPos builder)
+    {
+        final UnlockResearchEffect effect = colony.getResearchEffects().getEffect("Barracks", UnlockResearchEffect.class);
```
marchermans: Why is this an effect?!?!?!?
This is a research condition.

Conditions -> Research -> Effects....

marchermans: Oh no hold on i get it!
This is an effect that unlocks a given option.....

Well okey disregard the previous comment! -> But still this needs to be better named. `UnlockBuildingEffect` or `UnlockAbilityEffect` are more appropriately named for buildings and citizen feature locking. Cause right now i had to read 6 classes to understand what is going on.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/EntityAIKnight.java
```diff
@@ -111,12 +111,12 @@ protected int getAttackDelay()
...
     protected IAIState attackProtect()
     {
-        final int shieldSlot = InventoryUtils.findFirstSlotInItemHandlerWith(getInventory(),
-          Items.SHIELD);
+        final int shieldSlot = InventoryUtils.findFirstSlotInItemHandlerWith(getInventory(), Items.SHIELD);
 
         if (target != null && target.isAlive())
         {
-            if (shieldSlot != -1)
+            final UnlockResearchEffect effect = worker.getCitizenColonyHandler().getColony().getResearchEffects().getEffect("Shield Usage", UnlockResearchEffect.class);
```
marchermans: This is 6 classes further in the review and here i understood what UnlockResearchEffect does..... Not well named.... No comment explaining what this variable is doing and checking.

marchermans: BTW this can simply be a boolean check -> getResearchEffects.has("Shield Usage").

marchermans: Also Constants!

marchermans (changes requested):
Some fundamental problems.....

src/api/java/com/minecolonies/api/research/effects/AbstractResearchEffect.java
Raycoms: Wrong javadoc

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
Raycoms: k

src/api/java/com/minecolonies/api/research/effects/IResearchEffects.java
Raycoms: k

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
Raycoms: k

src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowSkeleton.java
Raycoms: Because we do not handle all buttons like this and it spams the console when we don't.

src/api/java/com/minecolonies/api/research/interfaces/IGlobalResearch.java
Raycoms: For now I wanted to do this 1 research = 1 effect, especially to make it easier to display it in the UI

src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
```diff
@@ -156,6 +157,38 @@
     public final ForgeConfigSpec.BooleanValue creativeResolve;
     public final ForgeConfigSpec.BooleanValue canPlayerUseResetCommand;
 
+    /*  --------------------------------------------------------------------------------- *
+     *  ------------------- ######## Research Settings ######## ------------------- *
+     *  --------------------------------------------------------------------------------- */
+
+    public final ForgeConfigSpec.ConfigValue<List<? extends String>> tactictraining;
```
Asherslab, line 164: This is definitely something that would be removed once Datapack support is added.

Asherslab: Also, is the capitalization supposed to be all lowercase?

src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
```diff
@@ -224,7 +257,7 @@ protected CommonConfiguration(final ForgeConfigSpec.Builder builder)
         townHallPadding = defineInteger(builder, "townhallpadding", 20, 1, 20000);
         townHallPaddingChunk = defineInteger(builder, "townhallpaddingchunk", 1, 1, 200);
         restrictColonyPlacement = defineBoolean(builder, "restrictcolonyplacement", false);
-        enableDynamicColonySizes = defineBoolean(builder, "enabledynamiccolonysizes", false);
+        enableDynamicColonySizes = defineBoolean(builder, "enabledynamiccolonysizes", true);
```
Asherslab, line 260: Is that going to cause issues for pre-existing worlds?

src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import com.minecolonies.api.colony.requestsystem.token.IToken;
+import net.minecraft.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+import static com.minecolonies.api.util.constant.Constants.PARAMS_ITEMSTORAGE;
+
+/**
+ * Interface for the IItemStorageFactory which is responsible for creating and maintaining ItemStorage objects.
+ */
+public interface IItemStorageFactory extends IFactory<FactoryVoidInput, ItemStorage>
```
Asherslab, line 15: only at the top of the review, so not entirely certain what ItemStorage is being used for but I got the general gist. Why're we using the ItemStorage class here? We don't need any of the same data for research criteria that we do for crafting. Take a look at the ItemPredicate class from vanilla, very useful class and might be better suited to our needs? especially once using datapacks.
my main issue is that once we move to datapacks these classes will most likely have no need

src/api/java/com/minecolonies/api/research/IGlobalResearch.java
```diff
@@ -0,0 +1,138 @@
...
+     * Human readable description of research.
+     * @return the description.
+     */
+    String getDesc();
+
+    /**
+     * Getter of the id of the research.
+     * @return the String id.
+     */
+    String getId();
```
Asherslab, line 66: Is there a reason we're using String IDs and not resource locations? would be better suited for when research isn't statically (code) defined. or even just for when other mods add it

src/api/java/com/minecolonies/api/research/IGlobalResearch.java
```diff
@@ -0,0 +1,138 @@
...
+     * Get the id of the parent IResearch.
+     * @return the string id.
+     */
+    String getParent();
+
+    /**
+     * Get the string name of the branch.
+     * @return the branch name.
+     */
+    String getBranch();
```
Asherslab, line 78: I also think branch names could be resource locations?

src/api/java/com/minecolonies/api/research/IGlobalResearch.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Get the depth in the research tree.
+     * @return the depth.
+     */
+    int getDepth();
+
+    /**
+     * Check if this research is an only child research.
+     * This means, after researching one child no other childs can e researched.
```
Asherslab: typo here btw

src/api/java/com/minecolonies/api/research/IGlobalResearch.java
```diff
@@ -0,0 +1,138 @@
...
+     * @param onlyChild the param to set.
+     */
+    void setOnlyChild(boolean onlyChild);
+
+    /**
+     * Check if this research has other childs.
+     * @param localTree the local tree of the colony.
+     * @return true if so.
+     */
+    boolean hasResearchedChild(@NotNull final ILocalResearchTree localTree);
```
Asherslab, line 104: this method name doesn't match the javadoc's description, something like `hasChildren` would make more sense, unless the javadoc is wrong

src/api/java/com/minecolonies/api/research/IGlobalResearch.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Add a child to a research.
+     * @param child the child to add.
+     */
+    void addChild(IGlobalResearch child);
+
+    /**
+     * Get the list of childs of the research.
+     */
+    ImmutableList<String> getChilds();
```
Asherslab, line 115: plural child is children

src/api/java/com/minecolonies/api/research/IGlobalResearchTree.java
```diff
@@ -0,0 +1,66 @@
...
+     * Add a research to the tree.
+     * @param research the research to add.
+     */
+    void addResearch(final String branch, final IGlobalResearch research);
+
+    /**
+     * Get the list of all branches.
+     * @return the list of branches.
+     */
+    List<String> getBranches();
```
Asherslab, line 40: yeah i definitely think branches should use ResourceLocation IDs not strings

src/api/java/com/minecolonies/api/research/ILocalResearch.java
```diff
@@ -0,0 +1,59 @@
+package com.minecolonies.api.research;
+
+import com.minecolonies.api.research.effects.IResearchEffectManager;
+import com.minecolonies.api.research.util.ResearchState;
+
+/**
+ * Interface defining how a local research at a colony is.
```
Asherslab, line 7: Could you refine this to describe the fact that this is for the current research item that is being worked on in the colony?

src/api/java/com/minecolonies/api/research/ILocalResearchTree.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param research the research to add.
+     */
+    void addResearch(final String branch, final ILocalResearch research);
+
+    /**
+     * Check if a branch already researched a level 6 research.
+     * @param branch the branch to check.
+     * @return true if so.
+     */
+    boolean branchAlreadyResearchedLevelSix(final String branch);
```
Asherslab: that's a long method name. could be more concise? `isBranchAboveLevel6`. better yet, that's a very specific method, could we instead do `getHighestResearchedLevel` and then compare it where it's needed to 6 or etc.

src/api/java/com/minecolonies/api/research/IResearchEffect.java
```diff
@@ -0,0 +1,34 @@
...
+     * Set the research effect.
+     * @param effect the value to set it to.
+     */
+    void setEffect(T effect);
+
+    /**
+     * Getter for the ID of the effect.
+     * @return the String id.
+     */
+    String getId();
```
Asherslab: ResourceLocation here

src/api/java/com/minecolonies/api/research/IResearchManager.java
```diff
@@ -0,0 +1,35 @@
+package com.minecolonies.api.research;
+
+import com.minecolonies.api.research.effects.IResearchEffectManager;
+import net.minecraft.nbt.CompoundNBT;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Research manager of the colony holding the tree and effects.
```
Asherslab, line 8: this javadoc makes it sound like only one colony "holds the tree and effects", I'd assume that isn't so?

src/api/java/com/minecolonies/api/colony/jobs/ModJobs.java
```diff
@@ -35,6 +35,7 @@
     public static final ResourceLocation SIFTER_ID         = new ResourceLocation(Constants.MOD_ID,"sifter");
     public static final ResourceLocation FLORIST_ID        = new ResourceLocation(Constants.MOD_ID,"florist");
     public static final ResourceLocation ENCHANTER_ID      = new ResourceLocation(Constants.MOD_ID,"enchanter");
+    public static final ResourceLocation RESEARCH_ID       = new ResourceLocation(Constants.MOD_ID,"researcher");;
```
Asherslab: should be `RESEARCHER_ID`

src/main/java/com/minecolonies/coremod/client/gui/WindowHutUniversity.java
```diff
@@ -0,0 +1,79 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.minecolonies.api.util.constant.TranslationConstants.*;
+import static com.minecolonies.api.util.constant.WindowConstants.*;
+
+/**
+ * Window for the lumberjack hut.
```
Asherslab: review the JavaDoc in this class. "Lumberjack"

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingUniversity.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * ClientSide representation of the building.
+     */
+    public static class View extends AbstractBuildingWorker.View
+    {
+        /**
+         * Milk Cows or not.
+         */
+        private boolean milkCows = true;
```
Asherslab: you seem to have left some methods from the cow herder building in here

src/main/java/com/minecolonies/coremod/research/GlobalResearchFactory.java
```diff
@@ -0,0 +1,89 @@
...
+import com.minecolonies.api.util.constant.TypeConstants;
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.nbt.ListNBT;
+import net.minecraftforge.common.util.Constants;
+import org.jetbrains.annotations.NotNull;
+
+import static com.minecolonies.api.research.util.ResearchConstants.*;
+
+/**
+ * Factory implementation taking care of creating new instances, serializing and deserializing RecipeStorages.
```
Asherslab: this is not for recipes

Asherslab (changes requested):
Overall, i'd like to see the string IDs changed to ResourceLocations, it's one of the few things i wish we'd done for the Jobs and Buildings, makes mods adding them easier

src/api/java/com/minecolonies/api/research/IResearchManager.java
Raycoms: Only one colony holds its own local tree

src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
Raycoms: yeah, it's all lowercase so we can load the values from here.

src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
Raycoms: No

src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
Raycoms: This is for an entire different usecase, it's so that we can serialize itemStorages better.
Nothing related directly to research.
We still will need this, always.

src/api/java/com/minecolonies/api/research/IGlobalResearch.java
Raycoms: I want to avoid resource location where they are not strictly necessary, it's a lot of code and storage overhead and extra comparisons etc needed somewhere where we only expect strings to be added. There is no "different mod require different locations" necessity here.

src/api/java/com/minecolonies/api/research/IGlobalResearch.java
Raycoms: same thing here. Resourcelocations are only useful when we want to localize it for our mod and we're afraid another mod uses the same id.

src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
Asherslab: okay not research related. i thought it was

src/api/java/com/minecolonies/api/crafting/ItemStorageFactory.java
marchermans: Still part of the api...

marchermans (changes requested):
This looks better.
Give it another go over, when it comes to java doc and the classes which you added to the api and then we should be good.

Asherslab (dismissed):

marchermans (approved):

PR #4370, opened 2020-02-25
Bump forge and structurize requirements

Nightenom:
Forge has a fix for patching since 31.1 so bump it too

Review please

### DIFF
--- a/src/main/resources/META-INF/mods.toml
+++ b/src/main/resources/META-INF/mods.toml
@@ -6,7 +6,7 @@
 # The name of the mod loader type to load - for regular FML @Mod mods it should be javafml
 modLoader="javafml" #mandatory
 # A version range to match for said mod loader - for regular FML @Mod it will be the forge version
-loaderVersion="[30,)" #mandatory (27 is current forge version)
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/mods.toml
```diff
@@ -38,7 +38,7 @@ MineColonies is a Colony/Town Simulator that adds many structures and NPC worker
     # Does this dependency have to exist - if not, ordering below must be specified
     mandatory=true #mandatory
     # The version range of the dependency
-    versionRange="[31,)" #mandatory
+    versionRange="[31.1.0,)" #mandatory
     # An ordering relationship for the dependency - BEFORE or AFTER required if the relationship is not mandatory
```
Raycoms, line 42: this requires 31.1.0 and not 31.1, afaik

Nightenom: [x,) stays for x and anything higher
basically x <= version

Raycoms: ah okay,

PR #4375, opened 2020-02-26
Adds stacktraces to error-level issues.

someaddons:
Adds stacktraces to error-level issues. Should help for stuff like the building is null issue. 
Also removed a case where we returned an existing colony on an attempt to create a new one, as followup logic is certainly aimed at a new colony and would corrupt the existing one.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/WindowBuildBuilding.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/WindowBuildBuilding.java
@@ -249,7 +249,7 @@ public class WindowBuildBuilding extends AbstractWindowSkeleton
             }
             else
             {
-                Log.getLogger().error("WindowMinecoloniesBuildTool: Need to download schematic on a standalone client/server. This should never happen");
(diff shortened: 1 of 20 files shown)


marchermans (approved):

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildBuilding.java
```diff
@@ -249,7 +249,7 @@ private void updateResources()
             }
             else
             {
-                Log.getLogger().error("WindowMinecoloniesBuildTool: Need to download schematic on a standalone client/server. This should never happen");
+                Log.getLogger().error("WindowMinecoloniesBuildTool: Need to download schematic on a standalone client/server. This should never happen", new Exception());
```
Asherslab, line 252: Could we possibly create a custom exception? just because "Exception" is pretty bland XD.

someaddons: hm don't think that matters, since the log message appears above the stacktrace so you see its minecolonies and which error

PR #4385, opened 2020-03-01
Inventoryutils issues

someaddons:
Fixes a multitude of inventory issues, from dupes to lost stacks.
Few cases of unhandled remainders, bugged comparisons etc.

These are all bugfixes for the current inventory utils, though I'll probably rework the class as whole aswell later, squishing it down to usecases and avoiding duplicate implementations/unused stuff

Review please


src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -874,14 +874,27 @@ public static boolean addItemStackToItemHandler(@NotNull final IItemHandler item
                 slot = itemHandler.getSlots() == 0 ? -1 : 0;
                 while (!ItemStackUtils.isEmpty(resultStack) && slot != -1 && slot != itemHandler.getSlots())
                 {
-                    resultStack = itemHandler.insertItem(slot, resultStack, false);
+                    resultStack = itemHandler.insertItem(slot, resultStack, true);
                     if (!ItemStackUtils.isEmpty(resultStack))
                     {
                         slot++;
                     }
+                    // result empty, we can insert the whole stack
```
Raycoms, line 882: and if we were able to insert a partial stack first and then the rest?
This should have worked fine up to now.

src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -1473,7 +1486,7 @@ public static boolean transferItemStackIntoNextFreeSlotInItemHandler(
                                                                            final int sourceIndex,
                                                                            @NotNull final IItemHandler targetHandler)
     {
-        ItemStack sourceStack = sourceHandler.extractItem(sourceIndex, Integer.MAX_VALUE, true);
+        ItemStack sourceStack = sourceHandler.extractItem(sourceIndex, Integer.MAX_VALUE, false);
```
Raycoms: better then just the method that returns a certain stack at the right index.

src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -1485,19 +1498,11 @@ public static boolean transferItemStackIntoNextFreeSlotInItemHandler(
             sourceStack = targetHandler.insertItem(i, sourceStack, false);
             if (ItemStackUtils.isEmpty(sourceStack))
```
Raycoms: return isEmpty here. No need for anything else after this succeeded.

src/api/java/com/minecolonies/api/util/InventoryUtils.java
someaddons: resultStack is the remainder already for the partial insert, which gets re-used in the loop till its empty or the loop failed. So when its empty we can fit the whole. But can probably simplify this a lot later.

The issue with this was that the return value lost information,  doing a partial insert combined with true/false return either leads to some information loss, or to having to modify the stack parameters size and hoping there is followup logic which is really hacky. So this is now all or nothing

marchermans: This is flawed logic. There are inventories from which you can't extract and not insert.

src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -1473,7 +1486,7 @@ public static boolean transferItemStackIntoNextFreeSlotInItemHandler(
                                                                            final int sourceIndex,
                                                                            @NotNull final IItemHandler targetHandler)
     {
-        ItemStack sourceStack = sourceHandler.extractItem(sourceIndex, Integer.MAX_VALUE, true);
+        ItemStack sourceStack = sourceHandler.extractItem(sourceIndex, Integer.MAX_VALUE, false);
```
marchermans: You are violating the IITemHandler contract here!

Always simulate first before performing any operation

marchermans (changes requested):
Please reevaluate your changes and possibly old code.
Because in a lot of places you violate the IITemHandler contract! First simulate then actually execute.

You might want to figure out an elegant way to always do this with any operation. Kind of like a wrapper around IITemHandler against which you code. Which then executes a simulation first and asks if this is the desired result before actually committing the change.

src/api/java/com/minecolonies/api/util/InventoryUtils.java
someaddons: you're saying there are inventories which take a not-empty stack into insert and then return an empty stack without beeing able to insert it?

src/api/java/com/minecolonies/api/util/InventoryUtils.java
someaddons: where do I find that IITemHandler contract? in the documentation I only see that input stacks are not to be modified, thats all

src/api/java/com/minecolonies/api/util/InventoryUtils.java
marchermans: Yes

src/api/java/com/minecolonies/api/util/InventoryUtils.java
marchermans: It is not really a contract.
It is the general rule that was setup when they where defined:

First test if you actually can modify the inventory
Then actually the modification when you are happy with it.

someaddons: ah I see, so for that we always want to do extract simulate -> insert simulate -> extract real -> insert real?

src/api/java/com/minecolonies/api/util/InventoryUtils.java
someaddons: hm but I think if insert with simulate works then insert without simulate should work aswell?

src/api/java/com/minecolonies/api/util/InventoryUtils.java
someaddons: btw I saw that forge(or maybe vanilla?) does insert simulate -> set stack for inventories

marchermans: Both work really.
Important is that you do a simulate first.
Set stack just overrides.

Forge does it in vanilla cause insert simulate -> set stack basically is the same.

marchermans: Personally I would create something like an Inventory interaction handler
Which records somehow the interactions thgat you want to do.
THen returns the simulated results, and allows you to tell it to commit the change.

someaddons: ye that would be nice to have, kinda like DB transactions

marchermans: Yes exactly.
Something that has the exact same signatures as IItemHandler (Except maybe the simulate flag) or something.

Raycoms: Shouldn't we able to code this in a utility function somehow?

someaddons: ye should be doable, although I'd leave that for the rework, since this was just supposed to be bugfixes

PR #4390, opened 2020-03-04
exp fixes

### LINKED ISSUE #4389: No experiance from raids
The issue is that raids don't drop any exp.
Found the relevant code, and it seems to be a bug. 
in addition the set exp value for the raiders is very low (1), maybe it should be higher:
1-small slime
5-zombie
10-blaze

can make pull request if you guys want it fixed.

adoxentor:
Closes #4389

-raids drop exp
-When guard is following you, gives you effect without particles
-Fixed infinite loop when citizen earns too much exp too quickly

Review please


src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -940,6 +941,10 @@ private void increaseLevel()
             {
                 this.levelExperienceMap.put(job.getExperienceTag(), new Tuple<>(entry.getA() + 1, entry.getB()));
             }
+            else
+            {
+                this.levelExperienceMap.put(job.getExperienceTag(),new Tuple<Integer,Double> ((int)MAX_CITIZEN_LEVEL,ExperienceUtils.getXPNeededForNextLevel(MAX_CITIZEN_LEVEL-1)));
+            }
```
Raycoms, line 947: Why is this needed?

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
```diff
@@ -915,7 +915,7 @@ public void setPlayerToFollow(final PlayerEntity player)
         if (this.getColony().getWorld() != null)
         {
             this.getColony().getWorld().getScoreboard().addPlayerToTeam(player.getScoreboardName(), new ScorePlayerTeam(this.getColony().getWorld().getScoreboard(), TEAM_COLONY_NAME + getColony().getID()));
-            player.addPotionEffect(new EffectInstance(GLOW_EFFECT, GLOW_EFFECT_DURATION_TEAM, GLOW_EFFECT_MULTIPLIER));
+            player.addPotionEffect(new EffectInstance(GLOW_EFFECT, GLOW_EFFECT_DURATION_TEAM, GLOW_EFFECT_MULTIPLIER,false,false));//no reason for particales
```
Raycoms, line 919: Good catch, you can remove the comment here.

src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenExperienceHandler.java
```diff
@@ -210,8 +211,16 @@ public void gatherXp()
...
+            double d1 = vec3d.lengthSquared();
+
+            if (d1 < 1.0D) {
+                addExperience(orb.getXpValue() / 2.0D);
+                orb.remove();
+                return;
+            }
+            double d2 = 1.0D - Math.sqrt(d1) / 8.0D;
+            orb.setMotion(orb.getMotion().add(vec3d.normalize().scale(d2 * d2 * 0.1D)));
         }
```
Raycoms, line 224: Awesome

src/api/java/com/minecolonies/api/entity/mobs/AbstractEntityMinecoloniesMob.java
```diff
@@ -280,17 +281,6 @@ public ILivingEntityData onInitialSpawn(
...
-    protected void onDeathUpdate()
-    {
-        if (!(this.getAttackingEntity() instanceof PlayerEntity) && (this.recentlyHit > 0 && this.canDropLoot() && world.getGameRules().getBoolean(GameRules.DO_MOB_LOOT)))
-        {
-            final int experience = ExperienceOrbEntity.getXPSplit(BARBARIAN_EXP_DROP);
-            CompatibilityUtils.addEntity(world, new ExperienceOrbEntity(world, this.posX, this.posY, this.posZ, experience));
-        }
-        super.onDeathUpdate();
-    }
-
```
Raycoms, line 293: This is very old code, didn't know there was a new way around this. Good job

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
adoxentor: it fixes a bug that when a guard collects too much exp it goes into infinite loop.

Raycoms: Where does this infinite loop come from?

adoxentor: https://github.com/ldtteam/minecolonies/blob/3f0ac6352d78fbfb93a4ec1f808c8e11152e4fea/src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenExperienceHandler.java#L131-L134

So when level up fails the loop continues, my fix is deleting the excess exp. anther idea is make all of the call tree return bool value so the loop can stop and handle the issue in the top level.

Raycoms: You added * returns true if level up succeeded, but no return value btw =D

someaddons: this fix work indirectly though, the while's citizenData.getLevel queries the value and the while loop breaks

Raycoms: ahhh

someaddons (approved):

PR #4393, opened 2020-03-06
Raiders rework for 1.15

someaddons:
Closes #
Closes #
Closes #

Originally this was to fix loading/unloading/counting issues with the raiders, but ended up reworking it a bit and creating a new event system which can be used for future colony events as well.

Whats new: We now have ColonyEvents, which are registered through a forge registry for them, so other mods can add events aswell. Colonyevents are managed through a new EventManager, which updates them and saves/reads them from nbt. Those events receive a regular tick from the colony,albeit pretty slow(each 500 ticks) to update their status and do whatever update logic they need to. Colony events can also register/unregister entities related to them and spawn temporary structures. For managing the structures the EventManager has an EventStructureManager. The structure manager allows spawning structures for a given event and takes a backup schematic of the surroindings. Once the event is removed(or was not loaded during colony load) the backup schematic is put back in place.

I've translated both the raider and pirate events over as colonyevents, which also allows us now to spawn multiple ones seperatly if we want to. Atm I've added that once a raid succeeds the biggest raid size it gets split up into an additional raid from another spawn pos. 

Both raid events now keep track of their entities and respawn them if those were not actually killed(onDeath triggered), but removed by other means, e.g. minecraft itself removing it, this also includes chunk unloads. The respawn is limited to the previous position, or a max distance of the original spawn position, so that when the colony is still active, but the raiders happen to be unloaded it tries to place them a bit closer in a loaded chunk.

The PirateEvent now has a maximum amount of pirates it can have active, depending on the amount of spawners remaining(spawnercount * 2 +10). It also now spawns additional pirates,depending on shipsize, without a player beeing close to the spawners, so that it always raids the colony.

The BarbarianEvent has a given Horde(as previously) which has a predetermined size of archers/melees/bosses. It actively keeps track of all and respawns any missing. Any additional mobs which try to register over the Horde's counter are killed off, this e.g. might happen with chunk unloading/loading.

Barracks got a little new feature, now that we can have more than one raid from different angles, which gives the raid mobs a glow effect so they are easier to find. This is called "Spies" and uses gold to hire those spies  into the enemy lines which then lets the players know the position.

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4393/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #4409, opened 2020-03-09
Fix postbox shape, fix crusher crafting

### LINKED ISSUE #4223: Collision box of postbox is a full block even so the postbox is only a half block
### Minecolonies version
Version: minecolonies-0.10.616-ALPHA-universal

### Expected behavior
It would be nice if the postbox's collision box matches it's visual appearance

### Actual behaviour
It's collision box is a full block which causes one to collide with it when it appears you could easily walk past it.

1. Place it next to a building
2. Walk past the building
3. ... stub your toe on the postbox

Raycoms:
Closes #4223 

- Fix postbox shape
- Fix crusher crafting on request
-

Review please


someaddons (dismissed):

src/main/java/com/minecolonies/coremod/blocks/huts/BlockPostBox.java
```diff
@@ -11,6 +17,11 @@
  */
 public class BlockPostBox extends AbstractBlockHut<BlockPostBox>
 {
+    private static final VoxelShape SHAPE_NORTH = Block.makeCuboidShape(0.0D, 0.0D, 0.0D, 16.0D, 16.0D, 8.0D);
```
marchermans, line 20: Use one and rotate on demand. Less to maintain in case it changes.

marchermans (changes requested):
Minor thingy.... But still... After that merge it.

PR #4421, opened 2020-03-11
Schematics115

lunadragnstar:
Fix hospital symbols


pixar02:
Could you check the wooden blacksmith 5 out. Some blocks are missing (above the arch at the patio)

PR #4442, opened 2020-03-15
Updating the PathFinding Heuristics.

marchermans:
This PR updates the path finding heuristics to be more concise and much more efficient.

### A* Background
In A* (The path finding algorithm used by us) the next evaluated node is determined by its node-value, called the f-value. This f-value is composed out of two factors. The c-value, known as the cost. This is the cost it would take an entity to walk from the starting position of the path (usually the current position of the entity) to the node in question. And the second factor is the h-value, known as the heuristic. This value is an indication for the algorithm, and indicates how much it would cost an entity using the path to walk from the current node to a destination. 
Now if the h-value is not accurate a lot of nodes will have the same f-value and as such the algorithm has to consider a lot of them, if it wants to make progress. As such the correct heuristic is very important.

### Current situation:
To calculate the heuristic we currently use: ` -(startEndDist - destination.manhattanDistance(pos))` this heuristic is not very usable in a 3D world, since it produces a lot of the same f-values, making the system search more blocks then it needs to. Also using the manhattan distance causes the citizens to walk in L-Shaped routes. Not going actually the shortest route: which would be diagonally.

### New situation:
In the new situation we use the squared euclidian distance, which produces a much more accurate heuristic h-value across the board. Additionally this now actually causes the algorithm to produce pathing results which go diagonally.

### Screenshots before:
image
image

### Screenshots after:
image
image


someaddons (dismissed):
looks good

src/api/java/com/minecolonies/api/util/DebugUtil.java
```diff
@@ -0,0 +1,15 @@
...
+public final class DebugUtil {
+
+    private DebugUtil() {
+        throw new IllegalStateException("Tried to initialize: DebugUtil but this is a Utility class.");
+    }
+
+    public static boolean hasReleaseMouse() {
+        Minecraft.getInstance().mouseHelper.ungrabMouse();
+        return true;
+    }
```
Raycoms, line 26: You want this in too? Then we should probably add some doc

Issue #4595, opened 2020-04-07
[Feature] more compostable stuff


ToMe25:
The composter can already compost alot of stuff the Vanilla composter can't handle, i would however suggest adding support for a few more things, some of which the Vanilla composter can compost and some more, that kinda just feel missing, as why should for example some seeds not be compostable?
Here the list of things i would suggest to make compostable:

- Pumpkin Seeds
- Melon Seeds
- Pumpkins?
- Carved Pumpkins?
- Grass
- Tall Grass
- Kelp?
- All 6 types of Leaves
- Sea Grass?
- Cactus
- Sugar Cane
- Vines
- Cocoa Beans?
- Fern
- Large Fern
- Sunflower
- Lilac
- Peony
- Lily Pad
- Sea Pickle?
- Wheat
- Brown Mushroom Block
- Red Mushroom Block
- Mushroom Stem
- Cake?
- Rabbits Foot?
- Fermented Spider Eye

The items on the list with a question mark are the ones i am not sure about, but felt like i had to include them for different reasons.

someaddons:
you can add additional composter items with the config

ToMe25:
I feel like atleast the other tall flowers should be compostable by default tho.

PR #4601, opened 2020-04-08
Fix duplicate config option luckyblocks and add compostables

ToMe25:
- Fix the luckyores config option being called luckyblocks thereby overriding the reald luckyblocks config option
- Add the suggested compostables from #4595 to the default config

If you think any of the added items shouldn't be compostable, or i shouldn't mess with that default value please let me know.

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4601/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #4665, opened 2020-04-17
don't add itself as provider

### LINKED ISSUE #4664: Ticking Entity Crashing the Game
### Minecolonies Version

Version: minecolonies-1.12.2-0.11.804-RELEASE-universal.jar / Twitch Launcher

### Expected behavior
My builder upgrading the farmers hut and the citizens requests not causing the game to crash.

### Actual behaviour
When I first stumbled upon the ticking entity error I realized that it was because of a certain citizen that would request an item. So I killed the citizen using a command and the problem went away until 5 minutes later another citizen was the entity causing the game to crash again. What I've come to conclude is that every time a worker would request something, immediately the game would crash and the worker with the request would be the ticking entity error that causes the crash. I've experimented with different workers by affecting them to cause a request and the system would crash, and I would see in the crash report that they would be the actual cause of the entity ticking error.

After the game crashes it takes me back to the launcher with this message : java.lang.StackOverflowError: Ticking entity

1. Have a builder working on an upgrade.
2. Make sure the builder does not have all the items to do the upgrade so he requests.
3. The game automatically crashes within seconds of the request because of the entity tick error.
4. ... Profit

Note: I've tried going to a back up colony save and the same problem occurs. I have restarted Minecolonies twice now due to crashing bugs that I couldn't solve unless I restarted the entire world.
### Logs: 
- latest.log: https://gist.github.com/ralphydez/cc8223aca4f09a4277ce48834a965065
- crashlog: https://gist.github.com/ralphydez/cad4a033ae42a05bba739aee7537922d

Raycoms:
Closes #4664

- Fix the provider
-
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/tileentities/TileEntityColonyBuilding.java
+++ b/src/api/java/com/minecolonies/api/tileentities/TileEntityColonyBuilding.java
@@ -455,6 +455,7 @@ public class TileEntityColonyBuilding extends AbstractTileEntityColonyBuilding i
                                        .map(world::getTileEntity)
                                        .collect(Collectors.toSet()));
                     providers.removeIf(Objects::isNull);
+                    providers.removeIf(p -> p instanceof TileEntityColonyBuilding && ((TileEntityColonyBuilding) p).pos.equals(pos));
                 }
 
                 final List<IItemHandler> handlers = providers.stream()


marchermans (approved):

ralphydez:
Hi, sorry im new to this whole thing,
what did you mean by don't add itself as provider?
Do I do that? and how?

Raycoms:
Nah, you don't worry

ralphydez:
Oh okay, do I just wait for the next update I assume?
Thank you again btw for your help!

Raycoms:
ye

PR #4694, opened 2020-04-21
Sound rework

Raycoms:
This is the final and a major sound overhaul.

We now generate sounds ids dynamically for all workers.
The sound IDs are all in the final sounds.json file.

Those are all set to point to a set of default sounds for male/female now.

Mods and resourcepacks can override them selectively if they opt to.
I will prepare a resourcepack with all our existing sounds already.

This deletes 1601 lines of codes as a tradeoff of extra json lines we will need anyway eventually.

Review please


src/api/java/com/minecolonies/api/colony/jobs/ModJobs.java
```diff
@@ -77,4 +77,45 @@ private ModJobs()
     {
         throw new IllegalStateException("Tried to initialize: ModJobs but this is a Utility class.");
     }
+
+    /**
+     * Get a list of all jobs.
+     * @return the list.
+     */
+    public static JobEntry[] getJobs()
```
marchermans: No use the registry to get all your registered jobs.
No static stuff.

src/api/java/com/minecolonies/api/sounds/BarbarianSounds.java
```diff
@@ -16,5 +16,6 @@
      */
     private BarbarianSounds()
```
marchermans, line 17: Why does this class still exist?

marchermans (changes requested):

src/api/java/com/minecolonies/api/sounds/BarbarianSounds.java
Raycoms: They are completely different sounds which are independent of the citizen sounds.

src/api/java/com/minecolonies/api/colony/jobs/ModJobs.java
Raycoms: I tried that, how do I do that without getting the placeholder and shit?

Raycoms: btw, I only want minecolonies jobs, because other mods have to register their own sounds

marchermans: You can do both perfectly fine. Filter on the minecolonies namespace and just filter out all jobs you do not care for. 

_Sent from my Honor 7X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

marchermans (dismissed):
It is completely possible todo this without a static array of jobs. Even if you just want the minecolonies ones. So please remove the static list.

PR #4753, opened 2020-04-25
Claim adjustment and some view fixes

someaddons:
Adapted claiming to no longer seperate into modes, but instead the config settings determine how it behaves now. We keep our 3 configs settings for
Initial claim size (size when creating a colony)
Max claim size (max distance where claims are allowed from the colony center)
Min colony distance (min distance at where colonies are allowed to place)

"Static Mode" now is setting initial and claim to the same value, "Dynamic mode" is otherwise.
The initial chunks are not unclaimed when a building is broken, so you no longer end up with a half-claimed colony as you do currently.

Fix: Old colony views on client now get removed when a colony is deleted, so they can create a new one.
Fix: Views not always send to all receivers as intended, when e.g. the building manager is dirty AND there are new subscribers.
Fix: Workorders no longer get sent as single message each to each player and not using dimension along colony id
Fix: Buildings claim area missing level 4 and defaulting to 0
Fix: Barracks claim area is calculated using its avg towerlevels, but not beeing calculated when a tower was built instead of the barracks itself
Fix: Barracks claiming a chunk at level 0
Fix: On placing a building a view is now send immediately instead of with a delay
Fix: Citizens swimming so slow that they'd get stuck with flowing water

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4753/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

marchermans (dismissed):
Nice!

marchermans:
Once the build completes

marchermans:
Then you can merge

someaddons:
thanks :) ye seems like I messed up some javadoc

marchermans:
Jep
With the rapid growth and the fact that people are starting to use the api/mod for plugins i wanted that to be nice and validated.

Hence it now testing for it.

someaddons:
ye think its a good idea to validate it, just gotta get used to running javadoc to check compiling :)

PR #4759, opened 2020-04-25
Fix Stock View sometimes not showing all items

### LINKED ISSUE #4758: Stock window not showing all items if multiple stacks of the same item are in the same rack
### Minecolonies version
Version: 0.11.880

### Expected behavior
I expect the Stock window to show the amount of all items of the same type added together.

### Actual behaviour
If more than one stack(doesn't have to be a full stack) of the same type of item is in the same rack, it only counts one of the stacks.
Under some circumstances it also shows more items of a type than there are, but i can't reproduce that yet.

1. Build a warehouse(i tested it with nordic and fortress level 2)
2. distribute a stack of items into multiple slots of the same rack
3. go to the stock window in the warehouse block
4. see only some of your items

- latest.log:[ByrpvEJA](https://pastebin.com/ByrpvEJA)
- crashlog: Not a Crash.

ToMe25:
Closes #4758
Closes #
Closes #

- Fix item handling in StockView to prevent it from only showing the amount if items from one stack per Item type per rack.
- Add the Hut block inventory to the stock view.
-

Review please


Raycoms:
Stack has a max size of 64, and for some things even less, you'll need to store this in an itemStorage object to fix this.

ToMe25:
> Stack has a max size of 64

In theory maybe, but it does work like this as can be seen here:

All these Hut blocks are in the same rack.

lexustec:
It should count together all Stacks.. so it shows ONLY one Itemtype per Stack.. so You know exactly how much You have in a glance

It is not a chest view where You have stacks it is like the View in Refined storage

ToMe25:
I moved the Fix to the TileEntityRack class as that seems to be the less hacky solution, tell me if i should moce it back.
> It should count together all Stacks

i know, it still does that, the problem before was that it only counted some stacks.

lexustec:
You mean it counted the Stacks twice ?

ToMe25:
No.
lets say you have two stacks of cobble in one rack, then it counted only one of them.
If you however have two stacks of cobble in different racks it counts both.

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -225,12 +225,14 @@ public void updateItemStorage()
...
+            final ItemStack stackCopy = stack.copy();
+            final ItemStorage storage = new ItemStorage(stackCopy);
             int amount = ItemStackUtils.getSize(stack);
             if (content.containsKey(storage))
             {
                 amount += content.remove(storage);
             }
+            stackCopy.setCount(amount);
             content.put(storage, amount);
         }
```
Raycoms:
```
int amount = ItemStackUtils.getSize(stack);
if (content.containsKey(storage))
{
amount += content.remove(storage);
}
content.put(storage, amount);
```
Why isn't this enough?

lexustec:
Okay ,.... i tried that btw.. this is funny maybe i lost sth while having issues with my GIT

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
ToMe25: because [this](https://github.com/ldtteam/minecolonies/blob/84fa0aaf06052529fb9c664f27a6f715c3b2bb2b/src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java#L150) only uses the stack, not the amount.
both in the current, and my changed version.

lexustec: that should work normally

Raycoms: it gets the amount from the stack and adds it to the amount in the map. That should be enough

ToMe25: Should i change it back to how it was in the second commit then? that didn't change TileEntityRack at all.

Raycoms: none of this should be necessary of the above.

ToMe25: does that mean you prefer leaving this alone like it was, and just changing the WindowHutAllInventory?

Raycoms: The above code is correct. It should not be the cause of any issue.

ToMe25: ok i changed it back.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java
```diff
@@ -147,7 +147,9 @@ private void updateResources()
 
                 for (final Map.Entry<ItemStorage, Integer> entry : storage.entrySet())
                 {
-                    items.add(new ItemStorage(entry.getKey().getItemStack(), entry.getValue(), false).getItemStack());
+                	ItemStack stack = entry.getKey().getItemStack().copy();
```
marchermans: This does not work officially, and even though it works in the current Mc version it is not officially supported. As such please change this to use itemstorage instances or use the proposed change by @Raycoms.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java
ToMe25: should i just change the items list to use ItemStorage instead of ItemStack then?

Raycoms:
okay, I was looking over all of this again and I think there are things we can improve:

https://github.com/ldtteam/minecolonies/blob/3f60f2e9d059d2f00f6d923b273936db8c7faba9/src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java#L169

Instead of filling a list and then filling the map in my link. We can move this map up to the beginning of the class and count up in this map from the beginning. This should also solve this issue.

ToMe25:
OK i will do that.

lexustec:
It happened after i changed the code from the original aproach with the new map   

```
if (rack instanceof TileEntityRack)
            {
                 int allSlots = ((AbstractTileEntityRack) rack).getInventory().getSlots();
                        Log.getLogger().info("Used Slots " + allSlots);
                        for (int i = 0; i < allSlots; i++)
                        {
                            if (!((AbstractTileEntityRack) rack).getInventory().getStackInSlot(i).isEmpty())
                            {
                                items.add(((AbstractTileEntityRack) rack).getInventory().getStackInSlot(i));
                            }
                        }


            }

```

that was the old aproach..

ToMe25:
@Raycoms you mean like this?

lexustec:
the change of allItems will break the view

ToMe25:
Again if absolutely no change to TileEntityRack is wanted, i can undo that and instead get the value of the map in WindowHutAllInventory and set the amount there.

ToMe25:
> the change of allItems will break the view

allItems? i am not messing with anything names allItems.

lexustec:
sorry i read it wrong

src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java
ToMe25: I did get rid of it now.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java
```diff
@@ -135,7 +135,7 @@ private void updateResources()
     {
         final List<BlockPos> containerList = building.getContainerList();
 
-        final List<ItemStack> items = new ArrayList<>();
+        final Map<ItemStorage, ItemStorage> storedItems = new HashMap<>();
         final World world = building.getColony().getWorld();
```
Raycoms, line 139: Yes better, and make this a Map<ItemStorage, Integer> and increment the quantity in the. And keep ItemStorage the same

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -231,6 +231,7 @@ public void updateItemStorage()
             {
                 amount += content.remove(storage);
             }
+            storage.setAmount(amount);
             content.put(storage, amount);
```
Raycoms: this should not be necessary.

ToMe25: ok i will remove it.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java
```diff
@@ -158,29 +165,25 @@ else if (rack instanceof ChestTileEntity)
...
-                storedItems.put(currentStorage, currentStorage);
-            }
-        }
-
-        final List<ItemStorage> filterItems = new ArrayList<>(storedItems.keySet());
+        final List<ItemStorage> filterItems = new ArrayList<>();
+        storedItems.forEach((storage, amount) -> {
+            storage.setAmount(amount);
+            filterItems.add(storage);
+        });
```
ToMe25, line 182: Is like this OK, or do i have to find another way to set the amount of the ItemStorage?

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
ToMe25: I removed it.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java
ToMe25: Did that.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutAllInventory.java
Raycoms: Should be fine like this.
Did you test this?

ToMe25: Yes i did.

ToMe25: If you want to suggest any more changes, i will look at it tomorrow, today its late enough :)

Raycoms: Looks fine for me now. Does this solve all the issues there were before?

Can you add an additional check to query the rack from the position of the building itself (it's not include din the list of containers, so it needs an additional line to add the stacks from it.

ToMe25: "Does this solve all the issues there were before?" idk, i don't know about many issues with this view, i just made this to fix the linked issue.
"Can you add an additional check to query the rack from the position of the building itself" I can do that, but i think lexustec said he is working on that, so i wouldn't put that in here.

ToMe25: "Can you add an additional check to query the rack from the position of the building itself"
I have a working version for the warehouse now, i will test it with some other buildings, and than hopefully push it.

lexustec:
saw it ..... nice fix :-)

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4759/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

ToMe25:
For some reason this currently isn't working for vanilla chests, i will work on fixing that now.

marchermans (dismissed):

ToMe25:
I looked at the problem with the chests some more, and i think there are three solutions for it, but i am not sure which one to use.
The solutions:
1. Ignore that this isn't working with chests as it didn't before this PR either
2. Create a custom packet to transfer the chest inventory to the client to read it
3. I am not sure, but maybe it would be possible to (mis?)use a vanilla packet to transfer the chest contents

Please tell me what the prefered solution would be

Raycoms:
That sounds weird, the chest contents should also exist on the client side, don't they?

ToMe25:
I haven't found anything that is able to get them clientside without a custom packet, and the clientside tileentity doesn't contain any items even when the chest is full.

Raycoms:
@OrionDevelopment did you know that?

ToMe25:
I also thought that is weird, but i couldn't find anything that is able to read this clientside.

marchermans:
No this is the normal state. All vanilla inventories do not sync their inventory. Only when a container is opened the container contents are synced.

ToMe25:
So should i just leave the chest problem be, or make a packet for that?
I don't think there is a vanilla packet for this.

Raycoms:
I'd say then remove the vanilla support code and leave it be. And we tell our schematic authors to not use chests at all anymore =)

ToMe25:
ok, but i think no chests will be a loss for the mod, as just one type of storage just always has places where it looks wrong.

lunadragnstar:
We primarily use racks, but sometimes we use chests and barrels in builds

ToMe25:
barrels already don't fully work, i think about making a pr for that tho...
but i probably wont as this doesn't supports chests now anyways.

lunadragnstar:
Is this specifically for the warehouse, or for all storage

ToMe25:
is what for the warehouse or all storage?

Raycoms:
@lunadragnstar it's for all buildings.

Raycoms:
@OrionDevelopment what is your opinion on this?

marchermans (approved):
Not the greatest of solutions. But alright.

PR #4797, opened 2020-04-29
Fix builder not seeing the new air types as air

### LINKED ISSUE #4795: Builder breaks cave air
### Minecolonies version
Version: 0.11.893

### Expected behavior
I expected the builder to ignore the new Cave Air block just like regular air and continue breaking blocks.

### Actual behaviour
If the miner encounters Cave Air in a place where he is meant to build something he will at the start in the block breaking phase break the cave air, which doesn't make that much sense.

1. Go into a cave
2. create a build job
3. hope that this cave is actually filled with cave air, and not regular air
4. see the builder breaking air

- latest.log:
- crashlog:

ToMe25:
Closes #4795
Closes #
Closes #

- Fix the builders handling of the new air types
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/util/StructureIterator.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/util/StructureIterator.java
@@ -226,10 +226,10 @@ public class StructureIterator
             case CLEAR:
                 return advanceBlocks(this.theStructure::decrementBlock,
                   structureBlock -> structureBlock.doesStructureBlockEqualWorldBlock()
-                                      || structureBlock.worldBlock == Blocks.AIR);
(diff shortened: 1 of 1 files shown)


Raycoms:
Was doing this concurrently: https://github.com/ldtteam/minecolonies/pull/4798
But I think the instanceof AirBlock is probably more generic

ToMe25:
ok seems like Raycomms is doing this but everywhere, closing this then.

Raycoms:
I was about to look into the builder water issues. If you want you can take that over?

ToMe25:
I don't know how to correctly handle water as it has to get removed somehow, but not like this, so no.

PR #4828, opened 2020-05-01
free build tool in recipes

Raycoms:
- Workers should not request build tool for craft request "that's free"
-
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -1,6 +1,7 @@
 package com.minecolonies.api.crafting;
 
 import com.google.common.collect.ImmutableList;
+import com.ldtteam.structurize.items.ModItems;
 import com.minecolonies.api.colony.requestsystem.token.IToken;
 import com.minecolonies.api.util.CraftingUtils;
 import com.minecolonies.api.util.InventoryUtils;
@@ -86,7 +87,7 @@ public class RecipeStorage implements IRecipeStorage
 
         for(final ItemStack stack: input)
         {
-            if(ItemStackUtils.isEmpty(stack))
+            if(ItemStackUtils.isEmpty(stack) || stack.getItem() == ModItems.buildTool)
             {
                 continue;
             }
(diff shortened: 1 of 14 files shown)


Asherslab:
we have the com.minecolonies.api.crafting.RecipeStorage#checkForFreeSpace method, is it not possible to use that to detect if *any* item is free? (instead of hardcoding for BuildTool)

Raycoms:
No, because a bucket of milk consumes the milk, that is different than what we want. (We still need 10 buckets of milk and not only 1)

Asherslab:
@Raycoms we could do something similar though? don't just check if the stack returned is empty but also check if the stack returned is identical to the one fed in (if returned identical, then we're good)

Raycoms:
Wondered about that too, but actually I wanted the buildtool to be completely free actually. Like, doesn't even need a request for it.

Asherslab:
Ahh, well could you also add that for any other items?

Raycoms:
Yeah, but I don't want to add this for any other item atm.

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -138,6 +139,10 @@ public boolean canFullFillRecipe(final int qty, @NotNull final IItemHandler... i
 
         for (final ItemStorage stack : items)
         {
+            if (stack.getItem() == ModItems.buildTool)
```
Asherslab: not sure this is necessary after your changes to `calculateCleanedInput` no?

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -238,6 +243,11 @@ public boolean fullfillRecipe(final List<IItemHandler> handlers)
 
         for (final ItemStorage stack : getCleanedInput())
         {
+            if (stack.getItem() == ModItems.buildTool)
```
Asherslab: same as previous

Asherslab (commented):
looks good otherwise

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
Raycoms: This is so that it doesn't get "discounted" because the tool is not in the inventory

Raycoms: Ah wait, we call cleaned input, you're right

Raycoms: I'll adjust it

Asherslab (approved):

someaddons (approved):

PR #4861, opened 2020-05-03
Fix javadoc

ToMe25:
- This PR mainly adds some missing javadoc.
- The only problems javadoc generation tools should be able to find after this are(what is actually found depends on the tool):
    1. Some javadoc implementations wont be able to handle link tags with nested classes.
    2. Some javadoc implementations will say something about html problmes, mainly with > chars.
- This pr also includes one(i think, maybe two) method return type getting changed from boolean to void.

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4861/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #4914, opened 2020-05-07
Fixed incorrect job level calculation

mcmanustfj:
Citizen job level should be primary + secondary/4, currently is (primary+secondary)/4.


Raycoms:
This was kinda on purpose, to make sure that the previous system (1-50 levels) is equivalent to the new system (2 - 2 * 99 levels). A better calc would probably be ((2xprimary+secondary)/3)/2

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=39549&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=39557&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

mcmanustfj:
y'all are welcome to balance it however you please, but this way you'll be balancing intentionally and not  based off a bug.  You might also consider doing primary/2+secondary/4. (or 2/3*p, 1/3 * s, or whatever)
I will say that with the bug fixed, buying people with diamonds is actually worth it.

Raycoms:
The code how it is currently in the PR would cause a lot of issues because the return value could go above 50, and that would issues with the current system. So, whichever of those is fine as long as it keeps it under 50 =P

mcmanustfj:
if the max level is 50, why is `MAX_CITIZEN_LEVEL` 99?  I guess that's used for skills?
If stats can go over 50 but job levels shouldn't, maybe something like doing `min(50, p+s/4)` would work?
While it makes sense now why you did (p+s)/4 (as that will max out at 50), I haven't ever seen stats get that high in any of my long-running saves, so it seems pretty difficult to ever get near that cap.  

Or maybe I'm just bad at the game, in which case something that maxes out at 50 when the two skills are at 99 makes sense.

Raycoms:
Yeah, it needs quite a while but after some times and high building/home levels, they start leveling fairly fast.

This "level" is currently more like a combined level of both stats. In the future we will remove this and use the stats separately to influence certain behaviour (Walking speed, building space, block breaking speed, crafting speed) etc. But for now we need the level 50 max to make sure things work without "divided by 0 exceptions".

So either we got the x+y/4 (will max at 50) or we will actually do a (x*2+y)/6 (which would then take the primary more seriously)

PR #4925, opened 2020-05-07
Feature/stripped lumberjack

JiaYow:
### Feature Suggestion
The new (awesome) Asian building style requires a fair bit of stripped wood.
Stripped variants of woods cannot be crafted, but must be made by right-clicking woods/logs with an axe. Which becomes old really quickly when the buildings require hundreds of them.

It makes sense for the lumberjack to be able to craft stripped versions of logs.

### Draft implementation
1. Lumberjack turns vanilla wood/log to stripped versions if a request for this exists in the request pipeline.
2. The lumberjack hut's GUI shows the list of recipes (i.e. the stripped versions), but you can't remove them and you also can't teach new recipes.

### Open considerations
1. Turning wood into its stripped version does not cost anything (but the original wood) currently. This is because I have no idea how to damage the axe properly (I looked at AbstractEntityAllInteract.java:234, and I don't understand how that would do anything). We could just make it cost 2 logs per stripped version :-P
2. It would be preferable if this was gated behind a university research. I have not found a way to update the building's recipe list onResearchCompleted (this event doesn't seem to exist). Therefore, one would have to reload the chunk/restart the server after researching to enable the feature. Which is worse than not gating the feature at all imho. How about linking this to the building level instead? There's an onUpdate-event for that.

Please review (thoroughly! I touched a lot of code in there with which I have no experience yet). Thanks :-)


src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowWorkerBuilding.java
```diff
@@ -96,6 +97,14 @@
 
     private String stateString = state ? DP_MODE_STATIC : DP_MODE_AUTOMATIC;
 
+    /**
+     * Defines whether or not the recipes of this buildings are read-only.
+     * Crafters with intrinsic recipes for example would override this to return true.
+     * @return Boolean stating if the teach-recipes button should be shown and Remove should be available.
+     */
+    protected boolean hasReadOnlyRecipes() {
```
Asherslab: While i agree the lumberjack shouldn't have recipes since it has normal ones, we probably shouldn't call this method `hasReadOnlyRecipes` and make it not show the recipes window, instead we should rename this to something more like `allowsRecipeModification` or something similar, in the future we may have huts that have both readOnlyRecipes and the ability to add recipes from the player

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
```diff
@@ -81,6 +82,23 @@ public WindowListRecipes(final IColonyView c, final BlockPos buildingId)
         updateRecipes();
     }
 
+    /**
+     * Constructor for the window when the player wants to assign a worker for a certain home building.
+     * Can additionally disable remove-buttons
```
Asherslab: javadoc here is off

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
```diff
@@ -126,6 +144,13 @@ public void updateElement(final int index, @NotNull final Pane rowPane)
                 final ItemIcon icon = rowPane.findPaneOfTypeByID(OUTPUT_ICON, ItemIcon.class);
                 icon.setItem(recipe.getPrimaryOutput());
 
+                if (!isRemovalPossible) {
```
Asherslab: can we make this per recipe? So that in the future we can have both readOnly and editable ones?

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractFilterableListIntrinsicCrafter.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static com.minecolonies.api.util.constant.BuildingConstants.CONST_DEFAULT_MAX_BUILDING_LEVEL;
+import static com.minecolonies.api.util.constant.NbtTagConstants.TAG_ID;
+
+/**
+ * Abstract class for all buildings which require a filterable list of allowed items.
+ */
+public abstract class AbstractFilterableListIntrinsicCrafter extends AbstractFilterableListBuilding
```
Asherslab: While the name is okay, i can already imagine myself and others getting confused by this name when searching for the class(es) that does this stuff
Can we rename the `Intrinsic` in these classes to Predefined or something similar?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
```diff
@@ -209,6 +215,47 @@ public Class getExpectedBuildingClass()
         return BuildingLumberjack.class;
     }
 
+    /**
+     * The lumberjack is a special worker.
+     * In his/her decision state, he/she will try to add lumberjack cycles
+     * If there's nothing left to craft, he/she will proceed with woodworking
```
Asherslab: maybe just use "they" here in the javadoc, or "it" lol

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
```diff
@@ -209,6 +215,47 @@ public Class getExpectedBuildingClass()
...
+     * @return the next state to go to.
+     */
+    protected IAIState decide()
```
Asherslab, line 227: This is missing @override

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
```diff
@@ -314,10 +361,6 @@ private IAIState findTree()
                                  copy.getOrDefault(SAPLINGS_LIST, Collections.emptyList()),
                                  worker.getCitizenColonyHandler().getColony());
             }
-
-            // Delay between area searches
-            setDelay(100);
-            return getState();
```
Asherslab, line 320: can you please make sure this isn't being run every tick or whatever now?

Asherslab (changes requested):
Overall good logic, 
the entire PR needs formatting done to it (brackets all over the place/etc)
Also please check over all your methods and add `@override` to the appropriate ones, i noticed a few missing but could have missed more

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=39565&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

Asherslab:
In response to the Considerations:
1. I see no reason to add extra cost, i feel like that's just a "benefit" of a crafter
2. I don't feel that gating this behind the university is a good idea, this is an extremely simple crafting thing, and given that most styles that use normal logs don't have such a restriction we shouldn't add that restriction to this single style

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLumberjack.java
```diff
@@ -95,6 +105,32 @@ public BuildingLumberjack(final IColony c, final BlockPos l)
         super(c, l);
 
         keepX.put(itemStack -> ItemStackUtils.hasToolLevel(itemStack, ToolType.AXE, TOOL_LEVEL_WOOD_OR_GOLD, getMaxToolLevel()), new Tuple<>(1, true));
+
+        if (recipes.isEmpty()) {
+            addStrippedWoodRecipe(Items.OAK_LOG, Items.STRIPPED_OAK_LOG);
```
Asherslab, line 109: I haven't looked at the Log code and etc but i would assume there's some way, either in the block or otherwise, to get the stripped version of a log from a log. Doing this would add better support for modded logs

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
JiaYow: I think this is a fragment of older code. These lines will now fall through to line 377 (the return), which will change the state to LUMBERJACK_NO_TREES_FOUND, which has a delay built-in.

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
JiaYow: How so? (Just for my information though. Since we're changing this to per-recipe, I'll rework this anyway)

Asherslab: Sorry, not that exact line, one above. 
"the player wants to assign a worker for a certain home building."

JiaYow: Ah lol, I just brain-afk copypasted it from the old constructor. Will fix.

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
Raycoms: You'd have to add this to RecipeStorage and then also add this to the serializer/deserializer this flag.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLumberjack.java
Raycoms: Agreed on that. We should probably do something like Items.Tags.Logs.getItems().forEach(addRecipe)

ravenbuilder934:
This is probably too late, but I kind of think the miller (carpenter) should craft stripped logs instead of the lumberjack because they already craft other wood things. Then the lumberjack could focus on chopping wood. Maybe keep the lumberjack though, since making stripped wood does require an ax.

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
JiaYow: Are RecipeStorages reused/shared using the RecipeManager? Is that maybe planned for performance reasons at one point? If so, coding readOnly into the recipe is bound to break something at some time (i.e. teaching the same recipe to two different huts). In that case, I would add builtInRecipes into the workerhuts, which would run in parallel with the current "playerRecipes". This would also align with the other suggestion about allowing/disallowing playerRecipes together with builtIn-recipes in the future. If recipes are never reused, I'll just add this flag to the recipe?

Raycoms: They are reused on a global level. The question is, do we want this recipe to be ever "alterable" ?

JiaYow: We're talking about the ability to remove the recipe from the building. I.e. we don't want the stripped recipes to be removed from the lumberjack. In extension, the same is true for the fletcher and the dyer btw.
But, if I happen to teach someone a recipe that another building has built-in (maybe that's something you guys are planning to use more in the future, say the sawmill can automatically and unremovably do the plank-recipe), then I should be able to remove that recipe from the building again.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLumberjack.java
JiaYow: Tbh, I can't find anything. The only mention of that in forge is in the item.AxeItem, and there it's also done manually:
`   protected static final Map<Block, Block> BLOCK_STRIPPING_MAP = (new Builder<Block, Block>()).put(Blocks.OAK_WOOD, Blocks.STRIPPED_OAK_WOOD).put(Blocks.OAK_LOG, Blocks.STRIPPED_OAK_LOG).put(Blocks.DARK_OAK_WOOD, Blocks.STRIPPED_DARK_OAK_WOOD).put(Blocks.DARK_OAK_LOG, Blocks.STRIPPED_DARK_OAK_LOG).put(Blocks.ACACIA_WOOD, Blocks.STRIPPED_ACACIA_WOOD).put(Blocks.ACACIA_LOG, Blocks.STRIPPED_ACACIA_LOG).put(Blocks.BIRCH_WOOD, Blocks.STRIPPED_BIRCH_WOOD).put(Blocks.BIRCH_LOG, Blocks.STRIPPED_BIRCH_LOG).put(Blocks.JUNGLE_WOOD, Blocks.STRIPPED_JUNGLE_WOOD).put(Blocks.JUNGLE_LOG, Blocks.STRIPPED_JUNGLE_LOG).put(Blocks.SPRUCE_WOOD, Blocks.STRIPPED_SPRUCE_WOOD).put(Blocks.SPRUCE_LOG, Blocks.STRIPPED_SPRUCE_LOG).build();
`

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
Raycoms: Tbf, that feature we're talking about would add it automatically to all and to new buildings. However the player would still be allowed to remove it.

(Similarly in the glassblower it is intentionally allowed to remove sand (to avoid requesting red sand to the warehouse).

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLumberjack.java
Raycoms: How does vanilla determine that one block can be made to the stripped version by using an axe on it? Hardcoded?

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
JiaYow: If that's the case, I'll just hide the teach-button on lumberjacks (if you even want that) and leave the rest alone? Or even better, I could refunction the teach-recipe to reteach the built-in recipes. How about that? 
BTW, adding to existing buildings didn't work with this method for me, but that might be because of the single player dev world ..

Raycoms: Yeah, I'd be fine with hiding the remove and teach button at the lj. There is nothing we'd have to teach the lj

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLumberjack.java
JiaYow: If I understand correctly, it simply uses the map I posted above:
`
Block block = BLOCK_STRIPPING_MAP.get(blockstate.getBlock());
      if (block != null) {
         PlayerEntity playerentity = context.getPlayer();
         world.playSound(playerentity, blockpos, SoundEvents.ITEM_AXE_STRIP, SoundCategory.BLOCKS, 1.0F, 1.0F);
`
Unfortunately, that map is protected and there's no Getter that I can see.
I could subclass a new AxeItem, and put a Getter for the map in there. Hacky though xD

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
JiaYow: So:
Option 1: Globally hide teach+remove on Lumberjacks, like it's done in the PR now.
Option 2: Show both teach+remove, like Fletcher+Dyer, but if the player removes the recipe, he/she has to rebuild the thing to recover.
Option 3: Refunction the Teach-Button to reset the built-in recipes to the initial state.

Admin's choice :)

Raycoms: I'd say option one, but the naming is probably bad that's why Asher complained. Better name it:

allowRecipeAlterations()

Raycoms: in the building that gets synched to the view

src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowWorkerBuilding.java
JiaYow: Renamed to `isRecipeAlterationAllowed()` (based on Raycom's suggestion + inspection) in d230c91.
Also fixed the brackets.

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
JiaYow: Fixed in d230c91

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
JiaYow: Renamed in d230c91, kept teach/recipelist logic as was previously, but added logic to hide the recipe-list-button too if there's simply no recipes available for this hut (no built-in + no way to teach).

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractFilterableListIntrinsicCrafter.java
JiaYow: Renamed in 7541035 to `AbstractFilterableListCrafter`, because there was really no reason to restrict this new type to only built-in recipes. Private+Public crafters that also show filterable lists are valid configurations. The lumberjack will now refuse to accept recipes himself. **HOWEVER** this means that the `AbstractFilterableListCrafter` now completely mirrors the code of `AbstractBuildingCrafter`. This is horrendeous code duplication, but I have no idea how this could be fixed without either diamond inheritance (won't do) or refactoring the whole system to use a composition-style pattern (out of the scope of this PR, honestly). If you have an idea, please let me know. Otherwise, it is what it is :-/

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
JiaYow: Changed in 7541035

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
JiaYow: Fixed in 7541035

JiaYow:
@ravenbuilder934: Having the sawmill do this was discussed a couple of times, but rejected. Reasons: Sawmill is gated behind university, but stripped logs should be accessible early. Also, as you said, lumberjack has an axe.

Devs, while testing this in Single Player, I crashed the request system a few times, and I can't pinpoint it. It seemed like there was a requesting loop or smth, and when I removed the dman-building, my console got spammed with "Request token not registered to ...". The dman would be unfireable (would stay a dman and basically just freeze). Trying to kill him crashes the game. I am unsure if this is because of my code, the dman's code, the postbox, or whatever. I have seen this exact thing on other servers already though, so I "hope" this is not my fault. Do you maybe know about this?

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=39641&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

ravenbuilder934:
> @ravenbuilder934: Having the sawmill do this was discussed a couple of times, but rejected. Reasons: Sawmill is gated behind university, but stripped logs should be accessible early. Also, as you said, lumberjack has an axe.
> 
Okay. What do you mean by sawmill is gated behind university?

JiaYow:
@ravenbuilder934 : You need to research the sawmill in the university before you can build it. It's not very difficult, but still a bigger obstacle than the other colony styles have right now.

ravenbuilder934:
@JiaYow Oh, the version I'm playing doesn't have the university, so I could just build the sawmill-that's why I was confused. Yeah, it makes sense for the lumberjack to do it.

src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowWorkerBuilding.java
```diff
@@ -148,7 +162,14 @@ private void changeDPState()
 
     private void recipeListClicked()
     {
-        @NotNull final WindowListRecipes window = new WindowListRecipes(building.getColony(), building.getPosition());
+        if (!isRecipeAlterationAllowed() && building.getRecipes().isEmpty()) {
```
Raycoms: bracket format is still an issue

src/main/java/com/minecolonies/coremod/client/gui/WindowHutLumberjack.java
```diff
@@ -77,12 +77,24 @@ public WindowHutLumberjack(final BuildingLumberjack.View building)
...
+    /**
+     * Defines whether or not players can change the building's recipe list.
+     * This is the case for most current buildings, but some buildings might only work on built-in recipes.
+     * It's recommended to turn this off for buildings that make no use of player-thaught recipes, to avoid confusion for new players.
+     * Turning this on will hide the "Teach recipes" button, hide the remove-buttons in the recipe list,
+     * and also hide the recipe list altogether if no recipes are present.
+     *
+     * @return true if player is allowed to alter  recipes, false if not
+     */
+    @Override
```
Raycoms: overriden method doesn't need the javadoc then. (javadoc only at origin method)

Raycoms: This should also not be a window function. The building should synchronize the value in the view. So that we can ask the building here

Raycoms: (Only require to override one function to achieve this behaviour and not 2)

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractFilterableListCrafter.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+        }
+
+        final Map<Predicate<ItemStack>, Tuple<Integer, Boolean>> toKeep = new HashMap<>(keepX);
+        toKeep.putAll(recipeOutputs.entrySet().stream().collect(Collectors.toMap(key -> (stack -> stack.isItemEqual(key.getKey().getItemStack())), Map.Entry::getValue)));
+        return toKeep;
+    }
+
+    @Override
+    public boolean canCraftComplexRecipes()
```
Raycoms, line 131: I guess we don't need those methods here (canCraft, canRecipeBeAdded, openCraftingcontainer)

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingLumberjack.java
Raycoms: I checked with forge and that's the only way to do it.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
```diff
@@ -243,7 +295,7 @@ private IAIState prepareForWoodcutting()
     {
         if (checkForToolOrWeapon(ToolType.AXE))
         {
-            return getState();
+            return START_WORKING; // Reset everything, maybe there are new crafting requests
```
Raycoms: comment on previous line please (also for line 314)

Raycoms:
Almost ready, only added some minor comments

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractFilterableListCrafter.java
JiaYow: The `AbstractFilterableListCrafter` inherits from `AbstractFilterableListBuilding`, not from `AbstractBuildingCrafter`. (The FilterableList one is tighter coupled with messages, so it seemed less error-prone to me to do it this way). Removing the aforementioned methods would lock this type out from complex recipes (and require subclasses to duplicate openCraftingContainer), which is not necessarily what we want. I'd leave it in for now - this whole thing (Abstract class for every type + their combinations) needs to be reworked in the long run anyway imho.

Raycoms: Ah okay, while it might not be necessary for the current lj, others might need it, true.

src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowWorkerBuilding.java
JiaYow: oh my world I've used the other style for years, this is killing me.
Fixed in c49066c.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutLumberjack.java
JiaYow: I removed all the override-JDocs in the files that I touched. Also, you are absolutely right about the window function, it's moved to the building now. Still need two overrides to keep it clean (source of truth in the building ->serialized to the view), but that's much better than before.
Fixed in f1bab0e.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractFilterableListCrafter.java
JiaYow: I did realize that `openCraftingContainer` was already useless ('cause 1:1 copy of its parent) in the original `AbstractBuildingCrafter`. I removed it from both there and from the new `AbstractFilterableListCrafter`. Fixed in f1bab0e.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java
JiaYow: Fixed in f1bab0e.

JiaYow:
Fixed a bug in 4b53f2a: Crafters only get gathered while they have tasks, which prevented the normal LJ loot from being gathered. It's hardcoded to `always gather` now, like normal gatherers do. Means, the dman might sometimes steal stuff from the LJ, but I have too little insight in the request system to prevent that from happening. In all likelihood, I would break something if I tried. The current solution might delay requests because the dman might have to redeliver resources, but generally works.

Raycoms:
Before I merge this, the most important question: Did you test this with the newest changes and does it work right?

JiaYow:
I've merged one more time with upstream and tested it. In my "laboratory setting" it works nicely. The only "obvious" problem is that the dman steals his stuff if he isn't back from logging in time to start crafting. The dman will bring him replacements though, so you will get the stripped items eventually.

But, given that I'm quite new to the code base and in the end had to touch quite a few things I was not familiar with, I think it would be best, this once, if someone else could test this too, so I can sleep better. I had hoped that this would be a smaller addition (like adding the recipe in the constructor as in the dyer), and I'd like to avoid crashing someone's colony because I didn't know I had to call onCraftingSuccess() manually or something.

Thanks!

someaddons (approved):
Looks nice

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/4925/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

marchermans (approved):

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=39954&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

PR #4968, opened 2020-05-11
Allow certain requests to have a minimum count

Raycoms:
This PR basically replaces the previous handling where specific workers were able to make requests that were partially fulfilled with a request specific handling. Thus worker can request something with a count and a min count, where the RS will try to fulfill the count and if this is not possible but the system has a bit less it will try to match the minCount.

I also removed some of the heavy streams with quicker for loops.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/IDeliverable.java
+++ b/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/IDeliverable.java
@@ -8,7 +8,6 @@ import org.jetbrains.annotations.NotNull;
  */
 public interface IDeliverable extends IRetryable
 {
-
     /**
      * Method called to check if a given stack matches this deliverable.
      * The first stack that returns true from this method is returned as a Deliverable.
@@ -21,9 +20,16 @@ public interface IDeliverable extends IRetryable
     /**
      * Method called to get the amount of items that need to be in the stack.
      *
-     * @return The amount of items that
+     * @return The amount of items.
      */
     int getCount();
+
+    /**
+     * Method called to get the minimum amount required to fulfill this request.
+     *
+     * @return the minimum amount.
+     */
+    int getMinimumCount();
 
     /**
      * Method to get the result of the delivery.
--- a/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Stack.java
+++ b/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Stack.java
@@ -20,6 +20,8 @@ public class Stack implements IDeliverable
     private static final String NBT_MATCHNBT    = "MatchNBT";
     private static final String NBT_MATCHOREDIC = "MatchOreDic";
     private static final String NBT_RESULT      = "Result";
+    private static final String NBT_COUNT       = "Count";
+    private static final String NBT_MINCOUNT    = "MinCount";
     ////// --------------------------- NBTConstants --------------------------- \\\\\\
 
     @NotNull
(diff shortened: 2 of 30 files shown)


LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=39906&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Stack.java
```diff
@@ -28,50 +30,58 @@
...
+    private int count;
+
+    /**
+     * The required count.
+     */
+    private int minCount;
 
     @NotNull
-    private ItemStack result = ItemStackUtils.EMPTY;
+    private ItemStack result;
```
marchermans, line 56: Please add the Empty initialization back. It just prevents errors and is just a safeguard no reason to remove it.

src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Stack.java
```diff
@@ -153,10 +153,16 @@ public static Stack deserialize(final IFactoryController controller, final Compo
         final boolean matchNBT = compound.getBoolean(NBT_MATCHNBT);
         final boolean matchOreDic = compound.getBoolean(NBT_MATCHOREDIC);
         final ItemStack result = compound.keySet().contains(NBT_RESULT) ? ItemStackUtils.deserializeFromNBT(compound.getCompound(NBT_RESULT)) : ItemStackUtils.EMPTY;
-        final int size = compound.getInt("size");
-        final Stack theStack = new Stack(stack, matchMeta, matchNBT, matchOreDic, result);
-        theStack.setCount(size);
-        return theStack;
+
+        int count = compound.getInt("size");
```
marchermans, line 157: Constant check.

src/api/java/com/minecolonies/api/colony/requestsystem/requestable/IDeliverable.java
```diff
@@ -21,10 +20,17 @@
...
+     * @return The amount of items.
      */
     int getCount();
 
+    /**
+     * Method called to get the minimum amount required to fulfill this request.
+     *
+     * @return the minimum amount.
+     */
+    int getMinimumCount();
```
marchermans, line 32: Please add a setter as well. Additionally use the given count as the default minimum count not 1.

If the builder then wants to override his special behaviour can then set the minimum count to what ever is handy.
Setting the default to one for every request type changes the behaviour.

marchermans (changes requested):
Implementation wise this is okey.
Api-wise not so much. The api of `IDeliverable` should have a setter for the default value.
By default the getMinimumCount should just return getCount() and the setter should noop.

If a requestable supports minimum counts then it can override the getter and possibly the setter and return a storable value.

Additional again by default the getter should return the count of the request unless the setter is invoked and is used to change the minimal value.

src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Stack.java
Raycoms: I fixed up the logic, now all constructors always initialize it correctly such that I'm getting an intellij warning too.

src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Stack.java
Raycoms: On purpose, this will be retired (old value) as soon as possible.

src/api/java/com/minecolonies/api/colony/requestsystem/requestable/IDeliverable.java
Raycoms: I wanted to avoid the previous behaviour we had with the setters all over the place. Thus, you initiate this with the value in the constructor (like all the other values) or you make a new one. No need for a setter here.

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=39975&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

marchermans:
Server is not starting up properly
Something with a recipe bug?

marchermans (approved):

PR #4973, opened 2020-05-11
Allow buildings to opt-out from deliveryman

JiaYow:
### Feature Suggestion
Currently, the deliveryman just randomly picks from all gatherable buildings.
This PR adds the option to opt-out from the deliveryman. The building will be left alone, except for when pickup requests (i.e. crafting complete) are sent by the building itself.
Be careful though, you now have to clean up after the building yourself.

### Implementation
1. Additionally to "Static" and "Automatic", there is now a "Pickup: Never" option. Click it.
2. Slightly changed the GUI of the buildings to better accomodate the new feature.
3. Selecting "Pickup: Never" will hide the Priority numbers / decrementors/incrementors to avoid confusion.

image

image

### Open Considerations
I had to change the current "isPriorityStatic()"-boolean to an enum to fit the third state. This means I had to modify the NBT-data where this is stored. I have written a mini-migration/failsafe into the code, so on NBT-load the current data should be loaded, and then on chunksave the new format should be serialized. Since screwing this up would mean a bunch of trouble, please take special note of `AbstractBuildingContainer:104` ff.


marchermans (dismissed):

src/api/java/com/minecolonies/api/colony/buildings/PickUpPriorityState.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Creates the enum from an int. Used for message deserialization.
+     * This is not DRY, but we're talking about 3 to maybe later 4 values here.
+     *
+     * @param intRepresentation The internal representation of the enum.
+     * @return The corresponding enum.
+     */
+    public static PickUpPriorityState fromIntRepresentation(int intRepresentation)
```
Raycoms, line 38: Why do we need the int representation, can't we pick the ordinal?

j-gruber-fraiss: Because enum ordinals in Java are not meant to be used in that way. I mentioned it in line 10 of the enum.

There's a couple of reasons for this, see https://stackoverflow.com/a/44654410/886407.

JiaYow: Because Java's ordinals are not meant to be used for such scenarios.
There's a couple of reasons for this, see https://stackoverflow.com/a/44654410.

Raycoms: Yeah, it definitely would lead to backwards compatibility problems if you insert values inbetween. Due to laziness all of our enums use the ordenal for persistency. 

I wonder if it is possible to abstract this a bit so that it could be used anywhere basically (so that we can avoid adding this boilerplate to all our enums).

JiaYow: Backwards compatibility isn't even my main concern. [This is.](https://www.tutorialspoint.com/java/lang/enum_ordinal.htm)
A pointer to the enum will return 0, not the value you would expect. This will undoubtedly cause a bug along the way.

As for boilerplate, there might be a way by using enum types inheritance and/or generic shenanigans. However, I feel this is slightly out of the scope of this PR^^

Raycoms: Sounds okay, although, whoever will create an enum next, will have to extract this to avoid copying the code just over.

Raycoms:
@JiaYow just missing the conflicts now

JiaYow:
This weird stripped lumberjack PR conflicted with mine. Resolved it :-P

Raycoms:
Okay, I'll have dinner in a bit, after the deploy of the test server ran through I'll merge this.

LDTTeam-Buildsystem:
This PR has been automatically built and hosted by our build system, during this process there was a crash, error, or otherwise, please check on this at: https://teamcity.minecolonies.com/viewLog.html?buildId=40004&buildTypeId=LetSDevTogether_Minecolonies_PullRequests_2_DeployTestServer

PR #5003, opened 2020-05-15
fix #4947

### LINKED ISSUE #4947: Guards stucks, maybe by glass panes
### Minecolonies version

Minecolonies version 0.11.933-ALPHA - but happens since months
Minecraft version 1.15.2
Forge version 3.11.77

### Expected behavior

Guards not stucks for ages

- 

### Actual behaviour

I have 21 guards in my colony and always the 4 guards in the screenshots stay stuck at the same place. If I recall them to the guard tower, the patrol again and stay stucks again in the same place, jumping for ages in front of my window made of glass panes.
I suspect that the glass pane is the reason.

- 

### Proposition to avoid the problem

As citizens are often stucks (for ex in fences, some slabs, or other), and in order not to have them killed by mobs because they try to find their path instead of attack, would it be possible to follow this kind of algorithm:
- If a villager is jumping in the same place for more than 10s, recall it automatically to its bed or hut

Raycoms:
Closes #4947

- Improves handling when too close to a glasspane/etc
-
-

Review please


someaddons (dismissed):

marchermans (approved):
Generally this is okey. However PVI is basically ready and this should be in there.
So I would rather not merge it, but if it is a critical bugfix you might want to port this to PVI as well.

Raycoms:
It's a rather critical one, so we want this in both for sure.

PR #5016, opened 2020-05-16
Adds lantern recipe to mechanic

PowerSchill:
Signed-off-by: Mark E. Schill <Mark.Schill@cmschill.net>

Closes #
Closes #
Closes #

-
-
-

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingMechanic.java
```diff
@@ -109,7 +109,8 @@ public boolean canRecipeBeAdded(final IToken token)
               || storage.getPrimaryOutput().getItem() == Items.JACK_O_LANTERN
               || (storage.getPrimaryOutput().getItem() instanceof BlockItem && ((BlockItem) storage.getPrimaryOutput().getItem()).getBlock() instanceof HopperBlock)
               || Tags.Items.STORAGE_BLOCKS.contains(storage.getPrimaryOutput().getItem())
-              || storage.getPrimaryOutput().getItem() == Items.ENCHANTING_TABLE)
+              || storage.getPrimaryOutput().getItem() == Items.ENCHANTING_TABLE
+              || storage.getPrimaryOutput().getItem() == Items.LANTERN)
```
marchermans: Good catch.... However this is kind of way to complicated.
We are now way past the "we can do this with select- or if-switches"-case and I would rather see a proper registry in the API for this, then these (what feels like 20 but is more like) 6 if switches.

I know this is a minor change, but all in all this is going a bit too far. Talk to @Raycoms or me @OrionOnline on discord to figure out what exactly we would need here.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingMechanic.java
Raycoms: I don't think we can solve this too easily.

Probably the only thing we can do is create a

"canBeCraftedByWorkerX" tag that we then assign in the data to all the blocks we want. But that would also be a mayor pain to handle so I want to avoid this for now.

I told him he can add it like this for now to the mechanist.

Raycoms: It's in general far out of the scope of this PR.

marchermans: Actually I think that is a great idea
Why is this out of scope?

PowerSchill: Wouldn't the tag option just be moving the same rules to a single class that is loaded at startup or is there an alternative idea to the rules themselves?

Raycoms: We would have to create 20 tags and apply them to 100 items, that's why it's auto of scope for the PR.

(1 tag per worker + 1 tag for each of the items the worker accepts).

Then, we'd have to check through all the workers and check if they have "single block" accepts somewhere and add tags accordingly.

marchermans: So in my eyes this would be 1 Tag per worker.
Not 1 Tag per worker + 1 tag per Item.

Raycoms: 1 tag per worker, and then we have to add this tag to all the items we want it in.

marchermans: that is a file per worker....

Raycoms: No, that is a file per vanilla item to have a tag afaik.

marchermans: No it is a file per tag. You list each item that you want to be part of the tag in that file.

Easy as pie. For example there is a file for forge:ingot which lists all ingots.

Raycoms: @PowerSchill can you for a test set up a tag for the Mechanic and add those items to it?

PowerSchill: So would this be a single class that loads on startup and iterates through all allowed results and add a tag that is checked for by the mechanic?

Raycoms: Check how datapacks handle item tags. Then you create a new item tag for the mechanist, and apply it to all the items he should be able to craft.

PowerSchill: Just did a quick read on datapacks and item tags. The downside is that it would be static and require manual update. For example if I didn't want to have to code this by hand I would want to create a script to create all of the json files. `storage.getPrimaryOutput().getItem().getRegistryName().getPath().contains("ice")`

Raycoms: Yes, those would be the exceptions that would stay behind because tags don't handle this as well.

The building would check for a) Own tag and b) Some of these exceptions.

But the tags allow packmakers to add more items to it, so that definitely does make sense to have.

PowerSchill: Hopefully this is what you had in mind.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingMechanic.java
```diff
@@ -91,40 +92,45 @@ public Skill getSecondarySkill()
     @Override
     public boolean canRecipeBeAdded(final IToken token)
     {
-        if(!super.canRecipeBeAdded(token))
+        ResourceLocation builder_products = new ResourceLocation("minecolonies", this.getJobName().toLowerCase().concat("_product"));
+        ResourceLocation builder_ingredients = new ResourceLocation("minecolonies", this.getJobName().toLowerCase().concat("_ingredient"));
+
```
Raycoms, line 97: Those have to be registered somewhere I believe and then be constants there, but the rest looks correct ys.

PowerSchill: It works with it the way it is unless you are pointing out something I am missing.

Raycoms: @OrionDevelopment I'm not 100% sure, wasn't there a registering steps with custom tags too?

marchermans (approved):
This looks much better. We should do this from now on out..

As far as I know this is fine. But we should ask the Forge guys.

Raycoms:
I double checked with forge and this seems to be fine. As a continunation. Could you set this up for all crafter workers? Maybe add two methods to the abstractBuildingCrafter to get those two resource tags instead of having to create them in each of the methods (to avoid duplicate code).

PR #5058, opened 2020-05-22
Feature/recipe tags

### LINKED ISSUE #5131: Plantation worker should be able to produce sugar
### Minecolonies version

Version: Head

### Expected behavior

- Plantation worker should be able to craft sugar from sugarcane, as he is also able to craft paper and books.

### Actual behaviour

- Sugar recipe is not teachable.

### LINKED ISSUE #5160: [Suggestion] Allow stone smelter to smelt netherrack into nether bricks
Version:  minecolonies-1.15.2-0.11.957-universal.jar

### Expected behavior

Allow stone smelter to smelt netherrack into nether bricks

### Actual behaviour

Currently cannot teach netherrack smelting recipe 

1. Teach stone smelter this recipe
2.  automate building of netherbrick-heavy builds
4.  ...Profit

### LINKED ISSUE #5161: [Suggestion] Allow stonemason to craft flower pot recipe
Version:  minecolonies-1.15.2-0.11.957-universal.jar

### Expected behavior

Allow stonemason to craft flower pot recipe from three bricks.
only one brick less than the brick block recipe.
useful as there are some blueprints with a lot of flower pots
Still seems relevant to his job

### Actual behavior

cannot currently teach stonemason flower pot recipe, 

1. Allow stonemason to delve into pottery
2. Get flower pots
3. ... Profit

PowerSchill:
Closes #5161
Closes #5160
Closes #5131

-  Introduces a system for adding and excluding recipes based on datapack tags
-
-

Review please


PowerSchill:
Fixed a few recipes but it should be good for review now.

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5058/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

Raycoms:
Have you tested this through with most of the crafters?

PowerSchill:
Yeah, I loaded up and built a test colony and went through all of the crafters and focused especially on the each of the items I specified in the tags. There may be something I missed but I made sure to test anything I could think of.

PowerSchill:
Also need to work on documentation for the wiki related to the datapacks and I will do some additional testing with other forge mods.

Raycoms:
great so :shipit:

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBaker.java
```diff
@@ -276,6 +279,12 @@ public void onColonyTick(@NotNull final IColony colony)
     @Override
     public boolean canRecipeBeAdded(final IToken token)
     {
+
+        ResourceLocation builder_products = new ResourceLocation("minecolonies", this.getJobName().toLowerCase().concat("_product"));
```
marchermans: Wrong naming variables.

marchermans: This logic can be abstracted into a protected method in the abstract building: canRecipeBeAddedBasedOnTags(final IToken token) which can then be reused through out the buildings even when the recipe addition logic is changed.

Please rework this as such, since this is a lot of copy pasting.

marchermans (changes requested):
More abstraction is needed. Extract the validation based on tags into a seperate protected method and invoke where needed. Potentially use a callback for finer control.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingCrafter.java
```diff
@@ -173,4 +176,55 @@ public static boolean canBuildingCanLearnMoreRecipes(final int buildingLevel, fi
     {
         return (Math.pow(2, buildingLevel) * EXTRA_RECIPE_MULTIPLIER) >= (learnedRecipes + 1);
     }
+
+    /**
+     * @param token
+     * @return whether the recipe can bee added based on tokens.
+     */
+    protected Optional<Boolean> canRecipeBeAddedBasedOnTags(final IToken token)
```
Raycoms, line 178: This seems like a duplicate. And why not just make it return true/false ?

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBaker.java
```diff
@@ -276,35 +275,36 @@ public void onColonyTick(@NotNull final IColony colony)
...
 
-        for (final IRecipeStorage recipe : BakerRecipes.getRecipes())
-        {
-            if (recipe.getPrimaryOutput().isItemEqual(storage.getPrimaryOutput()))
-            {
-                return false;
-            }
-        }
+        isRecipeAllowed = super.canRecipeBeAddedBasedOnTags(token);
+        if (isRecipeAllowed.isPresent()) {
```
Raycoms: the format is messed up in all these classes though.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingCrafter.java
PowerSchill: Because we need three cases:

- True means the recipe is allowed. 
- False means the recipe is explicitly denied. 
- Null means false unless a following rule allows it.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBaker.java
PowerSchill: What do you mean?

Raycoms: lala {

}

and not 

lala 
{

}

PowerSchill: Oh, is that "messed" up? :)

Raycoms: Compared to the default formatting yeah =D

PowerSchill: Looks like this was the only one

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingCrafter.java
Raycoms: But do we need it in the crafter + in the worker override?
(Also, we never return null in the method, that's why I asked why we don't return a boolean)

PowerSchill: Actually we probably don't need it in both places; I am testing that now. The actual `canRecipeBeAdded` method always returns true or false but the `canRecipeBeAddedBasedOnTags` needs to be able to say Yes, No, or "I don't care" before letting the builder-specific logic.

Raycoms: I mean the canRecipeBeAddedBasedOnTags also atm always is true/false and never null

Raycoms: or did you want to return nullable(null) instead of nullable(false) ?

PowerSchill: I return `Optional.ofNullable(false)` which is like Null but won't generate null pointer exceptions. You have to set it to something.

Raycoms:
```
    public static <T> Optional<T> ofNullable(T var0) {
        return var0 == null ? empty() : of(var0);
    }
```

Optional.ofNullable(false) is the same as Optional.of(false), you have to give it null

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBlacksmith.java
```diff
@@ -92,47 +91,37 @@ public Skill getSecondarySkill()
...
-        final ItemStack output = storage.getPrimaryOutput();
-        return output.getItem() instanceof ToolItem ||
-                output.getItem() instanceof SwordItem ||
-                output.getItem() instanceof ArmorItem ||
-                output.getItem() instanceof HoeItem ||
-                output.getItem() instanceof ShieldItem ||
-                Compatibility.isTinkersWeapon(output) ||
-                ingots == size ||
-                nuggets == size;
+        //return false;
```
Raycoms: commented out?

Raycoms:
Looks fine, but can you remove the commented out code parts?

Raycoms:
:shipit:

marchermans (approved):

PR #5067, opened 2020-05-24
Replaces randomized deliveryman gathering with targeted pickup-requests

JiaYow:
### Feature Suggestion

Currently, the deliveryman just randomly picks from all gatherable buildings.
This PR replaces the randomized mechanism with a pickup-request system, where workers will request pickups when they deem necessary (usually after having gathered a certain amount of items).
This significantly increases the efficiency of dmen. Most of their time is now spent dumping their own inventory in preparation for delivery requests (might be also improved in the future), not randomly wandering around aimlessly.

Also, the dman hut's request page was bugfixed, it now only shows the request that are actually in progress.

### Implementation
1. Weighted randomized gathering has been removed.
2. Workers request pickups after dumping their inventory, **if** certain conditions are met.
2a. The pickup priority must not be "Never".
2b. The dump must not be part of a crafting request (otherwise the dman would visit the building twice).
2c. There must be a worker employed at the hut to make the request in the first place.
3. The pickup modes "Static" and "Automatic" have been removed. There is now a global priority that is internally treated as the initial priority for the pickup request.
4. Pickup requests are scheduled based on an [aging scheduling algorithm](https://en.wikipedia.org/wiki/Aging_(scheduling)). Simply said, requests become more important as they become older, but start with the priority set in the building. For most players, just leaving the pickup priority at 1 (default) will suffice. Players can set the priority between 1 and 10.
5. The aging algorithm can set the priority to 11, thereby making very old requests always more important than new requests.
6. When a worker's inventory is full, it will always issue a priority-10-pickup-request.
7. When a deliveryman has nothing to do, it will loiter around in the warehouse.
8. There is a "Request pickup now" button that ... requests an immediate pickup. Technically speaking, it requests a pickup at priority **12**. You're all welcome.
image

### Open Considerations

The pickup ranges are configurable and might change over time. Depending on the colony size, the aging mechanism might raise every request to the maximum of 11 very quickly. If everything is important, nothing is important :-)
We'll see ... over time.

**Please make sure the RS-reset is done correctly in the PR.**


src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAICrafting.java
```diff
@@ -355,4 +356,10 @@ private int getRequiredProgressForMakingRawMaterial()
     {
         return PROGRESS_MULTIPLIER / Math.min(worker.getCitizenData().getJobModifier() + 1, MAX_LEVEL) * HITTING_TIME;
     }
+
+    @Override
+    public boolean isAfterDumpPickupAllowed()
+    {
+        return currentRequest == null;
+    }
```
Raycoms, line 364: Ahh nice, that's what you mean with this, yes, this works fine

Raycoms:
Looks good for me :shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5067/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

JiaYow:
Fixed the first "bug". While the inventory is full, the worker would have spammed pickup requests. This will now not happen anymore. There are still cases where a couple more pickups than intended are triggered, but in this case it's better to have a bit too many than too few. It's an edge-case anyway.

Raycoms:
Oh, there are ways to check if another request is open of the same type at this hut to avoid it entirely.

JiaYow:
Oh, yep, there's a better way. Fixed.

src/api/java/com/minecolonies/api/colony/requestsystem/requestable/deliveryman/AbstractDeliverymanRequestable.java
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return priority;
+    }
+
+    @Override
+    public void incrementPriorityDueToAging()
+    {
+        // The priority set by by the aging mechanism can actually exceed the maximum priority that requesters can choose.
+        // Worst case, the priority queue turns into a FIFO queue for really old requests, with new maximum-priority requests having to wait.
+        priority = Math.min(MAX_DELIVERYMAN_AGING_PRIORITY, priority + 1);
```
marchermans, line 31: Sadly this will never work.
The Delivery can not, and will never, have control over execution order of its tasks.

The RS handles ordering and as such the DMan should never alter them, since this might mess up request-trees, if one order gets executed before another. Yes you might have it working now, but this is fragile and the RS handles this internally already.

marchermans (changes requested):
Initial review: The prioritization has to be removed.
The RS handles ordering internally so that requests that need to be executed first are assigned first.
This is the RS's prime directive and not within the skope of the AI of the DMan to decide. It has no overview over all tasks let alone in what order they need to be executed in.

Yes you might have it working now, but this is fragile any change to the RS or a specific worker might break this.

Please remove.

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/PickupRequestResolver.java
```diff
@@ -0,0 +1,172 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.text.ITextComponent;
+import net.minecraft.util.text.TranslationTextComponent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Comparator;
+import java.util.List;
+
+public class PickupRequestResolver extends AbstractRequestResolver<Pickup>
```
someaddons, line 40: missing javadoc

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/PickupRequestResolver.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public boolean canResolveRequest(@NotNull final IRequestManager manager, final IRequest<? extends Pickup> requestToCheck)
+    {
+        if (manager.getColony().getWorld().isRemote)
+        {
+            return false;
+        }
+
+        final Colony colony = (Colony) manager.getColony();
+        final ICitizenData freeDeliveryMan = colony.getCitizenManager().getCitizens()
```
someaddons, line 64: How frequently is this called? might want to use a normal loop instead of streams

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/factory/PickupRequestResolverFactory.java
```diff
@@ -0,0 +1,65 @@
...
+import com.minecolonies.api.colony.requestsystem.factory.IFactoryController;
+import com.minecolonies.api.colony.requestsystem.location.ILocation;
+import com.minecolonies.api.colony.requestsystem.resolver.IRequestResolverFactory;
+import com.minecolonies.api.colony.requestsystem.token.IToken;
+import com.minecolonies.api.util.constant.TypeConstants;
+import com.minecolonies.coremod.colony.requestsystem.resolvers.PickupRequestResolver;
+import net.minecraft.nbt.CompoundNBT;
+import org.jetbrains.annotations.NotNull;
+
+public class PickupRequestResolverFactory implements IRequestResolverFactory<PickupRequestResolver>
```
someaddons, line 16: missing javadoc in this class too

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
```diff
@@ -521,70 +355,74 @@ else if (buildingToDeliver == null)
             extracted = true;
             final ItemStack insertionResultStack;
 
-            if (iTileEntityColonyBuilding.getBuilding() instanceof AbstractBuildingWorker)
+            // TODO: Please only push items into the target that were actually requested.
```
someaddons, line 364: outdated todo? its using ((IBuildingWorker) targetBuilding)::isItemStackInRequest

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
```diff
@@ -596,50 +434,33 @@ else if (buildingToDeliver instanceof TileEntityColonyBuilding)
...
+            // The current task has changed since the Decision-state.
+            // Restart.
+            return START_WORKING;
+        }
 
-                return gatherItems(request);
-            }
+        final Delivery delivery = (Delivery) request;
+        if (InventoryUtils.hasItemInItemHandler(worker.getInventoryCitizen(),
+          itemStack -> delivery.getStack().isItemEqualIgnoreDurability(itemStack)))
```
someaddons: think you want to use an ItemStackUtils comparing function for this since afaik requests are also nbt sensitive

src/main/java/com/minecolonies/coremod/network/messages/server/colony/building/ForcePickupMessage.java
```diff
@@ -0,0 +1,62 @@
...
+import com.minecolonies.api.colony.buildings.views.IBuildingView;
+import com.minecolonies.api.colony.requestsystem.requestable.deliveryman.Pickup;
+import com.minecolonies.coremod.network.messages.server.AbstractBuildingServerMessage;
+import net.minecraft.network.PacketBuffer;
+import net.minecraftforge.fml.network.NetworkEvent;
+import org.jetbrains.annotations.NotNull;
+
+import static com.minecolonies.api.colony.requestsystem.requestable.deliveryman.AbstractDeliverymanRequestable.MAX_DELIVERYMAN_PLAYER_PRIORITY;
+
+public class ForcePickupMessage extends AbstractBuildingServerMessage<IBuilding>
```
someaddons, line 19: javadoc here too please :D

someaddons (commented):
Looks good overall, just some small comments

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
JiaYow: I love how we all fall into that particular pit. No, isItemStackInRequest doesn't filter the items. It prevents those items from being _replaced_ by forceItemStackToItemHandler (which would happen if the inventory was full).

someaddons: oh lol :D

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
JiaYow: Blame @Raycoms. I only changed the formatting here, not the logic. Ray, what do you think?

someaddons: interesting, maybe thats one of the things throwing the dman off

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/PickupRequestResolver.java
JiaYow: This is Orion's code. I'm not touching that xD
https://github.com/ldtteam/minecolonies/blob/3004c9f66cc2474053828ce3a949d6695795944a/src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/DeliveryRequestResolver.java#L51

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
Raycoms: It's like those exclude predicates in vanilla to get entities

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
Raycoms: Since he got the exact delivery from the RS, just equals should be fine here yes. Maybe having one that matches with ignore durability but doesn't match the normal way here, is really throwing him off.

Raycoms: isItemEqual I mean.

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/PickupRequestResolver.java
someaddons: hehe xD

Raycoms: We can make a PR once going through all .stream calls and calculating if its worth having a stream there or if the overhead would be really high.

src/main/java/com/minecolonies/coremod/network/messages/server/colony/building/ForcePickupMessage.java
JiaYow: Fixed in b2b9ff1.

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/factory/PickupRequestResolverFactory.java
JiaYow: Fixed in b2b9ff1.

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/PickupRequestResolver.java
JiaYow: Fixed in b2b9ff1. Also added the an equivalent comment to DeliveryRequestResolver.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
marchermans: He need to check for nbt and meta equal. Item Equality is not enough at this stage.

Raycoms: Yes, isItemEqual does check that. While isItemEqualIgnoreDurability checks less.

marchermans: Okey yeah then it needs to be isItemEqual

JiaYow: Changed in 51e3213.

JiaYow: Okay, tried it one more time with aac342d.
It now uses `ItemStackUtils.compareItemStacksIgnoreStackSize`, same as `gatherIfInTileEntity`.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
```diff
@@ -450,7 +450,7 @@ private IAIState prepareDelivery()
 
         final Delivery delivery = (Delivery) currentTask.getRequest();
         if (InventoryUtils.hasItemInItemHandler(worker.getInventoryCitizen(),
-          itemStack -> delivery.getStack().isItemEqual(itemStack)))
+          itemStack -> ItemStackUtils.compareItemStacksIgnoreStackSize(delivery.getStack(), itemStack)))
```
Raycoms, line 453: Looks fine, @OrionDevelopment let's merge this?

PR #5116, opened 2020-05-30
Slightly better handling of pickup requests

JiaYow:
1. Pickup Requests are now only triggered if the worker has actually dumped items. This should reduce the frequency of "empty" runs (happens when the worker dumps without actually having transferred items into the inventory)
2. A mechanism was added to allow arbitrary scaling of pickup priorities. For example, by setting this to a quadratic function, a priority of 1 would still be 1, but a priority of 2 would internally be handled as 4, 3 as 9 etc. This effectively slows down the aging mechanism. **In this PR, the scaling is set to the identity function, because this seems to make most sense after all.** Reason being that in a quadratic function, a prio-1 request would have to be set back 169 times by delivery requests to reach a prio of 170 (and thus be "immune" against deliveries). This seems excessive. Currently, deliveries default to priority 13, so a prio-1 request can be set back 13 times in theory. That sounds enough.
3. The pickup priority is now visible in the deliveryman's hut. This may or may not be good to add to the production version, but it sure helps with debugging issues / bottlenecks for deliverymen.

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
@@ -307,10 +307,10 @@ public interface IBuilding extends ISchematicProvider, ICitizenAssignable, IBuil
      * The call will return false if a pickup request already exists, or if the priority is not within
      * the proper range, or if the pickup priority is set to NEVER (0).
      *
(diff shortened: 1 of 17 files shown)


Raycoms:
Code wise this looks fine, we'll play with that during the stream tonight to see how it behaves.

PR #5121, opened 2020-05-31
Add lighter tooltype translation

ToMe25:
- add lighter tooltype translation
-
-

Review please


JiaYow:
Isn't Flint&Steel a vanilla item that can be referenced directly?

ToMe25:
It is, but its like this for reasons like mod compat, i guess.

Raycoms:
Yeah, theoretically this is for any "lighter type". Atm we only support flint and steel but it's generally more generic.

someaddons (approved):

PR #5127, opened 2020-05-31
Add beekeeper

Minecraftschurli:
# Changes proposed in this pull request:
- add Beekeeper colonist

- [x] mechanics (harvesting/breeding)
- [x] hut recipe
- [x] hut block model & texture
- [x] scepter texture
- [ ] translations
- [ ] schematics (beekeeper hut has no special requirement)

Please review

PR #5185, opened 2020-06-09
add a Tag based request

Minecraftschurli:
this PR adds a Tag based request as proposed by @Raycoms in #5127 

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Tag.java
+++ b/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Tag.java
@@ -24,6 +24,7 @@ public class Tag implements IDeliverable
     /**
      * The tag.
      */
+    @NotNull
     private final net.minecraft.tags.Tag<Item> theTag;
 
     /**
@@ -48,7 +49,7 @@ public class Tag implements IDeliverable
      * @param tag   the required containing tag.
      * @param count the count.
      */
-    public Tag(final net.minecraft.tags.Tag<Item> tag, final int count)
+    public Tag(@NotNull final net.minecraft.tags.Tag<Item> tag, final int count)
     {
         this(tag, count, count);
     }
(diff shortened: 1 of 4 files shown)


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5185/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

marchermans (changes requested):
This implementation is not correct. Please check the Stack requestables it already has logic for this. Or it had logic for oredic support. If this is not ported it has to be added again. If it was ported this pr is a duplicate and not required.

Raycoms:
Basically what he means. The "Stack" request should be able to handle this too. It has an oreDict flag. If not, then it can be added to the Stack request.

Minecraftschurli:
I think it would be better to remove the oredict flag from the stack request and replace it with a dedicated tag request because a stack request with oredict checks all tags the item is in not just the specified one which can cause problems. For example with bees which check the flowers tag for breeding, but if the flower that's used for the request is added to another tag all items in that would match the request which would not work for the bees.

marchermans:
The stack request represents a request for a single stack under the given conditions. As such it is the right class and requestables to add this logic to. If that means that you would need to change some of its logic so that you can only pass a tag ID into it and use that directly, then so be it.

Conincidentally this would reduce the complexity and the overall size of the pr to about 20% of what you have right now. Less to maintain and manage is always some thing to strave for.

Minecraftschurli:
That doesn't make sense to me. That would make the stack request more complicated. For me it seems simpler and easier to maintain to split out that functionality into another request (the tag request). Also it would be confusing if you create a stack request but with a tag and not with a stack. Furthermore it would make the nbt serialisation/deserialisation way more complicated to have that functionality in the stack request. 
You are right in that the pr would be smaller but at the cost of complexity being much higher.

Raycoms:
We discussed this internally. :shipit:

PR #5217, opened 2020-06-12
Add the rallying banner

JiaYow:
The rallying banner allows players to summon their guards to them.

## Features
1. Sneak-Rightclicking on a guard building hut adds it / removes it from the rallying list.
2. Right-Clicking in the air opens the GUI, which allows the player to remove Guard Buildings from the list without sneak-rightclicking them.
3. Sneak-Rightclicking in the air or on the ground summons or dismisses the guards to the player.
4. While running to the player, guards will receive a massive speed boost (Speed2 at minimum, but up to Speed5 when highly trained).
5. While rallied, guards will glow in the colony's team color, and are immune to damage from the rallying player. The same is **not** true the other way around!
5. Guards will return to regular duty when dismissed, or when the banner is not in the player inventory anymore.
6. When an active banner is put back into the player inventory (grabbed from a chest, or moved from slot to slot), chances are that the rallying will proceed.
7. There is a new permission that allows players to rally guards of other people's colonies.

## Limitations
1. Guards will use a high amount of food while rallied. Currently it's calibrated to reduce the guard to 2 saturation points after 5.5 minutes of rallying.
2. Dismissing the guard while below 2 saturation will remove those 2 saturation, forcing the guard to eat.
3. Guards will only rally within the colony area. Attempting to move them out of the colony will dismiss them automatically.

## Possible future researches
- Less food consumption
- Strength buff
- Stamina buff
- No friendly-fire-damage from rallied guards to rallying players
- Teleportation on rally
- Increase of allowed rallying range
- Placement of banner to rally guards at a static location (the code is written with that in mind)
- Combat formations (e.g. toggle by shift-scrolling through formations)


Raycoms:
:shipit: 

This has been exhaustively tested on our dev server. So it should be fine.

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5217/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -493,16 +498,24 @@ public void updateAttackingPlayers()
...
+        return checkOrCreateTeam();
+    }
+
     /**
      * Check or create the team.
      */
-    private void checkOrCreateTeam()
+    private ScorePlayerTeam checkOrCreateTeam()
     {
         if (this.world.getScoreboard().getTeam(TEAM_COLONY_NAME + id) == null)
```
someaddons: I think teams are not per-world, so we probably should add colony id + colony dimension to avoid future problems

src/main/java/com/minecolonies/coremod/colony/permissions/Permissions.java
```diff
@@ -136,7 +136,7 @@ public Permissions(@NotNull final Colony colony)
         this.setPermission(Rank.OWNER, Action.TELEPORT_TO_COLONY);
         this.setPermission(Rank.OWNER, Action.RECEIVE_MESSAGES_FAR_AWAY);
         this.setPermission(Rank.OWNER, Action.CAN_KEEP_COLONY_ACTIVE_WHILE_AWAY);
-
+        this.setPermission(Rank.OWNER, Action.RALLY_GUARDS);
```
someaddons, line 139: I think we can add Officer's here too and you need to bump up the permissions version ontop on changes for backwards compat stuff

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIFight.java
```diff
@@ -84,9 +84,9 @@ public AbstractEntityAIFight(@NotNull final J job)
     {
         super(job);
         super.registerTargets(
-          new AITarget(IDLE, START_WORKING, 1),
+          new AITarget(IDLE, PREPARING, 1),
```
someaddons, line 87: Is start working unused now? if so remove the AITarget/function

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
```diff
@@ -140,6 +142,11 @@
      */
     private static final int GUARD_REGEN_INTERVAL = 40;
 
+    /**
+     * Interval between saturation losses during rallying. 1 saturation loss per interval.
+     */
+    private static final int RALLY_SATURATION_LOSS_INTERVAL = TICKS_SECOND * 10;
```
someaddons: Did you consider that they use up food too when fighting?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
```diff
@@ -537,13 +600,30 @@ private IAIState helping()
...
      */
     protected IAIState decide()
     {
         reduceAttackDelay(GUARD_TASK_INTERVAL * getTickRate());
+
+        final ILocation rallyLocation = buildingGuards.getRallyLocation();
+
+        if (rallyLocation != null || buildingGuards.getTask() == GuardTask.FOLLOW)
+        {
+            worker.addPotionEffect(new EffectInstance(GLOW_EFFECT, GLOW_EFFECT_DURATION, GLOW_EFFECT_MULTIPLIER, false, false));
```
someaddons, line 617: think we should check if a potion effect already exists before applying a new one

someaddons (commented):
Looks good overall, left a few comments. Btw what does happen when the player they follow dies? do they take that as previous rally location?

JiaYow:
> Looks good overall, left a few comments. Btw what does happen when the player they follow dies? do they take that as previous rally location?

The guards will currently return to their default behaviour. In code, the guard building checks if the banner is in the inventory of the player, and if it's not, the guards go home. When you pick up the banner again (or have soulbound or something on it), the guards will rally back to you, wherever you are.
Depending on the direction you want to take it, you can either have a research that allows you to place down the banner (thereby it doesn't matter when the player dies), or have a research that changes the `EntityLocation` to a `StaticLocation` onPlayerDeath(), thereby having the guards guard the player's graveyard. Both of those would be elegant solutions to this problem, for another PR :)

src/main/java/com/minecolonies/coremod/colony/Colony.java
JiaYow: I have not changed this, it was "always" this way.
According to https://minecraft.gamepedia.com/Scoreboard#NBT_format, it seems that the Scoreboard, and thereby the teams, are dimension-agnostic. I hope that colony IDs are unique across multiple dimensions as well?

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIFight.java
JiaYow: START_WORKING is still used for regeneration and when a weapon breaks on the guard etc. I just made it not be the initial state, because that means that after server restart the guards will always go home to their hut (`startWorkingFromOwnBuilding` is part of that state), which would potentially kill you when a server restart happens while you're rally/fighting.

In any case, Ray wanted to improve the guard AI anyway - it's a little bit "weird" at the moment, definitely not like more modern AIs are programmed. So he will probably change this too, if he wants to.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
JiaYow: I did. That is why the saturation reduction from this mechanism is always a fraction of your current saturation, ie. when you have low saturation, it won't deduct a lot, so it's not that punishing. Actually, this mechanism alone will only affect your saturation very little when you are below 3 points or so. The point of this code is mainly to force the saturation down to <2 after ~5 minutes of idling/walking, at which point unrallying will flatly set the saturation to 0, thereby forcing them to eat.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
JiaYow: The Follow-code didn't do that either, and as far as I understand, adding potion effects ontop of existing ones will simply extend the duration.
So, as far as I understand it, checking for the potion effect will only cost performance without doing anything meaningful.

src/main/java/com/minecolonies/coremod/colony/Colony.java
someaddons: colony ID's arent unique for dimensions, need dimension+colony ID. And ye I saw that it was like that before, but while you're on changing it we should probably add the dimension ID to it

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
someaddons: kk nice :D was just making sure you considered it

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIFight.java
someaddons: ah its still used fine then

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
someaddons: actually the applying is more expensive than checking for existing, since applying involves events fired for other mods which listen to potion stuff. And some of those mods checks are rather laggy, thats why xD

JiaYow: I understand, but simply adding the if-clause would cause stutter-glowing. The decide() function is only called once every 5 seconds. If the effect runs out 1 second after the last check, the guard will stop glowing for 4 seconds, and then light up again. Is it worth adding logic to check the remaining time of the effect vs. the time until the next check?

Raycoms: once every 5 seconds should be finne.

src/main/java/com/minecolonies/coremod/colony/Colony.java
JiaYow: Fixed in 1123994 and e54bba8.

src/main/java/com/minecolonies/coremod/colony/permissions/Permissions.java
JiaYow: Changed in 6a08fe3.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
JiaYow: I changed the penalty a little bit in e54bba8. It will now reduce the saturation even slower (after about 5 minutes you reach LOW_SATURATION (which is 3 currently), but the flat reduction on unrallying is also increased to LOW_SATURATION. So, they fight longer, but will more likely have to eat after rallying.

JiaYow:
Updated according to code review, plus prevent guards from eating during rallying.

Raycoms:
:shipit:

someaddons (approved):

marchermans (approved):

PR #5229, opened 2020-06-15
Fix #5156

### LINKED ISSUE #5156: Multiple colonists in the same bed
mc 957 I'm still seeing multiple colonists sleeping in the same bed.

ToMe25:
to fix multiple citiizens sleeping in one bed.

Closes #5156
Closes #
Closes #

- Check for other citizens going to the same bed, to prevent multiple citizens from sleeping in the same bed.
-
-

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingHome.java
```diff
@@ -512,12 +538,47 @@ public void setBuildingLevel(final int level)
...
+        return occupiedBeds.contains(bed);
+    }
+
+    /**
+     * Sets whether the given bed is occupied.
+     * 
+     * @param bed the head of the bed to set.
+     * @param occupied whether the bed is occupied.
+     */
+    public void setBedOccupied(BlockPos bed, boolean occupied) {
```
Raycoms: formatting

src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAISleep.java
```diff
@@ -278,4 +271,25 @@ private void goHome()
...
         }
     }
+
+    /**
+     * Sets the used beds occupied state.
+     * 
+     * @param hut      the hut in which the citizen is sleeping.
+     * @param occupied whether the bed should be occupied.
+     */
+    private void setBedOccupied(BuildingHome hut, boolean occupied) {
```
Raycoms: formatting

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingHome.java
```diff
@@ -512,12 +538,47 @@ public void setBuildingLevel(final int level)
...
+     * @param bed the head of the bed to check.
+     * @return whether the given bed is occupied.
+     */
+    public boolean isBedOccupied(BlockPos bed)
+    {
+        return occupiedBeds.contains(bed);
+    }
+
+    /**
+     * Sets whether the given bed is occupied.
```
Raycoms: There is a newDay call in here. On which we should reset this list.

ToMe25: yes, i didn't notice that method :)

Nightenom:
I'm not aware of how vanilla logic for beds works, but we should use that to block players from sleeping in occupied beds, instead of our mechanism

ToMe25:
afaik we already do that, but if the chunk gets unloaded, the vanilla mechanic gets reset, while ours shouldn't.

Nightenom:
villagers must have logic to prevent incorrect behaviour, they should be unloading in similar way to our colonies

Raycoms:
I believe villagers just get up all the time.

Nightenom:
done some research on occupied bed blockstate property and LivingEntity#wakeUp(), using occupied property should be enough and the only way we should use imho

Raycoms:
It doesn't work, on chunk reload the bed gets set to not occupied anymore.

ToMe25:
The only real difference between that and what we already did is the setBlockState flag, we use 3, setBedOccupied uses 4.

Nightenom:
that's quite a difference though :D

ToMe25:
It shouldn't change the unloading behavior tho.

Nightenom:
probably try swapping to proper methods provided by entity classes rather than this, what's the issue know btw?

ToMe25:
LivingEntity#startSleeping() actually uses a flag of 3 too.
only wakeUp uses 4.

Raycoms:
@Nightenom the issue is that several citizens sleep in the same bad if you reload chunks because the flag gets reset. So we have to store it additionally in the building.

Raycoms:
:shipit: 

Did you test it?

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5229/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

ToMe25:
I tested it, yes.
As there is however no reliable way to get them to choose the same bed(or atleast non i know of), something like this can't be 100% verified in a reasonable amount of time.

Raycoms:
It's pretty easy. Make a level 5 building. Destroy all beds but one. Then relog several times during a night =D

ToMe25:
even before this change, for me they didn't cooperate.
rarely a second one got in the bed, but never more.

ToMe25:
also even after relogging, they still sometimes tried to sleep in destroyed beds.

someaddons:
I think we could do this a lot simpler, by checking upon choosing a bed, the other citizens within the citizen home to which bed position they are going. Then the bed doesnt need to "know" at all and we don't need to store/keep data valid at all

ToMe25:
done.

Raycoms:
:shipit:

someaddons (approved):

PR #5266, opened 2020-06-21
Fixed bug that caused the banner to not work if it's renamed before its first use.

JiaYow:
Fixed bug that caused the banner to not work if it's renamed before its first use.

### DIFF
--- a/src/main/java/com/minecolonies/coremod/items/ItemBannerRallyGuards.java
+++ b/src/main/java/com/minecolonies/coremod/items/ItemBannerRallyGuards.java
@@ -435,14 +435,18 @@ public class ItemBannerRallyGuards extends AbstractItemMinecolonies
         if (!banner.hasTag())
         {
             final CompoundNBT compound = new CompoundNBT();
-            compound.putUniqueId(TAG_ID, UUID.randomUUID());
-            compound.putBoolean(TAG_IS_ACTIVE, false);
-
-            @NotNull final ListNBT guardTowerList = new ListNBT();
-            compound.put(TAG_RALLIED_GUARDTOWERS, guardTowerList);
             banner.setTag(compound);
         }
-        return banner.getTag();
+
+        final CompoundNBT compound = banner.getTag();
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/items/ItemBannerRallyGuards.java
```diff
@@ -435,14 +435,18 @@ public static CompoundNBT checkForCompound(final ItemStack banner)
         if (!banner.hasTag())
         {
             final CompoundNBT compound = new CompoundNBT();
+            banner.setTag(compound);
+        }
+
+        final CompoundNBT compound = banner.getTag();
```
Minecraftschurli, line 441: just use getOrCreateTag that does the same thing as you do here
```suggestion
        final CompoundNBT compound = banner.getOrCreateTag();
```

Minecraftschurli (commented):
seems that the previous code was a little derpy

src/main/java/com/minecolonies/coremod/items/ItemBannerRallyGuards.java
JiaYow: You are 99% right. The reason why I did it this way is because it's done this way literally in all other items too. 
Search in the project for `getTag()` and look at the Item-Classes. Then look for `getOrCreateTag()`. See what I mean?

I must assume that the devs have a reason to have it this way. For example, `hasTag()` takes into consideration `isEmpty`, while `getOrCreateTag()` does not. Maybe there has been a bug related to that in the past.

So, when the devs confirm that they simply didn't know about `getOrCreateTag()`, we should change that **everywhere**. Otherwise, I prefer streamlined code to individually smart code. You don't want future devs to think "Ah, the banner uses getOrCreate, the others use getTag(), there must be a reason for this."

@Raycoms?

Raycoms: I believe we've been using hasTag() since forever for all the other items and getOrCreateTag() didn't exist yet.

PR #5271, opened 2020-06-22
add Witch guard type (and alchemist worker?)

Minecraftschurli:
the alchemist is a colonist who supplies the colony with potions especially the witch who heals allies and harms foes in battle


Raycoms:
Oh, about this, I think those should be two different things.
A) A witch type guard you can select in guard/barrackstowers that throw harm/heal potions around.
B) An alchemist worker that creates potions on requests and might in free time walk around and apply random positive potion effects on players and citizens.

Minecraftschurli:
Yea that's what I meant too I'll just work on the alchemist first

Minecraftschurli:
I need suggestions:
What effects should the witch use to debuff enemies at which level?
example: nausea 1 for 0:30 at level 3

What effects should the witch use to buff allies at which level?
example: resistance 3 for 0:20 at level 5

Should the witch also wake up other guards?

Is the alchemist still needed if the witch has infinite potions?

ravenbuilder934:
The witch shouldn't have infinite potions, that's way too OP. Even if they did, the Alchemist's potions could be for the Healer (to heal injuries, not diseases) and for the Postbox's requests. Also, if the Alchemist doesn't have any requests, they should apply random positive potion effects on players and citizens (from ray, above).

Minecraftschurli:
The infinite potions were decided by ray and the chat in todays stream

ravenbuilder934:
oh ok, I didn't watch the entire stream

ravenbuilder934:
The witch shouldn't wake other guards, they're not military-type and consider themselves above that xD

ravenbuilder934:
Maybe the witch should have multiple helping and harming potions to choose from at each level, and which one they chose would be random. Like at Lvl 3 they could have 3 helping and 3 harming to choose from, or 3 total.

The potions should be actual splash potions, not made-up effects/lengths though.

Raycoms:
I'll close those two until they are PR ready.

PR #5290, opened 2020-06-25
Fixes

### LINKED ISSUE #5246: Requests created with some (not all) items already in the wharehouse are never finished.
Version:  minecolonies-1.15.2-0.11.960-universal

### Expected behavior

When builder request an item which has some already crafted in the warehouse and some that need crafting, the missing items are correctly crafted via the appropritate crafter and delivered to the builder.

### Actual behaviour

Wharehouse has 1 flower pot out of 6 required by the builder. He requests 6 (1 picked up and 5 to be crafted. All is good on the UI). However, it is never delivered to the first crafter (in this case, clay is never delivered to the stone smelter so he cannot craft bricks to allow the stonemason to craft pots). Some other requests that were created after the pots even got priorised and smelted super quickly by the stonesmelter and fully crafted correctly.

1- Have a request created by the builder which has items partially present in the warehouse.
2- Make sure you have all the raw items needed in the warehouse.
3- Wait for it to be done. You can even see other requests being done in the meantime through the same required crafter.

I did not take screenshots from the flower pots issue but I reproduced it with torch on the builder itself. The request is stuck this way. In this case, charcoal or sticks will never be delivered to the builder (same for clay on original flower pot issue). He will neither received the 1 torch present in the wharehouse.

image

This is the recipe the builder knows.

image

Just to prove items are in the wharehouse.

image
image

No crash or any message in the latest.log when I reproduce this problem.

### LINKED ISSUE #5272: Fletcher recipe list bug
minecolonies-1.15.2-0.11.961-universal.jar
structurize-0.10.324-ALPHA.jar

Fletcher building level 1. The recipe list is already populated with the recipe White Wool -> String x 4, multiple times. Trying removing them doesn't work. I didn't teach any recipe in the first place.

https://youtu.be/t5CLq7IQyfk

### LINKED ISSUE #5273: Restaurant doesnt work well
Each time I connect on my server, i see all citizen in the restaurant waiting for food.
I need to manually go to find food in each rack of the warehouse or in the worker huts and bring them to the restaurant myself.

citizen, especially guards grab all items they found and never put them back in the warehouse
they keep all eggs, feathers and steaks...very annoying

I put buckets in the cowboy hut....i find them back in the warehouse but without milk or cakes
backer seems to not do any bread

I use chunkloader to load my colony, I dont know if it can be linked to the problem

version : 
minecolonies-1.15.2-0.11.961-universal.jar

Raycoms:
Closes #5273
Closes #5272
Closes #5246

- Fix some recipe overflow issues
- Add missing recipes
- Stop increasing hunger/diseases when offline
- Fix some RS issues
- Adds some extra documentation

Review please


src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -1701,4 +1701,10 @@ public ColonyState getState()
     {
         return colonyStateMachine.getState();
     }
+
+    @Override
+    public boolean isActive()
+    {
+        return colonyStateMachine.getState() != INACTIVE;
```
someaddons, line 1708: why not == ACTIVE?

Raycoms: Because there were some other states too, that would be fine.

someaddons: ye there is UNLOADED which is the state when the colony doesnt meet the required amount of loaded chunks

Raycoms: Yeah, I wanted citizens that are loaded and the player is online still to consume food, etc

someaddons: kk then it does what you want

someaddons (approved):

PR #5312, opened 2020-06-29
Feature/buckets

Raycoms:
- Prefetch materials for builder
-
-

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingStructureBuilder.java
```diff
@@ -62,6 +60,11 @@
      */
     private Map<String, BuildingBuilderResource> neededResources = new LinkedHashMap<>();
 
+    /**
+     * The different possible buckets.
+     */
+    private Deque<Tuple<Map<String, Integer>, Integer>> buckets = new ArrayDeque<>();
```
marchermans: Instead of doing `Deque<Tuple<Map<String, Integer>, Integer>>` you should make this an actuall class with documentation what what is. So create something like `StructureBuildSection` and make this field a `Deque<StructureBuildSection>` Patch all your fields and methods accordingly, and do not forget to document.

marchermans (changes requested):
Generally a good change.
However the implementation with a Tuple of a Map of a String and an Interger and an Integer is terrible.
Nobody understands what is what, and this warrents a custom class, with documentation, toString() and the whole shabang.

Please modify this code to make it more maintainable.

marchermans (approved):
Much better.

PR #5314, opened 2020-06-29
Adds shears and furnace to worker's recipes

### LINKED ISSUE #5304: Stonemason won't always accept cobblestone as an ingredient
Version: 0.13.6-ALPHA

### Expected behavior

Stone, and Cobblestone to be valid ingredients for stonemason, and thus have the stonemason able to craft a furnace.

### Actual behaviour

Stone and Cobblestone are both tagged as stonemason_product, and I believe they should be stonemason_ingredient. As such, trying to add the recipe for the furnace fails. Cobblestone walls on the other hand work, as they are explicitly tagged as stonemason_product

1. Build a stonemason
2. Attempt to add furnace as a recipe
3. "Unable to store the blueprint...."
4. ... Profit

PowerSchill:
Closes #5304
Closes #
Closes #

-
-
-

Review please


src/main/resources/data/minecolonies/tags/items/blacksmith_product.json
```diff
@@ -2,6 +2,7 @@
   "replace" : false,
   "values" : [
     "#forge:nuggets",
-    "#forge:ingots"
+    "#forge:ingots",
+    "minecraft:shears"
```
Mekle001, line 6: Why put it here, rather than putting #forge_ingots in blacksmith_ingredient.json so that it picks up shears, flint and steel, and anything added by mods? I'm assuming ingots are in products for the purposes of compacting nuggets to ingots. But if you moved all of this to ingredient, you'd still get that. That change would open up lamps here, so you'd want to exclude torches from the blacksmith.

src/main/resources/data/minecolonies/tags/items/stonemason_product.json
```diff
@@ -8,6 +8,7 @@
     "#minecraft:stairs",
     "#minecraft:walls",
     "minecraft:bricks",
+    "minecraft:furnace",
```
Mekle001, line 11: So, I get resolving the lack of Furnace. What I don't get is why #forge:stone is a product, when that is a product of the stonesmeltery.  Cobblestone too, as there isn't a normal recipe for the stonemason to produce cobble. 

I expected both #forge:stone and #forge:cobblestone to move to stonemason_ingredient.json. At which point, many of these other items are no longer necessary. 

You also likely want to add redstone to the stonemason_ingredient_excluded.json

PowerSchill:
Don’t approve yet. I forgot to add iron bars.

src/main/resources/data/minecolonies/tags/items/blacksmith_product.json
PowerSchill: Because then you could craft anything with ingots like armor for example which is what the blacksmith is for or pistons instead of mechanic.

src/main/resources/data/minecolonies/tags/items/stonemason_product.json
PowerSchill: Using [stone.json](src/generated/resources/data/forge/tags/items/stone.json) as a reference you will see that cobblestone is not included but polished stones are. Also some mods make granite, diorite, and andesite craftable so this could include those. Also the stone smeltery can only handle furnace recipes so the mason handles all stone crafting recipes.

src/main/resources/data/minecolonies/tags/items/blacksmith_product.json
Mekle001: If you add ingots to the blacksmith ingredients, and redstone and torches to the blacksmith exclude, it seems like you'd get the exact set of things that you'd expect from the blacksmith, including iron bars.

src/main/resources/data/minecolonies/tags/items/stonemason_product.json
Mekle001: I'm not sure which file that hyperlink is supposed to be, as it appears to be a broken link.  My point above is that according to the tags, cobblestone/stone are products of the stonemason, instead of ingredients. Given that they aren't marked as ingredients, you have to add furnace to the products to be able to enable it. 

My proposal is above: add 2 things to stonemason_ingredient.json and 1 to stonemason_ingredient.excluded.json. I believe that fits the theming of the hut, and provides for a much wider array of recipes. I was thinking it might be good to also add #stonemason_product to stonemasion_ingredient.json, which should cut down on future maintenance here.

If you take my proposal, I believe there are several other things in stonemason_product.json that can be removed. 

As an aside, is there a way to provide custom versions of these files for packs? Whatever rules are in them can be broken by 'expert mode' packs

Mekle001: Oh! please consider moving minecraft:nether_brick to stonemason_ingredients, that would unblock nether_brick_fence.

src/main/resources/data/minecolonies/tags/items/blacksmith_product.json
PowerSchill: still have to deal with lanterns and rails and hoppers and minecarts and. tripwire

Mekle001: Lanterns are covered by the torches ingredient exclusion, rails are going to have to be a product exclusion, as the ingredients are the same as an iron sword.  For the tripwire, a product exclusion might be the easiest solution, as a planks exclusion would conflict with the wooden tools. 

If you can exclude mechanic_product from the blacksmith, then you easily solve moving things to the mechanic by adding them as a product there.

src/main/resources/data/minecolonies/tags/items/stonemason_product.json
PowerSchill: If you move cobblestone/stone to the ingredients just look at all of the exceptions you would have to add: https://minecraft.gamepedia.com/Cobblestone and https://minecraft.gamepedia.com/Stone.

Yes, the whole reason that we migrated to the use of item tags is because any pack owner or user can easily adjust them with json files and no code updates required.

Mekle001: I'm not seeing that many exceptions. I see an ingredient exception for redstone, and one I missed which is stick. 

That leaves comparator and stonecutter unaccounted for, which I'd argue that anywhere you have a redstone exception, a redstone torch should be there.  It might be nice to account for that by including #mechanic_ingredient as stonecutter ingredient exceptions. That does mean that tag marking has to make at least a second pass over things it's already marked, or some form of recursive evaluation.

PR #5323, opened 2020-07-01
Show schematic outlines while the player has the building tool in the main hand

JiaYow:
This PR makes it so the building tool will render the max-size building outlines while you have it equipped.

Goal is to 1. make it much easier to perfectly align buildings / avoid collisions, and 2. later on we can make the outlines pretty and have a kind of "blueprint" view of a colony. could look very cool from above. i'd like to know if it causes noticeable lag on the client-side for huge colonies.


someaddons:
our Building Tileentities do have the corner information, you don't need to sync it additionally at all. See TileEntityColonyBuilding te; te.getCornerPositions()

JiaYow:
> our Building Tileentities do have the corner information, you don't need to sync it additionally at all. See TileEntityColonyBuilding te; te.getCornerPositions()

The corners have no y information.

PR #5343, opened 2020-07-05
Fix Crafter Recipes

### LINKED ISSUE #5340: Crafters using the new Tag system sometimes don't honor the tags the way you'd expect..
Version: minecolonies-0.13.16-ALPHA-universal.jar

### Expected behavior

- Tags to evaluate products, and then ingredients so that product marking always takes precedence. 

### Actual behaviour

- Evaluated with exclusions of both products and ingredients and then inclusions of them both. 

1. Define Brick @ stonesmelter, and Stick @ Sawmill
2. Attempted to add "Clay Shingle Slab" to Sawmill (It's explicitly marked sawmill_product)
3. adding the recipe fails, because brick is marked "sawmill_ingredient_excluded" as it arguably should be. 
4. ... Profit

Mekle001:
Reorder Tag evaluations to be product and then ingredient, rather than exclude and then include
Add Terracotta to Stonesmelter
Add sawmill products to blacksmith exclusions

Closes #5340
Closes #

-
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
@@ -238,6 +238,12 @@ public abstract class AbstractBuildingWorker extends AbstractBuilding implements
             return Optional.of(false);
         }
 
+        // Check against allowed products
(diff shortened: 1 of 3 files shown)


Mekle001:
#5340

PR #5345, opened 2020-07-05
Show schematic outlines while the player has the building tool in the main hand - 2nd try

JiaYow:
This PR makes it so the building tool will render the max-level building schematics and their outlines when the active preview is nearby. This should make it easier to properly align neighboring buildings.


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5345/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/event/ClientEventHandler.java
```diff
@@ -77,18 +103,128 @@ public static void renderWorldLastEvent(@NotNull final RenderWorldLastEvent even
...
+                        }
+                    }
+
+                    final Blueprint blueprint = wrapper.getBluePrint();
+
+                    if (blueprint != null)
+                    {
+                        Mirror mirror = buildingView.isMirrored() ? Mirror.FRONT_BACK : Mirror.NONE;
+
+                        final TileEntity tileEntity = world.getTileEntity(currentPosition);
```
someaddons, line 183: Could we get a highlight on the Tileentity to render aswell? would be very useful for finding buried buildings.
Should be a simple RenderBox() or so call, the code for that exists in structurize

someaddons (commented):
Looks good overall

src/main/java/com/minecolonies/coremod/event/ClientEventHandler.java
Asherslab: i'd say it'd be simpler to bring up the marker for the anchorPos (which is the TE in any building schematic)

JiaYow: Highlighting the TE doesn't help with finding it, because if it's buried, the highlight will also be buried. By the time you find the highlight, you can see the hut anyway. I experimented with removing the depth check on the TE-highlight to make it shine through other solid blocks like a wallhack, but I couldn't make it work.
However, since the PR will render the full lvl5 building now (it works a bit different now than when you last saw it someaddons), it should be a bit easier to find the hut "intuitively".
Ashers, what do you mean with "bring up the marker"?

Asherslab: https://github.com/ldtteam/Structurize/blob/9dfc02da242ee5b6f7ac4ecb49e6722f19ad262f/src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java#L120

JiaYow: Ah yes. Then my comment stands, the highlight isn't really useful for finding the hut, especially if it's buried. Would be a whole different story if the highlight were wallhacky, but no idea how to accomplish that.

someaddons: the tag hightling in structurize actually can be seen through blocks hmm

someaddons: oh funny its actually only the text highlightning which goes through blocks xD

Raycoms:
There is a gl flag for that. It's used around the icons of the workers if
you want to know it exactly

src/main/java/com/minecolonies/coremod/event/ClientEventHandler.java
JiaYow: I tried reproducing the behavior of the icons of the workers, based on Ray's comment. It refused to work. The icons render as textures, and use `.depthTest(RenderState.DEPTH_ALWAYS)`, but no matter how much I fiddle around with this and culling, it just doesn't work. Maybe there's one or fifty other GL flags I need to set, I don't know.

Additionally, the icons of the workers are not visible through ghostblocks, so it wouldn't work for this PR anyway.

In short, sorry but I give up. Someone who understands rendering better than I do should take a look at this please, in a separate PR.

Raycoms: Okay, we'll leave it for someone else to do it for a future PR. It's not actually so directly connected to this.

marchermans (approved):

PR #5364, opened 2020-07-10
Add batching to miner resource request when mining

Mekle001:
- When the miner is mining, inflate resource requests up to 32 to minimize dman trips

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingStructureBuilder.java
```diff
@@ -669,7 +674,7 @@ public void checkOrRequestBucket(@Nullable final BuilderBucket requiredResources
                         break;
                     }
 
-                    worker.createRequestAsync(new Stack(itemStack.getItemStack(), requestCount, requestCount));
+                    worker.createRequestAsync(new Stack(itemStack.getItemStack(), Math.max(requestCount, resourceBatchSize), requestCount));
```
Raycoms: That looks simpler than what I would've expected even.
I would make the resourceBatchSize a getter which on default returns 1  (in the abstract building of the structure builders) and in the miner returns 10.

Such that here we request requestCount * getResourceBatchSize

Mekle001: I like it.

Raycoms: This way instead of requesting 32 planks and 32 fences, we request 30/40 (which is a better ratio considering what we need too)

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingMiner.java
```diff
@@ -168,6 +168,15 @@ public int getMaxBuildingLevel()
         return MAX_BUILDING_LEVEL;
     }
 
+    /**
+     * The Miner wants to get multiple nodes/levels worth of stuff when requesting.
+     */
+    @Override
+    public int getResourceBatchMultiplier() {
+        //Ask for 10x the resources if possible
```
Raycoms, line 177: bracket should go down.

marchermans (approved):

PR #5371, opened 2020-07-11
Add visible status indicators for citizens

someaddons:
-Add visible status indicators for citizens

Review please


src/api/java/com/minecolonies/api/entity/citizen/VisibleCitizenStatus.java
```diff
@@ -0,0 +1,48 @@
+package com.minecolonies.api.entity.citizen;
+
+import com.ldtteam.structurize.util.LanguageHandler;
+import com.minecolonies.api.util.constant.Constants;
+import net.minecraft.util.ResourceLocation;
+
+/**
+ * Enum for citizen status icons, resource location and translation
+ */
+public enum VisibleCitizenStatus
```
marchermans: This is not extensible.
Please refactor this in a way that it does not require the enum to be used.

In other words the statusses that a citizen can be in, need to be extendible via the API.

marchermans (changes requested):
Not extendable implementation is used.
Please adapt to allow for api users to modify the total amount of states

src/api/java/com/minecolonies/api/entity/citizen/VisibleCitizenStatus.java
someaddons: made it a simple class instead

marchermans (dismissed):

PR #5375, opened 2020-07-12
Beekeeper quickfixes

Mekle001:
I was having trouble getting the beekeeper to work, so I went squishing bugs of the non-bee variety. 

- Add some en-us string translations to help the UI make sense
-Fix the max hives to be non-zero at level one. 
-Fix the beekeeper losing track of hives as fast as they are added, and constantly complaining about no hives

Review please @Minecraftschurli

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
@@ -231,7 +231,7 @@ public class BuildingBeekeeper extends AbstractBuildingWorker
      */
     public int getMaximumHives()
     {
-        return (int) Math.floor(Math.pow(2, getBuildingLevel() - 1));
+        return (int) Math.floor(Math.pow(2, getBuildingLevel()));
     }
 
     public static class View extends AbstractBuildingWorker.View
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
@@ -231,7 +231,7 @@ public class BuildingBeekeeper extends AbstractBuildingWorker
      */
     public int getMaximumHives()
     {
-        return (int) Math.floor(Math.pow(2, getBuildingLevel()));
+        return (int) Math.floor(Math.pow(2, getBuildingLevel() - 1));
     }
 
     public static class View extends AbstractBuildingWorker.View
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
```diff
@@ -231,7 +231,7 @@ public boolean shouldHarvestHoneycombs()
      */
     public int getMaximumHives()
     {
-        return (int) Math.floor(Math.pow(2, getBuildingLevel() - 1));
+        return (int) Math.floor(Math.pow(2, getBuildingLevel()));
```
MatthiasMann: return 1 << getBuildingLevel();

marchermans: Yep. But it not really relevant, but a good idea todo anyway.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
Raycoms: I'd say no. While it is quicker. It makes this code much less readable about what we want to do.

Raycoms: We also have this everywhere like this exactly already as code.

Minecraftschurli: Not necessary because 2^0=1

Minecraftschurli (changes requested):
Looks good except for one thing

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
Mekle001: I have to agree with @Raycoms. The shift left is faster, but far less readable. If we want to switch to it, then I think that should be a separate change that hits all of the code to keep it consistent. There are 13 locations in the code today to be reviewed.

MatthiasMann: I only pointed out the left shift because it avoid the possibilities for rounding issues - esp when using Math.floor() instead of Math.round().

Mekle001: Yeah, I backed out the change at @Minecraftschurli's request, and then realized a few minutes ago that what I was seeing was likely due to the use of floor. I'm doing a bit of testing, and will likely remove the floor. None of the other 12 locations use floor.

Minecraftschurli (dismissed):
Looks good now

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBeekeeper.java
Mekle001: Ok, floor gone.

PR #5380, opened 2020-07-13
Add breeding toggle for herders

ToMe25:
Closes #
Closes #
Closes #

- Add a Button to the second page of the cowboy, shepard and chicken herder to disable them breeding the animals.
-
Here some images:

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/util/constant/NbtTagConstants.java
+++ b/src/api/java/com/minecolonies/api/util/constant/NbtTagConstants.java
@@ -482,6 +482,16 @@ public final class NbtTagConstants
     public static final String TAG_WAITING_TICKS     = "waitingticks";
 
     /**
+     * Herder breeding tag
+     */
+    public static final String TAG_BREEDING = "breeding";
+
+    /**
+     * Cowboy milking tag
+     */
+    public static final String TAG_MILK_COWS = "MILK_COWS";
+
+    /**
      * Beekeeper tags
      */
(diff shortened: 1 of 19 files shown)


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5380/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

someaddons (commented):
looks nice, could you add that turning off breeding also does turn off random sheep dying? Since that won't work when breeding is off

ToMe25:
I don't think thats a good idea, as it would either keep it turned off for when you re enable breeding, or would have to additionally store the old value, just to set it back as soon as breeding gets turned back on.
As long as this doesn't cause any issues i would leave the dying in the state it is when disabling breeding.

ToMe25:
If this does however cause any problems it might be worth it to temporarily disable random dying while breeding is disabled, tho i would still probably look at whether there is a easier solution for whatever problems it might cause.

someaddons:
well as is now a user could expect to have breeding off and random dying on

MatthiasMann:
Just gray out the UI button to indicate that this switch has no effect while breeding is disabled?

someaddons (approved):
Its fine, dying happens on shearing not breeding

ToMe25:
But random dying works while he isn't breeding them...
It just doesn't make much sense, but thats the players problem xD

PR #5383, opened 2020-07-13
Upgrade the cook, and a few more recipes

### LINKED ISSUE #5276: Cook not working because of 'missing items' 2
A reopening of #5187, because it still isn't fixed

Mekle001:
Closes #5276

- The cook will try to resolve furnace recipes (for food) without being taught. 
- The cook has a new assistant who does all the crafting, leaving the cook to focus on serving the colony, and basic cooking.
- The cook's assistant is able to be hired at building level 3
- The cook (and other crafters) will better handle secondary outputs for recipes. (1 pot for a full stack of output!)
- The cook will now stop 'auto smelting' food while working on specific recipes, to avoid stealing from itself
- Enable the crusher to convert/decompress clay blocks into clay balls - Bricks, bricks galore!
- The farmer is now capable of learning the composted dirt recipe
- All pressure plates are now crafted at the mechanic. As a 2x2 recipe, the builders can also craft for themselves.
- Miscellaneous cleanup of unnecessary imports, as a result of my previous changes.  

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCrusher.java
```diff
@@ -123,6 +127,30 @@ private void loadCrusherMode()
         }
     }
 
+    private void loadCurrentRecipes()
```
Raycoms, line 133: javadoc

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -175,6 +182,34 @@ public boolean canBeGathered()
...
+                    ImmutableList.of(new ItemStack(item, 1)),
+                    1,
+                    smeltingResult,
+                    Blocks.FURNACE);
+          
+                addRecipeToList(IColonyManager.getInstance().getRecipeManager().checkOrAddRecipe(storage));
+            }
+        }
+    }
+
```
Raycoms: I'm nt a fan at all of this. Adding a lot of automatic recipes slows down the colony (each time a request is executed all recipes have to be checked). Why do we need this explicitly?

Mekle001: In a vanilla world, we really don't. Once you add in modpacks, there are a lot of recipes that require cooked foods that can't be fulfilled today. As we discussed in Discord, I'll look at the temporary recipe via resolver.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -141,6 +138,50 @@ public boolean canBeGathered()
...
+    public IRecipeStorage getFirstRecipe(final Predicate<ItemStack> stackPredicate)
+    {
+        //First, do the normal check against taught recipes, and return those if found
+        IRecipeStorage storage = super.getFirstRecipe(stackPredicate);
+        if(storage != null)
+        {
+            return storage;
+        }
+
+        //If we didn't have a stored recipe, see if there is a smelting recipe that is also a food output, and use it.
```
Raycoms, line 225: I would only query this if ISFOOD

Mekle001: Given that we don't know what the item is until after the query, I'm not sure how to limit it. We could do a combined predicate, though I don't see any advantage to that, and this code is clearer.

Raycoms: I think there was one getFirstRecipe that gets an itemStack too. Where we could intercept.

Mekle001: I had done that, and pulled it, because it's used rarely, and at the abstract level it just wraps the predicate version.  That is actually what I did first, and it didn't actually let things work, because the predicate is used almost everywhere.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -77,6 +84,16 @@ public BuildingCook(final IColony c, final BlockPos l)
         keepX.put(FurnaceTileEntity::isFuel, new Tuple<>(STACKSIZE, true));
     }
 
+    public boolean getIsCooking()
```
Raycoms, line 101: javadoc

Raycoms:
:shipit: looks good now

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5383/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -73,6 +84,24 @@ public BuildingCook(final IColony c, final BlockPos l)
...
+        return isCooking;
+    }
+
+    /**
+     * Record the state of the assistant processing requests
+     * @param cookingState true if currently crafting
+     */
+    public void setIsCooking(final boolean cookingState)
+    {
+        isCooking = cookingState;
```
someaddons, line 112: hm I think this can cause issues, e.g. if the assistant stops working or is fired for some reason

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -90,6 +119,12 @@ public int getMaxBuildingLevel()
     @Override
     public IJob<?> createJob(final ICitizenData citizen)
     {
+        for (final ICitizenData leadCitizen : getAssignedCitizen())    {
```
someaddons: formatting

someaddons (commented):
Some minor things, we probably should also not show the teach recipe button before level 3, since we don't allow 3x3 recipes before and got no fitting error message when it fails

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
Mekle001: Good point. Will add citizen tracking too.

Mekle001:
@someaddons on the topic of disabling the button, I agree that would be nice. Right now as the PR stands, it will give the 'upgrade the building to add more recipes' error for levels 1 and 2, which I thought was reasonable.

someaddons (approved):

PR #5385, opened 2020-07-14
Fix dyer and concretemixer

### LINKED ISSUE #5378: Concrete Mixer doesn't deliver Concrete Powder on request
Version: 0.13.29

### Expected behavior

- Concrete mixer to mix powder, and hand off to deliveryman when done. 

### Actual behaviour

- Concrete Mixer makes the powder, and instantly starts placing it in water and converting it to concrete.

1. Build a concrete mixer
2. request some concrete powder
3. watch it retrieve all ingredients, mix, and then reset to try again because it all changed to concrete. 
4. ... Profit

Mekle001:
Closes #5378 

- Add non-placeholder UI to the Dyer, so that you can define fuel and thus actually craft green dye
- Stop concrete mixer from mixing any and all powder, Now it will only mix what is requested

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/WindowHutDyer.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/WindowHutDyer.java
@@ -1,0 +1,72 @@
+package com.minecolonies.coremod.client.gui;
+
+import com.google.common.collect.ImmutableList;
+import com.ldtteam.blockout.views.View;
+import com.ldtteam.structurize.util.LanguageHandler;
+import com.minecolonies.api.colony.IColonyManager;
+import com.minecolonies.api.crafting.ItemStorage;
+import com.minecolonies.api.util.constant.Constants;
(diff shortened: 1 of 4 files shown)


Raycoms:
Oh the concrete mixer part was on purpose. When some concrete dust is left in the building or a player puts it there he converts it even out of a request.

Mekle001:
@Raycoms I thought that might have been the intent. Unfortunately, the implementation was that if you requested 4 dust, by the time the deliveryman came to pick it up it was always mixed to concrete. The end result being that you ended up with a LOT of concrete, and only got your dust if you got lucky on timing. 

I was going to try to get the best of both worlds, but that quickly became far more complex and potentially dangerous given how well I know the systems today. My thought would be to take this, and then come back separately and add 'auto processing' as a feature, or a togglable in a future PR.

Raycoms:
It should be relatively easy tbf. We can in the hut just check if there is a pickup request scheduled for the building. If so, then don't process it automatically.

someaddons (approved):

Issue #1644, opened 2017-11-17
[Feature] Better scarecrow GUI, and Options


bob0bob:
It would be great in the scarecrow gui to be able to select the field size.  Sometime you don't need a 5x5 farm field.  Having smaller ones available would be great.   Seems weird but a selection os a 2x2, 3x3, 4x4.  Would be great.  When using pams harvest, there are items you don't need a bunch of 

Just a thought.

Zoidey:
Not a bad idea, another idea to add to it is maybe allow up to 4 or 9 crops. 

if you had a 4x4 farm you could have wheat, carrots, potatoes and melon in a corner each or something. I guess kinda like you can set whats in the NW, NE, SW and SE sections in the Forestry Multifarms.

Wissi:
I agree on the second idea. 
Having a farmer "manage" up to 4 crops per field would be awesome due to the great varieties of mods like Pam's harvest. 
Maybe 4 slots in scarecrow to divide the plot in 4 sections w/4 diff. crops.

Raycoms:
The field size I agree on, multiple crops I don't agree on, the reason we have fields and not one big field is that there is exactly one crop per field. If the player wants different types he has to add more farmer levels and or fields, else this is just tooo easy =D

winex:
well, with harvestcraft this is insane, yeah.
maybe just give guys 2x2 grid in the scarecrow and half/triple(don't)/quad the field. wouldn't be hard to implement i think, well, even if AI checks this every time (i think there is no list of field-blocks). and Scarecrow probably already has some range check.

or just from NW to SE and to use all field you need to place 4 seeds.
i like this!

Raycoms:
Closed in favor of joint issue.

PR #5386, opened 2020-07-14
Adjustable Field Size

ShadowProtocol:
Implements #1644

- The scarecrow field GUI has four new directional buttons
    - Tooltips help the player orientate themselves
- Buttons in scarecrow GUI control how large the worked plot size of the field is (buttons toggle 0 through 5)
- Farmer AI (EntityAIWorkFarmer) will skip over blocks outside the defined size without requiring a new worker tick each time

The new GUI (image 1) and neatly sized plot (image 2)

Review please


Raycoms:
Why doesn't it say the worker name anymore that is assigned to it?

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -432,6 +432,15 @@
   "com.minecolonies.coremod.gui.buildtool.hut.level": "Level %s",
   "com.minecolonies.coremod.gui.buildtool.decorations": "Decorations",
   "com.minecolonies.coremod.gui.buildtool.scans": "My schematics",
+  "gui.field.north": "Northward field size",
```
Nightenom: use modid in lang keys

ShadowProtocol:
> Why doesn't it say the worker name anymore that is assigned to it?

It does. The name will appear above the GUI once ScarecrowTileEntity#setOwner is called (I have not changed when setOwner gets called afaik). If there is no farmer, there is no name. Did you want a "No farmer assigned" placeholder?

Raycoms:
Nvm, I didn't see it on top of it. The position changed and I got confused xDD

src/main/java/com/minecolonies/coremod/client/gui/containers/WindowField.java
```diff
@@ -40,7 +55,9 @@
     /**
      * Tile entity of the scarecrow.
      */
-    private final AbstractScarescrowTileEntity tileEntity;
+    private final AbstractScarecrowTileEntity tileEntity;
+
+    private final int[] radii = new int[4];
```
someaddons, line 69: javadoc

src/main/java/com/minecolonies/coremod/client/gui/containers/WindowField.java
```diff
@@ -53,18 +70,66 @@ public WindowField(final ContainerField container, final PlayerInventory playerI
...
+
+
+    }
+
+    @Override
+    protected void init ()
+    {
+        super.init();
+
+        final int offset = 24;
```
someaddons: make a constant for this

src/main/java/com/minecolonies/coremod/client/gui/containers/WindowField.java
```diff
@@ -91,4 +156,118 @@ public void render(int x, int y, float z)
         super.render(x, y, z);
         this.renderHoveredToolTip(x, y);
     }
+
+    protected class DirectionalButton extends Button
```
someaddons, line 170: javadoc

someaddons (dismissed):
Looks good, just minor doc/lang key prefixes missing

src/main/resources/assets/minecolonies/lang/en_us.json
ShadowProtocol: Missed this before somehow. Anyways adding "com" and "coremod" is adding megabytes onto the lang files all up, and is not really current convention. So do you want "gui.minecolonies.field..." instead?
Also quite happy to do a PR to tidy up all these lang files (removing deprecation, sort in order, etc). Not that hard with project-wide find-and-replace.

Nightenom: that's up to you, ideally, it should start/contain modid to prevent collisions with everything else

Nightenom: btw do not sort them, and we should be adding them to the end of the file, removing is fine, beware of project replace as it usually replaces more than you need :D

ShadowProtocol: Alright then, gui.minecolonies it is. And I never use project replace without regex, to be ultra careful :)

Raycoms: Make it the default com.minecolonies.coremod.gui, those few extra bytes in exchange for consistency. Maybe we rename them in the future to a shorter one. But for now that's the full gradle modid.

ShadowProtocol: Not that gradle really has anything to do with the games I18n at all. I'll put them with the full thing then for now, but I would like to change them all in another tidy up PR if there's no big objections. It has no advantages being like this, and to me looks kinda messy. (and if someone's offering to clean up for free, wouldn't you take it? :D)

Raycoms: We can clean those up when we drop 1.15 support. If you do this now, this will lead to massive conflicts with 1.16 potentially =D

ShadowProtocol: Sure, that's fair.

marchermans (approved):
Somebody file fixed the damn Scarescrow typo.
:D

PR #5421, opened 2020-07-20
Fix hasItemStack in racks

Mekle001:
- Fix TileEntityRack::hasItemStack to compare not just key, but quanity
- Without this, the request system has all sorts of weird behavior, and stuck requests

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
+++ b/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
@@ -133,7 +133,8 @@ public class TileEntityRack extends AbstractTileEntityRack
     @Override
     public boolean hasItemStack(final ItemStack stack, final boolean ignoreDamageValue)
     {
-        return content.get(new ItemStorage(stack, ignoreDamageValue)) != null;
+        final ItemStorage checkItem = new ItemStorage(stack, ignoreDamageValue);
+        return content.containsKey(checkItem) && content.get(checkItem) >= stack.getCount(); 
     }
 
     /**


src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -133,7 +133,8 @@ public int getFreeSlots()
     @Override
     public boolean hasItemStack(final ItemStack stack, final boolean ignoreDamageValue)
     {
-        return content.get(new ItemStorage(stack, ignoreDamageValue)) != null;
+        final ItemStorage checkItem = new ItemStorage(stack, ignoreDamageValue);
+        return content.containsKey(checkItem) && content.get(checkItem) >= stack.getCount();
```
someaddons, line 137: don't you need content.get(checkItem).getAmount() ?

marchermans: This behaviour is also completely wrong.
The ItemStackRequest, and StackListRequest have flags for checking against tags, metadata and other values optionally.
This is completely ignored in this situation.

Please fix this.

marchermans (changes requested):

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
Mekle001: @someaddons content.get(checkItem) returns an integer which is quantity. So, no, you don't need getAmount. 

@OrionDevelopment this is core TileEntityRack behavior that I'm just restoring previous behavior to. The flags are applied at another layer, and have never been driven down to here.

marchermans: Okey.

marchermans: Okey.

Raycoms:
Pushed a htofix that reverts the change completely.

PR #5481, opened 2020-07-31
Guard improvements

someaddons:

Guards received 3 new researchable abilities:
- Arrow usage, consumes arrows and increases damage
- Arrow piercing, arrows pierce two targets. This one should help with ranger's arrows beeing caught by knights on the frontline aswell.
- Knockback aoe, Knights get an aoe knockback dealing a little damage and knocking back surrounding enemies. Chance to use on hit and has a cooldown of 30 seconds. Should lead to more balanced fighting and less many vs one situations.

Knights can use shields at any levels once researched.
Knights have higher hp scaling aswell as the scaling based on the skill instead of arbitrary worker level.
Knights no longer unintentionally scale in damage.
Rangers have higher base range.
Rangers distance now scales with Adaptability instead of job level.
Rangers no longer scale in health with their level, only with their buildings level.
Rangers now are able to shoot better when stationed ontop of buildings, gaining extra range, patrol to edges where they can see far and are chasing mobs into vision from above without using ladders.

Guards now can give other patrolling guards a temporary consumed patrol position, when encountering raiders so they help each other out better.
Fleeing citizens now call the two closest guards instead of one.

Guards now properly equip the armor from their inventory upon starting to fight. They also now dump their inventory occasionally when patrolling/guarding and upon dump they check their hut for armor/better armor and request any missing.

Guards combat now is active in almost all worker AI states, so that they always can react when getting hit. They do not search for targets though when doing non guard task stuff.

Added a more dynamic version of isOkayToEat, called canBeInterrupted where the AI itself determines if its possible be to interrupted. Dumping/eating/sickness all respect that call and do not start executing if not allowed to.

Interactions no longer constantly query the requests to show the interaction icon, instead they are created upon request creation.
Added new function for checking for open sync requests, which avoids having to map all tokens to requests and then checking for all if the citizen has the token as async, instead directly checks the tokens for async.
Request interactions now show the first item they need visually in the interaction, e.g. A ranger displaying the bow he needs.
Fix mercenaries using a nonexisting texture as head.
Tavern no longer instantly spawns a visitor after beeing built.

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -1148,6 +1162,26 @@ public boolean hasWorkerOpenRequestsFiltered(@NotNull final ICitizenData citizen
         return getOpenRequests(citizen).stream().anyMatch(selectionPredicate);
     }
 
+    @Override
+    public boolean hasOpenSyncRequest(@NotNull final ICitizenData citizen)
+    {
+        final Map<Integer, Collection<IToken<?>>> requestIds = getOpenRequestsByCitizen();
+        if (!requestIds.containsKey(citizen.getId()))
```
Raycoms: why not call the function hasWorkerOpenRequests for this part?

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -577,25 +548,32 @@ private IAIState lookForRequests()
                     );
                     return NEEDS_ITEM;
                 }
-                else
```
Raycoms: This here is important, we should re-request if the pickup failed.

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -1436,6 +1435,7 @@ public void callForHelp(final Entity attacker, final int guardHelpRange)
 
         long guardDistance = guardHelpRange;
         AbstractEntityCitizen guard = null;
+        AbstractEntityCitizen secondGuard = null;
```
Raycoms: Potentially this should be more generic. Make it a set or a list, etc. So that this can be extended if needed.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
someaddons: ye thats simpler

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobCanSee.java
```diff
@@ -0,0 +1,82 @@
...
+        Vec3d vec3d = new Vec3d(pos.getX(), pos.getZ() + searchingEntity.getEyeHeight(), pos.getZ());
+        Vec3d vec3d1 = new Vec3d(lookTarget.getPosX(), lookTarget.getPosYEye(), lookTarget.getPosZ());
+        return this.world.rayTraceBlocks(new RayTraceContext(vec3d, vec3d1, RayTraceContext.BlockMode.COLLIDER, RayTraceContext.FluidMode.NONE, searchingEntity)).getType()
+                 == RayTraceResult.Type.MISS;
+    }
+
+    @Override
+    protected boolean isPassable(@NotNull final BlockState block)
+    {
+        if (block.getBlock() == Blocks.LADDER)
```
Raycoms: This means they can go stairs up and down though, right?

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobWalkRandomEdge.java
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Walks to a random edge block nearby, a block next to air. Does not use ladders
+ */
+public class PathJobWalkRandomEdge extends AbstractPathJob
+{
+    private static final Direction[] directions = {Direction.NORTH, Direction.SOUTH, Direction.EAST, Direction.WEST};
```
Raycoms: There is already a horizontal directions array in vanilla you can query afaik

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
someaddons: shouldnt the normal AI handle that? Like if the request is completed and gets removed but it did not get the item it wanted, it naturally would run into re-requesting

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobCanSee.java
someaddons: ye stairs would work, you often got stairs as rooftop aswell so can't exclude those.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
Raycoms: Not necessarily very timely. I think it's a neat speed improvement to have it here without any backlash.

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobCanSee.java
Raycoms: Maybe we should add something like "max 5y difference" or so. So they don't climb walls down if it has stairs.

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobWalkRandomEdge.java
someaddons: hm only seeing a private one

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobCanSee.java
someaddons: ye gonna add that as destination condition

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
someaddons: ok

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
someaddons: hm I wanted to avoid having to sort a complete list and compare it all again

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobWalkRandomEdge.java
Raycoms: Direction.Plane.HORIZONTAL

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
Raycoms: could make it a map citizen id to Entity for quick access

someaddons: I'm sorting them by distance here, don't think a map would help with that :D

Raycoms: That should be relatively simple. Make it an ordered list by distance such that you can always just poll the last one out each round. That would probably make even the existing one easier.

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobCanSee.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    protected double computeHeuristic(final BlockPos pos)
+    {
+        return searchingEntity.getPosition().manhattanDistance(pos);
+    }
+
+    @Override
+    protected boolean isAtDestination(final Node n)
+    {
+        if (Math.abs(start.getY() - n.pos.getY()) > 6)
```
Raycoms: I don't see a problem going up, I think the problem is only going down imo

someaddons: but some might want to station them not on the highest but inbetween somewhere

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -1447,15 +1447,21 @@ public void callForHelp(final Entity attacker, final int guardHelpRange)
...
                 {
-                    guardDistance = tdist;
-                    guard = entry.getEntity().get();
+                    possibleGuards.add(entry.getEntity().get());
                 }
             }
         }
 
-        if (guard != null)
+        Collections.sort(possibleGuards, Comparator.comparingInt(guard -> (int) getPosition().distanceSq(guard.getPosition())));
```
Raycoms, line 1453: int i = 0; ;
while (i < CALL_GUARDS_CONSTANT && i < possibleGuards.size)
callToHelpAti?

src/main/java/com/minecolonies/coremod/entity/pathfinding/PathJobCanSee.java
Raycoms: Let's say I did that. I'd be fine for them to go a bit up to hit the target, but not to go much further down (compared to the guarding position)

marchermans (dismissed):

PR #5502, opened 2020-08-04
Add support for skills enhancing smeltercrafter behavior

Mekle001:
- PrimarySkill is now used to determine how many furnaces are possible to use of what the hut has
- Smeltercrafter will now break down jobs across the possible furnaces for faster smelting of smaller jobs
- SecondarySkill is now used to speed up furnace smelting. At minimum, SmelterCrafter furnaces burn ~10% faster and can be MUCH faster at higher skills. 

Review please


src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIRequestSmelter.java
```diff
@@ -63,7 +65,9 @@ public AbstractEntityAIRequestSmelter(@NotNull final J smelteryJob)
            * Check if tasks should be executed.
            */
           new AITarget(START_USING_FURNACE, this::fillUpFurnace, 1),
-          new AITarget(RETRIEVING_END_PRODUCT_FROM_FURNACE, this::retrieveSmeltableFromFurnace, 1));
+          new AITarget(RETRIEVING_END_PRODUCT_FROM_FURNACE, this::retrieveSmeltableFromFurnace, 1),
+          new AIEventTarget(AIBlockingEventType.AI_BLOCKING, this::accelerateFurnaces, 20)
+          );
```
Raycoms, line 70: possibly use the TICKS_SECOND constant here.
And the retrieve smeltable check, can we do this less often? every 5 ticks possibly?

PowerSchill:
Shouldn't the building hut level dictate the number of furnaces? An alternative suggestion might be to make the primary skill dictate how efficiently they use the fuel?

Mekle001:
The building hut sets how many furnaces are available, but a high hut low skill worker can't take full advantage of the superior tools. And a high skill worker and low level hut needs a better hut.  The number of furnaces per hut level varies widely by style. 

A maxed out skill effectively means 10 furnaces that can be used by the worker. I believe that's more than are available in any level 5 smeltercrafter.

Raycoms:
Yeah the building detect how much they have. The worker level how much they can actually use in parallel.

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5502/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PowerSchill:
Since it is a change to gameplay it probably needs to be called out.

>Hint: The higher the level of the Smeltery, the more furnaces the Smelter will be able to use. A higher level will also have a higher chance to double and even triple the ingot output per block of ore. The Smelter’s worker level will sometimes determine the chance to double or triple the ingot output of ores, according to the Smeltery level.

Raycoms:
Different smelter

PowerSchill:
>Hint: The higher the level of the Stone Smeltery, the more furnaces the Stone Smelter will be able to use.

Better? :)

Mekle001:
Some, this is the stonesmelter, the cook's assistant, the glassblower and the dyer. ;p

I'll talk to Raven

ravenbuilder934:
*wakes up* Hmmm? What? Who? Where?

someaddons (approved):

ravenbuilder934:
Cook's assistant? I thought the assistant only crafted.

PR #5572, opened 2020-08-14
Initial version of BetterRecipes

Mekle001:
- Add a check to normal crafting to possibly improve the recipe
- Adds support for recipe replacement, to avoid changing order of resolution
- Adds new tags to control when recipe modification can happen, requires explicit ingredient marking, and can have products excluded

While I believe this is well tested and basically ready to go, I left debug logging in for the moment, as I wanted to get feedback on the probabilities before I did the final cleanup. 

Review please


src/main/resources/data/minecolonies/tags/items/reduceable_product_excluded.json
```diff
@@ -0,0 +1,11 @@
+{
+  "replace": false,
+  "values": [
+    "#forge:stone",
+    "#forge:cobblestone",
+    "#forge:gravel",
+    "#forge:sand",
+    "#forge:ingots",
+    "#forge:storage_blocks"
+  ]
```
Raycoms, line 10: There is an overlap with the above one? I think only storage_blocks should be excluded?

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -265,6 +267,59 @@ public boolean canRecipeBeAdded(final IToken<?> ignored)
...
+        final double actualChance = Math.min(5.0, (baseChance * count) + (baseChance * citizen.getCitizenSkillHandler().getLevel(getPrimarySkill())));
+        final double roll = citizen.getRandom().nextDouble() * 100;
+
+        Log.getLogger().info(this.getJobName() + ": for " + recipe.getPrimaryOutput().getItem() + " improvement roll is " + roll + " with an actual chance of: " + actualChance);
+        if(roll <= actualChance && !ItemTags.getCollection().getOrCreate(reducableProductExclusions).contains(recipe.getPrimaryOutput().getItem()))
+        {
+            ArrayList<ItemStack> newRecipe = new ArrayList<>();
+            boolean didReduction = false;
+            Log.getLogger().info(this.getJobName() + ": attempting to improve recipe for: " + recipe.getPrimaryOutput().getItem());
+            for(ItemStorage input : inputs)
```
Raycoms, line 320: This seems to go through ALL ingredients to improve then. But it should pick one at random actually.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -265,6 +267,59 @@ public boolean canRecipeBeAdded(final IToken<?> ignored)
...
+    {
+        final double baseChance = 0.0625;
+        final ResourceLocation reducableIngredients = new ResourceLocation("minecolonies", "reduceable_ingredient");
+        final ResourceLocation reducableProductExclusions = new ResourceLocation("minecolonies", "reduceable_product_excluded");
+        final List<ItemStorage> inputs = recipe.getCleanedInput().stream().sorted(Comparator.comparingInt(ItemStorage::getAmount).reversed()).collect(Collectors.toList());
+
+        final double actualChance = Math.min(5.0, (baseChance * count) + (baseChance * citizen.getCitizenSkillHandler().getLevel(getPrimarySkill())));
+        final double roll = citizen.getRandom().nextDouble() * 100;
+
+        Log.getLogger().info(this.getJobName() + ": for " + recipe.getPrimaryOutput().getItem() + " improvement roll is " + roll + " with an actual chance of: " + actualChance);
```
Raycoms: This could cause logging spam

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -265,6 +267,59 @@ public boolean canRecipeBeAdded(final IToken<?> ignored)
...
+        if(roll <= actualChance && !ItemTags.getCollection().getOrCreate(reducableProductExclusions).contains(recipe.getPrimaryOutput().getItem()))
+        {
+            ArrayList<ItemStack> newRecipe = new ArrayList<>();
```
Raycoms: final List<ItemStack>

src/main/resources/data/minecolonies/tags/items/reduceable_product_excluded.json
Mekle001: The logic here is that we can reduce ingots and stone etc. when used as ingredients, unless the output is ingots or stone etc. 

This allows the stonemason to reduce the cobblestone in cobblestone stairs, but not have the crusher reduce the cobblestone in sand, as that's the responsibility of "Gilded Hammer" research. 

So, we exclude as a product the things the crusher makes, and both types of storage (ingots/storage blocks) to avoid duplication exploits.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Mekle001: Yeah, this currently shows up once per crafting job in my server. I intend to remove all the Log.getLogger()... statements before merge.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Mekle001: Good catch. Fix incoming

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Mekle001: It only improves one, and while it's not random, it will improve the one with the largest starting count.  It touches all of the ingredients to copy them to the 'new' recipe.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Raycoms: kk

src/main/resources/data/minecolonies/tags/items/reduceable_product_excluded.json
Raycoms: ahh okay

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Raycoms: kk

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -265,6 +267,59 @@ public boolean canRecipeBeAdded(final IToken<?> ignored)
...
 
+    /**
+     * Has a chance to reduce the resource requirements for the recipe in this building
+     * @param recipe the recipe we're possibly improving
+     * @param count the number of items (chances) 
+     * @param citizen The citizen, as the primary skill can improve the chances
+     */
+    public void improveRecipe(IRecipeStorage recipe, int count, ICitizenData citizen)
+    {
+        final double baseChance = 0.0625;
```
someaddons: could use some more constants for the numbers and strings used, instead of "minecolonies" use MOD_ID

Mekle001: Where is MOD_ID defined, so I add the right Import?

someaddons: import static com.minecolonies.api.util.constant.Constants.MOD_ID;

someaddons: doesnt VS code find that? :D

Mekle001: It didn't find it on it's own, since we already had a minecraft constants included. I found it by looking at other references.

PR #5581, opened 2020-08-14
Add notification for Recipe Improvement

Mekle001:
- Notification will be sent out to officers when a recipe is improved

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
@@ -349,13 +349,15 @@ public abstract class AbstractBuildingWorker extends AbstractBuilding implements
                 replaceRecipe(recipe.getToken(), IColonyManager.getInstance().getRecipeManager().checkOrAddRecipe(storage));
 
                 // Expected parameters for RECIPE_IMPROVED are Job, Result, Ingredient, Citizen
-                for(PlayerEntity player :colony.getMessagePlayerEntities())
-                {
-                    player.sendMessage(new TranslationTextComponent(RECIPE_IMPROVED + citizen.getRandom().nextInt(3),
+
+                final TranslationTextComponent message = new TranslationTextComponent(RECIPE_IMPROVED + citizen.getRandom().nextInt(3),
                         new TranslationTextComponent(citizen.getJob().getName().toLowerCase()),
                         recipe.getPrimaryOutput().getDisplayName(),
                         reducedItem.getDisplayName(),
-                        citizen.getName()));
+                        citizen.getName());
+                for(PlayerEntity player :colony.getMessagePlayerEntities())
+                {
+                    player.sendMessage(message);
                 }
             }
         }
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
@@ -47,6 +47,7 @@ import net.minecraft.util.Tuple;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.text.ITextComponent;
 import net.minecraft.util.text.StringTextComponent;
+import net.minecraft.util.text.TranslationTextComponent;
 import net.minecraftforge.common.util.Constants;
 import net.minecraftforge.fml.network.NetworkHooks;
 import net.minecraftforge.items.IItemHandler;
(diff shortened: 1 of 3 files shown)


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5581/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1602,7 +1602,10 @@
   "com.minecolonies.coremod.university.researchconcluded.0": "Your researchers just successfully concluded: %s",
   "com.minecolonies.coremod.university.researchconcluded.1": "The researchers in your colony just unravelled the mystery of: %s",
   "com.minecolonies.coremod.university.researchconcluded.2": "The Research about %s was successfully demystified by your researchers",
-
+  "com.minecolonies.coremod.crafters.recipeimproved.0": "Your %s just successfully improved their recipe for %s by reducing the need for %s",
+  "com.minecolonies.coremod.crafters.recipeimproved.1": "%4$s just successfully reducing the need for %3$s when they are making %2$s",
```
```suggestion
  "com.minecolonies.coremod.crafters.recipeimproved.1": "%4$s just successfully reduced the need for %3$s when they make %2$s.",
```

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1602,7 +1602,10 @@
   "com.minecolonies.coremod.university.researchconcluded.0": "Your researchers just successfully concluded: %s",
   "com.minecolonies.coremod.university.researchconcluded.1": "The researchers in your colony just unravelled the mystery of: %s",
   "com.minecolonies.coremod.university.researchconcluded.2": "The Research about %s was successfully demystified by your researchers",
-
+  "com.minecolonies.coremod.crafters.recipeimproved.0": "Your %s just successfully improved their recipe for %s by reducing the need for %s",
```
```suggestion
  "com.minecolonies.coremod.crafters.recipeimproved.0": "Your %s just successfully improved their recipe for %s by reducing the need for %s.",
```

ravenbuilder934 (changes requested):

marchermans (dismissed):

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -343,6 +347,16 @@ public void improveRecipe(IRecipeStorage recipe, int count, ICitizenData citizen
                     Blocks.AIR);
 
                 replaceRecipe(recipe.getToken(), IColonyManager.getInstance().getRecipeManager().checkOrAddRecipe(storage));
+
+                // Expected parameters for RECIPE_IMPROVED are Job, Result, Ingredient, Citizen
+                for(PlayerEntity player :colony.getMessagePlayerEntities())
```
someaddons: pretty sure we had a LanguageHandler function to send it to all players, should check that out

someaddons (dismissed):
looks fine

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Mekle001: I did, and it didn't work with the replacements correctly. I looked at the source for it in Structurize, and it looks overly complicated. We probably need to review a bunch of our LanguageHandler stuff, and look at where it makes sense to use MessageFormatter rather than String.format.

Nightenom: There should be two versions, one uses string.format, second is the big thing for weird formatting used back in old minecolonies and left for compatibility reasons (tho ray is still using it and I still hadn't got how it works)

Nightenom: also you definitely should reuse the message instead of creating it every time

Raycoms: You should be able to give the Languagehandler version the TranslateableTextComponent

Mekle001: Only one of the functions in LanguageHandler.java takes a player collection, sendPlayersMessage. That's the one that creates a chat component and didn't work correctly.  

sendMessageToPlayer sends to an individual player, by creating a stringTextComponent wrapped around a translateKeyWithFormat, which is a simple string.Format on the resulting key. 

Nothing in LanguageHandler takes a ITextComponent as it's only parameter. 

I think what I'm doing here is actually the most correct way to accomplish this (except the message reuse, which I'll fix).  If I use any of the current methods of LanguageHandler, I'm implicitly recreating the message every time anyway.

Mekle001: To be clear: I believe we need to add this functionality (in several methods) to LanguageHandler, and I will start on that change, once I figure out appropriate naming that doesn't create a prototype that conflicts with what is already there.

marchermans (approved):

PR #5598, opened 2020-08-17
Bring skill effects into alignment with plan for crafters

Mekle001:
- Set most crafters to have primary skill improve recipes and secondary skill effect speed
- Blacksmith and Fletcher are the reverse
- Make furnace burn faster for Smelter based on skill
- Fix crusher to actually resolve clay balls by removing old code
- Add Melon and coarse dirt to Farmer 
- Add new auto-resolve recipe for colored wool to white wool
    - This will pick the colored wool with the most stock in the warehouse to dye white
- Resolve multiple recipes with the same output at the same crafter
    - This will pick the recipe by figuring out what the majority ingredient is, and choosing the one with the most stock at the warehouse for that ingredient

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/buildings/IBuildingWorker.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/IBuildingWorker.java
@@ -208,13 +208,6 @@ public interface IBuildingWorker extends IBuilding
     Skill getSecondarySkill();
 
     /**
-     * Crafting Speed skill
-     * @return the crafting speed skill
-     */
-    @NotNull
-    Skill getCraftSpeedSkill();
-
-    /**
      *  Recipe Improvement skill getter
      * @return the recipe improvement skill
      */
--- a/src/api/java/com/minecolonies/api/colony/buildings/IBuildingWorker.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/IBuildingWorker.java
@@ -208,6 +208,20 @@ public interface IBuildingWorker extends IBuilding
     Skill getSecondarySkill();
 
     /**
+     * Crafting Speed skill
+     * @return the crafting speed skill
+     */
+    @NotNull
+    Skill getCraftSpeedSkill();
+
+    /**
+     *  Recipe Improvement skill getter
+     * @return the recipe improvement skill
+     */
+    @NotNull
+    Skill getRecipeImprovementSkill();
+
+    /**
      * Check if the worker is allowed to eat the following stack.
      *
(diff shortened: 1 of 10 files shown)


marchermans (dismissed):

src/api/java/com/minecolonies/api/colony/buildings/IBuildingWorker.java
```diff
@@ -207,6 +207,20 @@
...
+    @NotNull
+    Skill getCraftSpeedSkill();
+
+    /**
+     *  Recipe Improvement skill getter
+     * @return the recipe improvement skill
+     */
+    @NotNull
+    Skill getRecipeImprovementSkill();
+
```
Raycoms, line 216: this should probably go into the crafter building

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingDyer.java
```diff
@@ -196,6 +208,68 @@ public void requestUpgrade(final PlayerEntity player, final BlockPos builder)
...
+        if(recipe == null && stackPredicate.test(new ItemStack(Items.WHITE_WOOL)))
+        {
+            final ResourceLocation wool = new ResourceLocation("minecraft", "wool");
+            final HashMap<ItemStorage, Integer> inventoryCounts = new HashMap<>();
+
+            final Set<IBuilding> wareHouses = colony.getBuildingManager().getBuildings().values().stream()
+                                                    .filter(building -> building instanceof BuildingWareHouse)
+                                                    .collect(Collectors.toSet());
+            
+            final List<ItemStorage> woolItems = ItemTags.getCollection().getOrCreate(wool).getAllElements().stream()
```
Raycoms, line 225: I think we should wait for the multithreaded simulation solution before doing this. This can get extremely expensive. If this is executed every call. The same thing for the "inWarehouse" general solution. I think we should wait for the threaded solution first before we take this step.

Mekle001: It's actually pretty fast. In the dyer case it ONLY queries the warehouse when trying to produce white wool. And I just submitted a change to only query the warehouse in the general case when there are duplicate output recipes.

src/api/java/com/minecolonies/api/colony/buildings/IBuildingWorker.java
Mekle001: done.

PR #5601, opened 2020-08-18
[1.16/HotFix] CME When changing job.

marchermans:
Fixing the CME caused by premature optimization applied by @someaddons


someaddons:
got a stacktrace of the issue? Can't think of a place where we'd remove an AI during iteration

someaddons:
nvm, I see the goalselector immediately removes it

someaddons (approved):
Should PR this to 1.15 aswell

Raycoms:
Is this necessary? Can't we iterate over new ArrayList<>(goals) ? (I have this also in my current upcoming PR)

PR #5636, opened 2020-08-22
Feature/improvements

Raycoms:
- Adds rack rendering
- Adds avoidance of unnecessary calls to setActive in dman
- Add way to trace racks with item (Added a HighlightManager for this sake)

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAICrafting.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAICrafting.java
@@ -15,7 +15,6 @@ import com.minecolonies.api.util.InventoryUtils;
 import com.minecolonies.api.util.ItemStackUtils;
 import com.minecolonies.api.util.Tuple;
 import com.minecolonies.coremod.colony.buildings.AbstractBuildingWorker;
-import com.minecolonies.coremod.colony.buildings.AbstractBuildingCrafter;
 import com.minecolonies.coremod.colony.jobs.AbstractJobCrafter;
 
 import net.minecraft.block.Blocks;
@@ -36,7 +35,7 @@ public abstract class AbstractEntityAICrafting<J extends AbstractJobCrafter<?, J
     /**
      * Time the worker delays until the next hit.
      */
-    protected static final int HIT_DELAY = 20;
+    protected static final int HIT_DELAY = 10;
 
     /**
      * Increase this value to make the product creation progress way slower.
--- a/src/main/java/com/minecolonies/coremod/event/HighlightManager.java
+++ b/src/main/java/com/minecolonies/coremod/event/HighlightManager.java
@@ -1,0 +1,64 @@
+package com.minecolonies.coremod.event;
+
+import com.ldtteam.structurize.util.RenderUtils;
+import com.minecolonies.api.util.Tuple;
+import com.mojang.blaze3d.vertex.IVertexBuilder;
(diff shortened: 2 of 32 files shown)


marchermans (approved):

someaddons (approved):

src/main/java/com/minecolonies/coremod/event/HighlightManager.java
```diff
@@ -0,0 +1,64 @@
...
+            final long worldTime = Minecraft.getInstance().world.getGameTime();
+            for (final Map.Entry<String, Tuple<BlockPos, Long>> entry : new ArrayList<>(HIGHLIGHT_MAP.entrySet()))
+            {
+                if (entry.getValue().getB() <= worldTime)
+                {
+                    HIGHLIGHT_MAP.remove(entry.getKey());
+                }
+                else
+                {
+                    RenderUtils.renderBox(entry.getValue().getA(), entry.getValue().getA(), 0, 1, 0, 1.0F, 0.002D, event.getMatrixStack(), linesWithoutCullAndDepth.get());
```
Mekle001, line 58: Is this only highlighting a single rack, what about connected racks?

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAICrafting.java
```diff
@@ -36,7 +35,7 @@
     /**
      * Time the worker delays until the next hit.
      */
-    protected static final int HIT_DELAY = 20;
+    protected static final int HIT_DELAY = 10;
```
Mekle001, line 38: I believe this is a significant increase in crafting speed, not just a visual change. Is that the intent?

Mekle001 (approved):

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAICrafting.java
Raycoms: Yes

src/main/java/com/minecolonies/coremod/event/HighlightManager.java
Raycoms: We don't know about connected racks in this part of the code. So it's better to just show the half that contains it. If it's a connected rack it will display the half that contains it. So either way the player will click the right one

src/main/java/com/minecolonies/coremod/event/HighlightManager.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * A position to highlight with a unique id.
+     */
+    @Nullable
+    public static final Map<String, Tuple<BlockPos, Long>> HIGHLIGHT_MAP = new HashMap<>();
+
+    /**
+     * Render buffers.
+     */
+    public static final  RenderTypeBuffers        renderBuffers            = new RenderTypeBuffers();
```
Nightenom, line 35: use the one from render world last event, preferably also call the event from there

Raycoms: I didn't call it from there just for separation on concerns. I think I can't reuse it then either because I can't close it at both places I'd guess?

Nightenom: no need to separate it, RenderTypeBuffers are the reason the entire rendering is reuseable and doesn't need tesselator anymore, side note: we should be using vanilla instance but in current event design it's impossible

PR #5661, opened 2020-08-27
Add a config option to allow multiple taverns

ToMe25:
- Add a config option to allow multiple taverns in one colony
- as suggested in ldtteam/minecolonies-features#144
-

Review please

### DIFF
--- a/src/main/resources/assets/minecolonies/lang/en_us.json
+++ b/src/main/resources/assets/minecolonies/lang/en_us.json
@@ -73,7 +73,7 @@
   "minecolonies.config.builderinfiniteresources": "Infinite Resources",
   "minecolonies.config.builderinfiniteresources.comment": "Should the Builder and Miner build without resources? (This also turns off what they produce.) [Default: false]",
   "minecolonies.config.limittoonewarehousepercolony": "Warehouse Limit",
-  "minecolonies.config.limittoonewarehousepercolony.comment": "Should there be a max of 1 warehouse per colony?. [Default: true]",
+  "minecolonies.config.limittoonewarehousepercolony.comment": "Should there be a max of 1 warehouse per colony? [Default: true]",
   "minecolonies.config.limittoonetavernpercolony": "Tavern Limit",
   "minecolonies.config.limittoonetavernpercolony.comment": "Should there be a max of 1 Tavern per colony?. [Default: true]",
   "minecolonies.config.builderbuildblockdelay": "Builder Block Delay",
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -74,6 +74,8 @@
   "minecolonies.config.builderinfiniteresources.comment": "Should the Builder and Miner build without resources? (This also turns off what they produce.) [Default: false]",
   "minecolonies.config.limittoonewarehousepercolony": "Warehouse Limit",
   "minecolonies.config.limittoonewarehousepercolony.comment": "Should there be a max of 1 warehouse per colony?. [Default: true]",
```
```suggestion
  "minecolonies.config.limittoonewarehousepercolony.comment": "Should there be a max of 1 warehouse per colony? [Default: true]",
```

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -74,6 +74,8 @@
...
+  "minecolonies.config.limittoonetavernpercolony": "Tavern Limit",
+  "minecolonies.config.limittoonetavernpercolony.comment": "Should there be a max of 1 Tavern per colony?. [Default: true]",
```
```suggestion
  "minecolonies.config.limittoonetavernpercolony.comment": "Should there be a max of 1 Tavern per colony? [Default: true]",
```

ToMe25: I didn't read it, i just copy pasted it xD

Raycoms:
to be fair, not sure if we want that. I mean having multiple warehouses makes sense to some degree and it's a config value because it's currently not really stable. But multiple taverns?

src/main/resources/assets/minecolonies/lang/en_us.json
ravenbuilder934: My fault for not seeing the period in my language PR 😄

ravenbuilder934:
London alone has over 3,000 taverns/pubs. This would be more for late-game colonies, so a colonist doesn't have to walk 10-15 chunks to get a drink. 😛

ToMe25:
Tbf its kinda useless because you can't get new citizens as long as the only free beds are in taverns,
there just wasn't any comment on the issue, and also i though it would be nice as decoration so i made it.

someaddons:
the features are a gathering of suggestions, if you want to add one you should talk with us guys before. The tavern is currently intentionally designed about beeing 1 per colony, so no we do not want to add this

ToMe25:
I will close this then, should i close the issue too?

someaddons:
leave the issue open, we got a label for won't be added. Just so that feature suggestions of the same are stacked up

ToMe25:
ok, will add that to the issue then.

PR #5662, opened 2020-08-27
Show a tooltip on not yet researched hut blocks

ToMe25:
- Add a tooltip to the huts that need researching, and aren't researched yet in the colony the player last visited.

Review please


src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1386,6 +1386,7 @@
   "com.minecolonies.coremod.research.requirementnotmet": "Minimum research requirement not met!",
   "com.minecolonies.coremod.research.started": "Research successfully started!",
   "com.minecolonies.coremod.research.havetounlock": "Build a University and research this hut to unlock it!",
+  "com.minecolonies.coremod.research.crafted": "You just crafted the %s, to build it you need research it at the university first.",
```
```suggestion
  "com.minecolonies.coremod.research.crafted": "You just crafted the %s. To build it, you must first complete its research at the University.",
```

ToMe25: you could understand that as "you have to research everything before you can build this", as "complete the research" can mean "finish researching this one thing" or "research everything".

ravenbuilder934: I replaced 'complete the research' with 'complete its research.' Does that make more sense?

ToMe25: yes, this way i can't mean "research everything" anymore.

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5662/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

ShadowProtocol:
Might I suggest adding a line to the tooltip in red or something instead for this warning? This way it is obvious *before* you craft the hut, as well as it is more visible to the player in the crafting screen (the chat is mostly obscured, and in multiplayer may not be noticed or quickly pushed out by other messages). Also it's more permanent, so even if they do craft it, the tooltip will still be there to remind them that they cannot place it yet. It also saves more chat spam.

ToMe25:
I generally like the idea to add a tooltip, but i don't think there is a good way to implement that.
You could either 
1. add a permanent tooltip that stays even after researching the building, or
2. add a tooltip that gets removed once the building is researched.

### Problems:
1. I don't think 1. is a good idea, because having a permanent red tooltip under every building hut that needs to be researched even after it is researched is annoying and also might confuse people thinking they still have to research it.
Also more tooltips can obstruct things you want to see.

2. I have looked for that a bit, but i don't think there is a good solution to implement this,
the two closest ones i found would be to either check the colony for whether or not the block is researched every time the item is asked for its tooltips(or every few seconds, and cache it),
or to store it in the nbt and remove it for all players that are in the colony when it gets researched.
The second of those solutions has the big disadvantage that it would ignore all players that aren't online atm,
while the first would only show the tooltip while the player is in a colony.

If anything i would do a temporary tooltip that only shows while in a colony in which the item isn't researched yet, but even that isn't a great solution imo.
Note: all these solutions need a custom item class afaik.

ShadowProtocol:
Yeah I was only ever thinking of something dynamic like 2 with NBT that is queried in the ``addInformation`` in the item. The only differences made past this point is performance on when and how often you update the information. Checking the inventory once as a player enters a colony (including login), and once as a research is completed for those present in the colony at the time, should be fairly simple and not too taxing as the colony information is probably already present at those times.
If we did not want to go that far maybe just a grey "Requires x research" permanently, but an added line for the player that crafted it, esp. as they craft it.

ToMe25:
I forgot that there is something event like for entering the colony, i will try to implement that later today then.

ShadowProtocol:
Oh and yes this will need a custom class, but it's as simple as extending ``BlockItem`` and overriding the ``registerBlockItem`` in ``AbstractBlockHut``

ToMe25:
I think the tooltip system is fully implemented now.
Currently it doesn't show any tooltips while the player isn't in a colony, but if that is the preferred solution i could just show all the tooltips while not in any colony.

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
```diff
@@ -269,4 +279,25 @@ public void onBlockPlacedByBuildTool(
...
+     * @return true if this block needs to be researched before building its hut.
+     */
+    public void checkResearch(final IColony colony)
+    {
+        needsResearch = false;
+    }
+
+    @Override
+    public void registerBlockItem(IForgeRegistry<Item> registry, net.minecraft.item.Item.Properties properties) {
+        registry.register((new ItemBlockHut(this, properties)).setRegistryName(this.getRegistryName()));
```
Raycoms, line 320: formatting

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -1482,6 +1484,7 @@ public void addVisitingPlayer(final PlayerEntity player)
     @Override
     public void removeVisitingPlayer(final PlayerEntity player)
     {
+        ItemBlockHut.checkResearch(null);
```
Raycoms: this will be done on each chunk a player enters, might be a bit too much?

ToMe25: I think this will only be done once, when the player leaves the colony.

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
ToMe25: sorry, missed that one.

src/main/java/com/minecolonies/coremod/colony/Colony.java
Raycoms: We do the list checks a lot here. So I'm not sure if this is called on every entry and exit of a chunk.

ToMe25: I have looked at every method calling this, and to me it looks like this is only called if an entity leaves a chunk into one that is claimed by another colony(or none), and if a player leaves a dimension.

Raycoms: Then it's fine.

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
ToMe25: done.

someaddons:
think it looks nice now 👍  :)

hmm one thing maybe, could be a bit tedious xD:
Add constants for the "School" etc research strings, gets hard to keep track where they're used without

ToMe25:
There are 160 research strings, i am not going to add them all manually as constants.
I am pretty sure that can be automated with scripts, but i don't know how, so either someone who does automates it,
or its not going to get done xD

someaddons:
nah I would only do it for those you've re-used now xD

ToMe25:
That are only 21, i might do that, where should i put them?
I feel like its kinda a waste of time to do that manually tho, but it will be faster then learning how to do it with scripts xD

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
```diff
@@ -72,6 +78,11 @@
      */
     private static final VoxelShape SHAPE = VoxelShapes.create(0.1, 0.1, 0.1, 0.9, 0.9, 0.9);
 
+    /**
+     * Whether this hut is yet to be researched in the current colony.
+     */
+    protected boolean needsResearch = false;
```
ShadowProtocol, line 88: Remember that the ``Block`` classes are merely definitions, so this is defining whether research is needed for every ``AbstractHutBlock`` blockstate, everywhere, in every world. 
This information must be stored per ``ItemStack`` in NBT. Then the check needs to search through all ``ItemStack``s in the relevant player'/s inventory for a relevant Hut and update the stack NBT.

src/main/java/com/minecolonies/coremod/items/ItemBlockHut.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        super(block, builder);
+        this.block = block;
+    }
+
+    @Override
+    public void addInformation(ItemStack stack, World worldIn, List<ITextComponent> tooltip, ITooltipFlag flagIn)
+    {
+        super.addInformation(stack, worldIn, tooltip, flagIn);
+        if (block.needsResearch())
```
ShadowProtocol: You need to check this from the stack's NBT here, as it is the only thing that is actually representing the item "instance" the player is working with.

ShadowProtocol (changes requested):
Well this works for a singe single-player world, at least, but unfortunately it's a bit more complex. I'm aware that this was a significantly more difficult task than sending a chat message :)

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
ToMe25: I will look this up later, but isn't `addInformation` client side only?
In that case its perfectly fine if its the case for every instance of the block, because its only whether or not the tooltip gets rendered for this player.

ToMe25: I looked it up, and in `BlockItem` `addInformation` is `@OnlyIn(Dist.CLIENT)`, so i think it should be fine on servers too.
I will test whether it actually works on a dedicated server, but i don't know why it wouldn't.

ToMe25: Finally figured out why it might not work xD
Its because the colony check only works server side, but is nbt really the only solution?
Couldn't the client store it somewhere, and it wouldn't be that much of a performance hit.
The client needs to know the researches anywhere anyways, can't i just use that for the check then?

Raycoms: The research effects are synched to the client afaik

ShadowProtocol: I know that addInformation is Client only, that's not what I am referring to. It's the other stuff that isn't client side only that may cause problems, like the needsResearch variable on the server (which is not purely Client based). It will be reflected through all the other blockstates. Sorry if I'm not making sense, but this problem is probably not very obvious with only one player.

ToMe25: > this problem is probably not very obvious with only one player.

It is. I just tested it, and the tooltips just aren't showing up at all on a dedicated server.
> The research effects are synched to the client afaik

Thx, i will look for them.

ToMe25: The only thing not working with the new commit, is if you leave a colony into wilderness, it keeps all the tooltips, even after going into another world.

src/main/java/com/minecolonies/coremod/items/ItemBlockHut.java
ToMe25: I prefer solutions that don't require nbt on every itemstack, even the forge docu says something like "don't use itemstack nbt if not absolutely necessary".

Raycoms: Yeah, custom nbt data on stacks might cause numerous issues as they won't be accepted by the builder for example anymore

ToMe25:
I did more testing on a dedicated server, and in singleplayer, it seems to work fine now.

someaddons:
Shadow is right, the issue is that you do a player specific setting on the block itself. Probably its confusing since it is called "Block" but is actually the BlockType, not an object for each single block. So atm you do change it for everyone once one player enters the colony/checks research. How about we simply add a tooltip to the item stating "requires research to build"

ToMe25:
> So atm you do change it for everyone once one player enters the colony/checks research. How about we simply add a tooltip to the item stating "requires research to build"

No, i am not.
I checked this, and its only for one player, because it gets only set by the client side `ColonyView` object.
Also what is it about everyone thinking i don't know what the `Block` class is, i worked with minecraft before xD

ToMe25:
The `needsResearch` property is never synchronized between client and server, and is only ever used client side, so it is set for every instance of the block, but only for that one client.
And that is what i thought was the intended goal.

someaddons:
ah true on client side only it works

ShadowProtocol:
> The `needsResearch` property is never synchronized between client and server, and is only ever used client side, so it is set for every instance of the block, but only for that one client.
> And that is what i thought was the intended goal.

Yeah I picked up on that, but was considering an inter-world solution, but you've rigged that with events instead; that works. Some of those methods and the needsResearch property should be marked with the client side annotation, mainly for clarity. Then I think we're just about right here.

ToMe25:
I don't really like the side annotation xD
Last time i tried using it on a method that was only ever called client side, it crashed when marking it as client side only.

ShadowProtocol:
Well, hey, if it works with the annotations in, at least we know the server will never mess with it in weird incomprehensible ways. But the annotation usually works for me.

If you *really* don't want to put it in, at least indicate it in the Jdocs, because part of my initial confusion was caused by unclear sidedness.

ToMe25:
I just tested it with putting it on all the methods for one of the huts, and that seems to work, so i will add it to all the others.
However as soon as i add it to `needsResearch` the server crashes on startup.
Maybe i will figure out how to prevent that, if not i will just write it in the javadoc for `needsResearch` and add the annotation everywhere else.

ToMe25:
Seems like i accidentally ~~pinged~~ *mentioned* someone xD

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
```diff
@@ -12,13 +12,17 @@
 import com.minecolonies.api.tileentities.AbstractTileEntityColonyBuilding;
 import com.minecolonies.api.tileentities.MinecoloniesTileEntities;
 import com.minecolonies.api.tileentities.TileEntityColonyBuilding;
+import com.minecolonies.coremod.items.ItemBlockHut;
+import com.minecolonies.coremod.research.UnlockBuildingResearchEffect;
```
ShadowProtocol: This is introducing a circular dependency. Maybe some of the research stuff can be boosted into the API?

ToMe25: The Item is necessary in this class, because this class is the one that needs to register it, also moving the item to the api shouldn't cause any problems.
I am however not sure how to handle `UnlockBuildingResearchEffect`, as it would be possible to not use it in here by moving the `checkResearch(IColony, String)` method to a utility class.
I don't think there is a utility class where it would fit tho.

ToMe25: This should be solved now.

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1387,6 +1387,8 @@
   "com.minecolonies.coremod.research.requirementnotmet": "Minimum research requirement not met!",
   "com.minecolonies.coremod.research.started": "Research successfully started!",
   "com.minecolonies.coremod.research.havetounlock": "Build a University and research this hut to unlock it!",
+  "com.minecolonies.coremod.research.item1": "§cTo build the %s, you first need",
+  "com.minecolonies.coremod.research.item2": "§cto complete its research in the University.",
```
ShadowProtocol: This is all looking good. My final nitpick, more of an opinion:
Seeing as the building name is right above this message in the tool-tip, it doesn't need to be included again. Also I think we can get it on one line. My suggestion is "Requires research at the University for hut construction"
EDIT: depending on how long that is in monospace it might still be better on two lines. *shrugs* however you think.

ToMe25: While i would like a message that is short enough that it can reasonably be one line, i also like the name of the Hut being in the tooltip.
Not sure what i will do about that yet.

Raycoms:
:shipit:

someaddons (dismissed):

src/main/resources/assets/minecolonies/lang/en_us.json
ToMe25: I tried shortening the tooltip until it could reasonably fit one line, but the shortest i could get it to still needs two lines to match the few vanilla tooltips.

ShadowProtocol: The name of the hut in the message is just unnecessary repetition, thus it is the least important piece of information in the tool-tip. The player will simply gloss over that information. But hey if you really want it there I suppose it makes very little difference either way.

ShadowProtocol (approved):

src/main/resources/assets/minecolonies/lang/en_us.json
ToMe25: I tried removing it, but imo that makes it too generic to fit what it should say.
Imo it stops mainly saying "You need to research me" and instead mainly says "I am a generic tooltip most of these huts have", thats why i feel like it should stay, even if it is unnecessary information.

ToMe25: I think it might work if the tooltip was even shorter, but i don't think there is anything but the hut name that can be removed without changing the message.

ShadowProtocol: Fair enough, I can see that side of things.

I mean for super bonus points you could say which research is required instead, but I don't think that's worth the effort :smile:

ToMe25: That would be nice, but the `addInformation` method doesn't have access to that.
Also the block currently can't really get that information either, because the effect doesn't seem to store that.
And i don't think a solution that requires storing that in the block anyways is worth that much effort :)

PR #5669, opened 2020-08-28
Add auto-hire from training centers

Mekle001:
- Guard towers will now attempt to hire the best from the respective training centers before hiring unemployed
- There is a new button in the Guard towers to turn off trainee hiring, it defaults to on. 
- Training centers are now 'homes' for their workers
- Added bed registration to training centers, though not all schematics have beds yet.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
@@ -809,11 +809,18 @@ public abstract class AbstractBuildingGuards extends AbstractBuildingWorker impl
             return tightGrouping;
         }
 
+        /**
+         * Is hiring from training facilities enabled
+         * @return
+         */
         public boolean isHireTrainees()
         {
             return hireTrainees;
         }
 
+        /**
+         * Set whether or not to hire from training facilities
+         */
         public void setHireTrainees(final boolean hireTrainees)
         {
             this.hireTrainees = hireTrainees;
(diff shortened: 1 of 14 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
```diff
@@ -454,7 +470,38 @@ public PlayerEntity getPlayerToFollowOrRally()
...
+                trainingCitizens = colony.getCitizenManager().getCitizens().stream().filter(x -> x.getJob() instanceof JobArcherTraining).collect(Collectors.toList());
+            }
+            else if (this.getGuardType() == ModGuardTypes.knight)
+            {
+                trainingCitizens = colony.getCitizenManager().getCitizens().stream().filter(x -> x.getJob() instanceof JobCombatTraining).collect(Collectors.toList());
+            }
+
+            if(trainingCitizens != null && !trainingCitizens.isEmpty())
+            {
+                Comparator<ICitizenData> skillComparator = Comparator.comparingInt(e -> e.getCitizenSkillHandler().getLevel(job.getPrimarySkill()));
```
someaddons: actually you can do the whole find of job + find best skill logic with one single loop instead, that should be a lot better performance wise, since you only need to find the best skilled one

Mekle001: Yup. Cleaned up.

someaddons (dismissed):
looks nice :)

PR #5695, opened 2020-09-01
Fix hiring from training facility

Mekle001:
- fix comparison to be against job, rather than citizen
- skip letting the superclass assign citizens if we assigned one, so that if there is more than one, we'll catch it the next time around

This is what happens when you integrate a refactor based on review feedback, but don't fully test.

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5695/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

someaddons (approved):

PR #5699, opened 2020-09-02
Add Min/max level of building to custom recipes

Mekle001:
- Change validity check to take building rather than colony
- Check building level against min and max levels 
- Fix the recipe removal code to queue removals during load, and execute on first access after load 
- Tweak XP pickup code to force-pickup within the larger range if it's tried for 5 seconds and failed. 

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
@@ -3,7 +3,6 @@ package com.minecolonies.coremod.colony.crafting;
 import java.util.*;
 import java.util.stream.Collectors;
 import com.google.gson.JsonObject;
-import com.ldtteam.blockout.Log;
 import org.jetbrains.annotations.NotNull;
 
 import net.minecraft.util.ResourceLocation;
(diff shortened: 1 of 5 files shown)


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5699/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
```diff
@@ -85,6 +89,19 @@ public void removeRecipe(@NotNull final JsonObject recipeJson, @NotNull final Re
      */
     public Set<CustomRecipe> getRecipes(@NotNull final String crafter)
     {
+        if(!removedRecipes.isEmpty())
+        {
+            for(ResourceLocation toRemove: removedRecipes)
+            {
+                final Optional<HashMap<ResourceLocation, CustomRecipe>> crafterMap = recipeMap.entrySet().stream().map(r -> r.getValue()).filter(r1 -> r1.keySet().contains(toRemove)).findFirst();
```
someaddons, line 96: isnt that the same as recipeMap.get(toRemove)?

Mekle001: No, recipeMap is a <String, Map<ResourceLocation, CustomRecipe>> where the string is the crafter name, since that is primary access route. During removal, I don't know which crafter it's coming from, so we're retrieving the correct submap.

someaddons: ah true, had assumed its a resloc map

someaddons (approved):

PR #5726, opened 2020-09-06
Adds more dynamic difficulty to raids

someaddons:
Colonies now have an individual raid difficulty, increasing and decreasing with success of raids for a better balanced experience. Colonies start with the lowest difficulty. Difficulty scales between 20% and 200%.
Raids start happening based on a minimum raidlevel instead of plain citizencount now.

Extra difficulty utlities for raiders:
- Teaming up: raiders get put on the same scoreboard team so they cannot hurt eachother
- Movementspeed bonus: Raiders can get a bit faster, letting them catch up to players better (high difficulty)
- Archer arrows can get pierce effect
- Archers gain range through difficulty
- Melees get slightly increased hit range at high difficulty

Citizen homes and Tavern recruits now have spawntimes in relation to missing citizen count, allowing for easier recruiting.

Fix paths not properly getting their target position and reach target flags in the pathfinding.
Fix guards sometimes starting to eat mid-combat due to not resetting their fight timer
Fix a bug where randomPos sometimes returns non-air positions.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/EntityAIWalkToRandomHuts.java
+++ b/src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/EntityAIWalkToRandomHuts.java
@@ -1,28 +1,35 @@
 package com.minecolonies.coremod.entity.mobs.aitasks;
 
+import com.minecolonies.api.colony.buildings.IBuilding;
 import com.minecolonies.api.colony.colonyEvents.EventStatus;
 import com.minecolonies.api.colony.colonyEvents.IColonyEvent;
 import com.minecolonies.api.entity.mobs.AbstractEntityMinecoloniesMob;
 import com.minecolonies.api.util.BlockPosUtil;
 import com.minecolonies.api.util.EntityUtils;
+import com.minecolonies.api.util.ItemStackUtils;
 import com.minecolonies.coremod.MineColonies;
 import com.minecolonies.coremod.colony.colonyEvents.raidEvents.HordeRaidEvent;
 import com.minecolonies.coremod.entity.pathfinding.GeneralEntityWalkToProxy;
 import net.minecraft.block.*;
 import net.minecraft.entity.ai.goal.Goal;
 import net.minecraft.fluid.Fluids;
+import net.minecraft.inventory.EquipmentSlotType;
 import net.minecraft.pathfinding.Path;
 import net.minecraft.pathfinding.PathPoint;
+import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.Direction;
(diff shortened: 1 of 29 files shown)


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5726/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/EntityAIWalkToRandomHuts.java
```diff
@@ -167,6 +174,20 @@ public void tick()
 
         if (this.isEntityAtSiteWithMove(targetBlock, 2))
         {
+            entity.swingArm(Hand.OFF_HAND);
```
marchermans: Why was this change made?

marchermans (commented):
This looks decent.
It would be nice, if you could document some more of your changes.

src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/EntityAIWalkToRandomHuts.java
someaddons: ah forgot to add this part, it is for giving raiders a small chance to steal an item when reaching a building to give players some incentive to defend their buildings aswell and not only the citizens (e.g. the extreme of recall all at the townhall and block the exit during raids)

marchermans: This is not going to work
And will cause all kinds of issues.

It is a good idea in concept, but this will cause amazingly bad problems with the RS and citizens.
I for see a bunch of citizens getting stuck because of this, aborting processing or any of a bunch of other types of problems.

So please revert this.

someaddons: I don't mind removing it, ray didnt like it much either.
But it should cause no problems for the RS at all, else any kind of outside intervention like players/hoppers etc will break it

Raycoms: yeah, in theory it shouldn't matter. If anything it will expose AI weaknesses more likely.

Raycoms: But, I wouldn't add this yet, not without some other balancing (having raiders with items not fight back and try to walk back to camp or similar)

PR #5734, opened 2020-09-07
Feature/border announcement

PowerSchill:
Closes #
Closes #
Closes #
Changes proposed in this pull request:

    Adds player announcement regarding the change in colony size when a build is completed.
    Tells the player how to increase their colony size when they try and place a building outside the colony

Review please


src/main/java/com/minecolonies/coremod/util/ChunkDataHelper.java
```diff
@@ -334,11 +339,21 @@ public static void claimChunksInRange(
                 {
                     continue;
                 }
+                else
+                {
+                    areAllChunksAdded = false;
+                }
```
Raycoms: No need for else here, it's always else here because of the continue in the if.

someaddons (approved):

PR #5827, opened 2020-09-25
Allocate builder to decorations

ShadowProtocol:
* Decorations now transition to a build screen after placement
* The resources of the decoration can be seen, and the builder can be allocated
* A confirming message is sent to the players when a decoration is placed
* Decoration names have simplified names in some places (removed the long "schematics/decorations/" prefix)
- To be ported

Review please


ravenbuilder934:
If it has a build screen, does this mean it also has Deconstruct?

ShadowProtocol:
No. This is primarily so you can allocate a builder on placement. MineColonies code doesn't actually track the decoration after it has been completed, so a deconstruct would be impossible.

ravenbuilder934:
Ah, okay

src/main/java/com/minecolonies/coremod/network/messages/server/BuildToolPlaceMessage.java
```diff
@@ -301,7 +309,10 @@ private static void handleDecoration(
                 }
             }
 
-            colony.getWorkManager().addWorkOrder(new WorkOrderBuildDecoration(schem, woName, rotation, buildPos, mirror), false);
+            WorkOrderBuildDecoration woDeco = new WorkOrderBuildDecoration(schem, woName, rotation, buildPos, mirror);
+            if (!builder.equals(BlockPos.ZERO)) woDeco.setClaimedBy(builder);
```
Raycoms: can we get the parenthesis here?

src/main/java/com/minecolonies/coremod/network/messages/server/BuildToolPlaceMessage.java
```diff
@@ -104,6 +105,8 @@ public BuildToolPlaceMessage(
         this.state = state;
     }
 
+    public BlockPos getPos() { return pos; }
```
Raycoms: potentially this properly formatted too

Raycoms: To be fair. I think it is not a good idea to give the window the message. Properly better to just give the window all the data it needs in the constructor instead of giving it the message and adding a getter here.

ShadowProtocol: Fair enough, I was tossing that one up anyway. Personally I prefer passing the object to save the long argument list, especially seeing that we still need to construct the message anyway. This way the message is being instantiated in a single place. I could also just pass in the builder BlockPos as a single extra argument to the Window? 
EDIT: although I did forget to lock access for the builder variable back to private. Oh nvm I needed that access.

Raycoms: that also works

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5827/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildDecoration.java
```diff
@@ -0,0 +1,310 @@
...
+            return;
+        }
+
+        builders.clear();
+        builders.add(new Tuple<>(LanguageHandler.format("com.minecolonies.coremod.job.Builder") + ":", BlockPos.ZERO));
+        builders.addAll(colony.getBuildings().stream()
+                .filter(build -> build instanceof AbstractBuildingBuilderView && !((AbstractBuildingBuilderView) build).getWorkerName().isEmpty()
+                        && !(build instanceof BuildingMiner.View))
+                .map(build -> new Tuple<>(((AbstractBuildingBuilderView) build).getWorkerName(), build.getPosition()))
+                .collect(Collectors.toList()));
```
someaddons, line 139: could you sort the builders by distance here? and also on the Buid Building GUI?

someaddons (dismissed):
Looks good to me

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildDecoration.java
ShadowProtocol: I don't like doing that sort of thing simply because it is not immediately apparent when you look at the list, and the user would have to learn the order, rather than one they most likely know (the alphabet). And just IMO, the point is to allow them to select the builder, not have that choice calculated for them. That said I do understand why that would be neat and convenient, and does make sense. So I'm torn. Thoughts?

Raycoms:
We could display the distance in blocks together too?

ShadowProtocol:
> We could display the distance in blocks together too?

I suppose that's the best of both worlds. I'll give it a go

ShadowProtocol:
So displaying distance works fine (not that hard), but the problem is now the label in the dropdown is way too long (we were pushing it for length before this as well). There's a little bit of room to make the whole dropdown longer, but the button texture is not that long. So for the sake of simplicity on my part I'll just do the sort by distance.

Raycoms:
:shipit:

someaddons (approved):

PR #5839, opened 2020-09-27
Feature/partial chunkload

Raycoms:
- Keep the colony chunks loaded as long as an owner/officer etc is around.
- Fix some banner pathing issues
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
+++ b/src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
@@ -53,6 +53,7 @@ public class CommonConfiguration extends AbstractConfiguration
     public final ForgeConfigSpec.BooleanValue disableCitizenVoices;
     public final ForgeConfigSpec.IntValue     diseaseModifier;
     public final ForgeConfigSpec.BooleanValue generateSupplyLoot;
+    public final ForgeConfigSpec.BooleanValue forceLoadColony;
 
     /*  --------------------------------------------------------------------------- *
      *  ------------------- ######## Command settings ######## ------------------- *
@@ -406,6 +407,7 @@ public class CommonConfiguration extends AbstractConfiguration
         disableCitizenVoices = defineBoolean(builder, "disablecitizenvoices", false);
         diseaseModifier = defineInteger(builder, "diseasemodifier", 5, 1, 100);
         generateSupplyLoot = defineBoolean(builder, "generatesupplyloot", true);
+        forceLoadColony = defineBoolean(builder, "forceloadcolony", false);
 
         swapToCategory(builder, "commands");
 
--- a/src/main/java/com/minecolonies/coremod/colony/Colony.java
+++ b/src/main/java/com/minecolonies/coremod/colony/Colony.java
@@ -449,14 +449,15 @@ public class Colony implements IColony
         lastOnlineTime = currTime;
 
         updateChildTime();
-        checkForUnload();
+        updateChunkLoadTimer();
         return false;
     }
 
     /**
      * Check if we can unload the colony now.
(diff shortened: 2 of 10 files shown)


src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
```diff
@@ -406,6 +407,7 @@ protected CommonConfiguration(final ForgeConfigSpec.Builder builder)
         disableCitizenVoices = defineBoolean(builder, "disablecitizenvoices", false);
         diseaseModifier = defineInteger(builder, "diseasemodifier", 5, 1, 100);
         generateSupplyLoot = defineBoolean(builder, "generatesupplyloot", true);
+        forceLoadColony = defineBoolean(builder, "forceloadcolony", false);
```
someaddons, line 410: rename to chunk load colony, easier understandable

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -437,9 +449,51 @@ private boolean worldTickSlow()
         lastOnlineTime = currTime;
 
         updateChildTime();
+        checkForUnload();
         return false;
     }
 
+    /**
+     * Check if we can unload the colony now.
```
someaddons, line 457: I'd call this updateChunkLoadTimer() and mention that its updating the timer or unloading the chunks when the time runs out

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -448,6 +502,7 @@ private boolean worldTickSlow()
     private boolean worldTickUnloaded()
     {
         updateChildTime();
+        checkForUnload();
```
someaddons: move this to worldTickSlow instead, here its only updated when the colony is not loaded

someaddons (commented):
some little changes needed, but fine overall

src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
Raycoms: I wanted to avoid calling it that because people will confuse it with "my colony is always chunkloaded with it even I'm not in range or offline"

src/main/java/com/minecolonies/coremod/colony/Colony.java
Raycoms: It's in both. We need it in both, else it won't unload it when the player is not around. (offline or far away)

src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
someaddons: ok

src/main/java/com/minecolonies/coremod/colony/Colony.java
someaddons: oh true we want both

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -29,6 +29,9 @@
   "minecolonies.config.generatesupplyloot": "Supply loot",
   "minecolonies.config.generatesupplyloot.comment": "Whether to generate supply ship/camps in chests occasionally. [Default: true]",
 
+  "minecolonies.config.forceloadcolony": "Force Load Colony",
+  "minecolonies.config.forceloadcolony.comment": "Whether if part of the colony is loaded, all chunks should be kept loaded by force. [Default: false]",
```
```suggestion
  "minecolonies.config.forceloadcolony.comment": "If part of the colony is loaded by an owner/officer, the entire colony should be kept loaded. [Default: false]",
```

someaddons (approved):

marchermans (approved):

PR #5852, opened 2020-10-01
Break up big crafting jobs

Mekle001:
- Break crafting jobs into chunks based on the ingredients fitting in the inventory of the crafter

This resolves the problem with cases where a crafting job requires more slots to load the recipe than the citizen has available.  this becomes very visible when the recipe uses an ingredient with a maxStackSize of 1, and the output stacks higher than the number of inventory slots on the citizen. 

Today, we have a special recipe for level 3 bakery that uses water bottles, and making a stack of that recipe needs 32 of those, with a total slot requirement of 33 + 1 for output.

The solution is to recognize the max recipe iterations we can load inputs for, and instead of creating a single crafting job of the full amount, calculate the maximum size we can process, and break it up into batches of that or smaller to accomplish the task. 

Review please


marchermans (changes requested):
This pr needs a lot more documentation. What does this achieve and how? Why does this resolver behave differently then before. What adverse effects are there when this PR is pulled?

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/core/AbstractCraftingRequestResolver.java
```diff
@@ -242,20 +245,60 @@ protected boolean createsCraftingCycle(
...
+        final AbstractResearchEffect<Double> researchEffect =  manager.getColony().getResearchManager().getResearchEffects().getEffect(INV_SLOTS, AbstractResearchEffect.class);
+        final int maxSlots = (27 + researchEffect.getEffect().intValue()) - (27 + researchEffect.getEffect().intValue()) % 8;  // retaining 1 slot per row for 'overhead'
+
+        int recipeExecutionsCount = (int) Math.ceil((double) count / requestStack.getCount());
+        int minRecipeExecutionsCount = (int) Math.ceil((double) minCount / requestStack.getCount());
+        int batchSize = recipeExecutionsCount;
+        int totalSlots = Integer.MAX_VALUE;
+
+        //Calculate how many slots are needed, and figure out the maximum number of iterations we can load inputs for into inventory
+        while (totalSlots > maxSlots) {
```
Raycoms: formatting

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/core/AbstractCraftingRequestResolver.java
```diff
@@ -242,20 +245,61 @@ protected boolean createsCraftingCycle(
             return null;
         }
 
-        return createRequestsForRecipe(manager, craftableCrafting.getPrimaryOutput(), count, minCount);
+        return createRequestsForRecipe(manager, craftableCrafting, count, minCount);
     }
 
+    /**
+     * Create the crafting request entries for the overall request
+     * Will produce multiple, if the inputs don't all fit in the crafters inventory.
```
Raycoms: I think the word "input" might be what is confusing people about the doc. Better call it "ingredients" for a recipe.

Raycoms:
@OrionDevelopment the problem is that the worker tries to craft one entire stack of the recipe first. So it will try to fill up its inventory with the ingredients and will fail (because if it needs a lot of non stackable items for example). Thus, what is should do instead is calculate good batches.

@Mekle001 I just thought about another problem. When crafting shields for example, the input is stackable but the output isn't we need something for this too.

Mekle001:
@Raycoms Output stackability is interesting. Should be a relatively easy fix.

marchermans:
Okey this is much better documented

marchermans:
Once he adds the output logic I will review it again.

Raycoms:
:shipit: should be ready @OrionDevelopment 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5852/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/core/AbstractCraftingRequestResolver.java
```diff
@@ -242,20 +245,62 @@ protected boolean createsCraftingCycle(
...
         }
 
-        return createRequestsForRecipe(manager, craftableCrafting.getPrimaryOutput(), count, minCount);
+        return createRequestsForRecipe(manager, craftableCrafting, count, minCount);
     }
 
+    /**
+     * Create the crafting request entries for the overall request
+     * Will produce multiple, if the ingredients don't all fit in the crafters inventory. 
+     * @param manager
```
marchermans: If you add the javadoc, then also make it complete.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/core/AbstractCraftingRequestResolver.java
Mekle001: Done

marchermans (approved):

marchermans:
Build is failing. But overall Okey.

PR #5875, opened 2020-10-04
Fix evaluation of custom recipes

Mekle001:
- Make RecipeStorage.equals more accurate
- fix the research evaluation to properly identify effects. 

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/research/GlobalResearchTree.java
+++ b/src/main/java/com/minecolonies/coremod/research/GlobalResearchTree.java
@@ -94,4 +94,18 @@ public class GlobalResearchTree implements IGlobalResearchTree
     {
         researchTree.values().forEach(b -> b.values().forEach(IGlobalResearch::loadCostFromConfig));
     }
+
+    @Override
+    public String getEffectIdForResearch(String id)
+    {
+        for(String branch: this.getBranches())
+        {
+            IGlobalResearch r = this.getResearch(branch, id);
+            if (r != null)
+            {
+                return r.getEffect().getId();
+            }
+        }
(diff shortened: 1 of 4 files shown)


src/main/java/com/minecolonies/coremod/research/GlobalResearchTree.java
```diff
@@ -94,4 +94,18 @@ public void loadCost()
     {
         researchTree.values().forEach(b -> b.values().forEach(IGlobalResearch::loadCostFromConfig));
     }
+
+    @Override
+    public String getEffectIdForResearch(String id)
```
Raycoms: @NotNull final

Mekle001: It should be nullable, it will just return null in that case. But I'll add the final.

Raycoms: Then @Nullable

Mekle001: Shouldn't that be on the interface?

Mekle001: I think this should be resolved.

Raycoms: in the interface it says @NotNull though =D

Mekle001: Yeah, I decided it was safe to do that.

marchermans (approved):

PR #5877, opened 2020-10-05
Broaden out the new location display

Mekle001:
- Change from AbstractBuildingCrafter.View to IBuildingWorkerView, to ensure more buildings display

Review please


ravenbuilder934:
Which display is this?

Mekle001:
The tooltip on the request in the request tree that Raycoms added in the last several days.

Raycoms:
Oh there is a second file where we have this too

PR #5886, opened 2020-10-06
Rewrite of door/gate/trapdoor toggle AI

someaddons:
Rewrite of door/gate/trapdoor toggle AI, this lets us use the same AI for blocks with similar toggle behaviour.
The idea is the same as the fenceGateAI before: we check the parts of path for toggleable blocks in range, and save their positions(previously we could only use one position). If we found some blocks the AI becomes active, while active it randomly activates one of the blocks it found. With multiple blocks in range it eventually finds a combination which allows it to go through.

Pathing now allows single trapdoor blocks as passable.
Pathing now also adds the starting node to the path, previously only nodes with a parent were added.

Example of what they now make their way through:
image

Doors no longer have double door issues, which the vanilla AI has.
Gates no longer have double gate issues either
Citizens now are able to use trapdoors, e.g. when going down/up a ladder

Review please


MatthiasMann:
I have used open trap doors on the wall to block citizens from walking into my house - I assume with this PR this no longer works?
Is there another way to block pathing for a citizen while allowing the player to walk through it (idialy a special floor block which can be placed below doors)?

ravenbuilder934:
Does this close https://github.com/ldtteam/minecolonies-features/issues/149?

someaddons:
they only can walk through single trapdoors, if you stack two they'll ignore them

someaddons:
and yes it should resolve that, why is that even in features thats more a bug report xD

ravenbuilder934:
*shrugs* Ask Ray, he's the one who transferred it

src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAIInteractToggleAble.java
```diff
@@ -0,0 +1,254 @@
...
+     * Only trapdoors and fence gates
+     */
+    public static final Predicate<BlockState> GATE_TRAP = new Predicate<BlockState>()
+    {
+        @Override
+        public boolean test(final BlockState state)
+        {
+            final Block block = state.getBlock();
+            return block instanceof DoorBlock || block instanceof FenceGateBlock || block instanceof TrapDoorBlock;
+        }
```
Raycoms, line 59: Why aren't those just <Block> predicates if we don't need the state anyway?

Raycoms: "Block predicate" I mean

someaddons: wanted to give them the chance to also add state property conditions, like state.getProperty(OPEN) == true

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5886/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

Mekle001 (approved):

PR #5896, opened 2020-10-09
Feature/immutable storage

### LINKED ISSUE #5721: Request system asks for thousands of items
### Prerequisites 

### Context 

- Minecraft Version: 1.15.2
- MineColonies Version: 0.13.223
- Structurize Version: 0.13.64
- Related Mods and their Versions: forge ver. 31.2.36

## Expected behavior

Dyer requests only the amount of items required to craft the blocks/items requested by the builder.

## Actual behavior

Built a dyer, taught some recipies, dyer started requesting thousands of items.

image

1.  Create dyer
2.  Teach dyer recipies (in my case light gray dye from oxeye daisies, light gray wool from light gray dye and white wool)
3.  Ask builder to build something that requires light gray wool
4. Request system borks itself

### Logs

- latest.log: https://gist.github.com/Tenebranas/2d829d0af0063aef8a545a725574cb52 

- crashlog: 

### Notes

---
## Viewers

Raycoms:
Closes #5721

- Fixes smelter recipes increasing in cost exponentially
- Adds code to prevent this kind of bug in the future
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
@@ -1,0 +1,33 @@
+package com.minecolonies.api.crafting;
+
+import net.minecraft.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Immutable ItemStorage version.
+ */
+public class ImmutableItemStorage extends ItemStorage
+{
+    /**
+     * Creates an instance of the storage.
+     *
+     * @param storage the mutable itemstorage to create it from.
+     */
+    public ImmutableItemStorage(@NotNull final ItemStorage storage)
+    {
+        super(storage.getItemStack(), storage.ignoreDamageValue(), storage.ignoreNBT());
+        super.setAmount(storage.getAmount());
+    }
+
+    @Override
+    public ItemStack getItemStack()
+    {
+        return super.getItemStack().copy();
+    }
+
+    @Override
+    public void setAmount(final int amount)
+    {
+        throw new UnsupportedOperationException("Immutable instance of ItemStorage can't set value!");
+    }
+}
--- a/src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
@@ -20,12 +20,6 @@ public class ImmutableItemStorage extends ItemStorage
     }
 
     @Override
-    public ItemStack getItemStack()
-    {
-        return super.getItemStack().copy();
-    }
-
-    @Override
     public void setAmount(final int amount)
     {
         throw new UnsupportedOperationException("Immutable instance of ItemStorage can't set value!");
--- a/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -137,22 +137,23 @@ public class RecipeStorage implements IRecipeStorage
         final int neededMultiplier = CraftingUtils.calculateMaxCraftingCount(qty, this);
         final List<ItemStorage> items = getCleanedInput();
 
-        for (final ItemStorage stack : items)
+        for (final ItemStorage storage : items)
         {
+            final ItemStack stack = storage.getItemStack();
             final int availableCount = InventoryUtils.getItemCountInItemHandlers(
               ImmutableList.copyOf(inventories),
               itemStack -> !ItemStackUtils.isEmpty(itemStack)
-                             && ItemStackUtils.compareItemStacksIgnoreStackSize(itemStack, stack.getItemStack(), false, true));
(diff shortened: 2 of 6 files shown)


src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
```diff
@@ -0,0 +1,27 @@
...
+    public ImmutableItemStorage(@NotNull final ItemStorage storage)
+    {
+        super(storage.getItemStack(), storage.ignoreDamageValue(), storage.ignoreNBT());
+        super.setAmount(storage.getAmount());
+    }
+
+    @Override
+    public void setAmount(final int amount)
+    {
+        throw new UnsupportedOperationException("Immutable instance of ItemStorage can't set value!");
```
Asherslab, line 25: So question, could we make this throw an exception in dev, but simply log an error in prod? 
just so we don't get crashes at least during prod if it's somehow missed in dev.

otherwise looks great

src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
```diff
@@ -0,0 +1,27 @@
+package com.minecolonies.api.crafting;
+
+import net.minecraft.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Immutable ItemStorage version.
+ */
+public class ImmutableItemStorage extends ItemStorage
```
someaddons, line 9: its nice that the setamount can now be immutable, but wasnt the most common problem that the internal stack got modified? which we still can do with this version. should we maybe make an ItemStack wrapper for that case?

Raycoms: The actual problems I fixed were that the count of this storage were changed. I did check into ItemStack wrappers, but then thought that would become an overkill. It's because on creating a new ItemStack we gotta copy the old one, which needs to go in the constructor of the itemStack which itself calls the capability events which can be very slow if mods mess it up.

src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
Raycoms: I actually do want it to crash "for now". If this here happens there would've been data corruption and just logging could potentially result in infinite loops and stackoverflow.

src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
Mekle001: I know we'd talked about possibly doing some corruption checks on the retrieval of the item stack from the ItemStorage. It should be as simple as checking that if the amount is less than maxstack, the count == amount, etc.. 

At least then we'd have early warning that the internal stack was getting modified when it shouldn't be.

src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
Mekle001: To be clear, this is 'corruption' is in-memory only, at least in the scenarios we've seen so far, since it's been on non-persisted ItemStorages. Correct?

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -137,22 +149,23 @@ public boolean canFullFillRecipe(final int qty, @NotNull final IItemHandler... i
         final int neededMultiplier = CraftingUtils.calculateMaxCraftingCount(qty, this);
         final List<ItemStorage> items = getCleanedInput();
 
-        for (final ItemStorage stack : items)
+        for (final ItemStorage storage : items)
         {
+            final ItemStack stack = storage.getItemStack();
```
Mekle001, line 154: I get why you're doing this, and overall I love this change. I'm not looking forward to the merge conflict with my version that adds multiple outputs and predetermined secondary outputs based asking the IRecipe for them, etc.

Mekle001 (approved):

src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
Raycoms: Yes, the problem is that if we don't increase the count here, but loop over this and use the count in a loop, this would be an infinite loop. I want to prevent that chance.

src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
Raycoms: I was thinking about easy ways to detecting it, but couldn't find any easy way that couldn't break in valid cases.

src/api/java/com/minecolonies/api/crafting/ImmutableItemStorage.java
Mekle001: Understood, and agreed.

someaddons (approved):

marchermans (approved):

PR #5918, opened 2020-10-13
Added a recipe status as per ldtteam/minecolonies-features#152

### LINKED ISSUE #152: Displaying open crafting recipe slots for crafters
### Prerequisites 

- [x] I am running the latest **alpha** version of Structurize for my Minecraft version.

### Context 

- Minecraft Version: 1.15.1
- MineColonies Version: 0.13.195-Alpha

It can be tedious to know how many recipes you thought a crafter (I.E. sawmill) and how many are still open on them. Since it becomes tedious in higher level crafters due to having 100+ recipes thought.

## Feature description
Having number(s) displayed in the recipe GUI off crafters (I.E. sawmill) for how many recipes it has been thought, and how many new recipes it can still be thought. it could be a simple 5 out off 25 being displayed or something more fancy like a second tab on clipboard that shows all crafters and how many recipe slots that been taken.

### Notes or related things
It would also help with accessibility to newer players, since those might not know off the top from their head how many recipes can be thought per level.
---
## Viewers

GaeaKat:
Closes ldtteam/minecolonies-features#152

- Adds an x of y line to the recipes screen detailing how many recipes used and what the max number is.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
@@ -83,7 +83,7 @@ public class WindowListRecipes extends Window implements ButtonHandler
      * Life count.
      */
     private int lifeCount = 0;
-    
+
     /**
      * Constructor for the window when the player wants to see the list of a building's recipes.
      *
--- a/src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
@@ -4,13 +4,16 @@ import com.ldtteam.blockout.Pane;
 import com.ldtteam.blockout.controls.Button;
 import com.ldtteam.blockout.controls.ButtonHandler;
 import com.ldtteam.blockout.controls.ItemIcon;
+import com.ldtteam.blockout.controls.Label;
 import com.ldtteam.blockout.views.Box;
 import com.ldtteam.blockout.views.ScrollingList;
 import com.ldtteam.blockout.views.Window;
+import com.ldtteam.structurize.util.LanguageHandler;
 import com.minecolonies.api.colony.IColonyView;
 import com.minecolonies.api.crafting.IRecipeStorage;
 import com.minecolonies.api.util.constant.Constants;
 import com.minecolonies.coremod.Network;
+import com.minecolonies.coremod.colony.buildings.AbstractBuildingCrafter;
 import com.minecolonies.coremod.colony.buildings.AbstractBuildingWorker;
 import com.minecolonies.coremod.network.messages.server.colony.building.worker.AddRemoveRecipeMessage;
 import com.minecolonies.coremod.network.messages.server.colony.building.worker.ChangeRecipePriorityMessage;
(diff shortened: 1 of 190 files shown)


GaeaKat:
image
image
image

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingCrafter.java
```diff
@@ -181,6 +181,16 @@ public static boolean canBuildingCanLearnMoreRecipes(final int buildingLevel, fi
         return (Math.pow(2, buildingLevel) * EXTRA_RECIPE_MULTIPLIER) >= (learnedRecipes + 1);
     }
 
+    /**
+     * Returns the maximum number of recipes a building may have
+     * @param buildingLevel the building level
+     * @return maxinum number of recipes
+     */
+    public static int buildingMaxRecipes(final int buildingLevel)
```
Mekle001: If we're going to have a central place like this, we need to update the "canBuildingCanLearnMoreRecipes" code to use it, otherwise we have a chance of it getting out of sync.

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
```diff
@@ -147,6 +157,7 @@ public void updateElement(final int index, @NotNull final Pane rowPane)
     public void onUpdate()
     {
         updateRecipes();
+        recipeStatus.setLabelText(LanguageHandler.format("com.minecolonies.coremod.gui.workerhuts.recipestatus",building.getRecipes().size(),AbstractBuildingCrafter.buildingMaxRecipes(building.getBuildingLevel())));
```
Mekle001: You should not use a static accessor here for buildingMaxRecipe. There is a possibility not every building has the same number of recipes, so you need it to be overridable.  A good example of this is the lumberjack, where the recipe count and max recipes are always the same.

Mekle001 (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingCrafter.java
Mekle001: thinking more about this, you probably need a static and an instance version of this code. I think the instance should be _public int getMaxRecipeCount()_

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
MatthiasMann: And it needs to take the research for more recipe slots into account.

marchermans: Also constants!
All translationkeys go in TranslationKeys

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingCrafter.java
marchermans: We should never ever have a static version of this code.
Please refactor this so that it is building/worker specific.

There is too much context required to make this calculation even remotely feasible as a static method.

marchermans (changes requested):
Remove the static method.
It is not context sensitive enough

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingCrafter.java
Raycoms: Yeah, ask the building in the view sync for the max and synch it over instead of this additional static method

PR #5939, opened 2020-10-17
Feature/building rework 0.1

### LINKED ISSUE #5934: Tavern "Recall Citizens" doesn't work
### Prerequisites 

### Context 

- Minecraft Version: 1.15.2
- MineColonies Version: MineColonies Official pack v2.9
- Structurize Version: MineColonies Official pack v2.9
- Related Mods and their Versions: none

## Expected behavior
The button to recall citizens in the House GUI works, so I would assume the same button in the tavern works as well.

## Actual behavior
Nothing happens when I press it.

1. Build a tavern
2. House a couple people there
3. Press "Recall Citizens"

### Logs

This is on the 1.15 patreon server, unless you need client logs I can't get them :p

- latest.log: 
- crashlog: 

### Notes

---
## Viewers

Raycoms:
Closes #5934

This is the first part of the building rework.

This creates a building module list as well as some interfaces and abstract classes for building-modules.
It's a first attempt to modularize buildings.

Each building has the module list and might have several modules.
Modules depend on the hutblock and are queried from it.
Buildings switching to the modularized version will become "DefaultBuildingInstances" where only the modules define the behaviour.

Modules do not need a registry for now since they are queried on building creation and then serialized/deserialized.
This also allows us to maintain backwards dependency.

We might extend this to a registry in the future if we deem it necessary. I would like to avoid it during the transition period though (afterwards the building registry could be re-used for this sake for example).

This is also the first step where only the Tavern and HomeBuilding is modularized.
In the following step I will modularize the views and allow drawing one page per module (optional).
After that I will start modularizing the other building types.

I have broken this apart into multiple pull-requests to avoid too many merge conflicts and to allow to review the base mechanics decently.

Review please


src/main/java/com/minecolonies/coremod/client/gui/WindowHutTavern.java
```diff
@@ -8,19 +8,21 @@
...
 
 import static com.minecolonies.api.util.constant.TranslationConstants.COM_MINECOLONIES_COREMOD_GUI_HOME_ASSIGN;
 import static com.minecolonies.api.util.constant.TranslationConstants.COM_MINECOLONIES_COREMOD_GUI_WORKERHUTS_LEVEL_0;
+import static com.minecolonies.api.util.constant.WindowConstants.BUTTON_RECALL;
 
 /**
  * Window for the tavern
  */
-public class WindowHutTavern extends AbstractWindowBuilding<BuildingTavern.View>
+public class WindowHutTavern extends AbstractWindowBuilding<TavernBuildingModule.View>
```
Mekle001, line 25: How is this going to work when we have multiple modules that need to participate in the view? For instance, Crafter (job queue) and Smelter (Fuel)

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -343,6 +397,65 @@ public void onDestroyed()
...
+    {
+        for (final IBuildingModule module : modules.values())
+        {
+            module.removeCitizen(citizen);
+        }
+        super.removeCitizen(citizen);
+    }
+
+    @Override
+    public boolean assignCitizen(final ICitizenData citizen)
```
Mekle001, line 438: How do you see modules working together here? Specifically, the restaurant has two citizens with very different jobs. I expect that will be two modules, each with one citizen assigned. Is it First one wins? or do we expect a priority, and thus support promotion? If the cook dies, does the assistant cook become the cook?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/LivingBuildingModule.java
```diff
@@ -385,21 +362,15 @@ private void inheritLastName(@NotNull final ICitizenData child, final String par
...
-    @Override
-    public int getMaxInhabitants()
-    {
-        return getBuildingLevel();
-    }
-
     /**
      * Looks for a homeless citizen to add to the current building Calls. {@link #assignCitizen(ICitizenData)}
      */
     private void addHomelessCitizens()
```
Mekle001, line 287: Should this be part of LivingBuildingModule, or separate? Do you expect LivingBuildingModule to be used for towers and training facilities? If so, this then creates the potential to mix the two assignment types.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutTavern.java
Raycoms: Yeah, I commented that in the description. That is the next PR I want to make. This one was getting too big already.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
Raycoms: For now, It would be by addition order (so first cook than assistant, if cook dies, assistant stays assistant and new worker will become cook). However, yes, there is room for additional functionality in the future here.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/LivingBuildingModule.java
Raycoms: These places could override this module and create a SpecializedLivingBuildingModule that they will add for example.

src/main/java/com/minecolonies/coremod/client/gui/WindowHutTavern.java
Mekle001: Fair enough.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/LivingBuildingModule.java
Mekle001: Ok, while that sends us partway down the inheritance path again, it's small enough that seems reasonable.

Mekle001 (dismissed):

src/api/java/com/minecolonies/api/colony/buildings/modules/AbstractBuildingModule.java
```diff
@@ -0,0 +1,139 @@
...
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Abstract class for all modules. Has base methods for all the necessary methods that have to be called from the building.
+ */
+public abstract class AbstractBuildingModule implements IBuildingModule
```
marchermans, line 8: This class is weird AF.
This is just an empty class which literally makes no sense.

If you want this to be the default behaviour for those empty methods for a module, then make it the default implementation of the method in the interface not in this weird abstract class.

marchermans (changes requested):
I personally think that the Module system is not up to par.

While taking into account that this a first draft, the API should be relatively cleanly implemented with this first draft, since else other people might start building on top of this none complete module API, which will cause all kinds of problems

Design problems:
- Modules are not aware of their surroundings and it is not clear what happens if multiple modules define potentially incompatible behaviour.
- Modules define "counts", for the lack of a better word, (like max level etc) but there is no agreement on how the returned values should be processed. Are they added up, does the biggest value always win? 
- The way the current implementation is done (aka via a large common interface) still runs into the limitations of the current system. (Monolithic limitations). This major interface should be broken down into at least 4 different interfaces that define more specifically a modules intentions and behaviours.

All in all this PR is far from being somewhere where it can be merged, and the approach should in my eyes be again discussed on Discord, and then the scope should be shrunk to just the initial API on the first PR (preferably even Draft)

src/api/java/com/minecolonies/api/colony/buildings/modules/AbstractBuildingModule.java
Raycoms: Yeah, true, I forgot I can set the empty methods as defaults in the interface.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -32,6 +34,27 @@
 
 public interface IBuilding extends IBuildingContainer, IRequestResolverProvider, IRequester
 {
+    /**
+     * Check if the building has a particular module.
+     * @param clazz the module of the class to check.
+     * @return true if so.
+     */
+    boolean hasModule(final Class<? extends AbstractBuildingModule> clazz);
```
marchermans: No, this should be `Class<? extends IBuildingModule>` or what ever the root interface for a module is.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -32,6 +34,27 @@
...
+
+    /**
+     * Get the module with a particular class.
+     * @param clazz the modules class.
+     * @return the module or null of not existant.
+     */
+    @Nullable
+    IBuildingModule getModule(Class<? extends AbstractBuildingModule> clazz);
```
marchermans: See above

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -32,6 +34,27 @@
...
+
+    /**
+     * Register a specific module to the building.
+     * @param module the module to register.
+     */
+    void registerModule(@NotNull final AbstractBuildingModule module);
```
marchermans: See above

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
```diff
@@ -330,4 +332,10 @@ public boolean needsResearch()
     {
         return needsResearch;
     }
+
+    @Override
+    public void registerBuildingModules(final IBuilding building)
+    {
+
```
marchermans: You might want to consider firing an event here somehow.
This to allow external APIs to inject modules. It does not need to be implemented here yet, just a comment to mark it as a TODO somehow should suffice for this PR.

src/api/java/com/minecolonies/api/colony/buildings/modules/IAssignsCitizen.java
```diff
@@ -0,0 +1,24 @@
+package com.minecolonies.api.colony.buildings.modules;
+
+import com.minecolonies.api.colony.ICitizenData;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Interface for all modules that need special assignment handling.
+ */
+public interface IAssignsCitizen
```
marchermans: Needs to extends `IBuildingModule` else this makes no sense.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+public interface IBuildingModule
+{
+    /**
+     * Deserialize the module.
+     * @param compound the nbt compound.
+     */
+    default void deserializeNBT(CompoundNBT compound)
+    {
+
```
marchermans, line 10: Formatting is of here. Remove the not needed empty lines.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+import com.minecolonies.api.colony.buildings.IBuilding;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.PacketBuffer;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Interface describing the different modules.
+ */
+public interface IBuildingModule
```
marchermans: Make this extend `NBTSerializable` or whatever the name of the vanilla interface is, and do not declare your own custom methods.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+{
+    /**
+     * Deserialize the module.
+     * @param compound the nbt compound.
+     */
+    default void deserializeNBT(CompoundNBT compound)
```
marchermans: Remove this method, see interface declaration comment.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+    {
+
+    }
+
+    /**
+     * Serialize the module from a compound.
+     * @param compound the compound.
+     */
+    default void serializeNBT(final CompoundNBT compound)
```
marchermans: Remove this method, see interface declaration comment

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+    {
+
+    }
+
+    /**
+     * Serialization method to send the module data to the client side.
+     * @param buf the buffer to write it to.
+     */
+    default void serializeToView(PacketBuffer buf)
```
marchermans: We should declare a standard interface for serialization to and from `PacketBuffer` and treat it in the same way vanilla treats NBT with its `NBTSerializable` interface. Aka create an empty instance and then call deserialize on it.

I am pretty sure this interface was added when we added the PacketBuffer specific serialization for the RS.
This also makes it so that Views are not needed for modules, which is a huge improvement in maintainability.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    default void onBuildingMove(IBuilding oldBuilding)
+    {
+
+    }
+
+    /**
+     * Specific wakeup hook in modules.
+     */
+    default void onWakeUp()
```
marchermans: Again this has no bearing on a default module.
Move to a seperate interface.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+    default void onDestroyed()
+    {
+
+    }
+
+    /**
+     * Colony tick hook.
+     * @param colony the colony the tick is invoked from.
+     */
+    default void onColonyTick(@NotNull IColony colony)
```
marchermans: Again this has no bearing being on a default module. Some modules do not tick and as such do not need this.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+    default void onWakeUp()
+    {
+
+    }
+
+    /**
+     * On player entering hook.
+     * @param player the player that entered the building.
+     */
+    default void onPlayerEnterBuilding(PlayerEntity player)
```
marchermans: Again see other methods in this class.
The base module spec should only specific the _bare_ minimum to classify itself as a module.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
```diff
@@ -0,0 +1,109 @@
...
+    {
+
+    }
+
+    /**
+     * Specific dirty marking of modules (separate from building dirty).
+     */
+    void markDirty();
```
marchermans: Move to a seperate interface not related to modules, called `IHasDirtyState` or something.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModuleProvider.java
```diff
@@ -0,0 +1,15 @@
+package com.minecolonies.api.colony.buildings.modules;
+
+import com.minecolonies.api.colony.buildings.IBuilding;
+
+/**
+ * Blocks providing building modules.
```
marchermans: We should not limit this to just blocks, adapt the javadoc.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesBuildingModule.java
```diff
@@ -0,0 +1,23 @@
+package com.minecolonies.api.colony.buildings.modules;
+
+/**
+ * Interface describing the different modules.
+ */
+public interface IDefinesBuildingModule
```
marchermans: Javadoc completely misleading. And the name is wrong: `IDefinesCoreBuildingStatsModule` is more accurate.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesBuildingModule.java
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * Get the max building level a module allows.
+     * By default this is 5.
+     * @return the max level.
+     */
+    default int getMaxBuildingLevel()
```
marchermans: Nope.
This is up to the module.
Either make both methods have a default value or none.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesBuildingModule.java
```diff
@@ -0,0 +1,23 @@
...
+    {
+        return 5;
+    }
+
+    /**
+     * Get the max number of inhabitants this module allows.
+     * @return the modules max number of assigned citizens.
+     */
+    int getMaxInhabitants();
```
marchermans: See other method in this interface.

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -283,15 +283,7 @@ public boolean isPlantable(final ItemStack itemStack)
             }
 
             String[] split = string.split(":");
-            //todo, this is backwards compatibility, remove in 1.16
```
marchermans: This clearly states remove in **1.16** why was this removed in a PR that targets _1.15_

src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
```diff
@@ -137,8 +140,8 @@ public static void init(final RegistryEvent.Register<BuildingEntry> event)
 
         ModBuildings.home = new BuildingEntry.Builder()
                               .setBuildingBlock(ModBlocks.blockHutHome)
-                              .setBuildingProducer(BuildingHome::new)
-                              .setBuildingViewProducer(() -> BuildingHome.View::new)
+                              .setBuildingProducer((colony, blockPos) -> new DefaultBuildingInstance(colony, blockPos, "citizen", ModBuildings.home))
```
marchermans: This really really really feels like a code smell in my opinion. The building instance really should not care what building it is representing, neither should the modules which are inside it.

src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
```diff
@@ -298,8 +301,8 @@ public static void init(final RegistryEvent.Register<BuildingEntry> event)
 
         ModBuildings.tavern = new BuildingEntry.Builder()
                                 .setBuildingBlock(ModBlocks.blockHutTavern)
-                                .setBuildingProducer(BuildingTavern::new)
-                                .setBuildingViewProducer(() -> BuildingTavern.View::new)
+                                .setBuildingProducer((colony, blockPos) -> new DefaultBuildingInstance(colony, blockPos, "tavern", ModBuildings.tavern))
```
marchermans: See above.

marchermans (changes requested):
### Improvement
_First things first:_
This is an improvement compared to what it was before the monolithic design of the initial primary interface has been somewhat reduced, but this is still not mergeable.
The new _module_ interface do not extend the existing primary interface, and abstract classes are used where interface should be used.

### Required changes:
1. A lot more thought has to be put in, into how this system is going to behave. This literally has to be clear **now**, like when this is merged. What can it do, what can it not do? Who controls what? How are values processed? Have modules the option to overrule others? etc. All of this is completely not defined. And that will cause chaos along the road. Write all of this out somehere (preferrably in this PR using a somekind of readme file).
2) You have to figure out how you want this to be extensible in the future. Both from a Minecolonies perspective as well as from an external mod perspective. Put yourself in their shoes and then design something that treats minecolonies basically as an external mod. You do not need to implement all the related infrastructure at once, but a start should be visible here.

## TLDR:
This PR misses a vision for the future. It states a couple of times that registries are not needed, and that stuff will be added in the future. But it makes in no way clear what it is going to add. To that end, this PR feels very Minecolonies centric, and it looks extremely hard to extend or override functionality from a different module in the current spec.

Again this PR should probably go back to the drawing board stage. Design a system that works for everybody. Then first implement the basic mechanics and only then start converting buildings. Instead of implement something without some spec, already doing buildings, but only half since the views are still somehow in play. And all of this without a clear path for the future.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesBuildingModule.java
Raycoms: It's because maxBuildinglevel = 5 applies to 99% of our buildings. But the number of workers/inhabitants varies greatly. So it makes sense to have a default here but not in the other.

src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
Raycoms: The building instance absolutely does need it. Necessary for the schematics etc

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
Raycoms: The method signature we use is slightly different than NBTSerializable, so not possible

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModule.java
Raycoms: As I wrote above, views will be a part of modules. In a way that allows us to build nice GUIs from multiple module Views

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
```diff
@@ -0,0 +1,24 @@
...
+    default int getMaxBuildingLevel()
+    {
+        return 5;
+    }
+
+    /**
+     * Get the max number of inhabitants this module allows.
+     * @return the modules max number of assigned citizens.
+     */
+    int getMaxInhabitants();
```
marchermans: So I thought about this a bit.
I agree that these are indeed the core stats.
But I think we can flush out the API a bit here.
Hit me up on Discord we jump in voice and I will explain what my thoughts here were how we can make this a bit more precise.

src/api/java/com/minecolonies/api/colony/buildings/modules/IRegistersBlockModule.java
```diff
@@ -0,0 +1,20 @@
...
+
+import net.minecraft.block.BlockState;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Module type to register specific blocks to a building (beds, workstations, etc).
+ */
+public interface IRegistersBlockModule extends IBuildingModule
```
marchermans: Rename to `IModuleWithExternalBlocks`

src/api/java/com/minecolonies/api/colony/buildings/modules/IRegistersBlockModule.java
```diff
@@ -0,0 +1,20 @@
...
+{
+    /**
+     * Attempt to register a specific block at a specific module.
+     * @param blockState the state.
+     * @param pos the position.
+     * @param world the world.
+     */
+    void registerBlockPosition(@NotNull BlockState blockState, @NotNull BlockPos pos, @NotNull World world);
```
marchermans: Rename to `onBlockBuildInBuilding`

src/api/java/com/minecolonies/api/colony/buildings/modules/IRegistersBlockModule.java
```diff
@@ -0,0 +1,20 @@
...
+}
```
marchermans: You should add a method that returns all Blocks registered to this module.
Something like `getRegisteredPositions` or `getRegisteredBlocks(final IBlockReader world)`

src/api/java/com/minecolonies/api/colony/buildings/modules/IStoresDataModule.java
```diff
@@ -0,0 +1,28 @@
+package com.minecolonies.api.colony.buildings.modules;
+
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.PacketBuffer;
+
+/**
+ * Interface for all building modules that store additional data.
+ */
+public interface IStoresDataModule extends IBuildingModule
```
marchermans: Rename to: `IModuleWithData`.
Also extend the existing interface.

marchermans (changes requested):
This is getting really close, and I like it!
Good job.

Still not completely happy, contact me on discord.

src/api/java/com/minecolonies/api/colony/buildings/modules/IRegistersBlockModule.java
Raycoms: I thought about this, but each building holds the blocks differently for different purposes and sometimes even different sets of blocks. I don't think we need a getter that returns ALL of them

src/api/java/com/minecolonies/api/colony/buildings/modules/IModuleWithExternalBlocks.java
```diff
@@ -5,16 +5,24 @@
...
      * @param world the world.
      */
-    void registerBlockPosition(@NotNull BlockState blockState, @NotNull BlockPos pos, @NotNull World world);
+    void onBlockPlacedInBuilding(@NotNull BlockState blockState, @NotNull BlockPos pos, @NotNull World world);
+
+    /**
+     * Get the list of registered blocks.
+     * @return the list of positions of the blocks.
+     */
+    List<BlockPos> getRegisterBlocks();
```
marchermans: `getRegisteredBlocks()`

src/api/java/com/minecolonies/api/colony/buildings/modules/stat/IStat.java
```diff
@@ -0,0 +1,17 @@
+package com.minecolonies.api.colony.buildings.modules.stat;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A specific stat.
+ */
+@FunctionalInterface
+public interface IStat<N extends Number>
```
marchermans, line 9: Not the version shown in the LDTTeam Discord.
So question, is this intended?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/BedHandlingModule.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * The class of the citizen hut.
+ */
+public class BedHandlingModule extends AbstractBuildingModule implements IModuleWithExternalBlocks, IModuleWithData, IBuildingEventsModule
+{
+    /**
+     * List of all beds.
+     */
+    @NotNull
+    private final List<BlockPos> bedList = new ArrayList<>();
```
marchermans: You can make this a set instead of a list to make logics a bit easier.

src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAISleep.java
```diff
@@ -214,9 +215,9 @@ private void findBedAndTryToSleep()
             {
                 final IBuilding hut = colony.getBuildingManager().getBuilding(citizen.getHomePosition());
                 List<BlockPos> bedList = new ArrayList<>();
-                if (hut.hasModule(LivingBuildingModule.class))
+                if (hut.hasModule(BedHandlingModule.class))
```
marchermans, line 218: Seeing this structure makes my stumach crinch a bit.
I think we should add something like this to the building:
`
public Optional<T extends IModule> <T> getModule(Class<T> moduleClass)
`

On which we can then to chaining with things like `.map(IModuleWithExternalBlocks::getRegisteredBlocks).orElse(Collections.emptyList())`

marchermans (changes requested):
Looking very good.
Some minor nitpicking.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
```diff
@@ -0,0 +1,26 @@
...
+    default IStat<Integer> getMaxBuildingLevel()
+    {
+        return (prev) -> prev == 0 ? 5 : prev;
+    }
+
+    /**
+     * Get the max number of inhabitants this module allows.
+     * @return the modules max number of assigned citizens.
+     */
+    IStat<Integer> getMaxInhabitants();
```
someaddons, line 15: I thin inhabitants belong to assignments, as you cant have inhabitants without assignments

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
```diff
@@ -0,0 +1,26 @@
+package com.minecolonies.api.colony.buildings.modules;
+
+import com.minecolonies.api.colony.buildings.modules.stat.IStat;
+
+/**
+ * Interface describing core building stats.
+ * The first core stats module that is found in the building will define the values.
+ */
+public interface IDefinesCoreBuildingStatsModule extends IBuildingModule
```
someaddons, line 9: need a more descriptive name for this, do we need this at all? afaik building levels are tied to schematics

src/api/java/com/minecolonies/api/colony/buildings/modules/IModuleWithData.java
```diff
@@ -0,0 +1,35 @@
+package com.minecolonies.api.colony.buildings.modules;
+
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.PacketBuffer;
+
+/**
+ * Interface for all building modules that store additional data.
+ */
+public interface IModuleWithData extends IBuildingModule
```
someaddons: ISerializeableModule?

src/api/java/com/minecolonies/api/colony/buildings/modules/AbstractBuildingModule.java
```diff
@@ -0,0 +1,46 @@
+package com.minecolonies.api.colony.buildings.modules;
+
+import com.minecolonies.api.colony.buildings.IBuilding;
+
+/**
+ * Abstract class for all modules. Has base methods for all the necessary methods that have to be called from the building.
+ */
+public abstract class AbstractBuildingModule implements IBuildingModule
+{
```
someaddons, line 9: this class is kinda duplicated with IHasDirty, use either or?

src/api/java/com/minecolonies/api/colony/buildings/modules/stat/IStat.java
Raycoms: yeah, I played around a bit with it and I think this one was the one that is the easiest to handle. The other ones were a bit more complex which made the code more difficult to maintain

src/api/java/com/minecolonies/api/colony/buildings/modules/AbstractBuildingModule.java
Raycoms: No this one is the implementation of IHasDirty, so this one actually tracks the boolean

src/api/java/com/minecolonies/api/colony/buildings/modules/IModuleWithData.java
Raycoms: IPersistentModule I think is better, ye

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
Raycoms: Regarding this. We had the discussion a while back once to actually have the building check on the schematics to determine it's max level. But that'd be out of the scope of the PR, so I'd like to keep it for now here. I can add a todo here though.

src/api/java/com/minecolonies/api/colony/buildings/modules/stat/IStat.java
someaddons: what is that class even for? why do we turn int returns into complex objects which require unboxing?

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
Raycoms: true

Raycoms: The problem is that the tavern and home building use the same assigning module (as will many worker modules) but the actually max check is in the core module.

someaddons: could just move both into the assign one?

src/api/java/com/minecolonies/api/colony/buildings/modules/stat/IStat.java
Raycoms: Allows us to define how certain return values might be combined (If one module returns 5 max citizens and another one also 5, should they be added, replaced, math.max? This allows us to do this more generically

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
someaddons: I actually thought we'd keep the data which is core for a building, like schematic,level and such in the actual building object itself, as thats required and not modular

Raycoms: I moved the building level there now, yes

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
Raycoms: No, that would meddle with the entire system. Both buildings use different basis for the max (tavern is max 4, home is max 5 and 1 per building level).

src/api/java/com/minecolonies/api/colony/buildings/modules/stat/IStat.java
someaddons: hm I don't think we should combine return values of different modules, those usually then also have a different meaning

Raycoms: It makes sense in some cases: Allowing assignment of workers for example (each module allows to add one worker for example, someone could add a new module to a building with an additional worker). I could imagine many different use cases for combining values.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -32,6 +33,27 @@
...
+     * @return the module or null of not existant.
+     */
+    @Nullable
+    <T extends IBuildingModule> T getModule(Class<T> clazz);
+
+    /**
+     * Register a specific module to the building.
+     * @param module the module to register.
+     */
+    void registerModule(@NotNull final IBuildingModule module);
```
marchermans, line 55: I do not think this method should be in the API.
As stated before: building modules should be immutable once the building is constructed.
So I do not see a reason to have a registerModule method.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -32,6 +33,27 @@
...
+     */
+    boolean hasModule(final Class<? extends IBuildingModule> clazz);
+
+    /**
+     * Get the module with a particular class.
+     * @param clazz the modules class.
+     * @return the module or null of not existant.
+     */
+    @Nullable
+    <T extends IBuildingModule> T getModule(Class<T> clazz);
```
marchermans: I do not like nullables.
It can cause weird issues.

Use an optional here, which forces people to think about what todo when the module is not available in the expected building automatically.

marchermans: Might additionally collapse the hasModule into this method when switched to the optional.

Computationally it is identical.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
marchermans: I am going to argue that we should release that inhabitants <-> assigning logic in the near future.
We should add modules which allow assigning workers with specific jobs to a building. But the count should be up to the player in a certain degree.

Aka it should be possible to have 1 cook and 6 helpers or
2 cooks and 1 helper.

We should envision some kind of slot like mechanism where certain assignments of jobs use more slots, and where this method returns the amount of slots available.

marchermans (changes requested):
Some minor comments on the api surface still remain.
Looks good though.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
Raycoms: Because we handle IBuilding everywhere. As an alternative we could set up sth like "you can't add modules after the initial setup" with a throw or so.

Raycoms: But we should have this in IBuilding such that we don't have to give out the AbstractBuilding or so

someaddons: hm this reminds me a bit on the cook, which through upgrading a building gets added new functionality in terms of the assistant, so might need it since buildings can change through upgrades

Raycoms:
Yeah, I want to get this rework finally going, we should have those things all ready in the near future.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
someaddons: the way I see it is that we get specialized assignment modules, e.g. one for the Worker(s) where you can assign X workers, one for Living where you can have X ppl live in. And as I see that both of them would required a max you can put into the slots

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
Raycoms: In that case the building would get the module form the beginning, but the module wouldn't "work" unless the necessary conditions are met.

marchermans: @Raycoms you should build the list of modules. When the building instance is constructed..,....
For example somewhere in the building type or something.

The building should never have this.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
Raycoms: I don't see this as anything urgent though. There are 10 more urgent things on my list before sth like that.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
Raycoms: We do this on building instantiation. But we give the building as a hook to register modules on. I don't see a problem with that.

Raycoms: I can add to the javadoc: Must be called during building instantiation.

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
someaddons: no you need to flesh this out before just merging it, else we end up with you using it in 10 places and needing a big refactor after. Not saying you need to make worker/living modules now, but put a getMaxAssignees into the Assignment base module and just remove the getMaxinhabitants here. As far as I see this there is no reason at all for this interface to exist.

someaddons (commented):
btw isnt all the walking speed stuff in here dated?

src/api/java/com/minecolonies/api/colony/buildings/modules/IDefinesCoreBuildingStatsModule.java
Raycoms: For the time being, I was happy with how the system is currently defined, having this here centrally. A few examples of problems with assignment based method:

All buildings where people sleep and work have two assignment modules, one living, one working. So each would return 5 even though the max should be 5 building based, not module based.
Same thing for other methods. So this method is here to stay. If anything, we'd have a second method in the assignment with a getModuleMax()

someaddons (approved):

PR #5921, opened 2020-10-14
Delivery orders

someaddons:
Closes #5927
Closes #5923
Closes #5811

# Changes proposed in this pull request:
This adapts the way deliveries are assigned to deliveryman to a multi-factor comparison system.
The resolver for dmen now resides in a warehouse and resolves the request it gets to its delivermen in our case those of a single warehouse. So now we compare all assigned deliverymen of a warehouse to get the best fit for the request.
The resolver now ask the dmen for a score for adding a certain request and then compares the score to other dmen's scores. The score(lower is better) itself is combined of these factors:

- Closeness: How close the target/source of the new request is compared to an existing one, in relation to its total travel length.
- Priority(also contains age): Overtaking a request of a higher priority is worse.
- Pickup/Delivery alternating: When alternating between pickups and deliveries with close targets the score gets a little bonus.
- Index: The index we would add the new request at, higher the more request we overtake and the more request we have in total.

These result in lists on deliverymen of approx ~3 times closer deliveries, allowing the AI to also bunch up deliveries more nicer in the future.

Fixes visitors sometimes getting spawn on zero pos.
Fix raids not respecting distance to unloaded buildings
Fix shipsraids sometimes instantly despawning after killing spawners, now takes a day to do so.

Review please

PR #5940, opened 2020-10-17
Ignore consequences when visitors fall out of world

mulhollandms:
Quick fix for a problem I kept encountering. In singleplayer when I'm away from the colony, I'll often get messages saying all the visitors die from "outOfWorld". 

- When visitors fall out of the world (due to unloaded chunks, etc?), the consequences such as disabling visitors to your tavern and the death message sent to the player are ignored

Review please


ravenbuilder934:
This was already fixed in https://github.com/ldtteam/minecolonies/pull/5921, update

mulhollandms:
> This was already fixed in #5921, update

I'm not sure that change is particularly relevant in this case. It appears to only address the issue when it occurs in one or two ways such as a result of spawning/creating the visitor and only raises the log error when the xz position is set to (0,0). This was intended as more of a catchall. Nevertheless I've updated and will reopen if it persists.

PR #5975, opened 2020-10-22
Add a couple more blocks to deco phase

Mekle001:
- Add Banners to deco phase of build
- Add Wall signs to deco phase of build
-

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/5975/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
```diff
@@ -238,6 +238,14 @@ public boolean walkToConstructionSite(final BlockPos currentBlock)
         return worker.isWorkerAtSiteWithMove(workFrom, STANDARD_WORKING_RANGE) || MathUtils.twoDimDistance(worker.getPosition(), workFrom) < MIN_WORKING_RANGE;
     }
 
+    /**
+     * Checks for blocks that need to be treated as deco
+     */
+    protected boolean isDecoItem(Block block)
+    {
+        return block instanceof CoralBlock || block instanceof LanternBlock || block instanceof BannerBlock || block instanceof WallSignBlock;
```
marchermans: Yeah no, this is now officially way to much OR-InstanceOf combos.
Add a config entry todo this, or add a simple list based registry in the api, regardless this is not a great maintainable approach anymore.

marchermans: Or add a Tag?
That would probably be the best approach.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
Raycoms: Yeah, a tag is probably the best thing.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
```diff
@@ -147,6 +160,8 @@ protected AbstractEntityAIStructure(@NotNull final J job)
            */
           new AITarget(COMPLETE_BUILD, this::completeBuild, STANDARD_DELAY)
         );
+
+        decoItems = BlockTags.getCollection().getOrCreate(new ResourceLocation(MOD_ID, DECOBLOCK_TAG));
```
marchermans: This is not the way to do tags.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
Mekle001: Ok, I've updated this to static, to use a better pattern for the caching of this. Given that this is the way all of our tags access is in 1.15, I think it should stay consistent here and be updated to be 'right' in 1.16 using the TagUtils class.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
```diff
@@ -80,6 +83,16 @@
...
 
+    /**
+     * Tagname for decoration blocks
+     */
+    private static String DECOBLOCK_TAG = "decoblocks";
+
+    /**
+     * Tag of all the blocks that need to be in the deco phase of the builder
+     */
+    protected static Tag<Block> decoItems;
```
marchermans: No. As discussed moved the actual tag instance to a dedicated class in the API which will group all of these.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
Raycoms: Yeah, but that's the way we do this in 1.15 everywhere (also in several PR that were approved from you in the past). In the 1.16 port of this PR we can move this there

Mekle001: I think we all are now on the same page around what you're asking for Orion, with something like a ModTags module. 

I've been thinking about it overnight, and I realized we CAN NOT simply cache these returns, as they could be redefined at any time by a /reload command reloading a new datapack. 

This means that the proposed ModTags module isn't really following the rest of our Mod* pattern.

The "best" way I can currently see is a module with getters that lazy init the instances, and a dummy reload listener that clears the caches, with a complete sweep of the codebase to update to use it. Is that really a change we want to take in 1.15, or should it be done in 1.16 as a cleanup item?

Raycoms: There is an event to do that. So we can add an event listener to that ModTags class that handles that.

src/api/java/com/minecolonies/api/items/ModTags.java
```diff
@@ -0,0 +1,20 @@
...
+import net.minecraft.tags.Tag;
+
+public class ModTags
+{
+    public static Tag<Block>    decorationItems;
+    public static Tag<Item>     concretePowder;
+    public static Tag<Block>    concreteBlock;
+    public static Tag<Block>    pathingBlocks;
+
+    public static Map<String, Tag<Item>> crafterProduct;
```
marchermans: Initialize this straight up as a final hashmap.

src/main/java/com/minecolonies/apiimp/initializer/ModTagsInitializer.java
```diff
@@ -0,0 +1,137 @@
...
+    private static final ResourceLocation DECORATION_ITEMS = new ResourceLocation(MOD_ID, "decoblocks");
+    private static final ResourceLocation CONCRETE_POWDER = new ResourceLocation(MOD_ID, "concrete_powder");
+    private static final ResourceLocation CONCRETE_BLOCK  = new ResourceLocation(MOD_ID, "concrete");
+    private static final ResourceLocation PATHING_BLOCKS = new ResourceLocation(MOD_ID, "pathblocks");
+
+
+    @SubscribeEvent
+    public static void tagsUpdated(TagsUpdatedEvent event)
+    {
+        ModTagsInitializer.init();
```
marchermans: Not needed.
We use named tags as such we do not need this.

marchermans (changes requested):
Looking much better.
Some minor nitpickings, since this still will crash on 1.16 due to the named tag being pulled multiple times (aka during TagReload) That even is not needed anyway, since we use named tags which are only a wrapper around an inner tag object which gets automatically switched out.

src/main/java/com/minecolonies/apiimp/initializer/ModTagsInitializer.java
Raycoms: Apparently a static initializer is enough here.

marchermans (approved):

PR #5982, opened 2020-10-24
Do Not Merge

Justen-N:
A little stuck here. Config control is implemented, but I can't figure out how the request system would handle this. We want to autofill the request if the config is higher than the building's upgrade level, but I don't see how we were doing that before. Any pointers?
Addresses https://github.com/ldtteam/minecolonies-features/projects/1#card-42629540

-Create a graded config for the Builder Infinite Resources key
-Allow instance admins to control what level of free resources a building level gets. 
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
+++ b/src/api/java/com/minecolonies/api/configuration/CommonConfiguration.java
@@ -26,7 +26,7 @@ public class CommonConfiguration extends AbstractConfiguration
     public final ForgeConfigSpec.BooleanValue allowOtherDimColonies;
     public final ForgeConfigSpec.IntValue     citizenRespawnInterval;
     public final ForgeConfigSpec.IntValue     maxCitizenPerColony;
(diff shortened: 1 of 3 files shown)


Raycoms:
I guess I do need a good justification why this is needed =D
(Formatting is a bit off in the file btw)

Justen-N:
I can fix formatting.
Honestly. I picked an issue from the feature tracker and started working on it. I can see where it may be a balance issue. 
I could see a modpack builder or server admin setting this to say, level 2, to allow new players to quickly establish a working colony, especially if PvP is a part of the server. Allowing players to avoid some of the early micromanaging to catch up in 'personal' resources.

Justen-N:
And I just realized I grabbed an issue from the wrong tracker.

someaddons:
ye the features are possible suggestions of stuff what could be added, but many of it is not actually wanted to be added so its always a good idea to hop on discord and just ask us about the feature beforehand. This one seems rather weird, not sure why we would add level-based cheating, either you play survival or you do not

PR #5984, opened 2020-10-24
fix #5920

### LINKED ISSUE #5920: 1.16.3 Warehouse/Courrier issue
### Prerequisites 

- [x ] I am running the latest **alpha** version of MineColonies and Structurize for my Minecraft version.
- [ x] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure my issue is not covered there.
- [x ] I made sure that **this issue is not a duplicate of any existing issue.**

### Context 

- Minecraft Version: 1.16.3
- MineColonies Version: 0.13.348-BETA
- Structurize Version: 0.13.92-ALPHA

## Expected behavior

Couriers make deliveries from the warehouse to needed huts.

## Actual behavior

The couriers run back and forth within the warehouse. Every now and then they run out to pick something up from a hut, but never make deliveries to the huts.

Level 5 warehouse decided to upgrade storage with emeralds, then decided to hit the Sort button I had never noticed before. Since then the workers have done nothing. I know I have 4 emeralds left. When I look at the entire inventory of the warehouse using the Simple Storage Network I setup it shows only 4 emeralds. When I go looking in random racks I have seen a stack of 4 emeralds in at least 3 chests. I have noticed that this "duplicate" of items has occurred for other items in the warehouse as well but the total still shows correctly on the SSN.

1. 
2. 
3. 

### Logs

- latest.log: 
- crashlog: 

### Notes

---
## Viewers

Raycoms:
Closes #5920
Closes #
Closes #

- Refreshes combined inv list regularly.
- Fixes level 5 workers being able to work in the rain.
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
+++ b/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
@@ -56,11 +56,6 @@ public class TileEntityRack extends AbstractTileEntityRack
      */
     private int size = 0;
 
-    /**
-     * The combined inv wrapper for double racks.
-     */
-    private CombinedItemHandler combinedHandler;
-
     public TileEntityRack(final TileEntityType<? extends TileEntityRack> type)
     {
         super(type);
@@ -147,11 +142,6 @@ public class TileEntityRack extends AbstractTileEntityRack
         inventory = tempInventory;
         final BlockState state = world.getBlockState(pos);
         world.notifyBlockUpdate(pos, state, state, 0x03);
-
-        if (main && combinedHandler == null && getOtherChest() != null)
-        {
-            combinedHandler = new CombinedItemHandler(RACK, inventory, getOtherChest().getInventory());
-        }
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -95,6 +95,12 @@ public int getFreeSlots()
         return freeSlots;
     }
 
+    @Override
+    public void tick()
+    {
+        combinedHandler = null;
```
marchermans: Nope, figure out a better way todo this.
Neighbor update or what ever.

But this is not a good reason to make this TileEntity ticking, it will cause a massive performance regression and that is not acceptable for something like this, since this adds no functionality what so ever to our tile.

marchermans (changes requested):

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -438,11 +428,7 @@ else if (getOtherChest() != null)
             {
                 if (isMain())
                 {
-                    if (combinedHandler == null)
-                    {
-                        combinedHandler = new CombinedInvWrapper(inventory, getOtherChest().getInventory());
-                    }
-                    return LazyOptional.of(() -> (T) combinedHandler);
+                    return LazyOptional.of(() -> (T) new CombinedItemHandler(RACK, inventory, getOtherChest().getInventory()));
```
marchermans: Nope.
The idea of lazy optionals is that they should only be created when needed.

Meaning that these optionals should be effectively final as long as they are valid (aka as long as the block persists and his "neighbor"-state did not change.
Once any of these above properties change, you invalidate the old lazy optional and save a new one in a field which you then return in the getCapabilities method.

THis ensures that mods like AE etc can hold on to the itemhandler for as long as he is valid and only query for a new one once we invalidate the old one.

marchermans (changes requested):
Adapt the usage of LazyOptionals to follow guidelines.

marchermans (approved):

PR #5996, opened 2020-10-27
INV IMPROVEMENTS

Raycoms:
- Improves some of the inv handling
- Quick Rack Querying over expensive slot per slot
- Already return the building in the query of containers
- Adjust some old chest code in the AIs

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
+++ b/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
@@ -22,6 +22,7 @@ import net.minecraft.network.play.server.SUpdateTileEntityPacket;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.tileentity.TileEntityType;
 import net.minecraft.util.Direction;
+import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.Rotation;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.text.ITextComponent;
@@ -97,11 +98,11 @@ public class TileEntityRack extends AbstractTileEntityRack
     }
 
     @Override
-    public boolean hasItemStack(final ItemStack stack, final boolean ignoreDamageValue)
+    public boolean hasItemStack(final ItemStack stack, final int count, final boolean ignoreDamageValue)
     {
         final ItemStorage checkItem = new ItemStorage(stack, ignoreDamageValue);
 
-        return content.getOrDefault(checkItem, 0) >= stack.getCount();
+        return content.getOrDefault(checkItem, 0) >= count;
     }
 
     @Override
@@ -122,6 +123,29 @@ public class TileEntityRack extends AbstractTileEntityRack
                 return true;
             }
         }
+        return false;
+    }
+
+    @Override
+    public boolean hasSimilarStack(@NotNull final ItemStack stack)
+    {
+        final ItemStorage checkItem = new ItemStorage(stack, true);
+        if (content.containsKey(checkItem))
+        {
+            return true;
+        }
+
+        for (final ItemStorage storage : content.keySet())
+        {
+            for (final ResourceLocation tag : stack.getItem().getTags())
+            {
+                if (storage.getItemStack().getItem().getTags().contains(tag))
+                {
+                    return true;
+                }
+            }
+        }
+
         return false;
     }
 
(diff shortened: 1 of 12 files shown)


src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -720,19 +720,9 @@ public static int getItemCountInProvider(@NotNull final ICapabilityProvider prov
     public static int hasBuildingEnoughElseCount(@NotNull final IBuilding provider, @NotNull final ItemStorage stack, final int count)
```
Mekle001, line 721: There are several places that assume they will always get count from this, the MAXINT variant return being rare. 
Specifically, the AbstractBuildingWorker and the Dyer both use this function to figure out which ingredient is in largest supply.  Do we actually need the MAXINT result?

Raycoms: This method is kinda greedy, meaning that the moment we found enough we return. Thus, the returned count != the actually available count but just > the input count. Thus, if return is maxInt we know we got enough, else we know we got just as much as the return value specifies.

marchermans (dismissed):

Mekle001 (dismissed):

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -125,6 +126,29 @@ public boolean hasItemStack(@NotNull final Predicate<ItemStack> itemStackSelecti
...
+    @Override
+    public boolean hasSimilarStack(@NotNull final ItemStack stack)
+    {
+        final ItemStorage checkItem = new ItemStorage(stack, true);
+        if (content.containsKey(checkItem))
+        {
+            return true;
+        }
+
+        for (final ItemStorage storage : content.keySet())
```
someaddons, line 138: I'm wondering if we maybe want to sort by creative tab first and then by tag?

src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -1594,6 +1601,36 @@ public static boolean transferItemStackIntoNextBestSlotInItemHandler(
...
 
+    /**
+     * Method to swap the ItemStacks from the given source {@link IItemHandler} to the given target {@link IItemHandler}. Trying to merge existing itemStacks if possible.
+     *
+     * @param sourceHandler The {@link IItemHandler} that works as Source.
+     * @param predicate  the matching stack.
+     * @param targetHandler The {@link IItemHandler} that works as Target.
+     * @return True when the swap was successful, false when not.
+     */
+    public static boolean transferItemStackIntoNextBestSlotInItemHandler(
```
someaddons, line 1612: function name and doc don't match up

src/main/java/com/minecolonies/coremod/tileentities/TileEntityWareHouse.java
```diff
@@ -219,7 +210,7 @@ private TileEntity searchMostEmptySlot()
     {
         int freeSlots = 0;
         TileEntity emptiestChest = null;
-        for (@NotNull final BlockPos pos : getBuilding().getAdditionalCountainers())
+        for (@NotNull final BlockPos pos : getBuilding().getContainers())
```
someaddons, line 213: think you need to exclude the warehouse hutblock here?

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
Raycoms: I wanted to get rid of the creative tab searching because a lot of that info is client-side only and not "very" reliable.

src/main/java/com/minecolonies/coremod/tileentities/TileEntityWareHouse.java
Raycoms: why?

Raycoms: The warehouse block is also a valid chest to drop off

someaddons: thought we are not using it, but ok

Raycoms: We're using it, but it was broken in some cases. I fixed this in a recent PR

marchermans (approved):

Mekle001 (approved):

someaddons (approved):

PR #6004, opened 2020-10-29
Fix a few performance issues

someaddons:
Fixing a bunch of minor performance leaks, which though in total made for about ~30% of our citizens tick usage.
Fixed citizen inventories possibly running into issues with legacy armor slot code.
Fix NPE with inv utils, when it doesnt find the TE at the pos in world.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/inventory/InventoryCitizen.java
+++ b/src/api/java/com/minecolonies/api/inventory/InventoryCitizen.java
@@ -42,6 +42,11 @@ public class InventoryCitizen implements IItemHandlerModifiable, INameable
     private static final int ARMOR_SIZE = 5;
 
     /**
+     * Amount of free slots
+     */
+    private int freeSlots = DEFAULT_INV_SIZE + ARMOR_SIZE;
+
+    /**
      * The inventory. (27 main inventory, 4 armor slots, 1 offhand slot)
      */
     private NonNullList<ItemStack> mainInventory = NonNullList.withSize(DEFAULT_INV_SIZE + ARMOR_SIZE, ItemStackUtils.EMPTY);
@@ -163,7 +168,7 @@ public class InventoryCitizen implements IItemHandlerModifiable, INameable
      */
     public boolean hasSpace()
     {
-        return this.mainInventory.stream().limit(getSlots()).anyMatch(itemStack -> itemStack.isEmpty());
+        return freeSlots > 0;
     }
 
     /**
@@ -173,7 +178,7 @@ public class InventoryCitizen implements IItemHandlerModifiable, INameable
      */
     public boolean isEmpty()
     {
-        return !this.mainInventory.stream().limit(getSlots()).anyMatch(itemStack -> !itemStack.isEmpty());
+        return freeSlots == mainInventory.size();
     }
 
     /**
@@ -199,6 +204,7 @@ public class InventoryCitizen implements IItemHandlerModifiable, INameable
                 inv.set(index++, mainInventory.get(i));
             }
             mainInventory = inv;
+            freeSlots += futureSize - size;
         }
     }
 
@@ -253,7 +259,6 @@ public class InventoryCitizen implements IItemHandlerModifiable, INameable
     @Override
     public ItemStack insertItem(final int slot, @Nonnull final ItemStack stack, final boolean simulate)
     {
-        markDirty();
         final ItemStack copy = stack.copy();
         final ItemStack inSlot = mainInventory.get(slot);
         if (inSlot.getCount() >= inSlot.getMaxStackSize() || (!inSlot.isEmpty() && !inSlot.isItemEqual(copy)))
@@ -265,6 +270,8 @@ public class InventoryCitizen implements IItemHandlerModifiable, INameable
         {
             if (!simulate)
             {
+                markDirty();
+                freeSlots--;
                 mainInventory.set(slot, copy);
                 return ItemStack.EMPTY;
             }
@@ -279,6 +286,7 @@ public class InventoryCitizen implements IItemHandlerModifiable, INameable
         {
             if (!simulate)
             {
+                markDirty();
                 inSlot.setCount(inSlot.getCount() + copy.getCount());
             }
             return ItemStack.EMPTY;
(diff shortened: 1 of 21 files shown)


src/api/java/com/minecolonies/api/inventory/InventoryCitizen.java
```diff
@@ -41,6 +41,11 @@
      */
     private static final int ARMOR_SIZE = 5;
 
+    /**
+     * Amount of free slots
+     */
+    private int freeSlots = DEFAULT_INV_SIZE + ARMOR_SIZE;
+
```
Raycoms, line 43: I guess loading the inv from NBT loads things nicely too? Not sure armor slots should be dealt with as free slots? We practically only use armor slots for rendering.

someaddons: ye checked it and entities now have their own seperate equipment/hands inventory, so removed this legacy code

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/6004/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIUsesFurnace.java
```diff
@@ -64,7 +64,7 @@ protected AbstractEntityAIUsesFurnace(@NotNull final J job)
         super(job);
         super.registerTargets(
           new AITarget(IDLE, START_WORKING, STANDARD_DELAY),
-          new AITarget(START_WORKING, this::startWorking, TICKS_SECOND),
+          new AITarget(START_WORKING, this::startWorking, 100),
```
Mekle001: Can we use the supplier pattern that Ray put together here, and use TICKS_SECOND if there is a current recipe set? The AI goes back to START_WORKING constantly during actual processing, so a 5 second delay will be fairly noticeable between actions.

Mekle001 (changes requested):
Overall this looks pretty good, but I'm concerned about the visible performance hit on smeltercrafters of the change there.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIUsesFurnace.java
Mekle001: We could potentially have the supplier check both currentRecipe and the count of current furnaces burning. If we have no recipe, or burning furnaces, this could slow down quite a bit.

someaddons: What is the smeltercrafting doing that needs a higher update rate than every 5s?

Mekle001: This is going to hit between picking up results for each furnace. So, with 5 furnaces you're looking at almost 30 seconds to retrieve results from all of them.

Mekle001: That is just one example where we go back to START_WORKING in the AI. Most actions do.

Raycoms: Not considering probably walking to the furnace too.

Raycoms: Might be better to use a setDelay here then instead?

Mekle001: Walking depends on the schematic, there are quite a few where there is little to no walking during pickup. SetDelay could work, but you added the provider support for just such an occasion to get away from the SetDelays.

someaddons: he barly does any actions besides waiting for the furnace and sometimes taking out stuff, his action are good if they take some time then he seems less idle all day. Its supposed to be a human-like those aren't instant robot-like doing all tasks at once

Raycoms: Maybe a better question is, why is this call causing lag?

Mekle001: I'm not disagreeing, @someaddons. There will be delays in what I'm proposing, more than there are today. She's going to be slower to pick up new jobs, and to react to furnaces finishing. All I'm hoping for is that once she starts to work, she doesn't take breaks every step.

Mekle001: Heck, if we go to a provider, we can slow down this check by a LOT. As in, if we have a recipe and things are burning, then 20-30 seconds. If we just have a recipe, 1 second, if we have no recipe, 10-15 seconds.

someaddons: tbh I do not see delays between actions as breaks, to me its when the worker is "working" on sth that he takes some time. The reason why its relatively laggy are inventory searches

Raycoms: I think the #1 solution should be not reducing the tick rate. But simply adding to the startWorking method, after the walkToBuilding a check if there is request in the queue and if not, directly exit and not do the 100 checks we do.

someaddons: sure you can redesign it if you want, but it is not necessary because a 5s wait is totally fine

Raycoms: `The reason why its relatively laggy are inventory searches`

Ugh, why didn't you say so? that's an easy fix =P

Mekle001 (dismissed):

Mekle001 (dismissed):
Ok, I think we're good. I somehow missed that the supplier changes hadn't merged into the mainbranch. Once those are in, we can reconsider the timing.

PR #6013, opened 2020-10-31
Fix speed modifiers, fix raider damage scaling

someaddons:
Reworked a bit how we handle our entities speed a bit, we use the entities movement speed attribute as base speed, and any speedvalues passed along for certain pathing/move to operations are speedFactors multiplied ontop. In some cases we already use speed like this, in other cases we treated the given speedFactors as basespeed and caused issues. I've also renamed all the speed parameters to speedFactor for clarity.

Fixed raider difficulty scaling, it no longer scales base + scaling damage by difficulty, and also changed their damage to our custom damage source, which does not do additional world difficulty scaling with damage against players, as our raiders already do scale with world difficulty.

Review please


Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/6013/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

PR #6020, opened 2020-11-01
Fast/compost

Mekle001:
- Accelerate barrels based on Primary Skill level, similar to the way we accelerate furnaces.

Review please


src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIUsesFurnace.java
```diff
@@ -231,7 +231,7 @@ private IAIState accelerateFurnaces()
                 }
             }
         }
-        return getState();
+        return null;
```
marchermans, line 234: Is this even valid?
This should probably return some same default state.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/composter/EntityAIWorkComposter.java
```diff
@@ -92,6 +97,35 @@ public EntityAIWorkComposter(@NotNull final JobComposter job)
...
+                    {
+                        if (barrel.checkIfWorking() && !barrel.isDone())
+                        {
+                            barrel.tick();
+                        }
+                    }
+                }
+            }
+        }
+        return null;
```
marchermans, line 125: Again return some same default state here.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIUsesFurnace.java
Mekle001: Actually, there is a significant behavior difference between null and getState. Returning a valid state here stops further execution on that tick, where returning null allows further execution. I can add comments here, though I found that documentation in AbstractEntityAIBasic

src/main/java/com/minecolonies/coremod/entity/ai/citizen/composter/EntityAIWorkComposter.java
Mekle001: See previous comments

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIUsesFurnace.java
Raycoms: @someaddons can probably confirm

someaddons: ye null state return means no action, not null means we traverse into that state and do not check further transitions of this state in the current tick. I thought about changing it, but back then this was needed for backwards compat due to the way our AI is structured/used.

marchermans (approved):
Oke for now this is fine, we keep track of the `null` option in #6028, cause that is a major codesmell and should be avoided at all costs.

PR #6029, opened 2020-11-03
Invfixes

someaddons:
Improve citizen inventory free slots and stop damaging/modifying internal stacks from outside

Fix cap invalidate only upon change
Fix racks(and their chunk) beeing marked dirty from the buildings getCap call all the time
Fix capability invalidates to be like chests
Fix Building cap returning values when the TE was removed
Fix Enchanter doing animations serverside and not calling super tick
Fix CombinedHandler not having isItemValid implemented

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
+++ b/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
@@ -447,6 +447,11 @@ public class TileEntityRack extends AbstractTileEntityRack
     {
         if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY)
         {
+            if (lastOptional != null && lastOptional.isPresent())
+            {
+                return (LazyOptional<T>) lastOptional;
+            }
+
             if (single)
             {
                 lastOptional = LazyOptional.of(() -> inventory);
@@ -550,6 +555,7 @@ public class TileEntityRack extends AbstractTileEntityRack
             if (lastOptional != null)
             {
                 lastOptional.invalidate();
+                lastOptional = null;
             }
             super.setMain(main);
         }
@@ -563,6 +569,7 @@ public class TileEntityRack extends AbstractTileEntityRack
             if (lastOptional != null)
             {
                 lastOptional.invalidate();
+                lastOptional = null;
             }
             super.setSingle(single);
         }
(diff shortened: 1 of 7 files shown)


marchermans (dismissed):

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/6029/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -447,13 +445,13 @@ public void rotate(final Rotation rotationIn)
...
-        if (lastOptional != null)
-        {
-            lastOptional.invalidate();
-        }
-
         if (capability == CapabilityItemHandler.ITEM_HANDLER_CAPABILITY)
         {
+            if (lastOptional != null && lastOptional.isPresent())
+            {
+                return (LazyOptional<T>) lastOptional;
```
Raycoms: Ehm, shouldn't we avoid returning this cached value here?

marchermans: No why?

Raycoms: I have a feeling I'm talking to a wall. I discussed this 2 weeks ago and removed this caching stuff.

marchermans: Because you think it will break understood.
But I have no evidence that it will. It works for tons of other mods, even for vanilla. So come up with evidence that it will break.

He invalidates the cache when the single or double status changes. Which is the only relevant thing, in my eyes.

Raycoms: No no, not I think it will break, it did break in numerous cases and items were lost, not detected by dman, duped, etc

Raycoms: I'm talking with someaddons about this and forge does it completely different for chests too, (no simple cache like this) so we should probably copy this behaviour

marchermans: If that is the case, then go for it. I would follow the example of the chests and just stick with that.
I thought they cached it that is why I confused. If they do not then disregard my opinion.

Raycoms: their lazyoptional basically queries the tileEntity from the world =D

Mekle001 (dismissed):

marchermans (approved):

Mekle001 (approved):

PR #6067, opened 2020-11-12
Feature: fast pickup of specific items after crafting

### LINKED ISSUE #6035: Emptied milk buckets not being taken from Baker's storage when done with the Milk
### Prerequisites 

### Context 

- Minecraft Version: 1.15.2
- MineColonies Version: 0.13.415-ALPHA
- Structurize Version: 0.13.94-ALPHA
- Related Mods and their Versions: Everything from the MineColonies Official pack By: H3lay, Kostronor, D3miurge_, Wissin8r, Asherslab on Twitch (updated on the 22nd of October, 2020)

## Expected behavior
Once the baker is done making something that requires milk, the empty buckets are stored, then taken by a courier from their hut and put into the warehouse storage. Which are then delivered to the Cowboy to get more milk.

## Actual behavior
Once the baker has completed with the milk, the empty buckets are stored in the Baker's hut storage (main storage or racks) and stay there. No couriers ever go grab them. Even if I call in a courier to come pick up stuff, the buckets will stay in there. I have to manually grab them and drop them into a stash or give them to the cowboy manually.

1. Have an active warehouse and courier
2. Have a working cowboy with cows, enabled with "collect Milk" and make sure he has 3 buckets and wait for him to fill them with milk and that they are all put into storage at the warehouse
3. Have a level 2 baker, which knows how to make a cake 
4. Have a mailbox and request a cake and wait for it to be crafted and delivered.
5. Wait OR request buckets from the mailbox OR wait for the cowboy to request buckets.

The buckets will stay in the bakery storage unless you manually go there and request a pickup. Otherwise, they will never go and grab it. Even if set to 10/10 priority, it doesn't seem to be an item that is on the list of "needs to grab when unused". Probably because of the cowboy that also requires empty ones.

### Logs

None needed, no errors related to them in there.

### Notes
I've waited a little for many updates just to make sure, I've started my colony on 0.13.271-ALPHA and ever since I built the bakery, I've always had that problem. In the past, my buckets actually vanished, I couldn't find them anywhere in storage of all my colony and I checked every building / possible storage spot. I made 9 more (enough for 3 cakes per request), they no longer just vanish, but the problem of them staying stuck at the bakery persists.

---
## Viewers

Mekle001:
Closes #6035 

- This cleans up the code paths for retaining items during pickup, and makes it safer to call when the task queue is not empty
  - Set up the Baker to do fast pickups for empty buckets, to avoid getting stuck waiting for milk
- Ensure Minstock is honored during pickups, to avoid over-crafting

Review please


src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -1050,10 +1057,23 @@ else if (dumpOneMoreSlot())
...
+        {
+            building.createPickupRequest(scaledPriority(building.getPickUpPriority()), fastPickup);
+            fastPickup.clear();
         }
+
+
         return afterDump();
     }
 
+    public boolean addItemForFastPickup(ItemStack item)
```
Raycoms: jdoc

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/6067/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

marchermans (dismissed):
The whole approach here is backwards.
We stepped away from lists of stacks that need picking up in buildings and AIs three years ago.
And we are not going back to it.

This approach needs to be redesigned properly. Also I see no reason for this PR in the first place. What bug / awkward behaviour is fixed here?

Raycoms:
Okay, I've talked with @OrionDevelopment about this and he now understood why this is done this way. In the discussion about how we could handle it better, we did finally find a way to actually avoid the problem of a courier accidentally picking up the wrong items.

So, instead of having the targeted pickup, we can have a generic pickup and make sure that no outgoing deliveries are picked up.

That is, in: getRequiredItemsAndAmount() in the crafter building, we query all resolvers assigned to this building and query their delivery requests and add those to the toKeep map. (You can check in AbstractBuilding in the logic we've added to fulfill requests on player delivery to a building for the requests that a resolver might have how to get all the resolvers)

someaddons:
I'm not exactly sure what the use case here is, why and where do we need to pick up certain stuff right away? And if we init that in the building constructor how does that work with dynamically adding of recipes

marchermans:
The baker is basically the best and only example.
Ik can make cakes. Which have buckets as a result. When I added that system I basically made it so that nothing is picked up while crafting as to not confused the ai and prevent pick up of items which may or may not be in use by the crafter via the dman.

Now if the baker makes 10 cakes on request that results in 10 buckets laying around which could be used. At least that is what ray told me and that is wastefully and could be reused.

Personally I see no reason to change the current system but I was told it was required. I was however completely against the initial implementation with another list in the building that deals with stacks and items. 

I am willing to accept the pr of it fixes the root issue and only then.

someaddons:
ah so this is about intermediary items beeing required again, though in the case of cakes we need those buckets to be refilled with milk anyways right? To me it looks like it shouldnt do batch-crafting for stuff where it needs a reuseable thing, or alternatively should just view/use the reuseable item as an infinite count stack, where 1 can be used in places where it'd think it needs more

Raycoms:
The problem is not the batch crafting, but that those buckets will stay forever there at the baker and will potentially never be picked up (as long as the baker got an open task).

someaddons:
why would they not be picked up? does the baker prevent pickup up empty buckets?

Raycoms:
Crafters prevent pickup while they are crafting.

someaddons:
that seems wrong then, they should prevent pickup of items they need, not all. Else a busy crafter never releases finished products

Raycoms:
Either way, I've pointed out the solution to this problem above that will solve this =D

Mekle001:
I'm working on switching over to Ray's recommended changes.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -911,6 +912,13 @@ else if((forceReplace || newRecipe.getMustExist()) && duplicateFound != recipeTo
      */
     public String getJobDisplayName()
     {
+        if(jobDisplayName.isEmpty())
+        {
+            //This will only happen if getJobDisplayName is called before the first citizen is assigned. 
+            ICitizenData citizen = colony.getCitizenManager().createAndRegisterCivilianData();
+            jobDisplayName = this.createJob(citizen).getName();
+            colony.getCitizenManager().removeCivilian(citizen);
```
Raycoms: Might this here cause issues? @someaddons

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingSmelterCrafter.java
```diff
@@ -94,26 +94,31 @@ public int getMaxBuildingLevel()
...
                                 {
-                                    amount += recipeOutputs.get(itemStorage).getA();
+                                    amount += requiredItems.get(itemStorage).getA();
                                 }
-                                recipeOutputs.put(itemStorage, new Tuple<>(amount, false));
+                                requiredItems.put(itemStorage, new Tuple<>(amount, false));
                             }
                         }
-                    }
+                        final ItemStorage output = new ItemStorage(recipeStorage.getPrimaryOutput());
```
Raycoms, line 110: seems badly alinged

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
someaddons: hm not sure tbh, though this is a pretty weird solution to just get a name, seems like a hacky solution with a high overhead. Should be able to solve this much more smoothly

someaddons: I just checked and it seems to be not used at all? so could just remove it

Mekle001: This *is* used, in both PublicWorkerCraftingProductionResolver and PublicWorkerCraftingRequestResolver. In both of those cases, it shouldn't go through this codepath. The codepath that *might* hit his is the one in WindowTownHall.java, and then only once. 

This is the result of building.getJobName returning a non-localization key, and being needed by the crafting code. This is the only contact the building has today with the actual job, and to get the jobName localization key.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingSmelterCrafter.java
Mekle001: The problem is actually line 115. Fixing.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Raycoms: Couldn't we return a "not loaded" or so then?

someaddons: This is serversided and the stuff you're talking about is clientside

Mekle001: We could, but then that would show to the user, and they won't know how to fix it (assign a worker to the building). This is specifically the list of job names in the town hall. Right now, the town hall will show 0/1 for a blank job name if the building has never been assigned a worker before this change.

Mekle001: This code is serverside, and is serialized to the window for display clientside.

someaddons: where is this used in serialization? Unless intelliJ is lying to me its a function that never gets called

Mekle001: Great point, I just looked at the serialization again, and it's referencing the field. It should be calling this function. Fixing.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -1474,8 +1503,7 @@ public boolean createPickupRequest(final int scaledPriority)
         {
             for (final IToken<?> req : reqs)
             {
-                final IRequest<?> request = colony.getRequestManager().getRequestForToken(req);
-                if (request != null && request.getState() == RequestState.IN_PROGRESS)
+                if (colony.getRequestManager().getRequestForToken(req).getState() == RequestState.IN_PROGRESS)
```
Raycoms: Is this a good idea? I had the null check here on purpose.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Raycoms: I still don't like this since it's not side-effect free. I rather have it return empty for now, until we got a propper fix. (It's only a small UI error after all).

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
marchermans: I am pretty sure that null check is needed as well....
It threw exceptions previously but that was removed.

Mekle001: This is a bad merge/backout of the old code. Fixing.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Mekle001: removing for now

Raycoms:
:shipit:

marchermans (dismissed):

PR #6096, opened 2020-11-16
Version/1.16.3

kmanev:
Archery works with both old and new schematics. (either using haybale or target blocks as shooting targets).
Temporary fix for 1.16 until all schematics use target blocks.


Raycoms:
I would've preferred the updated schematic xD

PR #6108, opened 2020-11-18
Fixes wrong capitalization in the list of jobs in the town hall

ravenbuilder934:
Closes #
Closes #
Closes #

- Fixes wrong capitalization/etc in the list of jobs in the town hall (discussed in discord)
- I probably did this wrong
-

Review please


ShadowProtocol (changes requested):
The ``getJobName`` method is used for more than just as a name for translation for the user, it's essentially an internal id for the worker as well, so this will probably break things for current worlds. (Despite what I saw was discussed in chat)

I suggest a new getter method that forms the translation key from this job name.

ravenbuilder934:
> I suggest a new getter method that forms the translation key from this job name.

I have no idea what this is 😛

Raycoms:
okay, I checked where this is used, and we also use this in the custom recipe code @Mekle001 how does the recipe code handle with the few buildings that have a translation code as the getJobName? And can we change this with backwards compat?

Mekle001:
This is going to be a problem, since this is used in the tag names for valid recipes too.  I ran into this when I was working on something else recently.. My recommendation is to use building.getJobDisplayName() as it's initialized to job.GetName() by the entityAI. 

Without a significant rewrite with all things related to recipe tracking, we can't change the building getJobName to be the localization constant. It's the building "ID" for too many other things. The few places where it's a already a localization constant just happen to not need that part of the recipe system.

Mekle001 (changes requested):
Please don't change the return of getJobName, as it will break several things.  I will rename it to getJobShortName in the future. 
Instead, please change line 608 of WindowTownHall.java

Raycoms:
This looks like sth we should have a big "todo change in 1.17 to propper names".

ravenbuilder934:
Like this?

src/main/java/com/minecolonies/coremod/client/gui/WindowTownHall.java
```diff
@@ -605,7 +605,7 @@ private void createAndSetStatistics()
                 int max = ((AbstractBuildingWorker.View) building).getMaxInhabitants();
                 int workers = ((AbstractBuildingWorker.View) building).getWorkerId().size();
 
-                String jobName = ((AbstractBuildingWorker.View) building).getJobName().toLowerCase(Locale.ENGLISH);
+                String jobName = ((AbstractBuildingWorker.View) building).building.getJobDisplayName().toLowerCase(Locale.ENGLISH);
```
Mekle001: Close, not quite. Remove the building. out of this, since the code before that is the conceptual equivlent. You just need a direct replacement of getJobName with getJobDisplayName

ravenbuilder934: I think I fixed it

Mekle001: Yeah, that looks better. Now you can undo the other changes, and verify that it's doing what you want. :)

ravenbuilder934: I think I already undid the changes to the buildings. How do I test/verify it?

Mekle001 (dismissed):
That looks good now, as long as it works the way you expect in game.

Mekle001:
Ok, I ran a quick test, this is working as expected for most buildings. the only one not capitalized is Teacher, and that's because of some other code in WindowTownHall.java. I'd recommend fixing that separately.

ravenbuilder934:
How can I fix the teacher?

Mekle001:
Line 616 of WindowTownHall.java, it doesn't need the LanguageHandler.format part, or the () around the quotes.

ravenbuilder934:
Like that?

Mekle001 (approved):
Yes. All of them are correctly localized/capitalized in the townhall now.

ShadowProtocol (approved):
Yes, this will work fine.

Personally I still think this should be done dynamically from the short name (i.e. adding the com.minecolonies.coremod.job prefix onto a job name when requested) that way were not sending the entire translation key in packets, and it isolates language from code further. However it's done, yeah we should slap a TODO on this sometime.

PR #6126, opened 2020-11-20
Minor Config change

Justen-N:
- Refactor BuilderInfiniteResources key to Server Config
- Update JEI dependency, was three minor releases behind.

Intention: moving the debug and other such keys to the server config ought to reduce user confusion as to what the key is used for. 
I'm running into a runtime error from the JEI dependency while trying to build against minecolonies for some compatibility additions. I'm hoping that it is a mapping conflict, but I'm not sure. In any case, the JEI dependency runs fine at this level. 
Review please

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -7,5 +7,5 @@ forge_version=31.2.33
 mappings=20200606-1.15.1
 # Dependencies
 structurize_version=0.13.94-ALPHA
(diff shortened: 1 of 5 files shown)


Raycoms:
I don't think that is really necessary, we have that already done in 1.16, I think for now it would only cause confusion.

Issue #5218, opened 2020-06-13
[Bounty: $30] [Request] Aquaculture 2 Mod Support


Cacc22:
### Minecolonies version #

Version: 0.11.957

### Expected behavior

- Fisher will fish with Iron/Gold/Diamond/Neptunian rods.

### Actual behaviour

- Fisher will hold Iron/Gold/Diamond/Neptunian rods and clear notification of needing a rod, but will otherwise just stand around not fishing.

### Steps to reproduce the problem

1. Give Fisher an Aquaculture 2 fishing rod.
2. Participate in staring contest.

Cacc22:
Apologies for missing the Mod support pin, I wanted to add a bounty so I was more focused on that part -_-

ravenbuilder934:
Fixed

PR #6160, opened 2020-11-26
#5218 Support more varied fishing rods.  Also improves fisher mainhand equips.

gattsuru:
Targets Issue #5218.  I don't think it's a full enough to be worth the bounty (and I can't accept it), but may be enough for that user's purpose.

- Changed most tests for fishing rods from == Items.FishingRod to instanceof FishingRodItem.  This allows fishers to use Aquaculture Fishing Rods, and should allow support for most other modded fishing rods (or any other item that implements FishingRodItem, like the XercaMod GrabHook).  May eventually be worth adding a blacklist to avoid the next coming of the MFR "Fishing Rod", but most cases I could find were funny rather than dangerous.
- Converted fishing rod's getMiningLevel logic to test against durability and enchantment levels, instead of allowing any usage from any completed hut level.   In a vanilla+minecolonies setup, this basically just requires higher hut tiers for enchanted fishing rods.  With Aquaculture, Wood or Gold for Tier 1, Iron for Tier 2, and Diamond/Neptunium, with an extra hut tier required for enchantments.  This is based on durability, so it should support most common approaches for fishing rods, and falls back to T1 for unbreakable rods.
- Fishers who can't find a compatible fishing rod will swap to an air block, and will swap if the mainhand fishing rod doesn't match the first usable fishing rod in their inventory.  Doesn't handle case where a Fisher is fired and then has fishing rod removed from their inventory, but should be more consistent than before.

# Tests completed:
- Minecraft 1.16.4 + structurize-0.13.100-RELEASE, tested Fishers with no rod, an unenchanted wooden rod, and an enchanted III wooden rod at T1, T3, and T4 huts.
- Minecraft 1.16.4 + structurize-0.13.100-RELEASE + Aquaculture 2.1.6, tested Fishers with Wooden, Gold, Iron, Diamond, and Neptuneum fishing rods, and an enchanted Lure/Luck III Neptuneum Rod.  Placed warehouse and blacksmith to test request system and the fishers do ask for generic fishing rods, though I'm not sure I understand why the blacksmith picks a specific recipe to make if multiple are available.
- Minecraft 1.16.4 + structurize-0.13.100-RELEASE + Aquaculture 2.1.6 + XercaMod 1.16.3-1.1, tested Fishers with Grab Hook and Grabbing Grab Hook.  Worked and probably shouldn't, but just had the Fishers reeling in Aquaculture fish.

My apologies for the spam.  I'm a bit too used to svn, so making a mess with git.

Review please


src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -312,6 +314,20 @@ else if (ToolType.HELMET.equals(toolType)
                 return getArmorLevel(ArmorItem.getArmorMaterial());
             }
         }
+        else if (stack.getItem() instanceof FishingRodItem)
+        {
+            if(stack.getItem().isDamageable())
+            {
+                if(stack.getMaxDamage() < DURABILITY_MAX_WOOD_OR_GOLD)
```
Raycoms: Isn't there a registry where we could get that dynamically? Like maybe getting another tool and comparing the durability to make this more dynamic?

gattsuru: Because Fishing Rods are used much slower than other tools, Aquaculture gives them lower max durability: an Aquaculture Diamond Fishing Rod's max durability is 450 to the 1561 of vanilla's Diamond Pick/Axe/Shovel/Sword, and even a Neptunum Fishing Rod only goes to 1000.  The disparity is less extreme for iron, but at 125 the Aquaculture Iron Fishing Rod has lower durability than a vanilla Stone Pickaxe's 131.

Thermal Foundation's custom fishing rods more closely match their vanilla equivalents, but even they are a little off (TF Iron Fishing Rods at 255 and not 250, for example, which is conveniently just the wrong direction for munging it with greater or less than signs). I picked 400 so that steel and invar fishing rods (https://teamcofh.com/docs/1.12/thermal-foundation/fishing-rods/) would require a higher tier hut, for example, but that is admittedly a magic number.

Comparing to multiples of the vanilla fishing rod would better handle situations where someone just buffs the vanilla one, but net.minecraft.Items.FISHING_ROD.getMaxDurability() is deprecated and I'm not sure there's a way to get a generic ItemStack.FISHING_ROD.  net.minecraft.item.ItemTier would normally be great, but vanilla fishing rods don't use it to determine their uses, nor match its pattern.

I suppose using ItemTier's getMaxUses, adjusting it, and leaving a comment about the offsets is probably the best option.

EDIT: cleaned up code, swapped to ItemTier.Material.getMaxUses and an offset.  ThermalFoundation Silver is a stupid place to set the line from T1 to T2 since it hasn't even been ported to 1.16 yet, but they ended up in dungeon loot tables enough in 1.12 that it's probably better than most alternative points.  Also set a check against vanilla rods to always set them to T1 + enchantment value, just in case someone them ultra-durable without adding any other rod types.

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/6160/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

someaddons (dismissed):
looks good, please import our formatting schema from the documentation folder and reformat the changes 👍

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -479,6 +483,33 @@ else if (damageReductionAmount <= ArmorMaterial.DIAMOND.getDamageReductionAmount
...
+    /**
+     * Estimates the fishing rod tier from available durability and enchantment status.
+     * @param itemStack the tool to check.
+     * @return equivalent tool level.
+     */
+    private static int getFishingRodLevel(final ItemStack itemStack)
+    {
+        if(itemStack.getItem() == Items.FISHING_ROD)
+            return (1 + getMaxEnchantmentLevel(itemStack));
+        if(!itemStack.isDamageable())
```
Raycoms: We always add braces, even on those ifs

Raycoms (dismissed):

gattsuru:
Not sure why it's marked that review "dismissed".  I /think/ that covers all the style and format requirements.  Apologies on the brackets; I complain enough about students doing it that I should know better.

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -479,6 +483,42 @@ else if (damageReductionAmount <= ArmorMaterial.DIAMOND.getDamageReductionAmount
...
+     */
+    private static int getFishingRodLevel(final ItemStack itemStack)
+    {
+        if (itemStack.getItem() == Items.FISHING_ROD)
+        {
+            return (1 + getMaxEnchantmentLevel(itemStack));
+        }
+        if (!itemStack.isDamageable())
+        {
+            return 1;
```
marchermans: How sure are we this is a good idea.
I would say if th rod is not damageable it is a max hut level feature.

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -479,6 +483,42 @@ else if (damageReductionAmount <= ArmorMaterial.DIAMOND.getDamageReductionAmount
...
+        /**
+         *  Because fishing rods don't follow ItemTier values, these numbers require an offset.
+         *  Wood + 6 is required to allow T1 huts to use up to vanilla fishing rods equivalents.
+         *  Wood + 22 allows T1 huts to use up to ThermalFoundation Silver fishing rods.
+         *  Iron + 6 allows T2 huts to use up to ThermalFoundation Iron fishing rods or Aquaculture Iron,
+         *  while still requiring T3 for ThermalFoundation advanced alloys, Aquaculture Neptunium,
+         *  or Aquaculture diamond rods.
+         * */
+        final int rodDurability = itemStack.getMaxDamage();
+        if (rodDurability <= (ItemTier.WOOD.getMaxUses() + 22))
```
marchermans: These hardcored constants should be configurable from the config

marchermans (changes requested):

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
gattsuru: Fair.  I was thinking more for situations where it's the only craftable rod, but a) that should be done as an override to the vanilla one, b) an extreme edge case, and c) probably wouldn't be a bad place to make the fishing hut a higher-tier only choice anyway.

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
gattsuru: Ok.  Added them to the server-specific config as fishingRodDurabilityAdjustT1 and fishingRodDurabilityAdjustT2.  Hopefully MinecoloniesAPIProxy.getInstance().getConfig().getServer() is a reasonable way to get to that; I'm not sure the performance ramifications of the different access methods, but didn't break across a few attempts.

marchermans: This is fine.

src/api/java/com/minecolonies/api/configuration/ServerConfiguration.java
```diff
@@ -955,6 +957,9 @@ protected ServerConfiguration(final ForgeConfigSpec.Builder builder)
 
         dynamicTreeHarvestSize = defineInteger(builder, "dynamictreeharvestsize", 5, 1, 5);
 
+        fishingRodDurabilityAdjustT2 = defineInteger(builder, "fishingroddurabilityadjustt2", 6, -249, 250000);
+        fishingRodDurabilityAdjustT1 = defineInteger(builder, "fishingroddurabilityadjustt1", 22, -58, 250000);
```
Raycoms, line 961: negative values?

gattsuru: Comparing them to them ItemTier.WOOD.getMaxUses() and ItemTier.IRON.getMaxUses() for non-fishing tool.  The limits are set such that it won't go less than zero uses as the minimum acceptable use count for a tier.  Such a comparison wouldn't cause harm or a crash, it just wouldn't make sense, since items shouldn't have negative maxUses.

I don't think it's a common use case, but if someone disabled the vanilla fishing rod, had a rod at durability 30, and another at 50, and only wanted the dur-30 rod to be usable by T1 huts, they'd use an adjustT1 value between -10 and -29.

Raycoms: Couldn't it potentially go below 0 for rods with less durability than the vanilla ones?

gattsuru: The application logic applies the adjustments to the ItemTiers.WOOD and ItemTier.IRON enums, rather than to the instance of the ItemStack for the tool being tested:

> (rodDurability <= (ItemTier.WOOD.getMaxUses() + MinecoloniesAPIProxy.getInstance().getConfig().getServer().fishingRodDurabilityAdjustT1.get())

> (rodDurability <= (ItemTier.IRON.getMaxUses() + MinecoloniesAPIProxy.getInstance().getConfig().getServer().fishingRodDurabilityAdjustT2.get()))

To go below zero, another mod would have to overwrite the ItemTiers, but those are defined in Forge and don't have setters (and I don't think they can be accessed by the datapack system?).  And it still wouldn't break anything, just make it so the hut could only use vanilla rods.

Raycoms: okay, sounds good

Raycoms (approved):

marchermans (approved):

PR #6171, opened 2020-11-29
Small statemachine improvement

someaddons:
This changes the statemachine to no longer do hash lookups frequently when it already does know/has a reference to the result. Instead we do the lookup once the state changes and keep the reference to the current state's transitions.
Removed streams in the basic statemachine aswell.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/ai/statemachine/basestatemachine/BasicStateMachine.java
+++ b/src/api/java/com/minecolonies/api/entity/ai/statemachine/basestatemachine/BasicStateMachine.java
@@ -27,6 +27,9 @@ public class BasicStateMachine<T extends IStateMachineTransition<S>, S extends I
     @NotNull
     protected final Map<IStateEventType, ArrayList<T>> eventTransitionMap;
 
+    /**
+     * The current states list of transitions
+     */
     protected ArrayList<T> currentStateTransitions;
 
     /**
--- a/src/api/java/com/minecolonies/api/entity/ai/statemachine/basestatemachine/BasicStateMachine.java
+++ b/src/api/java/com/minecolonies/api/entity/ai/statemachine/basestatemachine/BasicStateMachine.java
@@ -10,6 +10,7 @@ import org.jetbrains.annotations.NotNull;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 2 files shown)


Raycoms (dismissed):

src/api/java/com/minecolonies/api/entity/ai/statemachine/basestatemachine/BasicStateMachine.java
```diff
@@ -27,6 +27,11 @@
     @NotNull
     protected final Map<IStateEventType, ArrayList<T>> eventTransitionMap;
 
+    /**
+     * The current states list of transitions
+     */
+    protected ArrayList<T> currentStateTransitions;
+
```
marchermans, line 35: Specific type instead of `List<T>`

src/api/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickRateStateMachine.java
```diff
@@ -31,6 +31,13 @@
      */
     private int tickRateCounter = 0;
 
+    /**
+     * Reference to our used global transition lists
+     */
+    private final ArrayList<ITickingTransition<S>> aiBlockingTransitions;
```
marchermans: Again specific type.

marchermans (changes requested):
Some code style is wrong here. Most notably the usage of concrete types.

src/api/java/com/minecolonies/api/entity/ai/statemachine/basestatemachine/BasicStateMachine.java
someaddons: hm there is a bunch of places in there where we're expecting that type though

someaddons: ye I can't change it anyway nicely

marchermans: Then refactor it.
You should never ever have a field with a concrete type. Especially ArrayList since it's api surface is identical to the normal list in all places that matter.

marchermans: What do you mean you can't change it anyway nicely?!?!??

someaddons: oh I think I just misunderstood, sure I'll change it to the list basetype :D

marchermans (approved):

PR #6201, opened 2020-12-05
Add loot tables to customrecipes

Mekle001:
- Add base functionality to recipestorage to store and generate loot from loot tables
- Add definition of loot tables & secondary outputs to custom recipes
- Add flint @ 10% rate to the Crusher's gravel recipes.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/crafting/IRecipeStorageFactory.java
+++ b/src/api/java/com/minecolonies/api/crafting/IRecipeStorageFactory.java
@@ -68,6 +68,11 @@ public interface IRecipeStorageFactory extends IFactory<IToken<?>, RecipeStorage
             throw new IllegalArgumentException("Eighth parameter is supposed to be a List<ItemStack> or Null!");
         }
 
+        if (context.length > MIN_PARAMS_IRECIPESTORAGE + 5 && context[8] != null && !(context[8] instanceof ResourceLocation))
+        {
+            throw new IllegalArgumentException("Eighth parameter is supposed to be a List<ItemStack> or Null!");
+        }
+
         final List<ItemStack> input = (List<ItemStack>) context[0];
         final int gridSize = (int) context[1];
         final ItemStack primaryOutput = (ItemStack) context[2];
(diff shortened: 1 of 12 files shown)


src/api/java/com/minecolonies/api/crafting/IRecipeStorageFactory.java
```diff
@@ -68,6 +68,11 @@ default RecipeStorage getNewInstance(@NotNull final IFactoryController factoryCo
             throw new IllegalArgumentException("Eighth parameter is supposed to be a List<ItemStack> or Null!");
         }
 
+        if (context.length > MIN_PARAMS_IRECIPESTORAGE + 5 && context[8] != null && !(context[8] instanceof ResourceLocation))
+        {
+            throw new IllegalArgumentException("Eighth parameter is supposed to be a List<ItemStack> or Null!");
```
Raycoms: either the message is off here, or sth else is wrong

Mekle001: Dangit. The message is off. Fixing.

src/main/java/com/minecolonies/coremod/colony/crafting/RecipeStorageFactory.java
```diff
@@ -207,6 +219,8 @@ public void serialize(IFactoryController controller, RecipeStorage input, Packet
         packetBuffer.writeInt(input.getSecondaryOutputs().size());
         input.getSecondaryOutputs().forEach(stack -> packetBuffer.writeItemStack(stack));
 
+        packetBuffer.writeResourceLocation(input.getLootTable());
```
Raycoms: won't this crash if null?

Mekle001: No, I don't think so. If so, I wouldn't be able to view any lists of recipes, and that works fine.

Raycoms: I think I still prefer a null check here

Mekle001: But we can't on the read, since this is a stream. By writing the null, the headers get written for the read to properly handle the null.

Raycoms: What I always do is, write a boolean of "false" if null and "true" if not, and if not, we read the resLoc

Mekle001: I thought that was encapsulated in the writeResourceLocation already?

Mekle001: Looking it up, its converts to string, and THAT is what could fail but a null string does get handled correctly since it starts with a length byte.

Mekle001: Null check in place.

marchermans:
I am not sure loot tables are a great idea here.....
Having a world in the recipe storage seems to be a really weird thing to me.

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/6201/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

marchermans (approved):

PR #6211, opened 2020-12-06
FriendlyFirePartialFix

tbase9:
ICitizenColonyHandler.java
- Fixed typo

CitizenColonyHandler.java
- Fixed typo

IGuardBuilding.java
- Removed checkIfGuardShouldTakeDamage() method from since it is no longer used in EntityCitizen.java and is referenced no where else.

EntityCitizen.java
- Changed instanceof check to PlayerEntity from ServerPlayer Entity
- Changed logic in handleSourceEntityForDamage to check if PlayerEntity is from the same colony as the EntityCitizen
- Changed logic in handleSourceEntityForDamage so EntityCitizen will not take damage from Player if there is currently a raid.
- Added logic so Guards will not take damage from player if they are currently following the player and replaced with logic in EntityCitizen.java

AbstractEntityAIGuard
- Added method to check if player is the damage target when their is a raid or the guard is following the player

EntityAIKnight
- Added call to AbstractEntityAIGuard to skip applying damage if the target is a player during a raid or a player that the guard is following.


tbase9:
I'm confident the Player->Guard changes should work. Less so about the Knight->Player aspect.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
```diff
@@ -1146,4 +1147,25 @@ public boolean canBeInterrupted()
...
+    /**
+     * Method which determines whether to apply attack actions.
+     *
+     * @return the boolean of whether or not to apply attack actions.
+     */
+    public boolean handleAttackExceptions(final Entity targetEntity)
+    {
+        if (targetEntity instanceof PlayerEntity)
+        {
+            if( worker.getCitizenColonyHandler().getColony() != null && worker.getCitizenColonyHandler().getColony().getPermissions().isColonyMember((PlayerEntity) targetEntity))
```
Raycoms, line 1156: maybe the check should here check for the guard attack permission

Raycoms:
I think we need to catch the player damage event and then catch the player damage, and check where the damage comes from, and if from a citizen, we have to do the checks.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/AbstractEntityAIGuard.java
tbase9: So you want to use this method, but change the method to work differently?

Raycoms:
Closed in favor of my own PR

PR #6249, opened 2020-12-15
Fluid Placeholder Block Placement

ShadowProtocol:
Follows up ldtteam/Structurize#234

- Allows builders to place Fluid Placeholder blocks as water or lava
    - would otherwise be placing a normal Fluid Placeholder block as is
- Uses the simpler method for adding placement handlers without clearing the original registry

Review please


ravenbuilder934:
Should the required Structurize version also be bumped?

Raycoms:
I already did

src/main/java/com/minecolonies/coremod/placementhandlers/PlacementHandlerInitializer.java
```diff
@@ -19,30 +19,15 @@ private PlacementHandlerInitializer()
...
+        PlacementHandlers.add(new GeneralBlockPlacementHandler(), PlacementHandlers.GeneralBlockPlacementHandler.class);
+        PlacementHandlers.add(new FieldPlacementHandler());
+        PlacementHandlers.add(new BuildingBarracksTowerSub());
+        PlacementHandlers.add(new BuildingSubstitutionBlock());
+        PlacementHandlers.add(new RackPlacementHandler());
+        PlacementHandlers.add(new WayPointBlockPlacementHandler());
+        PlacementHandlers.add(new ChestPlacementHandler());
+        PlacementHandlers.add(new FencePlacementHandler());
+        PlacementHandlers.add(new GatePlacementHandler());
+        PlacementHandlers.add(new NetherrackPlacementHandler());
```
Raycoms, line 31: Does this work nicely? The reason I had the complete override is to make sure the order stays like that.

ShadowProtocol: This way it preserves the original order in structurize, so now these mcol specific ones are stacked on top or overriding directly. It's in reverse order from before because its added to the start, not the end. I just tested it with a structure that uses most of those and it appears to work.

Raycoms: okay, sounds good

Raycoms:
:shipit: 

[Approved with PullApprove](https://pullapprove.com/ldtteam/minecolonies/pull-request/6249/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=ldtteam/minecolonies)

someaddons (approved):

PR #6285, opened 2020-12-22
Perf fixes

Raycoms:
- Cache reflection calls since those are slow
- Avoid streams and get first element from iterator.
- Avoid streams and do for loop

Review please


someaddons (dismissed):

marchermans (dismissed):

Raycoms:
@MatthiasMann I thought about that, but since the map is O(1) wouldn't the cost of the list declaration etc be almost the same?

MatthiasMann:
What do you mean with list declaration? Even a O(1) lookup costs time - in case of a hash map it needs to compute the hash value, compute the bucket index, traverse the chained bucket list (if there are hash collisions) and compare each key.

This is the same cost for get and containsKey - except that get also returns the value.
Storing the returned value in a local variable is completely free.

PR #6305, opened 2020-12-27
Improves Ship Raid Events, respells Babarian.

gattsuru:
- EventStructureManager was saving to paths including the full ResourceName of a dimension (eg schematicsstructbackup1 [minecraft : overworld]BlockPos{x=42, y=59, z=-208}.blueprint), which is not a valid file name on FAT32 and most Windows NTFS situations, as colons are a restricted character outside of POSIX emulation mode.  Changed structure to instead use ResourceName.namespace + ResourceName.path (eg schematicsstructbackup1minecraftoverworldBlockPos{x=42, y=59, z=-208}.blueprint).  Changed the load path management side to match.
 Added compat layer for those with existing old ship schematics floating around.

 New location is properly a path (eg schematics\structbackup\1\minecraftoverworld\BlockPos{x=42, y=59, z=-208}.blueprint)

This will also significant increase the number of raids for island or beach-side colonies for affected systems, as previously ship raids were being cancelled rather than rolling down to ground raids instead.

- Fixes babarian -> barbarian, checks for any locations using the old names.

- Fixes norsemen ships checking for biomes when spawned by command.  They will still check for water/ice regardless; if there's not a valid ship spawning location they'll do their ground variant instead.  NorsemenShipEvent natural spawns will still check for tundra biomes or the rare out-of-biome chance.

Review please


Raycoms:
One nitpick, we need a compat, for the ships that are already in the world pre updated and will be attempted to unload pos-updated

ravenbuilder934:
Also I noticed that the raid command didn't have a `pirate_ship_raid`, only `pirate_raid`, even though Norsemen have both ship and normal raids. Is this intentional?

gattsuru:
I can add it if useful, but it depends on the behavior you all desire.  Right now, norsemen_ship_raids will fall back to non-ship raids, and norsemen_raids will spawn a boat (uh, with the other fix) rather than just dumping them in the ocean.  If it'd be useful to have a way to only spawn the ground variants, I can add it pretty easily.

If I do that, I don't think I can practically limit pirate_raids and norsemen_ship_raids to their ship variants, though; either you have to cancel raid attempts that are too far from water, or you have to try respawning the event multiple times.   The former option is confusing, since especially raid tonight commands you'd get wrong feedback from the server message, and the latter could get extremely cpu and memory intensive or potentially go into infinite loop space.

Adding compat layer to successfully load old format files where they're present, rather than giving error messages, and also going to put the paths together so these file names don't drive me bonkers.

ravenbuilder934:
Maybe just do `norsemen_raid` then? If the ship raid can spawn on land and the normal raid can spawn on water (with a ship)?

gattsuru:
Ok.  Adds ground pirates to the event registry and gives them a distinct namespace, but excludes ground pirates and norsemen ship events from the commands suggestion builder and treats them as invalid (aka barbarians) if typed manually.  May eventually need extra sanity checks if that event registry gets used for anything else, but for now it's only relevant for save/unsave.

Added a compat layer to check for the old file format before looking to the new one.  Also changed the path for the new format such that the path portion actually is a path, eg (schematics/structbackup/1/minecraftoverworld/BlockPos{x=42, y=59, z=-208}.blueprint.

  Put a todo to remove the compat layer eventually,  since it does add extra file io and, even without the compat layer, it's just a couple log errors and a bit of work with an axe to remove.

src/main/java/com/minecolonies/coremod/colony/managers/EventStructureManager.java
```diff
@@ -156,8 +157,16 @@ public void loadBackupForEvent(final int eventID)
...
             if (entry.getValue() == eventID)
             {
-                final String backupPath = String.valueOf(colony.getID()) + colony.getDimension() + entry.getKey();
-                final String fileName = new StructureName("cache", "backup", Structures.SCHEMATICS_PREFIX + STRUCTURE_BACKUP_FOLDER).toString() + backupPath;
+                //TODO: remove compat for colony.getDimension()-based file names after sufficient time has passed from PR#6305
+                final String oldBackupPath = String.valueOf(colony.getID()) + colony.getDimension() + entry.getKey();
+                String fileName = new StructureName("cache", "backup", Structures.SCHEMATICS_PREFIX + STRUCTURE_BACKUP_FOLDER).toString() + oldBackupPath;
+
+                File blueprintFile = new File(fileName);
+                if(!blueprintFile.exists())
```
Raycoms: should be the other way around, first new backup path, if not exist try old

gattsuru: Ok.  Swapped to that approach, and also had it use the Structurize handlers for the check.

someaddons (approved):

Mekle001 (approved):

PR #6316, opened 2020-12-30
Rework the Sifter (phase 1) to be based on Recipes/Loottables 

Mekle001:
- Simplified the sifter and reused the crafting system to support it. 
  - Sifter no longer needs you to choose what to sift, it will sift anything siftable in it's building inventory., 
  - Sifter now supports prestock requests, which is the new way to request siftables to process
  - Sifter still supports a maximum number of blocks sifted per day, be sure to set this!
  - Sifter is now faster, and capable of being MUCH faster at higher levels. 
  - Sifter will rarely at higher primary skill levels, and only with iron/diamond meshes, produce more than one item from a single block
- Added support for diminishing returns on skill level, sifter is the first to use this for it's speed/luck
  - Early levels get a bonus, ie: skill 10 is an effective skill of 19, but skill 90 is effectively 99. 
- Added a chance for the baker's level 3 bread dough recipe to produce empty bottles.

This is only the first round of changes, and is effectively a drop in replacement as far as overall functionality. I plan on adding real items for the Sifter meshes, with durability that is effected by sifting. Will likely also add luck bonuses based on skills, and more complex loot tables (rare sand from gravel anyone?) 

Phase 2 will likely require the new research system also, as it allows for more complex conditions to trigger recipe availability. 

To be clear, Phase 2 will be a separate pull request after this one is in. This one stands alone. 

Review please


src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1282,6 +1282,7 @@
   "com.minecolonies.coremod.general.canttakeequipped": "The only armor you have is what you're wearing. Put it in your inventory if you want to give it to me.",
 
   "com.minecolonies.coremod.gui.workerhuts.siftersettings": "Daily limit",
+  "com.minecolonies.coremod.gui.workerhuts.sifterinfo": "Max %d :",
```
```suggestion
  "com.minecolonies.coremod.gui.workerhuts.sifterinfo": "Max %d",
```
If the max per day is before the colon, there's no need for it.

Mekle001: So, this displays as "Max %d : <textbox with>", so at level 1 it would be "Max 64 : [   0]" by default.

ravenbuilder934: Where the 64 is the max and the 0 is the current setting? I'd do something like `Daily Output 0/64` then

Mekle001: Well, the 0 is an input box so hard to put in the middle of a string like that. I expect this whole area to change further with Phase 2, possibly even remove this portion of the UI altogether.

ravenbuilder934: Ahh it's an input box. Two lines possibly?
`"com.minecolonies.coremod.gui.workerhuts.sifterinfo1": "Max: %d",
"com.minecolonies.coremod.gui.workerhuts.sifterinfo2": "Current: [fancy infobox stuff here]",`

Mekle001: Here's what it looks like on my test world: 
Imgur

ravenbuilder934: Kinda looks like you're changing the max imo

Mekle001: Ok, Updated to look like this: 
Imgur

ravenbuilder934: 👍

marchermans (approved):
Bit heavy with the loot context.
So we need to keep an eye on this, but implementation wise this looks good.

someaddons (approved):

PR #6325, opened 2020-12-31
More pathblocks

ravenbuilder934:
Closes #
Closes #
Closes #

- Adds a lot of new pathblocks
- I probably missed some
-

Review please


MatthiasMann:
I think stone brick stairs should stay - as it was the only way to make a path on hilly terrain so far and many colonies (including mine) use them.

someaddons:
I wanted to exclude the stairs as path blocks give a speed bonus, which doesnt fit going upstairs as that should be slower, pathing should already prefer stair blocks over others

someaddons:
ok think its better to do the speed modification another way, raven re-add the stairs please :D

MatthiasMann:
Would it be possible to check if the top side of the block is solid to determine speed? or would that cause issues with the grass path?

someaddons:
made it simpler, citizens just are a bit slower when changing y levels and dont get the path speed bonus there

Mekle001 (dismissed):

Mekle001 (approved):

PR #6410, opened 2021-01-18
1.16 mass rescan

ravenbuilder934:
Closes #
Closes #
Closes #

- Rescans all schematics that I could find in the schematic server and had an auto-scanner set up (and some that didn't have an auto-scanner). This is to fix issues with DataFixerSlayer as discussed in Discord.
- Also corrects some typos with schematic names: `pallisade` -> `palisade`, `barbacane` -> `barbican`, etc
-

Review please


Nightenom:
Just out of the curiosity what were the issues?

ravenbuilder934:
unknown

On the right

Nightenom:
Wow, this is truth versus the current blueprint in this repo? so I can take a look on what's going wrong

ravenbuilder934:
Yup, this is with a current blueprint that was last scanned in 1.15 and pasted with DataFixerSlayer running
@moleculor can tell you more

Nightenom:
building?

ravenbuilder934:
Affects all last scanned in 1.15, but I think the screenshot was taken with `fancy_corner_circle`? not sure

Nightenom:
Ah we have to ban this mod from structurize, I see the point, it literally removes the way our back-compatibility layer uses

ravenbuilder934:
Ban? As in not let Structurize load with it in?

Nightenom:
Ideally, or display a huge warning that the old blueprints won't load properly/entirely

ravenbuilder934:
Personally I'd rather go with the warning because people playing packs with both mods might not necessarily be using structurize at all
And this PR fixes almost all of the old schematics anyway

Nightenom:
yep, but the point is to not do everything-rescan every vanilla version

ravenbuilder934:
I think it'd be possible to do some sort of fully automatic rescan

> should be possible to write a converter command for that, which loads all schematics and saves them again with the datafixed blocks

quote from sam in discord

gattsuru:
RandomPatches has an optional behavior similar that can cause the same problem, if I understood Moleculor's reading properly.

I'd like to build a semi-automated tool to go through schematic files for various sanity checks, if that'd be acceptable, both for this purpose and some other benefits (getting familiarity with structurize, some sanity checks, getting list of top-five-blocks-used).  Won't be an immediate target, but if allowed would hopefully avoid this being a manual process for 1.17

Moleculor:
Both [Random Patches](https://www.curseforge.com/minecraft/mc-mods/randompatches-forge) (23 million downloads, but the DFU-killer is disabled by default) and [DataFixerSlayer](https://www.curseforge.com/minecraft/mc-mods/datafixerslayer) (Vazkii) do this. There may be other mods as well. And the mods are used for performance reasons. 

A single rescan should solve the issue, and possibly permanently (I make no promises)? If it ends up happening again between 1.16 to 1.17 (if they for some reason change the data structure *again*), I believe @gattsuru was mulling the idea of just writing a tool to do this rapidly, since the source code for the conversion process was made [available](https://github.com/Mojang/DataFixerUpper) by [Mojang](https://www.minecraft.net/en-us/article/programmers-play-minecrafts-inner-workings).

Nightenom:
still, we are not the only one using old blueprints, ppl are using their scans and this simply breaks everything that is not current version, so I would rather crash/yell in chat on world join

I get that it saves a bunch of memory, but it's also annoying for end-users (I don't care about us)

Nightenom:
> RandomPatches has an optional behavior similar that can cause the same problem, if I understood Moleculor's reading properly.
> 
> I'd like to build a semi-automated tool to go through schematic files for various sanity checks, if that'd be acceptable, both for this purpose and some other benefits (getting familiarity with structurize, some sanity checks, getting list of top-five-blocks-used). Won't be an immediate target, but if allowed would hopefully avoid this being a manual process for 1.17

planned in new structurize :D

ravenbuilder934:
> still, we are not the only one using old blueprints, ppl are using their scans and this simply breaks everything that is not current version, so I would rather crash/yell in chat on world join
> 
> I get that it saves a bunch of memory, but it's also annoying for end-users (I don't care about us)

Yeah let's do a warning too

Moleculor:
Oh it definitely impacts other mods as well, and annoys end users. I'm aware of [at least two](https://github.com/Vazkii/DataFixerSlayer/issues/2) other mods that killing DFU impacted, possibly others. They're also solving the issue by updating their data to 1.16.

*As* an end-user, however, I'll lean towards saying that if a mod **says** it's been ported to be compatible with 1.16, it's not *insanely* unreasonable to have the data present/included not slow things down simply because it was saved in 1.12 or 1.15 or whatever? (i.e. If it's ported to 1.16, it should be updated to 1.16 data?) Because from my vague understanding the options are either kill DFU or suffer a (likely small?) slowdown each time the server loads that particular old, un-updated content. 

The slowdown is likely unnoticeable in many situations, but I wouldn't be surprised if (for other mods) it impacts things like worldgen speeds.

Nightenom:
as I said in the dfSlayer issue:
> Well, I personally don't care about mod devs who can update their files "easily", the point with our mod (structurize/minecolonies) is that ppl have made their own schematics in older vanilla versions and we have to update them.

I have a plan of automatically overwriting the schematic file if it's old, but still, the DFU has to be there

Moleculor:
I also saw the comment about other users' schematics, which makes total sense.

Is there a way to split the difference? Detect old schematics, prevent [something] if schematicsAreOld && fakeDFUPresent? With a message about it?

Then users could potentially quit, shut things down, disable the DFU-killer, load the world, let your automated stuff do its magic, quit, shut everything down, reenable the DFU-killer, and live life happily?

ravenbuilder934:
I think the "automated stuff" in its current state has to run for each world load, not sure tho

Moleculor:
> I think the "automated stuff" in its current state has to run for each world load, not sure tho

If that were different, and it was a 'run and update once' kind of deal, I'd actually be more comfortable *as an end user* with Structurize flat out blocking game loading if a fake DFU was detected *until* it could make all the schematics be in a updated/functional/performant format. Then just let things load as normal.

But I'm not your typical end user, so I don't know which option would create more of a support headache for y'all.

Nightenom:
continue the conversation here if needed <https://github.com/ldtteam/Structurize/issues/264>

someaddons (approved):
I dont mind schematics beeing updated occasionally, as I wouldnt really trust with mojang keeping backwards compat forever

PR #6414, opened 2021-01-20
Feature/student hiring

Justen-N:
Closes #https://github.com/ldtteam/minecolonies-features/issues/192
- adds students, archers, and combat trainees to the hirable list. 
-
-

Review please


src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -200,8 +202,13 @@ protected void updateCitizens()
...
         citizens = colony.getCitizens().values().stream()
                      .filter(citizen -> !citizen.isChild())
-                     .filter(citizen -> (citizen.getWorkBuilding() == null)
-                                          || building.getPosition().equals(citizen.getWorkBuilding())).sorted(Comparator.comparing(ICitizenDataView::getName))
+                     .filter(citizen -> {
+                         return (citizen.getWorkBuilding() == null)
+                                 || building.getPosition().equals(citizen.getWorkBuilding())
+                                 || colony.getBuilding(citizen.getWorkBuilding()).getSchematicName().contains("library")
+                                 || colony.getBuilding(citizen.getWorkBuilding()).getSchematicName().contains("archery")
+                                 || colony.getBuilding(citizen.getWorkBuilding()).getSchematicName().contains("combatacademy");
```
gattsuru: Dunno if it's a huge deal here, but as a trick in the future, you can do "instanceof" on the Building for these type of checks and it's a lot less of a hassle than fighting with strings everywhere (especially for some of the cases where schematic names and class names are different).

Also lets developers do fun things with inheritance a bit simpler.

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -9,10 +9,12 @@
 import com.minecolonies.api.colony.IColonyView;
 import com.minecolonies.api.colony.buildings.HiringMode;
 import com.minecolonies.api.entity.citizen.Skill;
+import com.minecolonies.api.util.constant.BuildingConstants;
 import com.minecolonies.api.util.constant.Constants;
 import com.minecolonies.coremod.Network;
 import com.minecolonies.coremod.colony.CitizenDataView;
 import com.minecolonies.coremod.colony.buildings.AbstractBuildingWorker;
+import com.minecolonies.coremod.colony.buildings.workerbuildings.BuildingArchery;
```
gattsuru: Watch out for autoinserted import statements.  If you're using IntelliJ, I try to get in the habit of checking the Problems tab (alt-6 by default) for each changed file, since otherwise they'll end up collapsed out of view.  There's a lot of false positives there (it's _way_ too aggressive on typos, most obviously), but a quick way to check a bunch of files without having to scan them by eyeball.

I still end up committing unused logging imports, but at least it happens /less/.

gattsuru (commented):
More tips than a review:

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
Raycoms: This should be more generic. Create an interface like "IBuildingCanBeHiredFrom", add it to the views we want this in. And then just do this one check here.

Justen-N: Can do.

Justen-N: Alright. I'm obviously doing something wrong here. Because each building view for the three buildings in question is constructed by the AbstractBuildingWorker class, I extended that class into AbstractBuildingTraining, and made the view that AbstractBuildingTraining creates implement the IBuildingCanBeHiredFrom interface. Now, the code compiles and the check runs, but the instanceof isn't detecting the interface. I think. I'm going to go deeper into logging and debugging this, but figured I'd put this up there as a sheet anchor incase someone has an idea.

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
Justen-N: Extraneous imports removed in new commit.

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
gattsuru: You're in GUI world, so you probably do need to check against the BuildingWhatever.View.  See if : 

`|| colony.getBuilding(citizen.getWorkBuilding()) instanceof BuildingSchool.View 
|| colony.getBuilding(citizen.getWorkBuilding()) instanceof BuildingArchery.View
|| colony.getBuilding(citizen.getWorkBuilding()) instanceof BuildingCombatAcademy.View`

does what you need.

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -202,13 +201,10 @@ protected void updateCitizens()
...
-                     .filter(citizen -> {
-                         return (citizen.getWorkBuilding() == null)
-                                 || building.getPosition().equals(citizen.getWorkBuilding())
-                                 || colony.getBuilding(citizen.getWorkBuilding()).getSchematicName().contains("library")
-                                 || colony.getBuilding(citizen.getWorkBuilding()).getSchematicName().contains("archery")
-                                 || colony.getBuilding(citizen.getWorkBuilding()).getSchematicName().contains("combatacademy");
-                     }).sorted(Comparator.comparing(ICitizenDataView::getName))
+                     .filter(citizen -> (citizen.getWorkBuilding() == null)
+                             || building.getPosition().equals(citizen.getWorkBuilding())
+                             || colony.getBuilding(citizen.getWorkBuilding()) instanceof IBuildingCanBeHiredFrom).
```
Justen-N: So @gattsuru, this is what I mean by isn't working, we've changed how that check is implemented to this, but when the view is generated, I'm not sure that it is getting the interface.

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
Justen-N: I think I've got it. I thought they were implemented somewhere else, but they aren't. Thanks

src/api/java/com/minecolonies/api/colony/buildings/IBuildingCanBeHiredFrom.java
```diff
@@ -0,0 +1,5 @@
+package com.minecolonies.api.colony.buildings;
+//allows citizens from 'training' buildings to be added to hiring lists when implemented by the building.
+public interface IBuildingCanBeHiredFrom extends IBuildingWorkerView {
```
Raycoms: /** comment style (propper javadoc)
and 
{

} 

for the class

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -334,4 +336,12 @@ protected String createColor(final Skill primary, final Skill secondary, final S
         }
         return "";
     }
+
+    private boolean studentHiring(ICitizenDataView citizen) {
```
Raycoms: where is this called?

Justen-N: Dangit. thought I'd removed that. Was using it to debug, then refactored the code it contained.

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -19,15 +20,17 @@
...
 
+
 import java.util.*;
 import java.util.stream.Collectors;
 
 import static com.minecolonies.api.util.constant.TranslationConstants.COM_MINECOLONIES_COREMOD_GUI_HIRE_PAUSE;
 import static com.minecolonies.api.util.constant.TranslationConstants.COM_MINECOLONIES_COREMOD_GUI_HIRE_UNPAUSE;
 import static com.minecolonies.api.util.constant.WindowConstants.*;
-
+import static com.minecolonies.api.util.Log.*;
```
Raycoms: you can remove the util.Log and log4j imports then too

src/api/java/com/minecolonies/api/colony/buildings/IBuildingCanBeHiredFrom.java
```diff
@@ -0,0 +1,7 @@
+package com.minecolonies.api.colony.buildings;
+/**allows citizens from 'training' buildings to be added to hiring lists when implemented by the building.
```
Raycoms: Allows upper case and on the new line then. Check out some other classes for the style we're using in class based jadoc

Justen-N:
By the way, There is still a UI issue here. If a Student is included in the hire list, the hire button covers the student's last name. I don't fully understand the WindowPane system yet, so I have only completed the backend.

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
Justen-N: Done. Sorry for pinging you so much. My local version of this branch did something weird, so I'm working from the Github GUI.

src/api/java/com/minecolonies/api/colony/buildings/IBuildingCanBeHiredFrom.java
Justen-N: Done

Raycoms:
:shipit:

someaddons (approved):

PR #6429, opened 2021-01-23
Setup friendly fire fix

Raycoms:
Closes #
Closes #
Closes #

- Stops players from damaging citizens when there is a raid
- Stops citizens from damaging players when there is a raid
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/permissions/ColonyPermissionEventHandler.java
+++ b/src/main/java/com/minecolonies/coremod/permissions/ColonyPermissionEventHandler.java
@@ -26,8 +26,10 @@ import net.minecraft.block.ContainerBlock;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.MobEntity;
 import net.minecraft.entity.player.PlayerEntity;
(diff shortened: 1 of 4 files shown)


src/main/java/com/minecolonies/coremod/permissions/ColonyPermissionEventHandler.java
```diff
@@ -497,6 +500,19 @@ public void on(final ArrowLooseEvent event)
...
 
+    @SubscribeEvent
+    public void on(final LivingHurtEvent event)
+    {
+        if (event.getEntity() instanceof ServerPlayerEntity
+              && event.getSource() instanceof EntityDamageSource
+              && event.getSource().getTrueSource() instanceof EntityCitizen
+              && ((EntityCitizen) event.getSource().getTrueSource()).getCitizenColonyHandler().getColonyId() == colony.getID()
+              && colony.getRaiderManager().isRaided())
+        {
```
someaddons, line 520: I think we want to check the players permissions with the colony here, probably the guard attack one. Also a better function name + javadoc above would be nice :)

someaddons (approved):

PR #6503, opened 2021-02-06
Fix forester npe

vroni92:
Closes #
Closes #
Closes #

- Tell forester to walk to the tree if it has leaves left and is a nether tree to avoid work position being null
-
-

Review please


Raycoms:
:shipit:

PR #6540, opened 2021-02-14
JEI integration for querying crafting building support

uecasm:
This is an initial incomplete rough draft unpolished WIP sketch (and other such words) of an extension to the JEI integration within MineColonies to allow you to ask which buildings will allow you to craft certain items, either in-built or by teaching it the crafting recipe.

Presently, only the Mechanic has actually been implemented; I wanted to get some feedback on whether certain structural changes required to get this to work (see below) are acceptable or not before putting too much work into it.

## Features

image

Point at some item in JEI or your inventory and press R, and included among the crafting method tabs will be a tab for each colony building that can either directly craft it or can be taught to craft it (regardless of whether any current colony actually *has* been taught how to craft it or not -- or if any colonies exist at all).

It also contains a brief summary of the building/job or the kinds of things that it's expected to be able to craft, and a model of a citizen with that job.  (The model is randomised for each building whenever you join a world, including all the usual "new citizen" variations except for the medieval tag, which is never applied since there isn't currently a good way to know which tags are available.)

image

Additionally, if you click on the job title at the top, or select the hut block (anywhere, including on the left) and press U, it shows all potential recipes supported by that hut.

## Design changes that this required

* I needed a way to detect the list of allowable products; from the client side, when the particular building may not yet exist or indeed when there may be no colonies at all in existence.
    * Tags alone go a long way but are not quite sufficient, there are some special cases directly in the building code (e.g. to support mod-added minecarts without a data pack).  Unfortunately, this code is inaccessible without an existing colony building.
    * ~~So I have (partially) moved the relevant logic from the building to the hut block (as the hut block is the only one of the `BuildingEntry` class-triplet that exists on both sides without requiring any colony).  I'm open to putting it somewhere else but if the hut block isn't the right place then it would probably have to be a new fourth class, which would increase the burden to implementing new building types.  (The Job is another logical place for it, but this is also a little tricky to get at without a colony, although less so than the buildings.)~~
    * This would eventually result in removing the tag-searching code entirely from `AbstractBuildingWorker`; I haven't done that yet as the other job types still require it, but eventually that will go away as the rest are updated.
* ~~Similarly, it was necessary to move the job name to the hut block; this allows you to find the `JobEntry` from the block.  Again I haven't completely cleaned up some of the duplication in the building but that's relatively straightforward.~~
* I needed to make something public in `CitizenData` in order to be able to extract the list of supported skin tones.  Unfortunately `CitizenData` itself cannot be used for this without changing a lot of things to allow for a `null` colony.

## Future planned design changes (subject to change if there are objections)
* Change the interface ~~on the new hut block methods~~ to use `IColonyRecipe` instead of `IRecipeStorage` (and move the implementation of that somewhere less JEI-specific).
    * ~~Alternatively, modify `IRecipeStorage` to store inputs as `List<List<ItemStack>>` instead of only `List<ItemStack>`.~~
    * I left it as-is for now just to reduce the number of changes in the initial commit.
    * This will allow it to make decisions based on all alternative ingredient types (rather than picking one and hoping that it has compatible tags, as now -- which usually works fine but is not guaranteed to do so).~~
* Extend to all the other crafter types.  (If there's one that you think will be especially thorny, let me know so that I can try that one first to iron out any interface difficulties it may cause.)
* Also extend to the non-crafting-but-still-production types (e.g. apiary); not sure whether some of these might be trickier to identify, e.g. animal huts would probably have to check mob loot tables.

Please (informally) review; if there's any feedback or other things you want to discuss, that can either be covered here or by talking to me in the Discord; I'm Mirality there.


Raycoms:
I think moving the recipes to the blockHut is okay to acheive this. However, getJobId()
moving this to the blockhut doesn't work because for some jobs this is context based (teachers, guard towers, etc) and needs the building information to work.

I wonder if it might be easier to have a default constructor at the building only for JEIs sake or similar.
Definitely also needs @OrionDevelopment input.

Finally, I don't think ColonyRecipe should be necessary, neither a List<List<Recipe. The current list of input of the recipe storage could be easily used to build a two dim array. We could even add a list to IRecipeStorage with "getTwoDimInputList" or so.

uecasm:
> However, getJobId() moving this to the blockhut doesn't work because for some jobs this is context based (teachers, guard towers, etc) and needs the building information to work.

Teachers and guard towers aren't a problem because they're not crafting/production buildings, so they wouldn't show up in JEI anyway.

Though we could potentially index things the other way around if there might be future production buildings that do have multiple jobs with different crafting -- iterate through the registered jobs, then find the building for the job rather than the job for the building.  (Having said that, with the way that the recipe learning system works and the way the code was structured previously, it wouldn't have been feasible to have multiple jobs in the same building with different crafting options anyway, so I'm not sure this is a good justification.)

> Finally, I don't think ColonyRecipe should be necessary, neither a List<List<Recipe. The current list of input of the recipe storage could be easily used to build a two dim array. We could even add a list to IRecipeStorage with "getTwoDimInputList" or so.

Changing `IRecipeStorage` to accept `List<List<ItemStack>>` inputs was the other option.  This is the natural type that the recipe ingredients transform into.  (The current `List<ItemStack>` is sufficient for a recipe that has been manually selected by a player, as with the current learning system, but not one with abstract possible alternatives.)  I was more reluctant to do that as this is an existing API type and an existing persistent type, which will cause version compatibility issues if changed.  But it could be done.

Raycoms:
> > However, getJobId() moving this to the blockhut doesn't work because for some jobs this is context based (teachers, guard towers, etc) and needs the building information to work.
> 
> Teachers and guard towers aren't a problem because they're not crafting/production buildings, so they wouldn't show up in JEI anyway.

Yes, but that'd mean that the getJobId in the block would solely be for JEI purposes which is confusing.

> 
> Though we could potentially index things the other way around if there might be future production buildings that do have multiple jobs with different crafting -- iterate through the registered jobs, then find the building for the job rather than the job for the building. (Having said that, with the way that the recipe learning system works and the way the code was structured previously, it wouldn't have been feasible to have multiple jobs in the same building with different crafting options anyway, so I'm not sure this is a good justification.)
> 
> > Finally, I don't think ColonyRecipe should be necessary, neither a List<List<Recipe. The current list of input of the recipe storage could be easily used to build a two dim array. We could even add a list to IRecipeStorage with "getTwoDimInputList" or so.
> 
> Changing `IRecipeStorage` to accept `List<List<ItemStack>>` inputs was the other option. This is the natural type that the recipe ingredients transform into. (The current `List<ItemStack>` is sufficient for a recipe that has been manually selected by a player, as with the current learning system, but not one with abstract possible alternatives.) I was more reluctant to do that as this is an existing API type and an existing persistent type, which will cause version compatibility issues if changed. But it could be done.

Why can't this be like an additional method and or additional constructor rather than changing the API completely?

uecasm:
> Teachers and guard towers aren't a problem because they're not crafting/production buildings, so they wouldn't show up in JEI anyway.

To further clarify, the `hut.getJobId()` is allowed to return `null` if there's no sensible job that it can report (e.g. for the multi-worker case); these buildings would just never get a JEI crafting page.  The "real" building could still do whatever it wanted -- most would delegate to the hut but some could do their own different thing based on building-specific things.

uecasm:
> Why can't this be like an additional method and or additional constructor rather than changing the API completely?

Because if you're going to enhance it to support potential recipes then you *want* to change the API completely, to ensure that all the consumers deal with it correctly.  And it's redundant to store both data types since one is a strict superset of the other.

Also, most users of it construct it via a factory method that doesn't really seem to lend itself well to having multiple constructors.

But mostly it's because despite both being about recipes, they're not entirely related functions -- the primary purpose of one is to capture a single player-specified crafting recipe for use in actual server-side crafting, while the other is used to match ingredients from multiple possible recipes purely on the client side.  This is another reason why I made them separate types.

marchermans:
I think we should discuss this in Discord.
@uecasm You up for joining our dev discord?

And have a nice chat about this? Cause it requires some major changes to the inner workings of stuff.

But I like the end result, so most of them are probably worth making.

uecasm:
Sure.  As I said I'm already on the regular discord as Mirality.

marchermans:
Send you a friend request.

uecasm:
Alrighty, second pass.  (Still very WIP.)  The UI is unchanged from above, it's just backend changes.

* The block is back to normal; no changes there any more.
* I **did**, however, end up changing how modules are registered in `AbstractBuilding`.  They were registered into a map by concrete class, but I needed to find one based on interface without knowing the concrete class.  Interestingly, literally every single usage within `AbstractBuilding` itself was also doing the same thing, so I've just flattened it and `getModule` can now find a module by interface or super class, not just most-derived class.  Bonus: this also simplifies a lot of the code in `AbstractBuilding` itself.  The lists will be short enough that there shouldn't be any performance issues anyway.  It's even likely to be better.
* I've made an `ICraftingBuildingModule`.  Currently only the Mechanic has one, but the idea is that eventually all the crafting buildings (and possibly some of the non-crafting production buildings) would have one, and this will encapsulate all of the custom-crafting-related logic.  The concrete implementations of these are still building-specific and defined in the building class file.
* There are a few methods on the module that aren't actually called in the current code -- that was sort of an attempt to sketch out the general idea that all of the recipe-learning stuff (including the list of taught recipes itself, provided it can be moved without data loss) should eventually end up in there instead of in the building classes.  They might end up morphing a little as the dust settles, but it should suck in a lot of things that are currently in `AbstractBuildingWorker` and co.
* I haven't touched `IRecipeStorage`; it's still sufficient for storing single concrete learned recipes.
* I have kept `ColonyRecipe` (though it's a bit simplified), and also added `GenericRecipe` in the API, which is a wrapper used to make the actual decisions about which recipes are compatible.  It can be generated either from "potential" multi-alternative recipes as well as "concrete" `IRecipeStorage`.
    * I _might_ end up making `GenericRecipe` the "real" recipe type and morph what is currently `ColonyRecipe` into just a factory for `GenericRecipe`, since that's largely the role it currently has.

Along the way I noticed something slightly odd in `AbstractBuilding.getMaxInhabitants`.  As currently written, a building with no modules of any kind will return 1, a building with a `HomeBuildingModule` or a `TavernBuildingModule` will return whatever that says, and any other building (at least one module but not of those types) will return 0.  It's the first and last cases that seem suspicious; I don't know if that's intended or a bug, but it does seem peculiar.  I have not changed this behaviour as yet (though I did touch it as part of the module list change).

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -156,7 +156,7 @@
     /**
      * Set of building modules this building has.
      */
-    protected Map<Class<? extends IBuildingModule>, IBuildingModule> modules = new LinkedHashMap<>();
+    protected List<IBuildingModule> modules = new ArrayList<>();
```
Raycoms: I had it like this already, but then it failed in a few of the use cases I remember. Can you double check if the existing modules still work fine after this change?

Mekle001 (dismissed):
This is a fantastic start to modularization, and I look forward to the increased discoverability this brings to JEI. 

There will be some challenges when we get to things like the sifter, where it's loottable based, especially where the output changes based on attributes/level of the building. Secondary outputs may be enough, however. 

We may want to think about how to handle "tools" that burn durability during crafting too, as I'm about to enable that.

uecasm:
> There will be some challenges when we get to things like the sifter, where it's loottable based, especially where the output changes based on attributes/level of the building. Secondary outputs may be enough, however.

I was actually already thinking that the composter probably wouldn't use this system as-is -- instead we'd make a separate JEI category for the compost barrel itself (since the player can use that without the building), and the building wouldn't show up as its own tab like above but rather just as another way to use the barrel (on the left side).  This is because it's really its own entirely separate crafting mechanic, not related to the standard recipe crafting stuff.

I haven't looked into the sifter yet, but it sounds like it might be similar.

marchermans (dismissed):

uecasm:
Ok, I think the JEI parts (specifically) are now feature-complete.  Every building that seems like it has a crafting/smelting-related job has a JEI lookup now.

Let me know if anything seems odd or if there's a building/job not covered.

If we only want JEI integration, then we could stop here (though I'd want to clean up a few dangling bits).

Otherwise if we want to continue migrating crafting functionality into modules, the next step would probably be to look at migrating the recipe list and recipe learning support code, and removing these features from the buildings that don't really have any need for them.  Any other parts that ought to be in the crafting module instead of the building?

someaddons:
big enough as-is so I'd leave it at JEI integration. Better to do another PR lateron if one wants more changes

marchermans (dismissed):

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+import net.minecraft.network.PacketBuffer;
+import net.minecraft.util.JSONUtils;
+import net.minecraft.util.NonNullList;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.World;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class CompostRecipe implements IRecipe<IInventory>
```
Raycoms, line 23: This class def needs some more javadoc

src/api/java/com/minecolonies/api/crafting/CountedIngredient.java
```diff
@@ -0,0 +1,89 @@
...
+import net.minecraftforge.common.crafting.CraftingHelper;
+import net.minecraftforge.common.crafting.IIngredientSerializer;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class CountedIngredient extends Ingredient
```
Raycoms, line 19: Weird class name. Maybe "StackedIngredient" ? Also javadoc

src/api/java/com/minecolonies/api/crafting/GenericRecipe.java
```diff
@@ -0,0 +1,237 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static com.ldtteam.structurize.items.ModItems.buildTool;
+
+@SuppressWarnings({"LawOfDemeter", "ClassWithoutLogger", "NewMethodNamingConvention", "PublicMethodWithoutLogging", "ConditionalExpression"})
```
Raycoms, line 32: javadoc

Raycoms: And I think we can remove the SuppressWarnings, I'm removing that Intellij file in my PR. We have to have this nicely configured before adding it back in.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -309,6 +309,28 @@ public boolean canRecipeBeAdded(final IToken<?> ignored)
         return hasSpaceForMoreRecipes();
     }
 
+    /**
+     * @param token the recipe token
+     * @return whether the recipe can be added according to the crafting module (or false if there's no module)
+     */
+    protected boolean isRecipeCompatibleWithCraftingModule(final IToken<?> token)
+    {
+        return this.getModules(ICraftingBuildingModule.class).stream()
```
Raycoms, line 318: How often is this called, it might make sense to make this regular loops for performance reasons

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -309,6 +309,28 @@ public boolean canRecipeBeAdded(final IToken<?> ignored)
...
+                .map(crafting -> Optional.ofNullable(GenericRecipe.of(token))
+                        .map(recipe -> isRecipeCompatibleWithCraftingModule(crafting, recipe))
+                        .orElse(false))
+                .reduce(false, (cur, next) -> cur || next);
+    }
+
+    protected static boolean isRecipeCompatibleWithCraftingModule(@NotNull final ICraftingBuildingModule crafting, @NotNull IGenericRecipe recipe)
```
Raycoms, line 325: The second one needs some extra javadoc too

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -0,0 +1,114 @@
...
+import com.minecolonies.api.colony.jobs.IJob;
+import com.minecolonies.api.crafting.IGenericRecipe;
+import com.minecolonies.coremod.colony.buildings.AbstractBuildingCrafter;
+import net.minecraft.nbt.CompoundNBT;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Optional;
+
+@SuppressWarnings("ClassWithoutLogger")
+public abstract class AbstractCraftingBuildingModule extends AbstractBuildingModule implements ICraftingBuildingModule
```
Raycoms, line 16: Class needs more javadoc, Sub classes should probably be extra files.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBaker.java
```diff
@@ -316,4 +292,27 @@ public Window getWindow()
             return new WindowHutBaker(this);
         }
     }
+
+    public static class CraftingModule extends AbstractCraftingBuildingModule.Crafting
```
Raycoms, line 296: I think all the modules should be dedicated files rather than subclasses.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingConcreteMixer.java
```diff
@@ -341,4 +304,51 @@ public Window getWindow()
...
+        @Override
+        public boolean isRecipeCompatible(@NotNull final IGenericRecipe recipe)
+        {
+            return false;
+        }
+
+        @SuppressWarnings("ConditionalExpression")
+        @NotNull
+        @Override
+        public List<IRecipeStorage> getAdditionalRecipes(@NotNull final RecipeManager vanilla)
```
Raycoms, line 326: okay, now I got what we need the RecipeManager is used for here. Do we need that in so many cases?

src/main/java/com/minecolonies/coremod/compatibility/jei/LootTableAnalyzer.java
```diff
@@ -0,0 +1,198 @@
...
+                    final Item item = ForgeRegistries.ITEMS.getValue(new ResourceLocation(JSONUtils.getString(entryJson, "name")));
+                    final float weight = JSONUtils.getFloat(entryJson, "weight", 1);
+                    final boolean variableQuality = JSONUtils.getFloat(entryJson, "quality", 0) != 0;
+
+                    drops.add(new LootDrop(new ItemStack(item), weight / totalWeight, variableQuality));
+                }
+                else if (type.equals("minecraft:loot_table"))
+                {
+                    final ResourceLocation table = new ResourceLocation(JSONUtils.getString(entryJson, "name"));
+                    drops.addAll(toDrops(getLootTableJson(table)));
```
Mekle001, line 120: You're not taking weight into account here, nor modifying the returns appropriately. The nether gatherer I'm building which is based on the current sifter code calls through to a wide variety of sub loot tables, and this will produce a very bad set of probabilities. 

Frankly, I'm not comfortable with attempting to recreate the loot table logic for display, for reasons of performance and complexity, especially client side. I'd rather we calculate the drops on the server and include them in the recipes across the wire.

src/main/java/com/minecolonies/coremod/recipes/FoodIngredient.java
```diff
@@ -0,0 +1,76 @@
...
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.common.crafting.IIngredientSerializer;
+import net.minecraftforge.common.util.Lazy;
+import net.minecraftforge.registries.ForgeRegistries;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+public class FoodIngredient extends Ingredient
```
Raycoms, line 22: javadoc on all the new ingredient types

src/main/java/com/minecolonies/coremod/colony/crafting/SifterRecipe.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Represents a sifting recipe, consisting of information about the mesh to be used,
+ * and a list of input items and the results that they produce (via loot table).
+ *
+ * Note that it is legal for multiple recipes to exist with the same mesh type, but
+ * they are expected to have different input items in that case.  The mesh properties
+ * will be taken from one of them at random.
+ */
+@SuppressWarnings({"MethodReturnOfConcreteClass", "ClassWithoutLogger"})
+public final class SifterRecipe
```
Mekle001, line 25: This works great for the sifter, where the recipe doesn't yet have the mesh added to it, and mesh is the only pivot. Nether Gatherer recipes don't fit at all, and will be structured like the sifter was before this change. I'd rather see you focus on making that structure work well, as we'll have multiple crafters using it shortly.

Raycoms (changes requested):
The PR is pretty enormous and difficult to overview everything tbh. I had a few small comments above but overally it's a bit difficult to see the big picture.

I think a decent possibility would be:
a) Make the composter changes a seperate PR
b) Make this PR but leave the sifter and crusher changes out
c) After this PR then we can make a seperate PR for the sifter and crusher changes.

Having the sifter and crusher not directly in the JEI lookup is fine because they are very special cases (especially the sifter) and it makes it much easier to see the general API changes of the PR.

src/main/java/com/minecolonies/coremod/compatibility/jei/LootTableAnalyzer.java
uecasm: Well, currently there aren't actually any examples of a sub-loot-table.  So that could be looked into when that actually happens.

As for calculating it more directly on the server side, the main problem with that is that loot tables are not actually iterable, so you can't really get any answer better than "it might sometimes drop this item but it might be different in some way we didn't calculate" or by running a sample of say 1000 rolls on the loot table and estimating the results.  The more rolls, the more accurate, but obviously that's also a lot slower.  And you can still miss things, especially if the table includes context conditions.

This particular code is not comprehensive but it does suffice for all currently existing loot tables.  It also approximately matches what JER does, albeit simplified since we have simpler loot tables.

The best option would be to avoid loot tables like the plague and use recipe data that is more readily comprehensible instead, but I suspect that's not going to be an acceptable option.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBaker.java
uecasm: I didn't want to put them in separate files since they're just policy modifications on the base class rather than "real" classes.  Different defaults rather than new functionality.  Having them as separate files would just increase the clutter in the directory tree.

uecasm:
I could probably split some things back up into separate commits, so that you can see the changes individually there.  I had them like that before but merged it to avoid accidentally looking at older structure before the class names settled (which would be more confusing than helpful).  I'm not sure that separating into different PRs makes sense as the changes are fairly tightly integrated.

But you can still see some of the previous commits to get the basic idea, e.g. composting is at https://github.com/ldtteam/minecolonies/commit/524544a8566cd8686d38419fc3df7b5fc2833ac6.  Though the code there is not an exact match to the current code, the general shape is similar.

Raycoms:
> I could probably split some things back up into separate commits, so that you can see the changes individually there. I had them like that before but merged it to avoid accidentally looking at older structure before the class names settled (which would be more confusing than helpful). I'm not sure that separating into different PRs makes sense as the changes are fairly tightly integrated.
> 
> But you can still see some of the previous commits to get the basic idea, e.g. composting is at [524544a](https://github.com/ldtteam/minecolonies/commit/524544a8566cd8686d38419fc3df7b5fc2833ac6). Though the code there is not an exact match to the current code, the general shape is similar.

Ye, some things yes, but the sifter and crusher adjustments should be possible to be just pulled while maintaining the functionality, no? (Not visibly in JEI yet thugh).

And the composter was not very tight coupled I thought.

ravenbuilder934:
Conflict needs to be fixed

uecasm:
This is on hold anyway since it was requested to break it up into smaller chunks first.

uecasm:
Oh wait, who un-drafted it?  Let me fix that.

ravenbuilder934:
Closing for now

PR #6578, opened 2021-02-19
Update for new Structurize bricks

ravenbuilder934:
Closes #
Closes #
Closes #

- Add the new Structurize bricks (cream, beige, and brown) to the pathblocks tag
- Lets the new bricks be crafted by the Stonemason (and all their variants)
-

Review please


src/main/resources/data/minecolonies/tags/items/stonemason_ingredient.json
```diff
@@ -14,6 +14,9 @@
     "minecraft:polished_blackstone",
     "minecraft:polished_blackstone_slab",
     "minecraft:polished_blackstone_bricks",
-    "minecraft:basalt"
+    "minecraft:basalt",
+    "structurize:cream_bricks",
```
Mekle001: What is the recipe for these? Does it include dye?

ravenbuilder934: Beige bricks are bricks and gravel
Brown bricks are bricks and terracotta
Cream bricks are bricks and sandstone

Nothing includes dye. I added them to the ingredient tag so the stonemason could make their derivatives (beige brick stairs, etc). I added them to the product tag because I don't think bricks are included as an ingredient?

Mekle001: So, bricks are not included because all of the possible products from bricks are covered by the tags of #slabs, #stairs, etc in the product json. If the colored bricks have similar variants and they are properly tagged, it should work. I'll have to look in game, unless you have all of that handy.

ravenbuilder934: I think I did that for the `walls` tag, but not `slabs` or `stairs`

Mekle001: So, looking more at this, just adding minecraft:brick to the ingredients should make all of these craftable, and you shouldn't need to adjust either ingredient or product beyond that. If the stairs/etc aren't craftable, the right fix for those is to add them to the broader tags like minecraft:stairs.

ravenbuilder934: Done. I'll make a PR to structurize adding them to `minecraft:stairs` and `minecraft:slabs` as well

ShadowProtocol: The auto gen in my PR will do this automatically. So patience (it should be ready for review within 24 hrs) will save you doing even more things that I'll overwrite anyway,

Mekle001 (approved):

marchermans (approved):

PR #6585, opened 2021-02-20
New mystical site building

sboigelot:
All class and string have been updated from "wonder" to "mystical site" after request from reviewer.

# Intro & disclaimer

Hello, I'm new here, Tobugis on discord. I made this new building to learn a bit about the code base. I'm a dotnet dev and didn't do much java since uni long long ago. I tried to follow the same code style as the rest of the code but really feel free to tell me if I need to change anything.

This has not been requested anywhere but I just wanted it in my own game. It is a bit in line with the future plan of entertainment building but not much. If you do not think it has a place in the mod, there is no issue for me, just let me know 👍 

All balancing are welcome, and I'm not deeply attached to the schematics. So if you wish to propose any changes, feel free to do so!

- Introduce a new Building: the wonder. Once built, the wonder will slightly raise the citizen happiness by making them proud. (w: 1, value: 1 +(lvl building/2)
image

- Each level of the wonder requires a builder upgrade (same as for the citizen building)

- A research to unlock the building: "ambition" costing diamond
image

- A crafting recipe for the wonder hut block:
image

-A 3d model for the wonder hut block: 
image

- A set of schematics for wonder lvl 1 to 5 in the stone folder:
image

- A set of schematics for wonder lvl 1 to 5 in the nordic folder:
image
image

- 2 advancements:
image
image


ravenbuilder934:
Why did you implement this instead of an actual planned building?

src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
```diff
@@ -338,12 +345,36 @@ public ITownHall getTownHall()
...
 
+    @Override
+    public int getWonderMaxBuildingLevel()
+    {
+        int maxLevel = 0;
+        if(hasWonder())
+        {
+            for(final IWonder wonder : wonders)
+            {
+                int wonderLevel = wonder.getBuildingLevel();
```
Raycoms: I think you can simplify this by just doing wonder.getBuildingLevel() twice, it's just as expensive as making a new var

src/main/java/com/minecolonies/coremod/colony/buildings/BuildingWonder.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public BuildingEntry getBuildingRegistryEntry()
+    {
+        return ModBuildings.wonder;
+    }
+
+    /**
+     * The client side representation of the building.
+     */
+    public static class View extends AbstractFilterableListsView
```
Raycoms: This doesn't have to be FitlerableListsView. Since it doesn't have a special UI.

Raycoms:
> Why did you implement this instead of an actual planned building?

Everyone can implement what they want after all. However, of course it's always subject to the maintainers if it's accepted or not.

ravenbuilder934:
> > Why did you implement this instead of an actual planned building?
> 
> Everyone can implement what they want after all. However, of course it's always subject to the maintainers if it's accepted or not.

yes, of course. but it's more likely to be accepted if it's already planned. As it is I don't see the point of this building tbh

Raycoms:
> > > Why did you implement this instead of an actual planned building?
> > 
> > 
> > Everyone can implement what they want after all. However, of course it's always subject to the maintainers if it's accepted or not.
> 
> yes, of course. but it's more likely to be accepted if it's already planned. As it is I don't see the point of this building tbh

It's like a "happiness boosting building". For example for the asian the bathhouse could be converted to this.
Then those buildings could be points where citizens hang out etc.

Raycoms:
> > > > Why did you implement this instead of an actual planned building?
> > > 
> > > 
> > > Everyone can implement what they want after all. However, of course it's always subject to the maintainers if it's accepted or not.
> > 
> > 
> > yes, of course. but it's more likely to be accepted if it's already planned. As it is I don't see the point of this building tbh
> 
> It's like a "happiness boosting building". For example for the asian the bathhouse could be converted to this.
> Then those buildings could be points where citizens hang out etc.

After all, for later, when we got the pvp modes etc. We also wanted to have wonder-alike buildings like in AoE

ravenbuilder934:
> > > > Why did you implement this instead of an actual planned building?
> > > 
> > > 
> > > Everyone can implement what they want after all. However, of course it's always subject to the maintainers if it's accepted or not.
> > 
> > 
> > yes, of course. but it's more likely to be accepted if it's already planned. As it is I don't see the point of this building tbh
> 
> It's like a "happiness boosting building". For example for the asian the bathhouse could be converted to this.
> Then those buildings could be points where citizens hang out etc.

I understand that. But I also think a building like that should do more than just sit there (which is what it does atm if I'm understanding this right). I'd rather boost their happiness by having them hang out at the tavern after work, for example

sboigelot:
> > > > > Why did you implement this instead of an actual planned building?
> > > > 
> > > > 
> > > > Everyone can implement what they want after all. However, of course it's always subject to the maintainers if it's accepted or not.
> > > 
> > > 
> > > yes, of course. but it's more likely to be accepted if it's already planned. As it is I don't see the point of this building tbh
> > 
> > 
> > It's like a "happiness boosting building". For example for the asian the bathhouse could be converted to this.
> > Then those buildings could be points where citizens hang out etc.
> 
> I understand that. But I also think a building like that should do more than just sit there (which is what it does atm if I'm understanding this right). I'd rather boost their happiness by having them hang out at the tavern after work, for example

As I said, the reason I implemented it is because I wanted it in my game and now I have it as long as I use my fork to play. But of course, no worry if you don't think it's interesting for the mod. For now it only provide an extra challenge for the player and some passive happiness buff indeed.

src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
sboigelot: Updated

src/main/java/com/minecolonies/coremod/colony/buildings/BuildingWonder.java
sboigelot: Updated to extends AbstractBuildingView instead

marchermans (dismissed):
I like the concept but could you rename it to mystical site?

src/main/generated/resources/data/minecolonies/loot_tables/blocks/blockhutwmysticalsite.json
```diff
@@ -0,0 +1,25 @@
+{
+  "type": "minecraft:block",
+  "pools": [
+    {
```
ravenbuilder934: The file name is `wmysticalsite`?

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1555,6 +1560,7 @@
   "com.minecolonies.coremod.gui.townhall.happiness.security": "Guards in the Colony",
   "com.minecolonies.coremod.gui.townhall.happiness.social": "Social Factor",
   "com.minecolonies.coremod.gui.townhall.happiness.saturation": "Hunger",
+  "com.minecolonies.coremod.gui.townhall.happiness.mysticalsite": "Mystical Site",
```
```suggestion
  "com.minecolonies.coremod.gui.townhall.happiness.mysticalsite": "Mystical Sites",
```

ravenbuilder934 (changes requested):
The texture is still called `blockhutwonder`

And you really shouldn't be adding schematics. Each style has its own builder, and they're usually the only one who can touch that style (except for bug fixes).

src/main/generated/resources/data/minecolonies/loot_tables/blocks/blockhutwmysticalsite.json
sboigelot: Sorry, I corrected it, so much refactor that one escaped me

sboigelot:
> The texture is still called `blockhutwonder`
> 
> And you really shouldn't be adding schematics. Each style has its own builder, and they're usually the only one who can touch that style (except for bug fixes).

I updated the texture name.

About the schematics, should I delete them all? or keep at least the stone ones?

Raycoms:
Keep the schematics for now, the style authors can still adjust if necessary

sboigelot:
Great, they can feel free to re-do them completely :) No worry on my side

Raycoms:
Stone, Wooden and Nordic are "minecolonies styles" so many people worked on them, they don't really belong to anyone (nordic mostly Luna, but also a bunch of other people)

ravenbuilder934 (approved):
still disagree with the concept but ¯\_(ツ)_/¯

Raycoms:
@sboigelot can you hit me up on discord?

marchermans:
> The texture is still called `blockhutwonder`
> 
> And you really shouldn't be adding schematics. Each style has its own builder, and they're usually the only one who can touch that style (except for bug fixes).

@ravenbuilder934 In my eyes any building without a schematic is rejected by default. Yes the styles are up to the builders to adapt, but they are, at least in my eyes, second priority. The players are the first, as such when I add a building I add as many or as little styles as I want in the first PR so that players can at least play with them.

One other thing: Everybody can touch everybody styles. This mod is opensource and no one builder has the prorogative over a given style.

PR #6587, opened 2021-02-20
Enable tool retention/durability burn

Mekle001:
* Add support for tools to survive rather than be recreated during crafting
* Attempt to apply 1 damage to a tool each use during crafting. 
* Fix logic of addRecipe in Baker to not add furnace recipes if the precursor recipe wasn't added. 

This does NOT include full improvements to the RS to support asking for ingredients with durability.  Currently, it will check the building and suppress adding the request if there is enough durability.


src/main/resources/data/minecolonies/crafterrecipes/sifter/gravel.json
```diff
@@ -7,5 +7,16 @@
         "count": "1"
     }
   ],
+  "additional-output": [
```
Mekle001, line 10: These lists of possible outputs are NOT complete, they are a rough indicator of the sorts of things to be expected.  Some items require specific meshes, and some require certain luck/skill.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -265,7 +271,26 @@ public IRecipeStorage getFirstFullFillableRecipe(final Predicate<ItemStack> stac
...
     {
         final List<IItemHandler> handlers = getHandlers();
-        return storage.fullfillRecipe(this.getColony().getWorld(), handlers);
+
+        final AbstractEntityCitizen worker = this.getMainCitizen().getEntity().orElse(null);
+
+        if(worker == null)
+        {
+            // we shouldn't hit this case, but just in case...
+            return storage.fullfillRecipe(this.getColony().getWorld(), handlers);
```
Raycoms, line 280: can't we just return false or so?

Mekle001: This creates full backward compat, in case there is a scenario I missed, and the worker isn't set.

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -561,14 +612,26 @@ public ResourceLocation getRecipeSource()
 
     @NotNull
     @Override
-    public List<ItemStack> getSecondaryOutputs()
+    public List<ItemStack> getSecondaryOutputs(boolean includeTools)
     {
-        return secondaryOutputs;
```
Raycoms, line 566: I think we might want to separate tools from secondary outputs into a separate thing alltogether.
Either we treat them the same way as everything and just do a damage check on each item. Or we treat them differently and use two different methods.

Mekle001: This is a longer discussion, but I thought about that. I'll talk to you in discord about it.

src/main/resources/data/minecolonies/crafterrecipes/sifter/gravel.json
uecasm: I don't think it's a good idea to use additional-output for that, that's normally only for 100% guaranteed extra outputs (for conditional outputs, there's the loot table).

Mekle001: I hear what you're saying, but the loot tables aren't easily readable for display in the UI, especially when nested. We also likely don't want to display some of the items due to the conditions involved.  We need a way to declaratively state this are possible outcomes, even though the code will NOT Produce anything from them without a primary output defined. (What's a secondary output without a primary?)

So far,this is the only scenario I found where we need a display only list of items for output, and adding another list to custom recipe and every recipeStorage seems like overkill, when we have two sitting idle in this scenario. I didn't want to use the alternate list, as it implies creation of a classic single-output recipe in the RS.

marchermans (approved):

Issue #5842, opened 2020-09-28
Entry-Level Task Suggestions


Raycoms:
Hi, this is a suggestion board for people wanting to contribute to MineColonies. 
If you pick one of these, please create a new issue and assign it to yourself.

If you have other ideas about additions or questions regarding one of the issues, please contact us on our [discord](https://discord.minecolonies.com).

- [ ] Create Boar entity that spawns randomly around colonies, searches for fields to damage them (can dig out dirt).
- [ ] Create Goat entity that spawns randomly around colonies, searches for fields to damage them (can jump over fences).
- [ ] Create Hunter AI that has some dogs following him around that will hunt the entities above.
- [x] Extend the current Permission system to allow people to add additional ranks.
- [ ] Diverse mod compatibility requests: https://github.com/ldtteam/minecolonies-features/issues
- [ ] Other features from the link above (ask first to make sure it's actually planned)
- [x] Adjust the research system to use datapacks instead of statically registering it.
- [x] Fix research GUI to display arrows correctly
- [x] Adjust sifter drops to datapacks instead of defining it in the config.
- [x] Fix up hut block textures to use Minecraft base textures instead of our hardcoded ones.
- [x] Allow builder to build things that are placed with the shape tool
- [ ] Add special armor drops to Vikings
- [x] Add a special spear item Amazon raiders use and drop
- [x] Add a special medieval armor that only the blacksmith can create

Pyrocantaes:
Another question .. Guard towers have the option "recruit from Combat Academy/Archer" which is on by default.. however, when manually recruiting workers those within the Combat Academy/Archer are never shown as available.. you have to fire the from the academy/archer before you can recruit them to the guard tower.. is this just missing functionality or intent?

Asherslab:
missing functionality

ravenbuilder934:
October is over, should this be closed?

gattsuru:
![image](https://user-images.githubusercontent.com/39036909/100528313-88d7a500-31a9-11eb-95ea-9f357301a53e.png)

That includes conversion of existing recipes to json from their static assignments, as well as loading from user-defined per-world datapacks.

Lots of work to be done on this topic still before it's ready for a pull request, but put together a testbed of the core functionality.  I'd be grateful if people could sanity-check the json format, in particular -- it'll be a lot easier to add extra functionality now than a while down the road.

(especially since I was lazy and put together the initial export of existing researches in an ungodly mess of Java->csv->C#->bash script.)

[...]

PR #6606, opened 2021-02-24
Add a special medieval armor that only the blacksmith can create

sboigelot:
Last Entry-Level Task Suggestions #5842

"Add a special medieval armor that only the blacksmith can create"

Add a full set of medieval "plate" armor that only the blacksmith can create starting at lvl 4
The armor has a similar-ish damage reduction than diamond but only use iron+coal+leather (bonus of having lvl 4 blacksmith)

Guard using the armor set:
image

Blacksmith recipes:
image

Review please


sboigelot:
Sorry my picture for the recipe is incorrect, I remake it

sboigelot:
> Sorry my picture for the recipe is incorrect, I remake it

Ok the PR is ready now

Mekle001 (changes requested):
Yay! I'm glad to see this in progress. 
Overall, the changes as they stand look pretty good, but..

We need a better gate on access to this. Building level is fine to be part of it, we also need a research that needs to be done to unlock this. 

Please add a research to block this on.

ravenbuilder934:
I'm also a bit curious as to why coal is involved in its crafting.

sboigelot:
> I'm also a bit curious as to why coal is involved in its crafting.

Medieval plate armor were sometimes made out of steel. A common simplification of steel in mc is iron+coal.
I can of course change the recipe.

Raycoms:
An easy way to combine research + building level, make a research that requires blacksmith level 4

MatthiasMann:
It's a bit much to add a new item just for making the armor - but what about having the smelter make the steal?

Raycoms:
> It's a bit much to add a new item just for making the armor - but what about having the smelter make the steal?

Not easily possible, we don't have special smelter recipes yet.

sboigelot:
> Yay! I'm glad to see this in progress.
> Overall, the changes as they stand look pretty good, but..
> 
> We need a better gate on access to this. Building level is fine to be part of it, we also need a research that needs to be done to unlock this.
> 
> Please add a research to block this on.

I added the plate armor research in the combat tree to unlock the blacksmith recipes;
image

src/main/resources/data/minecolonies/crafterrecipes/blacksmith/platearmorhelmet.json
```diff
@@ -0,0 +1,25 @@
...
+      "count": 1
+    },
+    {
+      "item" : "minecraft:coal",
+      "count": 1
+    }
+  ],
+  "result": "minecolonies:plate_armor_helmet",
+  "count": 1,
+  "min-building-level" : 4,
```
Mekle001, line 20: Given that the research now enforces this, it's probably not necessary in the recipe definition, but it's also not critical.

Mekle001 (changes requested):
Ok, code is now looking good, and I love the new research.

However, all of the textures are showing as empty files.

Mekle001:
Oh, and find me on discord. I have some ideas for another iteration on top of this.

src/main/resources/data/minecolonies/crafterrecipes/blacksmith/platearmorhelmet.json
sboigelot: If the player has multiple blacksmith including some < lvl 4, this will prevent him to build the plate armors in the lower lvl blacksmith I guess

Mekle001: Fair point.

sboigelot:
> However, all of the textures are showing as empty files.

I'm lost about this - The textures are present and not empty on my fork, is this an issue with the auto PR?

https://github.com/sboigelot/minecolonies/blob/feature/1.16/plateArmor/src/main/resources/assets/minecolonies/textures/items/plate_armor_boots.png
https://github.com/sboigelot/minecolonies/blob/feature/1.16/plateArmor/src/main/resources/assets/minecolonies/textures/items/plate_armor_chest.png
https://github.com/sboigelot/minecolonies/blob/feature/1.16/plateArmor/src/main/resources/assets/minecolonies/textures/items/plate_armor_helmet.png
https://github.com/sboigelot/minecolonies/blob/feature/1.16/plateArmor/src/main/resources/assets/minecolonies/textures/items/plate_armor_legs.png

Mekle001 (dismissed):
Looks like the texture issue may actually be an issue in the diff. I'm going to assume so, since the images are viewable from the diff when you view file on them, even though the diff shows them as empty.

sboigelot:
I re-committed the files after a fresh open in paint.net & save.
It seems to have fix the diff

someaddons (commented):
Nice armor think this should fit our citizens well. One minor thing: could you reduce the material durability to 20? As this is significantly cheaper than diamond and we got some stuff to improve durability ourselves already

someaddons (approved):

PR #6610, opened 2021-02-25
Some fixes

someaddons:
Fix miner npe
Fix AI not toggling both door blocks
Add request update on hiring a worker

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
@@ -505,6 +505,8 @@ public abstract class AbstractBuildingWorker extends AbstractBuilding implements
             colony.getProgressManager()
               .progressEmploy(colony.getCitizenManager().getCitizens().stream().filter(citizenData -> citizenData.getJob() != null).collect(Collectors.toList()).size());
         }
+
+        updateWorkerAvailableForRecipes();
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -505,6 +505,8 @@ public boolean assignCitizen(final ICitizenData citizen)
             colony.getProgressManager()
               .progressEmploy(colony.getCitizenManager().getCitizens().stream().filter(citizenData -> citizenData.getJob() != null).collect(Collectors.toList()).size());
         }
+
+        updateWorkerAvailableForRecipes();
```
Raycoms, line 509: Maybe have a check for "complex recipes" so that this doesn't happen on private recipe add

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -610,6 +618,22 @@ public boolean addRecipe(final IToken<?> token)
         return false;
     }
 
+    /**
+     * Updates existing requests, if they match the recipes available at this worker
+     */
+    private void updateWorkerAvailableForRecipes()
```
Mekle001, line 624: We should consider making this public. We do this work in several other places today, and we should probably unify them. It's possible they are all in abstractbuildingworker, but I don't think so.

someaddons: I've checked the other places, and none of them want to update all recipes

Mekle001 (dismissed):

PR #6589, opened 2021-02-20
Code cleanup and other things

ravenbuilder934:
Closes #
Closes #
Closes #

# Changes proposed in this pull request:
- Fixes some models and textures: chicken herder hut, composter hut, concrete mixer hut, courier hut, enchanter's tower, glassblower hut, shepherd hut, warehouse. These are from Odd
- Fixes the IntelliJ Gradle configuration (the latest update broke it, at least for me)
- Fixes some typos in the build.gradle
- Bumps Forge and JEI versions in build.properties
- Removes a *lot* of unneeded imports
- Deletes unused assets
- Removes some dead code, constants, files, etc
- Fixes a couple typos in javadoc/etc
- Deletes `StructureTagConstants` and moves its constants to `SchematicTagConstants`
- Adds a lang string for the "Give Tool" button in the forester GUI (completes this TODO)
- Removes a couple TODOs that are already done
- Removes the duplicate forester crafting recipes for stripped logs (that are gated behind a nonexistent dummy research)
- Cleans up some other minor things (renames the `papper` textures to `paper`, fixes other typos in texture names, renames a constant to be in line with the others, etc)

------

### File size after: 29.8 MB

-------
Review please

PR #6615, opened 2021-02-26
Manual changes

ravenbuilder934:
Closes #
Closes #
Closes #

- Manual changes from #6589 
- Removes unused constants
- Moves the constants in StructureTagConstants to SchematicTagConstants
- Renames the constant LEATHER_LEVEL_RANGE to LEATHER_BUILDING_LEVEL_RANGE to fit the other constants
- Fixes some typos in jdocs and todos
- Bumps JEI, Structurize, and Forge versions
- Removes already-completed todo
- Completes TODO for the lumberjack "give tool" button
- Removes some dead lang strings (relating to the scan tool)
- Removes some dead code
- some research and mystical site lang fixes

Review please


src/api/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -49,7 +49,7 @@
      */
     private InventoryUtils()
     {
-        /*
```
Asherslab: these lines aren't javadoc. they should remain as they were. IDE even gives a warning about it

Asherslab (dismissed):
i don't see anything glaringly wrong.
If it compiles i dont care. 

Although, why are we bumping forge/jei versions? shouldn't do that unless we *need* to

ravenbuilder934:
Why not?

ravenbuilder934:
ah I see the problem with forge

Nightenom:
> i don't see anything glaringly wrong.
> If it compiles i dont care.
> 
> Although, why are we bumping forge/jei versions? shouldn't do that unless we _need_ to

imho We should bump forge regularly to ensure that ppl can run latest versions

Asherslab:
@Nightenom 
forge shouldn't be pushing breaking changes. if they do we can bump then. And they're pretty vocal about it when they do
but building for the lowest version we can means that players can use any version above that (as long as there is no breaking change)

marchermans (dismissed):

PR #6625, opened 2021-02-27
Make building modules accessible by parent class/interface as well as exact interface

uecasm:
This is how everything inside AbstractBuilding was treating them anyway.

- Changes building module lookup so you can find modules by base class or interface as well as concrete class.

Review please


Raycoms:
:shipit:

marchermans (approved):

PR #6626, opened 2021-02-27
Resourceful Bees compatibility

truenachtara:
- Adds compatibility with Resourceful Bees' tiered beehives. 
- Beekeepers are now able to harvest the correct, customized items from RB hives rather than always harvesting 3 vanilla honeycombs.
- build.gradle is updated to include RB for compiling but still does not require it as a dependency.

I've done my own testing, and I haven't encountered any issues. Cannot be made to work with RB bees working in vanilla hives as they do not properly store RB data.

Review please


build.gradle
```diff
@@ -99,6 +99,7 @@ dependencies {
     compile "com.ldtteam:datagenerators:0.1.38-ALPHA"
     
     implementation fg.deobf("mezz.jei:jei-1.16.4:${config.jei_version}")
+    compileOnly fg.deobf("curse.maven:resourcefulbees-384508:3206649")
```
marchermans: No.

Dependencies from the CurseForge maven are not allowed in this project.
Please provide a proper maven repository for this jar.

marchermans (changes requested):

build.gradle
truenachtara: This jar does not have a proper maven repository. I have contacted the devs about this.

marchermans: Then I am sorry, but this can't be merged.
The CurseForge maven is far from reliable and can't be properly proxied via our LDTTeam Buildsytem maven proxies.

ravenbuilder934: Why do you need it here at all?

truenachtara: I have code in another class that relies on RB to exist within the dev environment to compile properly. Is there a better solution?

marchermans: Depends.

I am not sure how Vanilla MC / Resourcefullbees exposes the items in the hive.

truenachtara: RB stores their items in NBT data and provides a method: TieredBeehiveTileEntity.getResourceHoneycomb() which removes one ItemStack from the beehive's storage, and returns it. I also need to import TieredBeehiveTileEntity so that I can check that it's one of RB's hives. Vanilla hives do *not* store items like this, as the only expected drop is honeycombs.

truenachtara: Maven is proper now

src/main/java/com/minecolonies/coremod/compatibility/resourcefulbees/ResourcefulBeesCompat.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import net.minecraftforge.fml.ModList;
+
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Stack;
+
+public class ResourcefulBeesCompat
```
Raycoms: this class and method should get some javadoc, else :shipit:

Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/compatibility/resourcefulbees/ResourcefulBeesCompat.java
truenachtara: will do in just a bit

ravenbuilder934:
If you bump Forge might as well bump it to the latest

someaddons (changes requested):
I'd like a bit different structure for the compat, so I went ahead and proposed one on your repo: https://github.com/truenachtara/minecolonies/pull/1

truenachtara:
latest forge has conflicts with mekanism last time i checked, that forge bump was likely accidental. forge 36.0.39 is the latest that will work with mek

someaddons (dismissed):

src/api/java/com/minecolonies/api/compatibility/Compatibility.java
```diff
@@ -29,6 +32,8 @@ private Compatibility()
         throw new IllegalAccessError("Utility class");
     }
 
+    public static IBeehiveCompat beeHiveCompat = null;
```
Raycoms: Can't we give this the default value already?

Raycoms: (javadoc please too)

someaddons: ye could do = new IBeehiveCompat(){}

marchermans (dismissed):
Generally approved. Just a question cause I do not see it properly handled anywhere: what happens if two different beehive compacts are loaded?

Raycoms:
> Generally approved. Just a question cause I do not see it properly handled anywhere: what happens if two different beehive compacts are loaded?

Yeah, if we ever run into the case of someone wanting to add a second compat for the same thing. They'll have to add a compat handler similarly to how we handle placements atm.

build.gradle
```diff
@@ -35,7 +35,7 @@ repositories {
         dirs 'libs'
     }
     maven {
-        url = "https://www.cursemaven.com"
+        url 'https://nexus.resourcefulbees.com/repository/maven-public/'
```
marchermans: This can be removed. The LDTTeam buildsystem maven on jfrog.ldtteam.com has this repo now mirrored and cached.

Please remove this before merging.

src/api/java/com/minecolonies/api/compatibility/Compatibility.java
Raycoms: After this small adjustment its ready to merge.

marchermans (dismissed):
Update the maven repositories.

src/api/java/com/minecolonies/api/compatibility/Compatibility.java
Raycoms: @truenachtara

marchermans: Yes this needs a default value.

src/api/java/com/minecolonies/api/compatibility/Compatibility.java
```diff
@@ -268,4 +273,17 @@ public static boolean isDynamicFamilyFitting(final BlockPos block1, final BlockP
...
+     * Get comps from a hive at the given position
+     *
+     * @param pos    TE pos
+     * @param world  world
+     * @param amount comb amount
+     * @return list of drops
+     */
+    public static List<ItemStack> getCombsFromHive(BlockPos pos, World world, int amount)
+    {
+        return beeHiveCompat.getCombsFromHive(pos, world, amount);
```
marchermans, line 287: Either populate the compat with something by default, or add a null check here.

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -864,4 +866,12 @@ private void discoverFreeBlocksAndPos()
             }
         }
     }
+
+    /**
+     * Inits compats
+     */
+    private void discoverModCompat()
+    {
+        ModList.get().isLoaded("resourcefulbees") ? Compatibility.beeHiveCompat = new ResourcefulBeesCompat() : Compatibility.beeHiveCompat = new IBeeHiveCompat() {};
```
marchermans: Wrong code style. Assignment style is used in Minecolonies.

marchermans (changes requested):

truenachtara:
What the heck happened to the maven repo? I just imported it into my IDE and now it won't build or anything. I even re-added the RB maven to see if that helps, and it still can't see RB stuff.

truenachtara:
Added some stuff, but I'm having problems with getting the IDE to recognize the existence of Resourceful Bees. I'm talking with the mod author to try to see if they changed something on their end

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
truenachtara: Not sure what this means, exactly.

Asherslab: `Compatibility.beeHiveCompat = ModList.get().isLoaded("resourcefulbees") ?  new ResourcefulBeesCompat() : new IBeeHiveCompat() {};`
That

For readability's sake, newlines after `?` and `:` are preferred

truenachtara: Ah. I didn't write that particular line, but I did expand that out anyway. Should be okay now, then.

src/api/java/com/minecolonies/api/compatibility/resourcefulbees/ResourcefulBeesCompat.java
```diff
@@ -0,0 +1,37 @@
...
+            {
+                //Take out all combs from hive and return them to the beekeeper
+                list.add(hive.getResourceHoneycomb());
+            }
+
+            return list;
+        }
+        else
+        {
+            list.add(new ItemStack(Items.HONEYCOMB, amount));
```
Raycoms: super call?

src/api/java/com/minecolonies/api/compatibility/resourcefulbees/NoBeeCompat.java
```diff
@@ -0,0 +1,19 @@
...
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class NoBeeCompat implements IBeehiveCompat {
```
Raycoms: why do we need this one, can't we use the interface impl just as well as the default?

truenachtara: You can't instantiate an interface.

src/api/java/com/minecolonies/api/compatibility/resourcefulbees/ResourcefulBeesCompat.java
truenachtara: what about a super call?

Raycoms: instead of list.add, return super.method

src/api/java/com/minecolonies/api/compatibility/resourcefulbees/NoBeeCompat.java
Raycoms: you should be able to do a new IBeehiveCompat { }

truenachtara: IBeehiveCompat is an interface, which is abstract. As such it's not allowed to contain a constructor and also is not allowed to be instantiated.

Technically the method in IBeehiveCompat should also be completely blank

Raycoms: The code that was there previously worked, you can make an anonymous instantiation of the class the way I said it.

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -838,4 +867,19 @@ private void discoverFreeBlocksAndPos()
...
+     */
+    private void discoverModCompat()
+    {
+        if (ModList.get().isLoaded("resourcefulbees"))
+        {
+            Compatibility.beeHiveCompat = new ResourcefulBeesCompat();
+        }
+        else
+        {
+            Compatibility.beeHiveCompat = new IBeehiveCompat() {};
```
Raycoms: No need for the else, we already got it instantiated

src/api/java/com/minecolonies/api/compatibility/resourcefulbees/NoBeeCompat.java
Raycoms: class can be deleted

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -801,6 +791,45 @@ private void discoverEnchantments()
         }
     }
 
+    /**
+     * Calculate the crusher modes from the config file.
+     */
+    private void discoverCrusherModes()
```
Raycoms: Somethign went wrong in this class

truenachtara: Weird. I didn't touch that and it works fine in dev

src/api/java/com/minecolonies/api/compatibility/resourcefulbees/NoBeeCompat.java
truenachtara: indeed it can.

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
truenachtara: OH I FIGURED IT OUT

You guys did something major with this class recently, so this PR is now recommending, essentially, reversion to the previous file. I'll try to update that bit later today so it's not recommending complete reversion

marchermans:
@truenachtara This is not compiling: https://buildsystem.ldtteam.com/buildConfiguration/LetSDevTogether_Minecolonies_PullRequests_2_BuildAndTest/73712?showLog=73712_368_368.432.472

truenachtara:
Problem is ModList isn't imported. I'm at work right now though so I can't fix it right away, but let me try something

PR #6635, opened 2021-03-01
Add a flag to mark colonies as abandoned allowing them to be destroyed.

ToMe25:
Closes #
Closes #
Closes #

- Make colonies destructible once they have no owner and no officer anymore.
- This is done by allowing all permissions except the following to everyone if the colony is fully abandoned:
    - GUARDS_ATTACK
    - CAN_PROMOTE
    - CAN_DEMOTE
    - EDIT_PERMISSIONS
    - MANAGE_HUTS
    - TELEPORT_TO_COLONY
    - EXPLODE
    - CAN_KEEP_COLONY_ACTIVE_WHILE_AWAY
    - RALLY_GUARDS
- This will automatically be undone should the colony somehow get an officer or owner again.
- I think this will automatically be detected for old abandoned colonies as well, but wasn't really able to test this fully.

Review please


src/main/java/com/minecolonies/coremod/colony/permissions/Permissions.java
```diff
@@ -766,6 +839,12 @@ public boolean removePlayer(final UUID id)
         if (player != null && player.getRank() != Rank.OWNER && players.remove(id) != null)
         {
             markDirty();
+
+            if (getOwnerName().equals("[abandoned]") && getPlayersByRank(Rank.OFFICER).isEmpty())
```
Raycoms: maybe this if here could be in general a method we call in all the method

ToMe25: I mean its very short and only called three times, but ok i will move it to its own method.

someaddons (approved):

PR #6645, opened 2021-03-02
Further Research Datagen-Related Fixes

### LINKED ISSUE #6656: Research-related log spam
### Prerequisites 

### Context 

- Minecraft Version: forge-1.16.4-35.1.4-installer.jar
- MineColonies Version: minecolonies-0.13.670-ALPHA-universal.jar
- Structurize Version: structurize-0.13.146-ALPHA-universal.jar
 1.741.346 furniture-7.0.0-pre19-1.16.3.jar
 8.109.657 ImmersiveEngineering-1.16.4-4.1.1-128.jar
 663.601 mcw-doors-1.0.1fix-mc1.16.4.jar
 4.309.452 worldedit-forge-mc1.16.3-7.2.2-dist.jar

1. start the colonie
2. 
3. 

Hi there

a few releases ago, you have changed something with the research System
since then, I see a huge amount of entries in my latest.log and debug.log
this is slowing down the performance within the colonie
my colonie was already full researched, but with the change, I had to do all the research again, they were all lost

Could you please have a look into my log´s, maybe you can see something about the reasons.

Today I started the colonie at [04Mär2021 10:31:10.570] and stopped at [04Mär2021 12:56:24.796]
the latest.log is right now  "04.03.2021  12:56   420.150.146 latest.log" big.

sorry i could not use gist for the logs, gist is crashing with that logfile size

### Logs

- latest.log: 
- crashlog: 

### Notes

---
## Viewers

gattsuru:
Closes #6656.

- Sets up Research Datagen to go into the `generated/data` directories, rather than `generated/assets`.
- Removes the now-irrelevant `generated/assets/researches` and `resources/data/researches` directories, so there's only one authoritative directory at `generated/data/researches`.  Gradle automatically scoops this into the jar in the right location, but can move this elsewhere if desired.
- Changes translation file generation in the Research Datagen to add directly to the en_us.lang.  The current approach mimics how Structurize does it, and should always transfer properly, but can strip out extra paragraph breaks or impact ordering.
- Adds some safeties for the home/citizen problem on the datagen side.
- Adds a translatedName for Taunt research.
- Fixes some edge cases with set names on research effects.
- Adds support for the `mandatory-building` requirement tag, which makes research require a single building at a specific level.  For now, this is only applied to the PlateArmor research, as Merkle pointed out that it would otherwise be possible to complete the research but not see any unlocked recipes from it.
- Fixes bug on Level 6 research unlocks.
- Fixes unrelated plantation bug.

Review please


src/main/java/com/minecolonies/coremod/generation/defaults/DefaultResearchProvider.java
```diff
@@ -163,6 +168,22 @@ public DefaultResearchProvider(final DataGenerator generator)
...
     }
 
+    @Override
+    public JsonElement getBaseLanguageJson()
+    {
+        JsonElement langJson;
+        try
+        {
+            langJson =
+              new JsonParser().parse(new FileReader(this.generator.getInputFolders().stream().findFirst().orElse(null).resolve("assets/minecolonies/lang/en_us.json").toFile()));
```
uecasm: Prefer using `event.getExistingFileHelper()` to get a helper class to read existing files by ResourceLocation rather than by direct path.

gattsuru: Is there a trick to doing so? It's giving errors about `@visiblefortesting` code in a production environment, ExistingFileHelper as a class is marked as Deprecated to be removed in 1.17, and when called in runData it doesn't seem like any of the minecolonies namespace is loaded into the resource manager.

uecasm: The class isn't deprecated, only one of the constructors is.  And you shouldn't be calling any of the constructors, you fetch the one that already exists inside the `GatherDataEvent`.  (Hence why I used an `event.` prefix.)

gattsuru: Ok.  And `gatherDataEvent.getExistingFileHelper().getResource()` works for things in the minecraft, realms, or forge namespace, but that's all that's loaded in this context: is there something that needs be done to have the mod namespace loaded in an instance of `GatherDataEvent`?

uecasm: In the data run command in build.gradle, swap `--input` to `--existing`.  Optionally, also add this at the end to make the Structurize resources visible too:
```
, '--existing-mod', 'structurize'
```

Raycoms:
:shipit:

marchermans (dismissed):

build.gradle
```diff
@@ -179,7 +179,7 @@ minecraft {
 
             property 'forge.logging.console.level', 'info'
 
-            args '--mod', 'minecolonies', '--all', '--output', file('src/main/generated/resources'), '--input', file('src/main/resources')
+            args '--mod', 'minecolonies', '--all', '--output', file('src/main/generated/resources'), '--input', file('src/main/resources'), '--existing', file('src/main/resources')
```
uecasm, line 185: I did say you could just replace `--input` with `--existing`; you don't need to duplicate the parameter and path.  But this is ok too.

src/api/java/com/minecolonies/api/research/AbstractResearchProvider.java
```diff
@@ -26,16 +27,16 @@
...
+    public static final String ASSETS_DIR = "data/" + Constants.MOD_ID + "/researches";
+    protected final GatherDataEvent event;
 
     /**
      * The abstract variant of a ResearchProvider, to register to fires during runData.
-     * @param generator  The DataGenerator to store the files.
+     * @param event  The GatherDataEvent to store the files.
      */
-    public AbstractResearchProvider(final DataGenerator generator)
+    public AbstractResearchProvider(final GatherDataEvent event)
```
uecasm: It's bad style to pass the event into the provider.  Pass the `DataGenerator` and `ExistingFileHelper` as separate parameters instead.

src/api/java/com/minecolonies/api/research/AbstractResearchProvider.java
```diff
@@ -360,7 +383,54 @@ public Research addBuildingRequirement(final String buildingName, final int leve
                 reqArray = new JsonArray();
             }
             JsonObject req = new JsonObject();
-            req.addProperty("building", buildingName);
+            // Temporary workaround for the discrepancy between schematic ID and modBuildings ID.
```
uecasm: Consider extracting this into a helper function rather than repeating it in multiple places.

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1965,6 +2081,7 @@
   "com.minecolonies.research.combat.flee.name": "Flee",
   "com.minecolonies.research.combat.flee.subtitle": "Sometimes it's better just to run.",
   "com.minecolonies.research.combat.hotfoot.name": "Hotfoot",
+  "com.minecolonies.research.combat.taunt.name": "Taunt",
   "com.minecolonies.research.combat.taunt.subtitle": "Your mother was a hamster and your father smelt of elderberries!",
   "com.minecolonies.research.combat.arrowuse.name": "Consume Arrows",
   "com.minecolonies.research.combat.arrowuse.subtitle": "They work better with ammo.",
```
uecasm: The research strings should no longer appear in the source en_us.json when they are being added by datagen.  While this shouldn't cause an actual bug (provided that the datagen is smart enough to replace existing values rather than simply appending), it will cause confusion down the line when someone tries to edit the wrong string and is confused when their changes do not take effect.

Conversely, however, the generated en_us.json and other files that are the result of runData should also be included in this commit.

uecasm: Unless the change to this file is because you are making the datagen directly write to this file instead of generating a replacement in the generated directory instead.  In which case: don't do that.  Revert the changes to this file, then delete the research strings, then change the output directory of the datagen.

gattsuru: Yes, right now it's feeding back into the same file as the origin.  I understand it's possible to produce a separate file, but doing so will result in two possible sources-of-true, in violation of the DRY rule, and especially with how many people you've got working on this project, that's a lot of potential for incident or frustration.  Is there a particular benefit to your approach, or is there a reason that Structurize wasn't doing it this way?

The generated datafiles are in `generated/data`, and were committed.  There's an unavoidable duplication of truth between the datagen and the jsons, but I can't come up with a good workaround on that.

build.gradle
gattsuru: I didn't add the original `--input` argument, so I'm a little worried about making gradle modifications that remove anything unless absolutely necessary.

uecasm: I've tested that in another mod and it works perfectly fine.  But as I said, while this is unnecessary it's not harmful, so it's ok to stay like this if you prefer.

src/main/resources/assets/minecolonies/lang/en_us.json
uecasm: As discussed on Discord, putting generated data into a non-generated folder is worse, because people will see "oh, here's a file in the hand-editing directory, let me change it there" and then their changes mysteriously vanish because they happened to be values that the datagen replaced.  With the way I suggested, at least there's the context that everything in the "generated" folder will get replaced on rundata, so you should never hand-edit them.  It's more consistent.

src/main/generated/resources/assets/minecolonies/lang/default.json
```diff
@@ -88,7 +88,7 @@
   "com.minecolonies.research.civilian.diligent.name": "Diligent",
   "com.minecolonies.research.civilian.studious.name": "Studious",
   "com.minecolonies.research.civilian.scholarly.name": "Scholarly",
-  "com.minecolonies.research.civilian.scholarly.subtitle": "Homework for the next decade... check!",
+  "com.minecolonies.research.civilian.scholarly.subtitle": "Homework for the next decade...check!",
```
uecasm: I think this looks better with the space, although it's not super important.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultResearchProvider.java
```diff
@@ -28,9 +36,9 @@
  */
 public class DefaultResearchProvider extends AbstractResearchProvider
 {
-    public DefaultResearchProvider(final DataGenerator generator)
+    public DefaultResearchProvider(final DataGenerator event)
```
uecasm: Wrong parameter name?

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -2078,7 +1764,6 @@
   "com.minecolonies.coremod.gui.workerhuts.school": "School",
   "com.minecolonies.coremod.gui.workerhuts.teacher": "Teacher",
   "com.minecolonies.coremod.gui.workerhuts.pupil": "Pupil",
-
```
uecasm: Now that this file isn't being stomped by datagen, it'd be nice to keep these line breaks (here and line 1757).

marchermans (approved):

src/main/generated/resources/assets/minecolonies/lang/default.json
ravenbuilder934: I have the opposite opinion (and actually removed the space myself in a recent, merged PR), but oh well

PR #6659, opened 2021-03-04
Disable chests, improve tool keeping

Raycoms:
- Disable Chests being recognized as containers
- Improve tool keeping to keep a better tool always too
-

Review please


Mekle001 (dismissed):

Raycoms:
I'll add some code to unregister chests if they are in the list too. Potentially on Courier pickup, so that they are emptied before unregistering them.

someaddons (approved):

PR #6674, opened 2021-03-06
Draft

someaddons:
An alternate suggestion to modules, splitting it by semantical modules instead of by implementation.


Raycoms:
As discussed, will close this here for now. We'll revisit the attempt when we got the system a bit further fleshed out to see if it is an improvement to the overall system.

Issue #6657, opened 2021-03-04
Bunch of issues I've met during playing


Nightenom:
List of random nitpicks/issues to fix:
_If you are not from dev team and want to comment please create a new issue and describe it properly_

- [x] 1) negative and weird percentage during requirements gathering phase, #6726
![image](https://user-images.githubusercontent.com/17338378/109997828-e6cbd980-7d10-11eb-8aa0-acf7fdaa4d84.png)
- [x] 2) ~lumberjack spinning when idling (when having no work)~ overridden by 9, #6678
- [x] 3) lumberjack restriction area has either bad chat info or does not work properly - not sure if it's 3d or 2d, if 3d then bad chat info, if 2d then he does not detect trees correctly - should be probably 2d from level y, #6678
- [x] 4) chest are still considered as storage but citizens do not consider them when searching items
- [x] 5) lumberjack should have separate toggles for planting and chopping - requires small gui redesign (box should have two lines probably: itemicon (left-aligned), first line sapling name, second line two toggles - chop and plant
- [x] 6) citizen should have stronger ai task for doors of citizen huts (houses), the current one is lenient cause of chunk-loading, special handling or stronger task should be used so doors are always closed during night
#### edit 1:
- [x] 7) remove alpha from tool textures, #6726
![image](https://user-images.githubusercontent.com/17338378/110027671-91a0bf80-7d32-11eb-9997-28c441afa54f.png)
- [ ] 8) minecolonies build tool requires having hut block to display its schematic, this is rather problematic when using preview sharing since other players are required to have the block too otherwise their build tool resets the schematic to decos (to raven - I consider this to be a bug)
#### edit2:
- [x] 9) (overrides 2) lumberjack looping in no_tree_found and start_working stages when no tree found, #6678
a) searchIncrement is never resetted
b) [this code](https://github.com/ldtteam/minecolonies/blob/version/1.16.3/src/main/java/com/minecolonies/coremod/entity/ai/citizen/lumberjack/EntityAIWorkLumberjack.java#L335) is problematic in setTree because the setTree doesn't take restriction in account thus lumber just spins between restriction area and building
c) waitBeforeCheckingAgain is not working as intended, the AI never pauses if no_tree_found state is hit
[...]

PR #6675, opened 2021-03-06
Fix colony border rendering

Nightenom:
Progress towards #6657

- fixes y shifting and color bugs
image

Review please

### DIFF
--- a/src/main/resources/META-INF/accesstransformer.cfg
+++ b/src/main/resources/META-INF/accesstransformer.cfg
@@ -53,3 +53,6 @@ public net.minecraft.client.multiplayer.ClientAdvancementManager field_192803_d 
 public net.minecraft.block.DoorBlock func_196426_b(Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;Z)V # DoorBlock playsound
 
 public net.minecraft.entity.Entity field_213326_aJ # eyeHeight
+
+# vbo draw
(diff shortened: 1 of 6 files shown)


someaddons:
just gave it a try, it doesnt seem to render anything for me for some reason: 
image

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -52,4 +52,7 @@ public net.minecraft.client.multiplayer.ClientAdvancementManager field_192803_d
 
 public net.minecraft.block.DoorBlock func_196426_b(Lnet/minecraft/world/World;Lnet/minecraft/util/math/BlockPos;Z)V # DoorBlock playsound
 
-public net.minecraft.entity.Entity field_213326_aJ # eyeHeight
\ No newline at end of file
+public net.minecraft.entity.Entity field_213326_aJ # eyeHeight
+
+# vbo draw
```
marchermans, line 57: Why would you need this at here?

marchermans (commented):
Generally oke. The rendering seems fishy to me with the at?

src/main/resources/META-INF/accesstransformer.cfg
Nightenom: I could probably just copy that function but I think using the vanilla is safer cuz of vanilla change
usage: uploading vbo repeatedly without using bufferbuilder every frame

someaddons (approved):
works now 👍

PR #6691, opened 2021-03-08
Move composting to vanilla-style datapack recipe and include JEI lookup.

uecasm:
Bite-sized initial part split off from #6540 

- Move compost recipes to datapack rather than a mix of tag and code
    - Retains existing division in "strength" of certain items, makes it easier to customise for particular items if desired.
    - Adds a third strength tier (better than the previous ones, only needs 8 items to fill) with almost nothing in it by default (but not actually nothing, because that annoys Forge)
    - Splits food items between all three strength tiers based on their saturation
- Only treat items as food if they have at least some healing+saturation (should prevent cook glitching out on some modded food)
- JEI lookup support for compost barrel (and by extension, composter hut)

Review please


uecasm:
If you absolutely hate the use of the vanilla RecipeManager, I'm open to restructuring things to avoid it.  But it does seem sensible to leverage standard Minecraft things where possible.

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
```diff
@@ -0,0 +1,158 @@
...
+    @Override
+    public IRecipeType<?> getType() { return TYPE; }
+
+    @NotNull
+    @Override
+    public ResourceLocation getId() { return this.id; }
+
+    public Ingredient getInput() { return this.input; }
+
+    public int getStrength() { return this.strength; }
```
Raycoms, line 72: At least some doc on the public getters would be nice

Raycoms (dismissed):
Okay, this was easy to review now. Just missing some javadoc here and there.

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1532,6 +1532,9 @@
   "com.minecolonies.coremod.info.barracks.2.name": "FAQ",
   "com.minecolonies.coremod.info.barracks.2": "Now let's cover some frequently asked questions:\n\n Q 1. Can I have more than 1 Barracks?\n   A 1. Yes, but the Barracks Tower levels are based on the Barracks they are attached to. So even if you have one Barracks that is level 5, if the second Barracks is only level 1, then that Barracks can only have a single level 1 tower.",
 
+  "com.minecolonies.coremod.jei.composter": "Uses compost barrels to turn various items into compost or dirt (look up usages for the compost barrel to see more).  At higher skill levels, they can work faster than you can.",
```
```suggestion
  "com.minecolonies.coremod.jei.composter": "Uses compost barrels to turn various items into compost or dirt (look up usages for the compost barrel to see more). At higher skill levels, they can work faster than you can.",
```
don't need two lines below it either, one is fine

ravenbuilder934 (commented):
Nitpick

Mekle001 (dismissed):

marchermans (changes requested):
This needs to be reworked.

Since this is now an IRecipe with a related DataPack link this needs to use an IResourceReloadListener to listen for changes to datapacks, and should just be handled via the vanilla recipe manager in the first place, instead of via the CompatibilityManager

uecasm:
But it is handled via the vanilla RecipeManager.

Raycoms:
> But it is handled via the vanilla RecipeManager.

Yes, it's handled via the RecipeManager, but not reloaded on datapack change, right?

uecasm:
I assumed that the CompatibilityManager reloaded on datapack change.  If it doesn't, that's a bug in that, since most of what it does is tag-based.

Raycoms:
> I assumed that the CompatibilityManager reloaded on datapack change. If it doesn't, that's a bug in that, since most of what it does is tag-based.

No, it doesn't reload at the moment.

Raycoms:
> > I assumed that the CompatibilityManager reloaded on datapack change. If it doesn't, that's a bug in that, since most of what it does is tag-based.
> 
> No, it doesn't reload at the moment.

Also not sure if that would work out fine.  Most of it comes from the oredict time pre reloading tags mid game.

Raycoms:
:shipit:

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
```diff
@@ -0,0 +1,182 @@
...
+
+    public static class Serializer extends ForgeRegistryEntry<IRecipeSerializer<?>>
+            implements IRecipeSerializer<CompostRecipe>
+    {
+        public static final Serializer INSTANCE = new Serializer();
+
+        static
+        {
+            INSTANCE.setRegistryName(CompostRecipe.ID);
+        }
```
marchermans: Please use a getter here instead of a direct field `getInstance()` is better here.

src/api/java/com/minecolonies/api/crafting/CountedIngredient.java
```diff
@@ -0,0 +1,104 @@
...
+    @NotNull
+    @Override
+    public IIngredientSerializer<? extends Ingredient> getSerializer()
+    {
+        return Serializer.INSTANCE;
+    }
+
+    public static class Serializer implements IIngredientSerializer<CountedIngredient>
+    {
+        public static final Serializer INSTANCE = new Serializer();
```
marchermans: Again use a getter here for the singleton.
Additionally this is missing the private constructor.

src/main/java/com/minecolonies/coremod/recipes/FoodIngredient.java
```diff
@@ -0,0 +1,167 @@
...
+                    maxHealing.map(healing -> food.getHealing() < healing).orElse(true) &&
+                    minSaturation.map(saturation -> food.getSaturation() >= saturation).orElse(true) &&
+                    maxSaturation.map(saturation -> food.getSaturation() < saturation).orElse(true);
+        }
+    }
+
+    public static class Serializer implements IIngredientSerializer<FoodIngredient>
+    {
+        public static final Serializer INSTANCE = new Serializer();
+
```
marchermans: Again use a getter here instead of a final field.
Additionally this is also missing the private constructor.

src/main/java/com/minecolonies/coremod/recipes/PlantIngredient.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public IIngredientSerializer<? extends Ingredient> getSerializer()
+    {
+        return Serializer.INSTANCE;
+    }
+
+    public static class Serializer implements IIngredientSerializer<PlantIngredient>
+    {
+        public static final Serializer INSTANCE = new Serializer();
+
```
marchermans: See other cases like this.

marchermans (changes requested):
Aside from some codestyle issues, this is completely missing the datagen part of the changes made here.

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
uecasm: This is the preferred style recommended by both Minecraft and Forge for recipe serializers.

Using a final field is also safer than using a `getInstance()` method, so that would reduce code quality.

src/api/java/com/minecolonies/api/crafting/CountedIngredient.java
uecasm: Again, this is the standard style.  And it is stateless so there is no harm in allowing public construction; the singleton is for convenience rather than required for correctness.

ravenbuilder934:
Conflicts need to be fixed

Raycoms:
Just missing the conflicts and :shipit:

marchermans:
Still missing datagen?!? I see no commit since I rejected this. So no. Not ready to merge!

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
marchermans: It might be the standard style forge and minecraft use. But not minecolonies. Static final fields are considered to be harder to unit test then static methods due to compiler inlineing.

uecasm: Another irrelevant argument.  There's no need to mock anything out, it can be tested in place without complication.  And the MineColonies source has no unit tests anyway.

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -492,6 +477,16 @@ public boolean isFreePos(final BlockPos block)
...
+    @Override
+    public void invalidateRecipes(@NotNull final RecipeManager recipeManager)
+    {
+        // TODO: this should probably also invalidate anything using tags too (which is most things)
+        //       although technically there's a different event for tag reloads
+
+        compostRecipes.clear();
+        discoverCompostRecipes(recipeManager);
+    }
+
```
uecasm, line 492: This is called automatically on datapack reload to regenerate the cache, as I already described in the Discord.

uecasm:
I usually rebase changes to avoid an untidy history, and because it produces cleaner diffs client-side, but since that's apparently harder to follow with the way that GitHub presents diffs, I'll stop doing it.  Rest assured, as shown above, datapack reloading has been implemented.

Raycoms:
> I usually rebase changes to avoid an untidy history, and because it produces cleaner diffs client-side, but since that's apparently harder to follow with the way that GitHub presents diffs, I'll stop doing it. Rest assured, as shown above, datapack reloading has been implemented.

I think he meant datagens for the existing json file

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
Raycoms: Translating this here: MineColonies tries to maintain a consistent style which might not be the default in Minecraft or Forge (similarly as our bracket style). So it makes sense to try to have the new code being added follow our default style as well.

uecasm:
> I think he meant datagens for the existing json file

That was not previously requested.  I could add it but I don't see any current need for that.  Datagens are useful for mass generation but these are only ever going to be a small number of hand-edited recipes, since they're mostly tag-based anyway.

Raycoms:
> > I think he meant datagens for the existing json file
> 
> That was not previously requested. I could add it but I don't see any current need for that. Datagens are useful for mass generation but these are only ever going to be a small number of hand-edited recipes, since they're mostly tag-based anyway.

Looking into the jsons, they seem difficult to datagen properly too, since they're mostly mixes of tags/items

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
uecasm: Style consistency is good (which is why I have tried to make all my code follow the brace style used elsewhere, despite it being unconventional for Java).  But rigorous adherence to a style checklist that contravenes good code design is not good.

There are some advantages to `getInstance()` -- the primary one is lazy instantiation, which is quite valuable in many cases, especially in a modloader context -- but not this case, since it's a stateless utility class, and it does impose a needless cost.  Unit testing is not a valid argument because either way it's static and can't be mocked out anyway (not that there is any need to) and because there aren't any.  Otherwise, a static final field is safer because it guarantees that the instance can't be tampered with (outside of extreme cases), and is automatically thread-safe (neither of which apply to getInstance).

There is technically an even better way to implement singletons in Java (via enums) but it's more obscure and the potential confusion didn't seem worth it since it wasn't used elsewhere in the codebase already.

uecasm:
> Looking into the jsons, they seem difficult to datagen properly too, since they're mostly mixes of tags/items

That's not really a problem.  It'd either just mean using a CompoundIngredient in the generator (that's what an array of ingredients is) or structuring the builder such that it does so internally.  Or (more likely) making it generate multiple separate recipes instead of using compound ingredients -- they were mostly a convenience to avoid needing many similar json files, which is less of an issue with generated files.

So if people prefer that approach, I am happy to do it instead.  I just don't think it adds anything particularly useful, though.

src/api/java/com/minecolonies/api/crafting/CompostRecipe.java
Raycoms: Through Enums is definitelty not worth it. Since the access seems solely within class (correct me if I'm wrong) I see no problem of having it potentially private and if we ever need to access it from the outside we create the getter.

uecasm: Almost, but not quite.  Registration of the serializer is the external usage, which happens over in CommonProxy.  It is technically possible to make that entirely private by having multiple internal subscribers to the mod event instead, but that's also a bit dodgy.

uecasm: For lazy instantiation you can use a static final `Lazy<T>` field, which gives the best of both worlds.

Having said that, there is another legitimate use of getInstance -- which is how it is currently used throughout much of existing MineColonies -- that of allowing abstraction of the storage of the instance, for example by storing it inside another singleton instead of directly (e.g. how most stuff ends up indirected through the API instance).  That is a good design -- but it's also one that's not applicable here.

uecasm:
Bump.

Raycoms (dismissed):

marchermans (changes requested):
Still not following the **_MINECOLONIES_** code style.
This will not get merged untill it does.

We do not care how Minecraft or Forge formats stuff, we care about how we format stuff, and untill PRs follow our guidelines, including for singleton accessors, PRs do not get merged.

uecasm:
I'm not going to make the code objectively worse (non-final field, pointless conditional branch, lack of thread safety) to satisfy an imaginary style checklist tick that is completely irrelevant for this use case.  I have standards.

This instance is **never** going to be accessed outside of the two places (one internal, one in mod registry) that currently access it.

Raycoms:
> I'm not going to make the code objectively worse (non-final field, pointless conditional branch, lack of thread safety) to satisfy an imaginary style checklist tick that is completely irrelevant for this use case. I have standards.
> 
> This instance is **never** going to be accessed outside of the two places (one internal, one in mod registry) that currently access it.

I mean, it's basically only a private constructor to hide the public implicit one, and a private static final field with the constructor with a static getter on the class.

It will still be final and the code stays 99% the same, the only thing is that the field is final and gets accessed through a public getInstance method

uecasm:
Well, ok, _that_'s less efficient but not actually offensive.  But that didn't sound like what was being requested.  I still think it's pointless but fine, I can make that change.

Raycoms:
> Well, ok, _that_'s less efficient but not actually offensive. But that didn't sound like what was being requested. I still think it's pointless but fine, I can make that change.

I checked and that's how all the other getInstances are handled in our mod

uecasm:
No, most of them are non-final lazy-branched.  Which is what I was objecting to for this case.  But whatever.

src/api/java/com/minecolonies/api/crafting/CountedIngredient.java
```diff
@@ -69,12 +69,16 @@ public CountedIngredient(@NotNull final Ingredient child, final int count)
...
     public IIngredientSerializer<? extends Ingredient> getSerializer()
     {
-        return Serializer.INSTANCE;
+        return Serializer.getInstance();
     }
 
     public static class Serializer implements IIngredientSerializer<CountedIngredient>
     {
-        public static final Serializer INSTANCE = new Serializer();
+        private static final Serializer INSTANCE = new Serializer();
```
marchermans, line 77: Yes this is what I want to see, and what I requested.

If there are classes you found which do not follow this pattern, send them to me.
Allthough I would find this very weird, since i have this pattern as a FileTemplate in my IDE, and it would be weird if as such Singletons with other structures were created, but hey, I might have missed them during reviews.

marchermans (approved):

src/api/java/com/minecolonies/api/crafting/CountedIngredient.java
uecasm: `FactoryVoidInput` is the only one that does it with a public `INSTANCE` field, at least.  There may be other cases with different names.  But there are many other places where the `getInstance()` refers to a non-final field.

PR #6692, opened 2021-03-08
Sort items by activeness making it easier to see selected items

RyTechro:
- Sorting filterable items by active items. In practice this will for instance sort the fuel page of the restaurant or the baker to show the enabled items at the top, this gives a quick overview of what items are active.

New pull request to fix broken local config

Review please


Raycoms (approved):

Raycoms:
Does this work nicely with enabling and disabling too? (Does it also persist when relogging)

RyTechro:
> Does this work nicely with enabling and disabling too? (Does it also persist when relogging)

Yes, been playing with it today and found no issues. Persistent with relogging and does only resort the items onOpen of the gui so that the key is correct even if you search and enable multiple at once

someaddons (approved):

PR #6693, opened 2021-03-08
Fix Research "Open To LAN" Support

gattsuru:
- Send packets to IntegratedServer clients, not just DedicatedServer ones, so long as the IntegratedServer has been opened to other users.

I don't like using port number to check IntegratedServer mode, but as far as I can tell, all the seemingly-related stuff like DedicatedServer.onlineMode values are more related to authentication servers.

Review please


Nightenom:
Why do you use such a weird structure at all?
Sending network messages in singleplayer is completely fine

src/main/java/com/minecolonies/coremod/datalistener/ResearchListener.java
```diff
@@ -92,7 +93,7 @@ protected void apply(@NotNull final Map<ResourceLocation, JsonElement> object, @
         // For dedicated servers, send to any connected players.  On startup, this will be no-one.
         // But it is possible to reload data packs live, and while not supported it's something to handle.
 
-        if(server instanceof DedicatedServer)
+        if(server instanceof DedicatedServer || (server instanceof IntegratedServer && server.getServerPort() > 0))
```
Raycoms: Yeah, I think just doing a !world.isRemote check here should be just fine imo

marchermans (changes requested):
Use world.isRemote

src/main/java/com/minecolonies/coremod/datalistener/ResearchListener.java
gattsuru: I'm not sure if there's a good world available.  The EventHandler case has event.getPlayer().world.isRemote(), but it's already already checked if the event.getPlayer() is a ServerPlayerEntity and returns false anytime the underlying event is fired (since PlayerLoggedInEvent fires from a server initializing a connection to a player).  For ResearchListener, doing something like server.getWorlds() seems like it's not correct, either.

In practice, I think these events only fire on the server side, so if there's no concerns about extraneous network messaging or performance given that they only happen once or twice a play session, can just simplify the checks.

Nightenom: Do not check for sides unless you have to split calls from client and server because of data/logic, network is completely side safe

Raycoms: Oh wait, this is on the server always, if you have the server object, you must be on the server, so a null check should be enough?

gattsuru: Yeah, from a not-crashing perspective the null check's all you need for ResearchListener.  It's a bit of extra RAM and CPU for single-player instances, since they end up rebuilding the tree twice, but that's probably premature optimization.

I've got a few concerns about order-of-actions logic stuff, but the worst case is seeing autostart+instant research announcements triggering twice on server startup and even that may not be possible, so not worth the ugliness.

src/main/java/com/minecolonies/coremod/event/EventHandler.java
```diff
@@ -420,11 +421,7 @@ public static void onPlayerEnterWorld(final PlayerEvent.PlayerLoggedInEvent even
                 }
             }
 
-            MinecraftServer server = ServerLifecycleHooks.getCurrentServer();
-            if(server instanceof DedicatedServer)
```
Raycoms, line 424: If this here was a problem, I'm quite sure you can do player.world.isRemote

gattsuru: Even on dedicated servers (where all players are remote), that always returns false.  Apparently intentionally, despite what one might expect.

It's not an issue, just philosophically bugged me.

Raycoms: No, that's not true, on dedicated servers, there is a "ServerPlayerEntity" and a "ClientPlayerEntity", so depending on the site it's exactly what you want

marchermans:
I checked you code. You can use world.IsRemote 100% of the time. There is no reason todo these weird checks here.

src/main/java/com/minecolonies/coremod/event/EventHandler.java
Raycoms: If it returns always false in this case, this means this code is only called on the server side (still makes sense to do a !player.world.isRemote check)

gattsuru: This is already wrapped in `if (event.getPlayer() instanceof ServerPlayerEntity)`.  I can add the additional check if you want it for code cleanliness purposes.

Raycoms: Then it's not needed at all

gattsuru: Aye.  So it's gone now.
The point was to try and avoid sending the unnecessary message for situations where the client and server shared JVM, rather than sidedness, but if that's not necessary or premature optimization, there's no reason to bother.

Mekle001 (approved):

PR #6687, opened 2021-03-08
Make Sifter use real items as meshes

Mekle001:
# Changes proposed in this pull request:
- Create new "mesh" items with durability for the sifter to use, updating and adding recipes to support it
- Fix tool damage to work when the tool isn't in the citizen's inventory. 
- Make custom recipes add at the top of the list, so later levels take precedence
- Make the prestock code ignore damage, to not request new meshes as soon as the current one is damaged
- Add Achievements tree to University, for hidden/autostart/instant researches to unlock things like recipes
- Wire up meshes to block on new "Achievements" in the university

The meshes are made by other crafters as such: 
- String Mesh at the Fletcher with 500 durability
- Flint Mesh at the Stonemason with 1000 durability
- Iron Mesh at the Blacksmith with 1500 durability
- Diamond Mesh at the Mechanic with 2000 durability

I still need to remove unnecessary UI elements

PR #6719, opened 2021-03-11
Simplify crushing.

uecasm:
Another bite-size piece extracted from the JEI PR.  This simplifies the crusher mode calculation, since it already has all the data in the crafterrecipes anyway.  (Possibly as a future extension this could get a similar treatment as #6687 and just craft automatically without mode selection, although that has some downsides too.)

- Calculates crusher modes automatically from datapack recipes instead of config.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCrusher.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCrusher.java
@@ -1,7 +1,7 @@
 package com.minecolonies.coremod.colony.buildings.workerbuildings;
 
+import com.google.common.collect.ImmutableMap;
 import com.ldtteam.blockout.views.Window;
-import com.minecolonies.api.blocks.ModBlocks;
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.IColony;
 import com.minecolonies.api.colony.IColonyManager;
@@ -9,17 +9,15 @@ import com.minecolonies.api.colony.IColonyView;
 import com.minecolonies.api.colony.buildings.ModBuildings;
 import com.minecolonies.api.colony.buildings.registry.BuildingEntry;
 import com.minecolonies.api.colony.jobs.IJob;
-import com.minecolonies.api.colony.requestsystem.StandardFactoryController;
+import com.minecolonies.api.colony.requestsystem.token.IToken;
 import com.minecolonies.api.crafting.IRecipeStorage;
 import com.minecolonies.api.crafting.ItemStorage;
 import com.minecolonies.api.entity.citizen.Skill;
-import com.minecolonies.api.util.constant.TypeConstants;
 import com.minecolonies.coremod.Network;
 import com.minecolonies.coremod.client.gui.WindowHutCrusher;
 import com.minecolonies.coremod.colony.buildings.AbstractBuildingCrafter;
 import com.minecolonies.coremod.colony.jobs.JobCrusher;
 import com.minecolonies.coremod.network.messages.server.colony.building.crusher.CrusherSetModeMessage;
-
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.network.PacketBuffer;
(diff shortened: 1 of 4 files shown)


uecasm:
Bonus: there was a clay block to clay balls custom recipe already, which it didn't look like the crusher AI could ever actually do, since it wasn't in the configured modes (unless I missed something).  Now it automatically appears in the mode selector like everything else.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCrusher.java
```diff
@@ -94,25 +95,23 @@ public BuildingCrusher(final IColony c, final BlockPos l)
      */
     private void loadCrusherMode()
     {
+        this.recipes.clear();
+        checkForWorkerSpecificRecipes();
```
Raycoms, line 99: does this here load the recipes into "this.recipes" ?

ravenbuilder934:
> Bonus: there was a clay block to clay balls custom recipe already, which it didn't look like the crusher AI could ever actually do, since it wasn't in the configured modes (unless I missed something). Now it automatically appears in the mode selector like everything else.

They could do it on request, which iirc is intentional

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCrusher.java
Mekle001: This concerns me, as this.recipes is a persisted collection, with a lot of behavioral semantics around it. While I think it would work for now, I'd rather see us just use what is in this.recipes.  What is the bad behavior this clear is fixing?

Mekle001 (commented):
Most of this makes sense, and is in alignment with some of the discussed roadmaps for the crusher. I'm still going through with a fine tooth comb, as I want to be sure that the crusher's dual-mode nature is retained. (one mode where it crushes infinitely, and the mode where it crushes on demand based on RS asks). To be frank, I personally never run the crusher in infinite crushing mode anymore, as it's hard to balance the different output, and using warehouse prestocks, you can accomplish the same thing in a much more manageable way. 

I'd actually argue there may be value in removing the infinite crushing mode all together.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCrusher.java
Raycoms: This is cleared because datapacks can be reloaded for the crusher I believe

Mekle001: We already handle datapack changes in the recipe system. checkForWorkerSpecificRecipes does full maintenance, and runs on colony tick.

uecasm: Yes, it reloads this.recipes.  I'm not entirely sure why, but without this specific reload I found hundreds of repeats of the same recipes stored in there (as well as garbage recipes that returned null after trying to resolve the token).  This fixes all of that.

It does have the downside of generating an essentially random order of the recipes and not remembering any order that the player has set, but at least as long as the crafting is mode-based this doesn't matter.

uecasm:
> I'd actually argue there may be value in removing the infinite crushing mode all together.

Yes, that's what I meant by "craft automatically without mode selection".  But this PR doesn't do that, it retains the mode selection; the only behavioural changes are to reset the recipe list (since it went very weird) and to show the clay blocks -> balls in the mode selection too.

Raycoms:
> > I'd actually argue there may be value in removing the infinite crushing mode all together.
> 
> Yes, that's what I meant by "craft automatically without mode selection". But this PR doesn't do that, it retains the mode selection; the only behavioural changes are to reset the recipe list (since it went very weird) and to show the clay blocks -> balls in the mode selection too.

I think we can remove the modes in this PR even, we now have minimum stock in the warehouse to ask for those kind of things, we don't need infinite crushing anymore.

uecasm:
Someone could do that in a follow-up or alternative PR maybe, but that sounds like a significant change to the AI behaviour that I'm not sufficiently confident to make myself.

(Although, FWIW, it looks like on-demand crafting bypasses the daily limits, which might be a balance problem?  Unless they're not intended for that, although as the limit increases with building level it seemed likely.)

Raycoms:
> Someone could do that in a follow-up or alternative PR maybe, but that sounds like a significant change to the AI behaviour that I'm not sufficiently confident to make myself.
> 
> (Although, FWIW, it looks like on-demand crafting bypasses the daily limits, which might be a balance problem? Unless they're not intended for that, although as the limit increases with building level it seemed likely.)

kk, no problem, yes on demand might surpass the limits, that was intended at the time. Now with minimum stocks this whole behaviour is not necessary anymore.

Mekle001 (approved):

PR #6722, opened 2021-03-11
Improve Research Branch Customizability

gattsuru:
- Changes the WindowUniversity display to use a ScrollingList, so to better support 8+ research branches.
- Adds support for the `branch-type`, `subtitle`, `sortOrder`, and `hidden` properties to branch-level jsons.
- `sortOrder` and `subtitle` are similar to their effects in research jsons. Branches with a higher `sortOrder` will be placed lower on the University Window, branches with a `subtitle` will have a tooltip on the selection list.
- `hidden` will prevent the display of a research branch if all of its primary researches are set to `hidden`, and none of those researches have their requirements fulfilled.  Branches where all primary researches are set to `hidden` but the branch itself is not will display a disabled ButtonImage with the text marked out and a tooltip listing the unlock options.
- `branch-type` allows wider styling or logical changes to a full branch.  The only currently supported types are `"default"` (current behavior) and `"unrestricted-tree"`, which is meant to support Mekle's use case where a branch needs University-specific capabilities not readily managed through vanilla advancements (either by having colony-wide effects or requirements, possibly occurring when players are offline, needing strictly enforced hierarchy, or wanting to require an item submission) by removing the time and university-level limitations.  Down the road, I could see potential uses for a couple stylings that use nested ScrollingList instead of a tree-style dragzoomview, and there are likely other more esoteric uses.
- Cleans up the related implementation logic to use Actual Objects instead of parallel hashmaps.
- Fixed "Heavily Loaded" not properly having the `immutable` property

Review please


src/api/java/com/minecolonies/api/research/ResearchBranchType.java
```diff
@@ -0,0 +1,43 @@
...
+     * or where requirement or reward is tied to a colony-wide behavior instead of a player-specific one.
+     */
+    ACHIEVEMENT("achievement");
+
+    /**
+     * The string tag representation of the Branch-Type.  Used both in logic, and in datapack JSON.
+     */
+    public final String tag;
+
+    private ResearchBranchType(String tag) {
```
Raycoms: formatting

src/api/java/com/minecolonies/api/research/ResearchBranchType.java
```diff
@@ -0,0 +1,43 @@
...
+     * Default branch using Tree-based view.  Enforces University Level, limits to one level-six research, shows research level headers, etc.
+     * All branches without an explicit type, or with an invalid explicit type, will act as DEFAULT.
+     */
+    DEFAULT("default"),
+    /**
+     * Achievement-style branch using Tree-based view.  Does not enforce university level, allows multiple researches above the university's max level,
+     * does not show research level requirement in headers, does not show research level time headers if branch-time is under 0.01.
+     * This is not a replacement for vanilla advancements: use only where University behaviors are important (eg: time, item cost, parent requirement enforcement),
+     * where requirement complexity is too broad to fit the Advancement system, where exclusive branches or colony-wide announcements are required,
+     * or where requirement or reward is tied to a colony-wide behavior instead of a player-specific one.
```
Raycoms: I'm not happy with calling this "achievement" still. Can we make this maybe "unrestricted" or so?

src/main/java/com/minecolonies/coremod/research/GlobalResearchBranch.java
```diff
@@ -0,0 +1,201 @@
...
+    @Override
+    public int getSortOrder(){return this.sortOrder;}
+
+    @Override
+    public ResearchBranchType getType(){return this.type;}
+
+    @Override
+    public boolean getHidden(){return this.hidden;}
+
+    public GlobalResearchBranch(final ResourceLocation id)
```
Raycoms, line 96: javadoc for the two constructors

src/api/java/com/minecolonies/api/research/ResearchBranchType.java
gattsuru: "unrestricted-tree" ok?  I'm kinda expecting to eventually find need for other layouts using the unrestricted behaviors.

ravenbuilder934: Could call it Unlockables, i think that's what Mekle is doing in his PR

ravenbuilder934: I prefer Achievements tbh. I feel like that tells players its purpose the fastest, and imo Unrestricted isn't very player-facing

Raycoms: Disagree, Achievements has a reserved meaning, this is not 'achievement'

uecasm: Unlockables, Features, Bonuses, Specials.... it might be easier to pick a name if there was more explanation of possible intended uses.  I know that Mekle was intending to use this for unlocking mesh crafting in other huts based on the sifter hut level, are there some other examples?

Raycoms: "Achievements" don't affect anything, so calling it achievements is kinda off. "Unlockables" "Bonuses" "Specials" would all work. For now it's only the mesh thing that will use it so far. So that could be "Feature" or "Bonus" or "Special" but, I think we fare better with "Unlockables" because it's more generic

gattsuru: "Unlockables" is kinda awkward since it could conceivably not have any ResearchEffects on it at all (either by intent, by using research-ids for recipe unlocks, or if some external mod only uses research-ids).  This trait doesn't require or really have much impact on effects; it's focused on removing general restrictions for starting or displaying research.  But it works, is less awkward than "unrestricted-tree", so have swapped from "unrestricted-tree" to it. 

The thought is that this would be useful for: 
* potential multiplayer-focused Expert-style modpacks like Enigmatica or Magiculture, where a large number of potential huts and a variety of recipes would be blocked by requirement that certain building or item requirements were completed, including key components of entry to other mods.
* a guided-tour style focus, such as in Harvest Crossing or early phases of skyblocks, where the researches aren't having much actual effect, but the requirements to achieve them points you toward the underlying gameplay loop for that stage of the game.
* A research branch where research has both positive and negative effects (more practical once the more complicated effect system is in place).  Ie, a chain of research might increase archer damage but decrease their movement speed, with magnitude increasing each level, and it be intended that the colony owner select or even change this as the colony undergoes different demands (possibly even outside ones, like some modpacks that change spawn rules or AI for monsters over time, like Manufactio).

And then there could potentially be variants using a scrollinglist or whatever implementation is used for the guidebook, but otherwise shared most of the logic, for cases that needed more text space or intended a more linear presentation.

This particular string is also specifically json- and code-side so the only time users should see it is if working on a datapack or with the api.

Raycoms:
Has a little conflict

PR #6726, opened 2021-03-11
Tool Texture Cleanup, Claim command messages, max limit to Builder progress-needed

gattsuru:
- from #6657, protects against the negative percentages on builder resource-related UIs, caused by situations where neededResources was getting filled up while the work order itself was only partially filled.
- from same issue, cleans up the alpha channels on the guard tool, lumberjack restriction tool, and permission scepter.  The minecolonies sceptergold isn't the one that gets actually used in all cases, but changed it anyway; will do the same in Structurize as a separate PR.
- adds a separate message for the Claim command in "false" mode, to better distinguish between claim and unclaim commands.  Fixes a related typo.

Review please


ravenbuilder934:
Beekeeper scepter?

gattsuru:
Ah, thanks, missed that one.  Beekeeper Scepter and the Scrolls added.

ravenbuilder934 (dismissed):

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingStructureBuilder.java
```diff
@@ -333,7 +333,7 @@ public void serializeToView(@NotNull final PacketBuffer buf)
                 }
 
                 buf.writeString(desc);
-                buf.writeDouble(workOrderBuildDecoration.getAmountOfRes() == 0 ? 0 : qty / workOrderBuildDecoration.getAmountOfRes());
+                buf.writeDouble(workOrderBuildDecoration.getAmountOfRes() == 0 ? 0 : (qty < workOrderBuildDecoration.getAmountOfRes() ? qty / workOrderBuildDecoration.getAmountOfRes() : 1.0));
```
Raycoms: Were you able to reproduce this happenin?

gattsuru: Yes, though not controllably enough to find the underlying reason getAmountOfRes was only partially filled.  Very large builds where the builder had to clear a lot of spaces and other modifications to the build area occurred seemed most likely to have it happen, but even with slower AI config it didn't always show up even when working the same build at the same location.

Raycoms: I tried hunting this down numerous times but couldn't reproduce it. It's weird because the amoutnOfRes is only set once, and that is after the scanning step is done. And it is hooked in, that if a rescan is done, and the new qty is bigger, it will adopt the new quantity.

gattsuru: It definitely can get called multiple times out of expected order in weird situations.  One example case: building a townhall1 in a spruce forest biome, with a builder2, warehouse1, four colonists, build order issued before the builder had complete builder2, all acacia style.  First setAmountOfRes sets from 0 to 8, and only requests eight torches.  At this time, the progress value is correct(ish) at 0%.

Builder begins clearing phase, cuts out trees, ferns, and ground clutter.  When the clearing phase is complete, the builder recalculates the qty value with normal resource requirements for the building, but does not yet issue a new setAmountOfRes().

image

At this point, you'd get a large negative number on the display -- in this case, progress would be set to 68.25 (546 / 8), and the builder hut would show -6725%, without this fix.  There's probably some deeper AI fault causing that first request for just the torches, but replicating it's been tricky enough that I'm not sure it's worth tracking down further.

Unfortunately, since the calculation isn't stored, closing and reopening the world resets the value (to a saneish number), so I can't just upload the world file.

gattsuru: Raycoms traced the problem down to, if I'm understanding it correctly, requestMaterialsState() not updating setAmountOfRes when in the SOLID, DECO, or ENTITIES state, and that seems a far more elegant fix.  Removed the extra ternary, and pushed his fix instead.  At least cases where I'd replicated the problem previously seem to be reliably giving non-crazy numbers.

Raycoms:
:shipit:

someaddons (approved):

ravenbuilder934 (approved):

PR #6736, opened 2021-03-13
Datagen for sifter

uecasm:
Sifter recipes, on the other hand, are an excellent candidate for datagen.

- Converts sifter recipes to datagen, with separated loot tables.
- Fixes transposed iron/dirt and diamond/dirt sifting recipes.
- Improves hand animation for sifter while crafting.

Review please


src/main/java/com/minecolonies/coremod/entity/ai/citizen/sifter/EntityAIWorkSifter.java
```diff
@@ -99,27 +103,36 @@ protected IAIState sift()
...
-            }
+            worker.setHeldItem(Hand.MAIN_HAND, inputItem);
         }
-
-        if(!meshItem.isEmpty() && (ItemStackUtils.isEmpty(worker.getHeldItemMainhand()) || ItemStackUtils.compareItemStacksIgnoreStackSize(worker.getHeldItemMainhand(), meshItem, false, true)))
+        if (!meshItem.isEmpty() && (ItemStackUtils.isEmpty(worker.getHeldItemOffhand()) || ItemStackUtils.compareItemStacksIgnoreStackSize(worker.getHeldItemOffhand(), meshItem, false, true)))
         {
-            worker.setHeldItem(Hand.MAIN_HAND, meshItem);
+            worker.setHeldItem(Hand.OFF_HAND, meshItem);
         }
```
uecasm, line 136: This makes the sifter hold the mesh in their offhand and the block being sifted in their main hand (and still swinging their main hand), which I think leads to a better animation.  (And the individual loot tables no longer care which hand the mesh is in, so that's not a compatibility issue.)

I've also removed the inventory search since the held items (as far as I can tell) are just for display purposes, so it doesn't matter whether it finds the "real" damaged mesh or the pristine mesh from the recipe.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
```diff
@@ -0,0 +1,364 @@
...
+                        .item(Items.POTATO, 1)
+                        .item(Items.PUMPKIN_SEEDS, 1)
+                        .item(Items.MELON_SEEDS, 1)
+                        .item(Items.BEETROOT_SEEDS, 1)
+                        .item(Items.DARK_OAK_SAPLING, 1)
+                        .item(Items.ACACIA_SAPLING, 1)
+                        .build()),
+
+                new SifterMeshDetails(ModItems.sifterMeshDiamond, 5, new LootTableBuilder()
+                        .bonusRolls(0.025f)
```
uecasm: I've kept this value from the original loot table, but I suspect that it's a typo, since all of the other diamond mesh tables have a bonus roll of 0.035 instead.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
```diff
@@ -0,0 +1,364 @@
...
+                        .bonusRolls(0.035f)
+                        .empty(40)
+                        .item(Items.REDSTONE, 20)
+                        .item(Items.IRON_NUGGET, 20)
+                        .item(Items.COAL, 20)
+                        .item(Items.LAPIS_LAZULI, 10)
+                        .item(Items.IRON_INGOT, 2)
+                        .item(Items.GOLD_INGOT, 2)
+                        .item(Items.EMERALD, 2)
+                        .item(Items.DIAMOND, 2)
```
uecasm, line 134: In a few places I've changed "2.5" from the original json files to "2" because loot table weights are only permitted to be integers.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
```diff
@@ -0,0 +1,364 @@
...
+                            .lootTable(new ResourceLocation(MOD_ID, "recipes/" + name))
+                            .minBuildingLevel(mesh.getMinBuildingLevel())
+                            .build(consumer);
+                }
+            }
+        }
+    }
+
+    // there is a LootTableProvider but it's a bit more of a pain to work with
+    private class SifterLootTableProvider implements IDataProvider
```
uecasm, line 331: I wasn't sure whether it was preferred to use the vanilla LootTableProvider and LootTable.Builder or the LootTableJson from the LDT datagenerators lib; I'm currently using the latter since it allows enumeration, which the other does not, and because the existing block loot generator uses it.  One downside of this is that LootTableJson does not provide a way to name the loot pools.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/sifter/EntityAIWorkSifter.java
uecasm: I'm not sure if the "don't set if already the correct item" checks are strictly necessary (I would think that would be checked inside the function), but I kept them since they were used in the original code.

Mekle001: The reason for the inventory search (and no, it didn't work yet) was to support possible enchantments on the meshes, like Unbreaking, where the pristine one in the recipe didn't have that enchantment.  The "don't set" checks are because there aren't a lot of checks. I found this out the hard way working on the smeltercrafter, where I set it everytime, and the block flickered in their hand.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
Mekle001: Yes, it's a typo.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
Mekle001: Are you sure about that? I thought I had traced down to the underlying implementation and found it a floating point when I did that.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
Mekle001: I'm not sure loot pool names is a big deal, and enumeration definitely has it's advantages. I think this is fine.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
```diff
@@ -0,0 +1,364 @@
...
+                {
+                    final String name = mesh.getName() + "/" + inputEntry.getKey().getRegistryName().getPath();
+                    final Stream<Item> loot = mesh.getLootTable().getPools().stream()
+                            .flatMap(pool -> pool.getEntries().stream())
+                            .filter(entry -> entry.getType().equals(EntryTypeEnum.ITEM))
+                            .map(entry -> ForgeRegistries.ITEMS.getValue(new ResourceLocation(entry.getName())));
+
+                    CustomRecipeBuilder.create(ModJobs.SIFTER_ID.getPath(), name)
+                            .inputs(Stream.of(new ItemStack(inputEntry.getKey()), new ItemStack(mesh.getMesh())).collect(Collectors.toList()))
+                            .secondaryOutputs(Stream.concat(Stream.of(mesh.getMesh()), loot)
```
Mekle001, line 319: It may be worth adding an additional filter here for the very low probability items.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
```diff
@@ -0,0 +1,364 @@
...
+                        .empty(40)
+                        .item(Items.CACTUS, 15)
+                        .item(Items.SUGAR_CANE, 15)
+                        .item(Items.GOLD_NUGGET, 15)
+                        .item(Items.COCOA_BEANS, 15)
+                        .build())
+
+                ).collect(Collectors.toList()));
+
+        inputs.put(Items.SOUL_SAND, Stream.of(
```
Mekle001, line 173: Did we also want to add SOUL_SOIL?

Mekle001 (changes requested):
Overall looks good. You called out a typo or two you didn't fix, so let's get those fixed.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/sifter/EntityAIWorkSifter.java
uecasm: True, an enchanted mesh would be a visible difference -- but then an enchanted mesh won't actually work with the code as written (and as before) anyway, because that's not an NBT match to the recipe.  Worry about that when it's actually implemented, probably. 😄

Having said that, I don't think it's too immersion-breaking to show the unenchanted mesh in their hands, it's mostly just to indicate which class of mesh they're using to make sure that the recipes are in the correct order in the hut.  (On that note, something I noticed both before and after these changes is that the recipes don't seem to be in any particular order -- notably string meshes appear both above and below flint meshes.  I thought you said that you had wanted them to be sorted with better meshes first?  I didn't investigate that, but it doesn't appear to be happening.)

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
uecasm: Ok, I'll fix that.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
uecasm: Yes, I'm sure.  The wiki docs for vanilla loot tables specify that both weight and quality are integers, and this also matches the implementation of `ItemLootEntry` in vanilla code and `EntryJson` in the datagen lib.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
uecasm: My only hesitation there is that the Forge documentation explicitly states that while names aren't required by vanilla they are required by Forge for the purposes of one of the events that lets mods edit tables.  Having said that, it does appear to actually work as is, so perhaps they're just treating it like a vanilla table and giving it an automatic name.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
uecasm: To hide them, you mean?  I'm not sure that's needed, there's nothing below 1% probability and it is useful to know that there is at least _some_ chance of producing an item there.  Otherwise the couriers will take it away and you have no idea where it's coming from.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
uecasm: I did have that in my previous PR, but you didn't in yours, so I've left it out for the moment.  It can be in a follow-up.

I previously had it using the soul sand table (which was just a one-liner at the time, but in this version you'd have to duplicate the whole code block and it'd create separate tables); and on further reflection it might be better giving it a different loot table anyway, perhaps something more like the dirt table but dropping nether saplings?  Or maybe that should be nylium instead?  Something for you to think about, anyway, particularly for the nether colony.  😄

uecasm: Some of the other weights might need a little tweaking in a later PR, after further experimentation with the current values.  I suspect that if you do any dirt sifting you're going to be drowning in wheat seeds before too long.  There's only so much the farmer and composter can take. 🤣

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
uecasm: Well, I guess a few things are under 1%.  But not by much.

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
uecasm: Fixed.

Mekle001 (approved):

src/main/java/com/minecolonies/coremod/generation/SifterCraftingProvider.java
Mekle001: Fair enough

someaddons (approved):

PR #6742, opened 2021-03-14
Create modular windows

Raycoms:
- Create modular windows
- Finish modularization of filterable lists
-

Review please


src/main/java/com/minecolonies/coremod/client/gui/AbstractModuleWindow.java
```diff
@@ -0,0 +1,77 @@
...
+    public AbstractModuleWindow(final IBuildingView building, final String res)
+    {
+        super(res);
+        final Random random = new Random();
+        int offset = 0;
+        if (building.getAllModuleViews().size() > 0)
+        {
+            final ButtonImage image = new ButtonImage();
+            image.setImage("minecolonies:textures/gui/modules/tab_side" + (random.nextInt(3) + 1) + ".png");
+            image.setPosition(-20, 10 + offset);
```
Nightenom, line 36: stop using negative positions

src/main/java/com/minecolonies/coremod/client/gui/AbstractModuleWindow.java
```diff
@@ -0,0 +1,77 @@
...
+
+            offset += image.getHeight() + 2;
+
+            this.addChild(image);
+            this.addChild(iconImage);
+
+            PaneBuilders.tooltipBuilder().hoverPane(iconImage).build().setText(new TranslationTextComponent("com.minecolonies.coremod.gui.maintab"));
+        }
+
+        for (IBuildingModuleView view : building.getAllModuleViews())
```
Nightenom, line 55: this seems to be in single column, using list might be better, you can hide the scrollbar if needed (needs bo change tho)

src/main/java/com/minecolonies/coremod/client/gui/ItemListModuleWindow.java
```diff
@@ -0,0 +1,216 @@
...
+    {
+        updateResources();
+    }
+
+    @Override
+    public boolean onKeyTyped(final char ch, final int key)
+    {
+        final boolean result = super.onKeyTyped(ch, key);
+        filter = window.findPaneOfTypeByID(INPUT_FILTER, TextField.class).getText();
+        updateResources();
```
Nightenom: only update if filter changes, otherwise any key type triggers update

Nightenom (dismissed):

src/main/java/com/minecolonies/coremod/client/gui/AbstractModuleWindow.java
Raycoms: For now we want this to be a single column, and if we have more one day, extend it to extend the other side on the other side of the UI

src/main/java/com/minecolonies/coremod/client/gui/AbstractModuleWindow.java
Raycoms: So, I have to increase the overall window size then?

src/main/java/com/minecolonies/coremod/client/gui/AbstractModuleWindow.java
Nightenom: then create a proper handler

src/main/java/com/minecolonies/coremod/client/gui/AbstractModuleWindow.java
Nightenom: yep

src/main/java/com/minecolonies/coremod/client/gui/AbstractModuleWindow.java
Raycoms: This is the proper place for this.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModuleView.java
```diff
@@ -21,4 +22,22 @@
      * @return this module itself.
      */
     IBuildingModuleView setBuildingView(final IBuildingView buildingView);
+
+    /**
+     * Get the matching window for the module.
+     * @return the window.
+     */
+    Window getWindow();
```
marchermans, line 33: You will need an @OnlyIn here or on the view.....
Or optionally a Supplier of a Supplier structure....

This won't work on a dedicated server, or at least is very dangerous.

marchermans (changes requested):
Look good, i have some sidedness concerns with the Window being exposed this way.

src/api/java/com/minecolonies/api/colony/buildings/modules/IBuildingModuleView.java
Raycoms: Aren't views always only client side?

someaddons: might be not due to the way we're constructing buildings now in our buildings entry right

uecasm: It'd probably be a good idea to have client-side-only modules in a different package from the server-side-only modules, at least. It's easier to spot when something is improperly importing across sides when they're segregated.

Raycoms: We're only giving it a producer in the entry, not an actual view. But I agree with @uecasm I'll make it a seperate folder

Raycoms: I double checked, this works fine in SMP too. But I moved it into seperate folders

PR #6791, opened 2021-03-21
Enchanter loottable

uecasm:
- ~~Specific recipe inputs can be explicitly tagged to ignore NBT (e.g. since Ancient Tome gets some NBT when handled by a player, but it still should be valid for crafting either way).~~
- ~~Items can be tagged in general to ignore NBT for most uses (can be overridden in specific cases); this is a little more dangerous, but useful where there's no great loss if the NBT is lost or merged and the items should be treated as the same.  Particularly handy for items that get tagged on `inventoryTick` purely for display purposes.~~
- Enchanter uses recipe system and loot tables instead of config.

Please review (especially the AI changes) ... it's probably simpler to review individual commits rather than all at once.

Note that as currently implemented the loot drops should be identical to previous implementation, but there are some issues with that which we should probably discuss and likely alter (which I'll post as a followup), hence why this is posted as a draft.  (And also because due to an unmerged dependency this won't actually build exactly as is.)


build.properties
```diff
@@ -8,4 +8,5 @@ mappings=20201028-1.16.3
 # Dependencies
 jei_version=7.6.1.71
 structurize_version=0.13.148-ALPHA
+datagen_version=0.1.48-ALPHA
```
uecasm: Note this version doesn't exist yet.  It's specified in anticipation of ldtteam/DataGenerators#2 being merged.

src/main/generated/resources/data/minecolonies/crafterrecipes/sifter/diamond/dirt.json
```diff
@@ -6,12 +6,12 @@
...
     },
     {
-      "item": "minecolonies:sifter_mesh_diamond"
+      "item": "minecolonies:sifter_mesh_diamond{Damage:0}"
     }
   ],
   "additional-output": [
     {
-      "item": "minecolonies:sifter_mesh_diamond"
+      "item": "minecolonies:sifter_mesh_diamond{Damage:0}"
```
uecasm: These changes weren't entirely intended, and I haven't specifically tested them as yet (but I don't expect them to cause issues since the sifter overrides things anyway).  It was a side effect of adding NBT support to `CustomRecipeBuilder`.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultEnchanterCraftingProvider.java
```diff
@@ -0,0 +1,403 @@
...
+                .item(enchantedBook(Enchantments.LOOTING, 1), 50)
+                .item(enchantedBook(Enchantments.POWER, 1), 50)
+                .item(enchantedBook(Enchantments.PROJECTILE_PROTECTION, 1), 50)
+                .item(enchantedBook(Enchantments.PROTECTION, 1), 50)
+                .item(enchantedBook(Enchantments.PUNCH, 1), 50)
+                .item(enchantedBook(Enchantments.RESPIRATION, 1), 50)
+                .item(enchantedBook(Enchantments.SHARPNESS, 1), 50)
+                .item(enchantedBook(Enchantments.SMITE, 1), 50)
+                .item(enchantedBook(Enchantments.SWEEPING, 1), 50)
+                .item(enchantedBook(Enchantments.UNBREAKING, 1), 50)
```
uecasm, line 65: These are copied from the previous configuration, but there are some issues that we might want to address:

1. They're all given equal probability.  The enchantment registry defines a "rarity" that in turn determines a weighting value that we should perhaps be applying to influence this instead.
2. Some vanilla enchants are entirely missing from this list:
    * Soul Speed
    * Efficiency
    * Silk Touch
    * Fortune
    * Luck of the Sea
    * Lure
    * Loyalty
    * Impaling
    * Riptide
    * Channeling
    * Multishot
    * Quick Charge
    * Piercing
3. Which also highlights an unfortunate side effect of doing things this way (rather than looking up enchants at runtime): no support for modded enchants (without datapack) and even vanilla enchants have to be added manually before they're supported.  Though the advantage is that we can more deliberately filter and alter probabilities of specific enchants this way.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultEnchanterCraftingProvider.java
```diff
@@ -0,0 +1,403 @@
...
+                .item(enchantedBook(Enchantments.POWER, 5), 1)
+                .item(enchantedBook(Enchantments.PROJECTILE_PROTECTION, 5), 1)
+                .item(enchantedBook(Enchantments.PROTECTION, 5), 1)
+                .item(enchantedBook(Enchantments.PUNCH, 5), 1)
+                .item(enchantedBook(Enchantments.RESPIRATION, 5), 1)
+                .item(enchantedBook(Enchantments.SHARPNESS, 5), 1)
+                .item(enchantedBook(Enchantments.SMITE, 5), 1)
+                .item(enchantedBook(Enchantments.SWEEPING, 5), 1)
+                .item(enchantedBook(Enchantments.UNBREAKING, 5), 1)
+                .item(enchantedBook(ModEnchants.raiderDamage, 2), 1)
```
uecasm, line 350: Another issue with this at present is that it violates the max level specified by the vanilla enchants -- for example:
* Aqua Affinity and Flame should only have level 1
* Punch only goes to level 2
* Respiration and Depth Strider only go to level 3
* Protection (all types) only go to level 4
(there are others)

This is not the first mod to allow "stronger" enchants than are possible via normal enchanting and crafting, and most will either cap their effect at the level they're supposed to (despite appearing higher) or will have an enhanced effect, which may or may not be intended.  But it's worth mentioning to consider if we want to fix this, or if it is intended.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultEnchanterCraftingProvider.java
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+        @Override
+        protected void registerRecipes(@NotNull final Consumer<IFinishedRecipe> consumer)
+        {
+            final CompoundNBT tag = new CompoundNBT();
+            tag.putBoolean(NbtTagConstants.TAG_IGNORE_NBT, true);
+
+            final List<ItemStack> tome = Collections.singletonList(new ItemStack(ModItems.ancientTome, 1));
+            tome.get(0).setTag(tag);
```
uecasm: It was necessary to ignore NBT in the recipe since the Ancient Tome has three different states (never handled by a player vs. last handled on a day when there was or wasn't a raid) which are considered incompatible by default.

I suspect (given some other reports on Discord) that the research system might have a similar issue with this item (although that might have only been on an older version).  (And perhaps we might want to use this tag on some other recipes -- IIRC there was a similar compatibility issue with Quark and clocks/compasses.)

uecasm:
I have tested the AI and it appears to work, but I might have missed something subtle.

build.properties
Raycoms: Just merged, should be available soon

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultEnchanterCraftingProvider.java
Raycoms: Efficiency should be added with the same probabilities and levels as sharpness.
Multishot can be as rare as mending
Silk Touch the same as mending
Fortune can be as rare as mending, but with increasing levels ( 1 at level 3, 2 at level 4, and 3 at level 5)
Quick Charge could be the same as sweeping

The rest we can leave out for now.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultEnchanterCraftingProvider.java
Raycoms: That is intended

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -98,11 +99,17 @@ public ItemStorage(@NotNull final ItemStack stack)
...
-        this.shouldIgnoreNBTValue = false;
+        this.shouldIgnoreNBTValue = getDefaultIgnoreNBT(stack);
         this.amount = ItemStackUtils.getSize(stack);
         this.creativeTabIndex = stack.getItem().getCreativeTabs().stream().filter(Objects::nonNull).map(g -> g.index).collect(Collectors.toList());
     }
 
+    private static boolean getDefaultIgnoreNBT(@NotNull ItemStack stack)
+    {
+        //noinspection ConstantConditions
+        return stack.hasTag() && stack.getTag().getBoolean(NbtTagConstants.TAG_IGNORE_NBT);
```
marchermans: You might want to consider making this a Tag.
Since then other mods can add the compatibility directly, incase they also have items with this behaviour.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/enchanter/EntityAIWorkEnchanter.java
```diff
@@ -408,4 +418,52 @@ private void resetDraining()
     {
         return BuildingEnchanter.class;
     }
+
+    private static class RecordingItemHandler implements IItemHandler
```
marchermans: This might actually be usefull in other locations.
Could you put this in the API as a proper class?

Also: There is IModifyableItemHandler, which has a direct set method, as such you might rather want to make sure that is covered also somehow.

src/main/java/com/minecolonies/coremod/generation/LootTableBuilder.java
```diff
@@ -0,0 +1,90 @@
+package com.minecolonies.coremod.generation;
```
marchermans, line 1: Should be part of the DataGen lib.
Or use the default one provided by Minecraft itself.

src/main/java/com/minecolonies/coremod/generation/LootTableJsonProvider.java
```diff
@@ -0,0 +1,54 @@
+package com.minecolonies.coremod.generation;
```
marchermans, line 1: Should be part of the DataGen lib.
Or use the default one provided by Minecraft itself.

marchermans (changes requested):
Some minor nitpickings.

But the rest looks really good.

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
uecasm: I'm not quite sure what you mean by this.  It is passed through the tag?

src/main/java/com/minecolonies/coremod/generation/LootTableJsonProvider.java
uecasm: This can't be in the DataGen lib since it depends on Minecraft types.

src/main/java/com/minecolonies/coremod/generation/LootTableBuilder.java
uecasm: Could do, although this is mostly a convenience wrapper for use in the MC-only datagen process, so it's a bit borderline.

As previously mentioned I'm not using the MC builder directly since it's write-only and the sifter datagen needs to be able to read back the data.

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
uecasm: Oh, you mean a datapack item tag?  That could be useful too, although it's a much bigger stick (this one affects only specific recipes, that would affect all recipes).

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultEnchanterCraftingProvider.java
gattsuru: Yeah, there were a couple inconsistencies due to this issue on research for More Scrolls, has been fixed.  Research now assumes that anything without explicit NBT data can accept any NBT state for item costs and reset costs, and ItemStorage can handle that, even if there's not much in InventoryUtils to help with it.

The Quark problem goes deep into Structurize, and unfortunately it's also an issue with dyeable armor and banners, and there's not such an easy separation there between cases where we don't care about NBT and those we do.

src/main/generated/resources/data/minecolonies/crafterrecipes/sifter/diamond/dirt.json
```diff
@@ -6,12 +6,12 @@
       "item": "minecraft:dirt"
     },
     {
-      "item": "minecolonies:sifter_mesh_diamond"
+      "item": "minecolonies:sifter_mesh_diamond{Damage:0}"
```
Mekle001: This worries me, as we don't want to match to Damage:0 here.

Mekle001 (changes requested):
Let's not include NBT specifiers for Damage:0, especially when we don't want to match on Damage.

src/main/generated/resources/data/minecolonies/crafterrecipes/sifter/diamond/dirt.json
uecasm: It was added due to the automatic support for any NBT values, and that being the default for a fresh ItemStack.

I was a bit worried about it at first as well, but I've tested it with a sifter in-world and they'll still happily do crafting operations with damaged meshes, so the code is overriding things in the right ways.  (But it would have had to, because specifying the item without NBT would have never matched any real-world item either.)

So the only effect this actually has is on the preview item, but there's no visual difference and it's the same as what it would default to the instant you put it into an ItemStack anyway.  So I think it's fine as is, other than producing a little clutter in the JSON.

uecasm:
The Damage:0 specifier doesn't appear to cause any harm, and it would be massively complicated to leave it out (especially for any item that has more than just a damage value).  So I'm inclined to leave it as is.

src/main/generated/resources/data/minecolonies/crafterrecipes/sifter/diamond/dirt.json
Mekle001: Ok.

src/main/resources/data/minecolonies/tags/items/crafting_ignores_nbt.json
```diff
@@ -0,0 +1,10 @@
+{
+  "replace": false,
+  "values": [
+    "#minecraft:banners",
+    "minecraft:clock",
+    "minecraft:compass",
+    "minecolonies:ancienttome",
+    "minecolonies:colony_banner"
+  ]
```
Raycoms: Can we add the vanilla books you can write inside to this too?

src/main/generated/resources/data/minecolonies/crafterrecipes/sifter/diamond/dirt.json
marchermans: All of these look majorly suspicous to me.
But what the hell, i checked Mc does not handle this any better so what ever...... Mojang should have come up with a better way of doing this to be honest.

marchermans (dismissed):

Mekle001 (dismissed):

src/main/generated/resources/data/minecolonies/crafterrecipes/sifter/diamond/dirt.json
uecasm: Completely agree.

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -98,11 +114,17 @@ public ItemStorage(@NotNull final ItemStack stack)
     {
         this.stack = stack;
         this.shouldIgnoreDamageValue = false;
-        this.shouldIgnoreNBTValue = false;
+        this.shouldIgnoreNBTValue = getDefaultIgnoreNBT(stack);
```
someaddons: We really can't/shouldn't make certain items get ignored in comparisons of all uses of ItemStorage, afterall it is used in far more places than just crafting. I do not think this is a good way to handle these things

someaddons (dismissed):

src/main/resources/data/minecolonies/tags/items/crafting_ignores_nbt.json
uecasm: I'm a bit reluctant to do that, as (unless I'm missing something) that would allow it to consume books the player has written important notes inside and happened to leave in their warehouse (and some players use their warehouse as primary storage, so that isn't too far-fetched).

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
uecasm: It's used in crafting and the request system.  That specific line that you highlighted only takes effect for itemstacks that have been specially tagged, which can only ever happen in recipes so cannot affect anything other than crafting, and doesn't affect any other usage of the same item, so it is entirely safe.

The item tag is more broad and will apply to every usage of that item (which is why I initially didn't have it in until Orion suggested it), but the specific items that are in the tag should always be "safe" to ignore NBT on for all purposes; when they do acquire NBT it's never anything important and just causes incompatibilities with recipes and other usage without good reason.  It's certainly more dangerous, if someone adds inappropriate things to the tag, but hopefully nobody does, or they quickly learn the error of their ways.  But the global nature does make it less likely for these known-problematic items to actually cause problems.

And if you really want to perform a comparison including NBT on these items in the code, you can still force it.

someaddons: It is used a lot in storage e.g. racks aswell as multi-purpose item comparisons ranging through a broad variety of use cases, just slamming an ignore into all of these for one use case doesnt make a lot of sense.

someaddons: Also I dont not think an ignore all nbt of specific items is a good solution of your problem either, I'd rather see a one sided hasRequiredNBt

src/main/resources/data/minecolonies/tags/items/crafting_ignores_nbt.json
Raycoms: fair

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
Raycoms: Now, one thing we have to be aware off, and we might want to have a comment on, is how someaddons said it, since we use this for matching, this can result in "stack compression" such that nbt data of items is multiplicated/lost.

I do think that all the items in the tag are totally find with that, but we should probably comment this on the tag

uecasm: Yes, stack compression is definitely a thing, but as Raycoms said for these specific items it should cause no harm.  I'm happy to add a dire warning to the tag JSON to that effect; what's the preferred syntax for that in a way that doesn't break the tag?  (I can invent one that works, but I'd prefer to use an established syntax if there is one.)

src/main/java/com/minecolonies/coremod/entity/ai/citizen/enchanter/EntityAIWorkEnchanter.java
```diff
@@ -228,36 +230,43 @@ private IAIState enchant()
...
-            {
-                final int openSlot = InventoryUtils.getFirstOpenSlotFromItemHandler(worker.getInventoryCitizen());
-                if (openSlot == -1)
-                {
-                    //Dump if there is no open slot.
-                    incrementActionsDone();
-                    progressTicks = 0;
-                    return IDLE;
-                }
+            List<RecordingItemHandler> handlers = getOwnBuilding().getHandlers().stream()
```
someaddons: using a RecordingItemHandler seems like quite the hacky workaround for the fullfillrecipe not giving back the required information, how about having a fullfillment which returns more information instead?

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
uecasm: And FWIW the tag is not intended to solve only this one use case.  These particular items have been perennial thorns in the side of all the use cases, including storage.

someaddons: I'm aware that there are issues caused by our currently complete/exact nbt comparisons, but just plugging the compat code for crafting into an all-purpose comparison storage is just a bad place for it

src/main/java/com/minecolonies/coremod/entity/ai/citizen/enchanter/EntityAIWorkEnchanter.java
uecasm: That's certainly another valid option, although it has its own share of awkwardness by requiring changing the signature of a **lot** of existing code, and of having multiple output values from a single method (which are certainly doable, but is uglier).

I'm happy to change it to do that if there's a general preference for that approach, but only in that case.

someaddons: I guess we also could have an alternate which provides more information on return instead? Or doesnt do the full craft and insert and instead just consume & return stacks

Raycoms: I think an alternative one that gives back a "CraftResult" object with the existing return value + the additional return is probably the best way.

uecasm:
Also bump.

Raycoms:
For the particular use-case of the ancient tomes, did you try to use the existing request system "Stack" request with the ignore nbt flag? It makes sense to use that and keep the more generic handling for a PR on the side since there is still need for discussion on how to handle this properly.

marchermans:
Did I miss it or is the datagen missing here?

uecasm:
> For the particular use-case of the ancient tomes, did you try to use the existing request system "Stack" request with the ignore nbt flag?

No, because I don't really know what that is.  But as far as I could tell, the recipe `getFirstFullFillableRecipe` was hardcoded to only ever do NBT-specific matching on all items, so even if it requested tomes without NBT it then couldn't actually use them.  (I did get the AI stuck in a loop like that several times before I settled on "fixing" the NBT issue.)

> It makes sense to use that and keep the more generic handling for a PR on the side since there is still need for discussion on how to handle this properly.

I can remove the tag implementation and go back to just flagging this one specific recipe to ignore NBT if you like.  It was Orion who suggested moving it to a tag.  Having said that, none of the items listed in the tag should cause any kind of trouble with the possible exception of banners, which do risk losing customisations.  I'd also be happy to just remove banners from the list until we decide how else to deal with them.  (But in my experience so far, having banners work is more valuable than losing customisations to them.  Granted I have not specifically tested if this helps the builder to use wrong-NBT banners or not.)

> Did I miss it or is the datagen missing here?

The datagen is in DefaultEnchanterCraftingProvider.  It might have been collapsed in the diff by default since it's a new file.

Raycoms:
> > For the particular use-case of the ancient tomes, did you try to use the existing request system "Stack" request with the ignore nbt flag?
> 
> No, because I don't really know what that is. But as far as I could tell, the recipe `getFirstFullFillableRecipe` was hardcoded to only ever do NBT-specific matching on all items, so even if it requested tomes without NBT it then couldn't actually use them. (I did get the AI stuck in a loop like that several times before I settled on "fixing" the NBT issue.)
> 
> > It makes sense to use that and keep the more generic handling for a PR on the side since there is still need for discussion on how to handle this properly.
> 
> I can remove the tag implementation and go back to just flagging this one specific recipe to ignore NBT if you like. It was Orion who suggested moving it to a tag. Having said that, none of the items listed in the tag should cause any kind of trouble with the possible exception of banners, which do risk losing customisations. I'd also be happy to just remove banners from the list until we decide how else to deal with them. (But in my experience so far, having banners work is more valuable than losing customisations to them. Granted I have not specifically tested if this helps the builder to use wrong-NBT banners or not.)
> 
> > Did I miss it or is the datagen missing here?
> 
> The datagen is in DefaultEnchanterCraftingProvider. It might have been collapsed in the diff by default since it's a new file.

An alternative "easy" solution could be adding a custom recipe to the hut, that transforms the item with nbt to the default tiem instance.

marchermans:
I checked this code.
I like this solution.

He is right. The recipes don't use a stack implementation for obvious reasons. So adaptations to ItemStorage are indeed needed. 

I want the tag to be there since this makes so much sense in this location to use.

I also found the datagen class now. Not sure why that slipped through my fingers the last time.

All in all good pr. Bit heavy, but this works.

marchermans (dismissed):

uecasm:
> An alternative "easy" solution could be adding a custom recipe to the hut, that transforms the item with nbt to the default tiem instance.

Well, the downside of that is that the recipe list UI would be more confusing.  And it'd either have to be two recipes or it'd only work on days there isn't a raid.  And that doesn't really help with the items affected by Quark, since the data is not predictable.

FWIW, I did leave the original method in here too -- you can put an NBT tag on a specific item in a crafterrecipe to make the recipe ignore NBT on that item only in the recipe context.  The item tag is more flexible and powerful, but that could still be used in cases where it's too dangerous for the item tag.  I don't think this is one of them.

uecasm:
Since it seemed unpopular, I've reverted the NBT ignoring changes, so now this just makes use of Mekle's `ItemStorage` changes instead to ignore only for these specific recipes.  Clocks and compasses will have to be solved another way again.

Mekle001 (approved):

Mekle001:
Overall I like this a lot. The RecipeStorage changes only get us partway to where we eventually need to be, as I believe we need the option to not insert into inventory during recipe fulfillment, so we'll have a bit more refactoring, or a new behavioral modifier to pass through.

marchermans (approved):

Raycoms (approved):

PR #6794, opened 2021-03-21
Hotfix/recipebimap

Mekle001:
- Fix Hashing of RecipeStorage
- Truncate save and limit load of recipes in huts (my concrete mixer had over 200k!)
- Cache the global version of customrecipes

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -384,11 +384,11 @@ public class RecipeStorage implements IRecipeStorage
     public int hashCode()
     {
         return Objects.hash(cleanedInput, 
-            primaryOutput.getItem(), 
-            primaryOutput.getCount(), 
-            intermediate, 
-            gridSize, 
-            hashableItemStackList(alternateOutputs), 
+            primaryOutput.getItem(),
+            primaryOutput.getCount(),
+            intermediate,
+            gridSize,
+            hashableItemStackList(alternateOutputs),
             hashableItemStackList(secondaryOutputs),
             hashableItemStackList(tools));
     }
--- a/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -12,6 +12,7 @@ import com.minecolonies.api.util.InventoryUtils;
 import com.minecolonies.api.util.ItemStackUtils;
 import com.minecolonies.api.util.constant.TypeConstants;
 import net.minecraft.block.Block;
+import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.loot.LootContext;
 import net.minecraft.loot.LootParameterSet;
@@ -382,22 +383,29 @@ public class RecipeStorage implements IRecipeStorage
     @Override
     public int hashCode()
     {
-        int result = cleanedInput.hashCode();
-        result = 31 * result + primaryOutput.hashCode();
-        result = 31 * result + (intermediate != null ? intermediate.hashCode() : 0);
-        result = 31 * result + gridSize;
-        if(recipeSource != null)
-        {
-            result = 31 * result + recipeSource.hashCode();
-        }
-        if(recipeType != null && !(recipeType instanceof ClassicRecipe))
-        {
-            result = 31 * result + recipeType.hashCode();
-        }
+        return Objects.hash(cleanedInput, 
+            primaryOutput.getItem(), 
+            primaryOutput.getCount(), 
+            intermediate, 
+            gridSize, 
+            hashableItemStackList(alternateOutputs), 
+            hashableItemStackList(secondaryOutputs),
+            hashableItemStackList(tools));
+    }
 
-        result = 31 * result + alternateOutputs.hashCode();
-        result = 31 * result + secondaryOutputs.hashCode();
-        return result;
+    /**
+     * Convert a list of itemstacks into something hashable
+     * @param items List of item stacks to convert
+     * @return hashtable of items and counts
+     */
+    private Map<Item, Integer> hashableItemStackList(List<ItemStack> items)
+    {
+        Map<Item, Integer> hashableList = new HashMap<>();
+        for(ItemStack item: items)
+        {
+            hashableList.put(item.getItem(), item.getCount());
+        }
+        return hashableList;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -382,22 +383,29 @@ public boolean equals(final Object o)
...
+        for(ItemStack item: items)
         {
-            result = 31 * result + recipeType.hashCode();
+            hashableList.put(item.getItem(), item.getCount());
         }
-
-        result = 31 * result + alternateOutputs.hashCode();
-        result = 31 * result + secondaryOutputs.hashCode();
-        return result;
+        return hashableList;
```
marchermans: Nope.

ItemStacks still need to be accurate.
See the JVM Spec on equals/hashCode.

Also to make this accurate, you need some combination of: Item, Damage and NBT.

marchermans (changes requested):

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
Raycoms: @OrionDevelopment we discussed this once already, we hash only item, like we do it in ItemStorage, and use equals for the fine tuned matching. That works very well.

marchermans: Why?
Like what is the reason todo that?

src/main/java/com/minecolonies/coremod/colony/requestsystem/management/manager/StandardRecipeManager.java
```diff
@@ -91,7 +93,7 @@ public void read(@NotNull final CompoundNBT compound)
         for (int i = 0; i < list.size(); i++)
         {
             IRecipeStorage recipe = StandardFactoryController.getInstance().deserialize(list.getCompound(i));
-            if (recipe != null)
+            if (recipe != null && !recipes.containsKey(recipe.getToken()) && !recipes.containsValue(recipe))
```
Raycoms: value check should be enough

src/main/java/com/minecolonies/coremod/colony/requestsystem/management/manager/StandardRecipeManager.java
```diff
@@ -26,7 +28,7 @@
     /**
      * Map of all recipes which have been discovered globally already.
      */
-    private final BiMap<IToken<?>, IRecipeStorage> recipes = HashBiMap.create();
+    private final BiMap<IToken<?>, IRecipeStorage> recipes = HashBiMap.create(4096);
```
Raycoms, line 33: 4096? debug code?

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -576,8 +576,12 @@ else if (workersTagList.getCompound(i).keySet().contains(TAG_WORKER_ID))
             final IToken<?> token = StandardFactoryController.getInstance().deserialize(recipesTags.getCompound(i));
             if (!recipes.contains(token))
             {
-                recipes.add(token);
-                IColonyManager.getInstance().getRecipeManager().registerUse(token);
+                final IRecipeStorage storage = IColonyManager.getInstance().getRecipeManager().getRecipes().get(token);
+                if (IColonyManager.getInstance().getRecipeManager().getRecipeId(storage).equals(token))
```
Raycoms: if (!Objects.equals(IColonyManager.getInstance().getRecipeManager().getRecipeId(storage), token))

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -382,22 +383,29 @@ public boolean equals(final Object o)
...
+    }
+
+    /**
+     * Convert a list of itemstacks into something hashable
+     * @param items List of item stacks to convert
+     * @return hashtable of items and counts
+     */
+    private Map<Item, Integer> hashableItemStackList(List<ItemStack> items)
+    {
+        Map<Item, Integer> hashableList = new HashMap<>();
```
someaddons, line 413: seeing as the hashing now seems to get pretty extensive, does the Recipestorage ever change? if not  we could just calc it once

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -599,7 +603,7 @@ public CompoundNBT serializeNBT()
         compound.put(TAG_WORKER, workersTagList);
 
         compound.putInt(TAG_HIRING_MODE, this.hiringMode.ordinal());
-        @NotNull final ListNBT recipesTagList = recipes.stream()
+        @NotNull final ListNBT recipesTagList = recipes.stream().limit(this.getMaxRecipes() * 10)
             .map(iToken -> StandardFactoryController.getInstance().serialize(iToken))
```
Raycoms, line 607: for anyone wondering, this is a limitter for the future, to avoid huge leaks, x10 should still be big enough to detect the leak for the player

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
Raycoms: theoretically that should be possible. But afaik that's "bad style". Unless the object really is immutable

someaddons: well we're using it relatively frequently right

someaddons: er actually it gotta keep the same hash, else we run into issues with the map thinking we got the old hash while we dont

Raycoms: The one inside the map yes, the ones we push around outside not

Mekle001: RecipeStorages are effectively immutable. All of their settings happen in the constructor. Caching should be safe.

someaddons (approved):

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
uecasm: Wouldn't it make more sense to use a `List<ItemStorage>` instead?  The main point of ItemStorage was to provide more sane equals/hash/large-count behaviour, I thought.

Raycoms: I thought about that as well, but itemStorage doesn't include "count"

PR #6797, opened 2021-03-21
Fix oak stairs causing sawmill to learn oak-oak timberframes

Mekle001:
- Include the output in prereq checking

This causes stairs/slab etc to match the timberframe recipe and add it to the recipe list

Review please


ravenbuilder934 (dismissed):

ravenbuilder934 (dismissed):

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -186,6 +191,10 @@ public RecipeStorage(final IToken<?> token, final List<ItemStack> input, final i
         {
             if (ItemStackUtils.isEmpty(stack) || stack.getItem() == ModItems.buildTool)
             {
+                if(stack.getItem() == ModItems.buildTool)
+                {
+                    usesBuildTool = true;
```
Mekle001: While tracking this isn't critical to this particular scenario, it allows the "full" input check during the custom recipe check without falling back to the uncleaned inputs and recleaning them each time. It also can be ordered  first in the evaluation, and becomes a great short-circuit for the way we use prereq recipes today (upgrading a normal to a multi-output) since all of our recipes today use a build tool in the inputs.

Mekle001: If we remove the usesBuildTool tracking, and leave the additional getPrimaryOutput comparison, it solves the observed problems, but leaves us with a possibility of inputs matching when only one side has the build tool.

PR #6798, opened 2021-03-22
Disease Config File Improvements, Better Compatibility Manager Reloading, Research Fixes

### LINKED ISSUE #6784: Removing disease from server config crashes save file
### Prerequisites 

### Context 

- Minecraft Version:
- MineColonies Version:
- Structurize Version:

## Expected behavior

when removing a disease from the server config, one of the 2 cases are reasonable:
1. the colonists currently with that disease is cured
or 2. the colonists that were sick are still sick but no one else can get that disease 
## Actual behavior

The save file crashes when the world is loaded. Re-adding back the config fixes the issue.
once everyone is cured, disabling the config works and the save no longer crashes.

1. get some colonists to have small pox
2. save and quit the world
3. remove small pox from config
4. launch the world
5. crash 

### Logs

- latest.log: 
- crashlog: 

### Notes

---
## Viewers

gattsuru:
Closes #6784

- Sets illnesses to be removed if no matching disease string is available.  Such a cure won't result in immunity ticks against other diseases, however.
- Prevents crashes when attempting to sicken a citizen when all diseases are removed from configuration file.
- Tries to have `CompatibilityManager` reload properly between different connections.  ~~To do so, sets `CompatibilityManager` to load during `FMLServerStartedEvent` on dedicated server and single-player mode from `FurnaceRecipe.java`, and adds a `ClientPlayerNetworkEvent.LoggedOutEvent` into `ClientEventHandler` to force invalidation of tag and config-file based Compatibility.  This isn't perfect -- connected clients still use the `onClientTick` approach, and this can often fire between the `ClientPlayerNetworkEvent.LoggedOutEvent` and the actual logout.  But trying to reuse the `invalidateRecipes()` call from a reload listener fired repeatedly and had serious performance impact on connecting to a world or running `/datapack enable/disable`, and most other places to run discover() do not reliably have block and item tags loaded for the smeltable list, while the worst-case performance impact here is limited to an extra call when leaving the server.~~ Vastly cleaner implementation, details below.
- ~~Changes gradle file to match changes in https://github.com/ldtteam/Structurize/pull/318~~ (which was apparently also already done, nevermind).  Does a few improvements to match changes to datagen for Structurize 156+ (which also cleans up the code anyway).
- Sets Healing Cream cost to 8 carrots.  Derp.

Review please


ravenbuilder934:
If the structurize version is ready to be bumped now, might as well bump it

gattsuru:
> If the structurize version is ready to be bumped now, might as well bump it

There were a _lot_ of unrelated changes outside of the Research Datagen side, and while most were just adding `get()`, some of them were really janky (`return block instanceof FenceBlock || block instanceof FenceGateBlock || block == ModBlocks.CACTI_BLOCKS.getRegisteredBlocks().get(3).get() || block == ModBlocks.CACTI_BLOCKS.getRegisteredBlocks().get(4).get()`).

  If it saves Orion or ShadowProtocol time, I can push those as a separate PR, but it's fussy enough I don't want to be sneaking it through with other stuff, and if they've got a version already prepared it's going to be better code -- there's a lotta parts I'm not entirely confident even in their purpose to do regression testing, most seriously the changes to (aka removal of) the placeholder block.  The research stuff was small enough and discrete enough that it could play fine with either both current and older versions of Structurize.

gattsuru:
Ok, summary : ModTags wasn't working consistently (or, afaict, at all) on remote clients.  `TagCollectionManager.getManager` and `ItemTags` seem to only return vanilla and forge tags long after a world load is complete.  This also means ItemTagUtils isn't reliable on a remote client for modded tags.  The only reliable tag sources I've found on remote clients are `ClientWorld.getTags()` and variants of `TagsUpdateEvent.getManager`.  Wasn't an issue until now, since we essentially only used tags on a logical server-side, but there's a lot of cases where that may not be the intended case, especially as more and more goes into the data pack system, and if any calculations for the JEI plugin need to be or should be done client side.

I've hooked ModTags to use and store (as `supplier`) the most recent `TagsUpdateEvent.VanillaTagTypes.getManager`.  This fires on connection and on datapack reload, on both client and server, so as long as we only use ModTags to access non-vanilla tags it seems reliable.  Unfortunately, this event happens early enough on server start or remote connection that FurnaceRecipes and their utility predicates aren't loaded, so we have to tie into other client and server matters tied to FurnaceRecipes. 

Tested against single player, dedicated server, and open to lan mode, both on a single machine and across a simulated crappy internet connection, both with just Minecolonies + Struct and with ATM6.

I have left the current implementation of `florist_flowers` to only include vanilla (non-wither-rose) flowers.  The config setting previously included `"biomesoplenty:small_flowers"`, and I think it makes sense to readd them with some behavioral changes, but there's some _different completely unrelated_ undefined results with the `"required":false` syntax for tags I haven't resolved yet (see rant in discord), and it's not required for core functionality.

This is stylistically awkward.  I tried to keep using the style and management for Mojang's behaviors, but there's a lot of alternative approaches to several components, building the event handlers most obviously, and I'm not sure what version is Right.  TagsUpdateEvent.VanillaTagTypes is also strictly limited to the four vanilla types (of which we only use blocks, items for now).  TagsUpdateEvent.CustomTagTypes also includes at least potion and enchantment tags; if those would be useful (either for the enchanter, or the future alchemist job), it's one word to change.  _Probably_ missed an immutable somewhere.

These changes aren't equivalent for the various CompatibilityManager lists.  Some are intentionally so: Wither Roses are no longer included for plantables, and plantables as a whole are opt-in.  Others are different but _may_ be correct changes: vanilla suspicious stew and tomestone lollipops were previously excluded from edibles (and allItems); if that was an intentional exclusion, I'll swap back.  But the diff for ATM6 was pretty much limited to those.

gattsuru:
Changed the layout and event handlers for the Tag issue.  They're all now in `TagWorkAroundEventHandler.java`.  Because they apply in different contexts, they're set up as different static inner classes -- I have _absolutely_ no preference on that as a solution versus using `onlyin` versus making a TagWorkAroundFolder with a separate class for each; tell me if the preferred style is different.  Tried to expand documentation and comments to describe the scope of the issue in TagWorkAroundEventHandler.  I can add more if desired or necessary, and if I get the description to the point where I'm confident enough to make a forge issue I might add a link to it.

Tests in Minecolonies alone and ATM6 gave sane results.  The ores list is trimmed down very severely due to removal of `forest_flowers`, foods and edibles gained suspicious stew and lolipops. Otherwise, consistent in single-player, dedicated server, remote client, and data pack reloads.

This approach orphans `UpdateClientWithRecipes`.  It's plausible that may be needed or be useful in the future for other tasks, so I've left it untouched.

Raycoms (approved):

someaddons (approved):

PR #6821, opened 2021-03-27
Fix deletion of all recipes

Mekle001:
- Remove global existence check while deserializing buildings. The global collection is often empty at this stage
- Add code to remove no longer existing custom recipes from crafters

The second change I was testing when I fixed the first. Both are very simple. 

I'll do a follow up PR to re-instate the global existence check as a once-run during checkForCrafterSpecificRecipes, so it runs on the first colonytick handled by the building. 

Review please


Raycoms:
:shipit:

someaddons (approved):

PR #6830, opened 2021-03-28
Add colony export command

someaddons:
Adds a colony export command, which exports a small debug world folder of the colony, its data and its region files. It is intended to allow easier debugging of certain worlds, esp when looking at a certain colony on a big server world.

Includes these folders of the original world, region is limited to only chunks of a certain colony
image

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/util/BackUpHelper.java
+++ b/src/main/java/com/minecolonies/coremod/util/BackUpHelper.java
@@ -27,6 +27,8 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
@@ -418,6 +420,7 @@ public final class BackUpHelper
     {
         final MinecraftServer server = ServerLifecycleHooks.getCurrentServer();
         final File topworldDir = server.func_240776_a_(FolderName.DOT).toFile();
+        final File minecraftDir = new File(topworldDir.getAbsolutePath().replace(topworldDir.getPath(), ""));
 
         final String worldname = topworldDir.getParent().replace("." + File.separator, "");
         final String minecoloniesZipDir = worldname + File.separator + "minecolonies";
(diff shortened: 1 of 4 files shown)


Raycoms:
:shipit:

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -408,6 +408,7 @@
   "com.minecolonies.command.raidnow.success": "Raid started for colony %s, take care!",
   "com.minecolonies.command.raidtonight.success": "A raid on colony %s is coming tonight!",
   "com.minecolonies.command.loadbackup.success": "Successfully loaded backup.",
+  "com.minecolonies.command.export.success": "Exported colony to zip: %s",
```
ravenbuilder934, line 411: and if it's *not* a success?

Mekle001:
Can we include a list of mods and their versions in the export?

src/main/resources/assets/minecolonies/lang/en_us.json
someaddons: then the %s has the error message

someaddons:
hm for which usecase do you want a list of mods added? the data gathering here pretty much includes no mod data

uecasm:
List of mods would be useful for the "oh, you're using mod X, that causes bug Y" cases.

someaddons:
but that would normally be before asking the player to provide a world no?

uecasm:
If they remember to tell you, or if they've previously shown logs, sure.  But that's probably not always the case, and more info never hurts, especially if it's easy to capture like the modlist should be.  (Including recent logs automatically could be useful too, but that's probably harder after the fact unless you can just copy the existing logfile based on known relative path.)

someaddons:
logs and mod list probably would be done the same way via paths n filenames, the loaded mods dont have the filename version info

marchermans:
There is a way to get that via the ModList.

someaddons:
Ye the modlist was the first thing I checked, it doesnt have actual filenames just mod.toml versions, so only mods who provided one would show. Just doing it via files now as its similar to the latestlog anyways

src/main/java/com/minecolonies/coremod/util/BackUpHelper.java
```diff
@@ -382,4 +409,164 @@ public static void loadColonyBackup(final int colonyId, final RegistryKey<World>
...
+    /**
+     * Exports a certain colony and its part of the world and colony data data to a zip
+     *
+     * @param colony colony to export
+     * @return file Path and name
+     */
+    public static String exportColony(final IColony colony)
+    {
+        final MinecraftServer server = ServerLifecycleHooks.getCurrentServer();
+        final File topworldDir = server.func_240776_a_(FolderName.DOT).toFile();
```
Nightenom, line 422: just side comment, using nio Path would be appreciated as it have things like toAbsolutePath, resolve, normalize and relativize which would remove the replace (and File.separator) hacks below

someaddons: yea not worth the effort, would have to do it in the whole file as we use it in other places too

PR #6847, opened 2021-04-01
Improve raider spawn points to be more variable

someaddons:
Improve raider spawn points to be more variable
Fix pirates only rarely spawning because liquid wasnt considered a good spawn pos

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/modules/TavernBuildingModule.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/modules/TavernBuildingModule.java
@@ -30,7 +30,9 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.text.TranslationTextComponent;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 
 import static com.minecolonies.api.entity.ai.statemachine.tickratestatemachine.TickRateConstants.MAX_TICKRATE;
 import static com.minecolonies.api.util.constant.Constants.TAG_COMPOUND;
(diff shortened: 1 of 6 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/modules/TavernBuildingModule.java
```diff
@@ -173,7 +175,11 @@ private void spawnVisitor()
 
         newCitizen.getCitizenSkillHandler().init(recruitLevel);
 
-        BlockPos spawnPos = BlockPosUtil.findAround(building.getColony().getWorld(), building.getPosition(), 1, 1, bs -> bs.getMaterial() == Material.AIR);
```
Raycoms, line 176: I wonder if we want a util for this "findCitizenSpawnPos" and use that then everywhere, also for recall etc

src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
```diff
@@ -461,23 +489,52 @@ private BlockPos findSpawnPointInDirections(final BlockPos center, final Directi
...
                     }
                     else
                     {
                         break;
                     }
                 }
-            }
 
+                // Direction two
+                if (random.nextBoolean())
```
Raycoms, line 507: I wonder if there is a good way to encapsulate this into a method, seems to be almost 4 times the same code

src/main/java/com/minecolonies/coremod/colony/buildings/modules/TavernBuildingModule.java
someaddons: ye we could add some predefined to the utils for common cases

src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
someaddons: wondered too, but difficult as its a conditioned assign to two different vars

PR #6856, opened 2021-04-04
Citizen Cap improvements, adjust Research GUI to match configs

### LINKED ISSUE #6813: Builder completion -6383% complete on resource scroll.
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: 0.14.48-ALPHA
- Structurize Version: 0.13.48-ALPHA

## Problem description
Builder is reporting a completion percentage that does not seem possible on the resource scroll.

This was a dark oak citizen4 to citizen5 upgrade.

https://i.imgur.com/zUINZNK.png

The builder is working properly though and progressing through the upgrade. I originally saw the percentage as something like -7500% but didn't get a screen capture until he was already at -6300%.

### Notes or related things
I haven't been able to reproduce this.

---
## Viewers

* Add a thumbs-up to the issue to vote for it. This helps the request become more visible to the team and doesn't clutter the comments.

gattsuru:
Closes #6813

- Allows Unlockable branch autostart research to start automatically without a corresponding university equal or greater to its depth.
- Sets CITIZEN_CAP researches to default higher, and then controls the display on the client side to match the serverconfig range better.  This should allow anything in the serverconfig scope to be valid without an additional data pack.  Very low (<100) max citizen settings will still be a little goofy, but this seems the best compromise.
- Removes display of research with only WORKING_IN_RAIN effect on servers with the serverconfig workersalwaysworkinrain on, for colonies that haven't started/completed the research.
- These are both just done in the GUI.  I'd _rather_ fix things in the ResearchListener, but it fires before config files are read from disk (or can readily be read from disk).  The Official Forge answer was to move those settings into the data pack, and I can do that if preferred... but it seems even jankier for this specific case.
- Sets a floor for localProgress in the BuilderView, to prevent large negative numbers from showing up.  Whatever cause remains for this is even trickier to reproduce than the last version of the problem, but I was able to replicate it twice in current versions.
- Sets CitizenManager's spawn conditions to prevent non-force spawns over the citizen cap, rather than just when the current citizen count equals the max citizen count.

Review please


Raycoms:
:shipit:

PR #6858, opened 2021-04-04
Adds wooden mystical site

ravenbuilder934:
Closes #
Closes #
Closes #

- Found in the schem server 😄 
- Appears to be the same as stone, but it works for both styles imo.
-

Review please


Raycoms:
If it's exactly the same, we don't need it imo

ravenbuilder934:
true

PR #6862, opened 2021-04-04
Convert ItemStack to ItemStorage in input

Mekle001:
- Convert the list of ItemStack into ItemStorage for RecipeStorage input

This is the first step toward enabling a wider array of behavioral control in RecipeStorage, as we can now use ItemStorage as a carrier for matching information

Review please


src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -182,14 +182,15 @@ public RecipeStorage(final IToken<?> token, final List<ItemStack> input, final i
     {
         final List<ItemStorage> items = new ArrayList<>();
 
-        for (final ItemStack stack : input)
+        for (final ItemStorage stack : input)
```
Raycoms: maybe rename this to storage, to avoid confusion in the future in the code

src/api/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -215,17 +216,17 @@ public RecipeStorage(final IToken<?> token, final List<ItemStack> input, final i
     private List<ItemStack> calculateSecondaryOutputs()
     {
         final List<ItemStack> secondaryStacks = new ArrayList<>();
-        for (final ItemStack stack : input)
+        for (final ItemStorage stack : input)
```
Raycoms: same here

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
```diff
@@ -142,7 +142,9 @@ public void updateElement(final int index, @NotNull final Pane rowPane)
                     {
                         if (i < recipe.getInput().size())
                         {
-                            rowPane.findPaneOfTypeByID(String.format(RESOURCE, i + 1), ItemIcon.class).setItem(recipe.getInput().get(i));
+                            final ItemStack displayItem = recipe.getInput().get(i).getItemStack().copy();
```
Raycoms: do we need the copy here?

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSawmill.java
```diff
@@ -115,9 +116,9 @@ public boolean canRecipeBeAdded(final IToken<?> token)
 
             double amountOfValidBlocks = 0;
             double blocks = 0;
-            for (final ItemStack stack : storage.getInput())
+            for (final ItemStorage stack : storage.getInput())
```
Raycoms: same here

src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
```diff
@@ -52,7 +50,8 @@ public ItemStorage getNewInstance(@NotNull final ItemStack stack, final int size
     public CompoundNBT serialize(@NotNull final IFactoryController controller, @NotNull final ItemStorage storage)
     {
         final CompoundNBT compound = new CompoundNBT();
-        @NotNull final CompoundNBT stackTag = new CompoundNBT();
+        @NotNull CompoundNBT stackTag = new CompoundNBT();
+        Log.getLogger().info("Writing new ItemStorage: " + storage.getItemStack().getDisplayName().getString());
```
Raycoms: debug logging, we should probably remove

src/main/java/com/minecolonies/coremod/client/gui/WindowListRecipes.java
Mekle001: Probably not

src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
Mekle001: Oops. Missed that one

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -393,12 +393,12 @@ public void improveRecipe(IRecipeStorage recipe, int count, ICitizenData citizen
...
                     reducedItem = input.getItemStack();
                     reducedItem.setCount(input.getAmount() - 1);
-                    newRecipe.add(reducedItem);
+                    newRecipe.add(new ItemStorage(reducedItem));
                     didReduction = true;
                 }
                 else
                 {
-                    newRecipe.add(input.getItemStack());
+                    newRecipe.add(new ItemStorage(input.getItemStack()));
```
uecasm: Why does this take an ItemStorage, convert it to an ItemStack, then back to an ItemStorage?  Shouldn't you just be copying the ItemStorage with its existing settings intact?

src/api/java/com/minecolonies/api/util/constant/NbtTagConstants.java
```diff
@@ -562,6 +562,11 @@
     public static final String TAG_GUARD_NEARBY  = "guardnearby";
     public static final String TAG_RESPAWN_POS   = "nextrespawnpos";
 
+    /**
+     * Compound tag for the stack.
+     */
+    public static final String TAG_STACK = "stack_store";
```
uecasm: This is a different string from what it's replacing (which was just "stack")... is that going to cause persistence data loss?  Or is it only used in non-persistent contexts?

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
Raycoms: Yeah true, this should probably make a real copy

uecasm: There might be other places too (I didn't look very hard); probably should be suspicious of anywhere that uses `getItemStack` to ensure it's not doing this sort of pattern.

Mekle001: This is likely the only place this is happening. We don't have a copy on ItemStorage today. That having been said, I'll add one, since future changes will require it.

src/api/java/com/minecolonies/api/util/constant/NbtTagConstants.java
Mekle001: It was never used before this change. ItemStorageFactory was effectively dead code before this change, as it wasn't referenced anywhere other than to register it, as far as I could find, and we don't persist ItemStorage anywhere that I can find. Racks, the most common usage of them store ItemStacks, and wrap them in ItemStorage for indexing purposes.

Mekle001: Looking at it, I'm going to revert it back for safety, as the reason this changed doesn't actually exist in the final form of the code.

Raycoms:
:shipit:

marchermans (approved):

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -251,4 +251,25 @@ public int getRemainingDurablityValue()
...
+    public boolean isEmpty()
+    {
+        return ItemStackUtils.isEmpty(stack) || amount <= 0;
+    }
+
+    /**
+     * Make a copy of the ItemStorage
+     * @return a copy
+     */
+    public ItemStorage copy()
```
someaddons, line 269: Could we add a .toImmutable too? should be useful

someaddons (approved):
I like it, a good step into the right direction

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
Mekle001: I'll look at that in the next round of changes.

PR #6870, opened 2021-04-05
Improve possible server stall from large config values on colony creation

someaddons:
 Improve colony creation range checks to no longer load a lot of chunks
 Improve error output when colony creation fails

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/ColonyManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/ColonyManager.java
@@ -298,7 +298,7 @@ public final class ColonyManager implements IColonyManager
     }
 
     @Override
-    public boolean isTooCloseToColony(@NotNull final World w, @NotNull final BlockPos pos)
+    public boolean isFarEnoughFromColonies(@NotNull final World w, @NotNull final BlockPos pos)
     {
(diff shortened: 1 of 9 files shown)


src/main/java/com/minecolonies/coremod/network/messages/client/CreateColonyMessage.java
```diff
@@ -118,8 +118,11 @@ else if (spawnDistance > MineColonies.getConfig().getServer().maxDistanceFromWor
             }
         }
 
-        if (colony == null || !IColonyManager.getInstance().isTooCloseToColony(world, townHall))
+        if (colony != null && IColonyManager.getInstance().isTooCloseToColony(world, townHall))
         {
+            LanguageHandler.sendPlayerMessage(sender, "com.minecolonies.coremod.gui.colony.denied.tooclose", colony.getName());
+        }
```
Raycoms, line 125: no return in this?

src/main/java/com/minecolonies/coremod/colony/ColonyManager.java
```diff
@@ -303,9 +300,17 @@ public IColony getColonyByPosFromDim(final RegistryKey<World> registryKey, @NotN
     @Override
     public boolean isTooCloseToColony(@NotNull final World w, @NotNull final BlockPos pos)
```
Raycoms: this method name is confusing, we should maybe make this return true, remove the ! in the lower one, and then adjust the handling ones too, while on it xD

PR #6890, opened 2021-04-09
Research UI Updates

gattsuru:
- Fixes play/cancel/check icons on the research UIs not accepting clicks.
- Fixes the time labels for the tree gui's top showing incorrectly calculated hour estimates.
- Extends the research name label, at least enough for Captain of the Guard and What is this Speed.
- Changes text color for unfulfilled requirements and warning text to a lighter salmon pink instead of the dark red.
- Overrides hovertext on research requirement item icons, to more precisely describe the requirement and change textcolor to salmon/aqua depending on whether it is fulfilled/unfulfilled.

Right now, this includes building and research requirements on the main tooltip for the research, but item requirements only show on mouseover for their specific icons.  I'm not sure that's the right compromise between clarity and too-much-text; it's pretty easy to change either direction, but a quick attempt at including everything made multi-item researches look a little much.
image

Review please


src/main/java/com/minecolonies/coremod/client/gui/WindowResearchTree.java
```diff
@@ -349,11 +349,12 @@ private void drawTreeBackground(final ZoomDragView view, final int maxHeight)
             final Text timeLabel = new Text();
             timeLabel.setSize(TIME_WIDTH, TIME_HEIGHT);
             timeLabel.setPosition((i - 1) * (GRADIENT_WIDTH + X_SPACING) + GRADIENT_WIDTH / 2 - TIME_WIDTH / 4, TIMELABEL_Y_POSITION);
+            final double hoursTime = ((Math.pow(2, i - 1) * IGlobalResearchTree.getInstance().getBranchData(branch).getBaseTime()) * BASE_RESEARCH_TIME) / (BASE_RESEARCH_TIME * 2);
```
Raycoms: I feel like we should have a utility method for this somewhere

gattsuru: `(Math.pow(2, i - 1) * IGlobalResearchTree.getInstance().getBranchData(branch).getBaseTime()) * BASE_RESEARCH_TIME)` could probably get wrapped into one, either in `BranchData.getBaseTime` or elsewhere.  It gives a weird unit, though (effectively 1/288 of an hour).  Not sure if there's a better place for it.

Raycoms: getBranchData(branch).getHoursTime(i), maybe?

gattsuru: I can add that as well, but this is the only spot that actually checks for hours directly.  The in-progress research estimates all do their math through immediate comparisons to current progress and then convert it into hours, because doing the division first seems like it'd be asking for IEEE745 weirdness.

Raycoms: Hmm, true, let's keep this here then for now. Just keeping in mind that when we eventually need this anywhere else then we make a method for it.

Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/client/gui/WindowResearchTree.java
gattsuru: If it's likely to come up, or even be useful to others interacting through the API, it's pretty trivial to put in place.  And it definitely does help to have all of these math calls moved into one spot, especially since we may want to have alternative calculations some time in the future (if only for external data packs).

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1353,6 +1353,7 @@
   "com.minecolonies.coremod.research.limit.onemaxperbranch": "Only one max-level research can be unlocked per branch.",
   "com.minecolonies.coremod.research.limit.immutable": "This research cannot be undone.",
   "com.minecolonies.coremod.research.limit.maxeffect": "MAX",
+  "com.minecolonies.coremod.research.limit.requirement": "Requirement : %s %s",
```
ravenbuilder934, line 1356: Why a space before the colon?

gattsuru: Mostly just that it kinda bled together with the tips of the t, but not attached to it in the slightest.

someaddons: maybe a different word than requirement fits better, as we have
Requires: Enchanter level 3 as requirement already. Maybe `Research Cost:` or just `Cost:`

someaddons (approved):

PR #6910, opened 2021-04-11
Remove optional for modules, and add optional optional method

Raycoms:
Instead of using flat optionals for the modules. This allows to do both to reduce code complexity and improve performance.

-
-
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/buildings/views/IBuildingView.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/views/IBuildingView.java
@@ -216,7 +216,7 @@ public interface IBuildingView extends IRequester
      * Get the first module view matching the class.
      * @param clazz the class to match.
      * @param <T> the type of module returned.
-     * @return the module or an empty optional.
+     * @return the module.
      */
     @NotNull
     <T extends IBuildingModuleView> T getModuleView(Class<T> clazz);
@@ -226,7 +226,7 @@ public interface IBuildingView extends IRequester
      * @param clazz the class of the module.
      * @param modulePredicate the predicate to match.
      * @param <T> the optional type.
-     * @return an optional of the matching module (could be empty).
+     * @return the module.
      */
     @NotNull
     <T extends IBuildingModuleView> T getModuleViewMatching(Class<T> clazz, Predicate<? super T> modulePredicate);
--- a/src/api/java/com/minecolonies/api/colony/buildings/views/IBuildingView.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/views/IBuildingView.java
@@ -219,7 +219,7 @@ public interface IBuildingView extends IRequester
      * @return the module or an empty optional.
      */
     @NotNull
-    <T extends IBuildingModuleView> Optional<T> getModuleView(Class<T> clazz);
+    <T extends IBuildingModuleView> T getModuleView(Class<T> clazz);
 
     /**
      * Get the module view matching a certain predicate.
@@ -229,7 +229,7 @@ public interface IBuildingView extends IRequester
      * @return an optional of the matching module (could be empty).
      */
     @NotNull
-    <T extends IBuildingModuleView> Optional<T> getModuleViewMatching(Class<T> clazz, Predicate<? super T> modulePredicate);
+    <T extends IBuildingModuleView> T getModuleViewMatching(Class<T> clazz, Predicate<? super T> modulePredicate);
 
     /**
      * Get a list of all modules matching a specific class.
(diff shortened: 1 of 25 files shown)


Raycoms:
> doesn't this conflict with the @NotNull annotation?

Yes and No, this should only ever be null, if something went wrong badly. So, this should then crash.

someaddons (dismissed):

src/api/java/com/minecolonies/api/colony/buildings/views/IBuildingView.java
```diff
@@ -219,7 +219,7 @@
      * @return the module or an empty optional.
```
uecasm: Javadoc no longer correct.

src/api/java/com/minecolonies/api/colony/buildings/views/IBuildingView.java
```diff
@@ -229,7 +229,7 @@
      * @return an optional of the matching module (could be empty).
```
uecasm: And here.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -159,12 +157,26 @@ protected AbstractBuilding(@NotNull final IColony colony, final BlockPos pos)
     @Override
     public boolean hasModule(final Class<? extends IBuildingModule> clazz)
     {
-        return getFirstModuleOccurance(clazz).isPresent();
+        return getFirstModuleOccurance(clazz) != null;
```
uecasm: This is bad style on a `@NotNull` method.  It works, but it will annoy the code analysis.

src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAISleep.java
```diff
@@ -226,7 +226,7 @@ private void findBedAndTryToSleep()
                 List<BlockPos> bedList = new ArrayList<>();
                 if (hut.hasModule(BedHandlingModule.class))
                 {
-                    hut.getFirstModuleOccurance(BedHandlingModule.class).ifPresent(module -> bedList.addAll(module.getRegisteredBlocks()));
+                    hut.getFirstOptionalModuleOccurance(BedHandlingModule.class).ifPresent(module -> bedList.addAll(module.getRegisteredBlocks()));
                 }
```
uecasm: It's not necessary to have a `hasModule` check when using `ifPresent`.

src/main/java/com/minecolonies/coremod/network/messages/server/colony/building/TriggerSettingMessage.java
```diff
@@ -70,7 +70,7 @@ public void onExecute(
     {
         if (building.hasModule(SettingsModule.class))
         {
-            building.getFirstModuleOccurance(SettingsModule.class).ifPresent(m -> m.with(new SettingKey(this.value.getClass(), this.key), this.value));
+            building.getFirstModuleOccurance(SettingsModule.class).with(new SettingKey<>(this.value.getClass(), this.key), this.value);
         }
```
uecasm: Whereas this one used to be redundant and now isn't -- but it would be better to write it using `ifPresent` instead of `hasModule`.

src/main/java/com/minecolonies/coremod/entity/ai/minimal/EntityAISleep.java
uecasm: (In fact most places that use `hasModule` probably shouldn't.)

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
Raycoms: Yeah, I know, but it's more lightweight than wrapping it before that in an optional to do an isPresent check. So I'm fine to live with this.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingComposter.java
```diff
@@ -77,7 +77,7 @@ public BuildingComposter(@NotNull final IColony c, final BlockPos l)
     {
         super(c, l);
 
-        keepX.put((stack) -> this.hasModule(ItemListModule.class) && this.getModuleMatching(ItemListModule.class, m -> m.getId().equals(COMPOSTABLE_LIST)).isItemInList(new ItemStorage(stack)), new Tuple<>(Integer.MAX_VALUE, true));
+        keepX.put((stack) -> this.getModuleMatching(ItemListModule.class, m -> m.getId().equals(COMPOSTABLE_LIST)).isItemInList(new ItemStorage(stack)), new Tuple<>(Integer.MAX_VALUE, true));
```
uecasm, line 80: Just removing the `hasModule` is not equivalent, unless you're certain it always will have one.  (On pain of null reference.)

someaddons: we are indeed certain that we have one, as this building is defined with one

marchermans:
Since this adds a lot of bad code style:

I would like to see numbers on all the claims with regards to performance in this PR.

PR #6943, opened 2021-04-17
Unlock item tooltip

### LINKED ISSUE #6900: Annoyances I found with the sifter
### Prerequisites 

- [ ] I am running the latest **alpha** version of MineColonies and Structurize for my Minecraft version.

### Context 

- Minecraft Version:
- MineColonies Version: 

minecolonies-0.14.77-BETA-universal.jar
--

- Structurize Version:

## Expected behavior
https://wiki.minecolonies.ldtteam.com/source/workers/sifter
Would tell me how do meshes for sifters can now be obtained, and which citizens craft them

## Actual behavior

They don't, and I can't figure out how to make them as JEI doesn't show their recipes.
Also when I apply the diamond mesh into the sifter via creative, a bunch of announcements appeared in chat telling me my researchers unlocked the research for all the meshes, and after I re-checked, there isn't any research for them at all.

1. 
2. 
3. 

### Logs

- latest.log: 
- crashlog: 

### Notes

---
## Viewers

gattsuru:
Closes #6900

- Adds a "show-tooltip" ~~and "show-jei"~~ property to crafterrecipes.
- Transfers crafterrecipes ~~with either of these traits set to true~~ to remote client, ~~and stores them inside a separate recipeOutputMap.  Documents in code which recipe map is populated based on sidedness. ~~  This should help with some of the JEI plugin work, ~~though it will require some refactoring to move to the new recipeOutputMap.  But hesitant to load a filtered list into the single main recipeMap, since it's really easy to end up with inconsistencies.~~
- Properly sets CustomRecipeManager to clear at start of data reload listener (eg, no longer having recipe data from previous worlds, or before a data pack disable).
- Adds a client event handler to show tooltips for all items with the "show-tooltip" property, and to handle research unlocks for hut blocks.
- Removes addInformation-override based tooltips for research on AbstractHutBlocks, and for crafter origin from baked goods and gates.  Item-specific matters like effects and gate behavior remain in the item-specific addInformation call, but this should be more manageable for customized environments.
- Cleans up some research-related code and comments.
- Provides a way to collect the research (or researches) tied to a given effect.  _This may be null, if no research with the effect is loaded._
- Provides a fast way to check if an item, ItemStack, or ItemStorage can be made with any crafterrecipe.

I'm not perfectly happy with the UI result, so feedback on that would be appreciated.  Current tooltip behavior is:
- Red text when in a colony that does not have a building of the required level, or the research, or when outside a colony when one's owned colony does not have the requirements.
- Aqua text when within a colony that does have the research or required building of the required level, or when outside of a colony but when your owned colony has the research or required building of the required level.  I'm not sure if it makes sense to have the research ones disappear when they're fulfilled, instead.
- Gray text for simple information, such as where the recipe doesn't require research or a specific building level.
~~Similarly, not sure if getting an icolony based in proximity is the right choice, or if it should go by ownered colony, or only when within the colony borders.~~ Updated re Raycoms.

Review please


src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1748,7 +1747,9 @@
   "com.minecolonies.coremod.item.milky.bread.gui": "Removes potion effects",
   "com.minecolonies.coremod.item.golden.bread.gui": "Instantly heals 2 hearts",
   "com.minecolonies.coremod.item.chorus.bread.gui": "Teleports you to the surface",
-  "com.minecolonies.coremod.item.bread.available.gui": "Only available from a MineColonies Bakery",
+  "com.minecolonies.coremod.item.available.gui": "Crafted by a MineColonies %s",
+  "com.minecolonies.coremod.item.buildlevel.gui": "Crafted by a MineColonies %s level %s or higher",
+  "com.minecolonies.coremod.item.requiresresearch.gui": "Requires Research : %s",
```
ravenbuilder934: Why the spaces around the colon?

uecasm:
In what circumstance do you imagine *not* setting `show-jei` to true?  It is intended that JEI be able to view all recipes (and even things that aren't formally recipes).

(i.e. it should probably be true by default, so that you modify the recipes you *don't* want visible rather than the reverse, since showing them should be the norm.)

Raycoms:
instead of distance I'd make two checks only
a) Within a colony?
b) No? Owns a colony? No? -> Write the requirement as if it wasn't met

gattsuru:
> In what circumstance do you imagine _not_ setting `show-jei` to true? It is intended that JEI be able to view all recipes (and even things that aren't formally recipes).

Central case, 470KB of structurize recipes that have a normal Crafting RecipeManager equivalent.  It's possible to parse that through the server for the MineColonies-internal stuff, but it's a little difficult to distinguish it from cases where we wouldn't want  to strip things out (eg, using altOutputs as a sort of currency or to gate unlocks, rather than the frame-specific behavior it's used for internally).

Beyond that, some crafterrecipes are set with invalid or unreachable targets (such as `"research-id" : "dummy-research"` or `"min-building-level" : 99`) specifically to remove them from huts that had learned the recipe beforehand, when replacing it with a different recipe.  These shouldn't be displayed in JEI; as they're not valid recipes.  But while they can be parsed on the receiving side (at least for `building-levels`), it's not really coherent to do so.

uecasm:
> 470KB of structurize recipes that have a normal Crafting RecipeManager equivalent

Yes, but the custom recipe is "better" because it collapses to a single recipe with alternate outputs, so doesn't take up a billion JEI pages.  There's a reason that the JEI branch hides the vanilla recipes rather than the custom recipes.

Also, if it's just the sawmill structurize recipes, there's a datagen for that so you can trivially add properties there if needed.  Though as above I'm not convinced that's the right direction.

> Beyond that, some crafterrecipes are set with invalid or unreachable targets (such as "research-id" : "dummy-research" or "min-building-level" : 99) specifically to remove them from huts that had learned the recipe beforehand, when replacing it with a different recipe.

Actually those recipes no longer exist at all; Mekle already made some changes that remove deleted recipes without needing to leave behind deliberately-invalid ones.  (The code to deal with them only still exists in the JEI PR because it was written before those changes were introduced and I haven't updated it yet since I've been splitting it into smaller PRs instead of the monolith.)

gattsuru:
> Yes, but the custom recipe is "better" because it collapses to a single recipe with alternate outputs, so doesn't take up a billion JEI pages. There's a reason that the JEI branch hides the vanilla recipes rather than the custom recipes.

Ah, ok.  I _was_ misreading those `removeIf` lines, then; I thought it was throwing out the custom recipe version, not the vanilla table one.  That's an issue.

> Actually those recipes no longer exist at all; Mekle already made some changes that remove deleted recipes without needing to leave behind deliberately-invalid ones.

Sifter ones still exist, but if that's the case, yeah, can get rid of a lot of the cruft.  Need to think about ways to reduce the packet transmission size, still, but that may just be something clients have to deal with.

uecasm:
> Sifter ones still exist

That's probably my bad, the PR that updated these recipes was part of what prompted Mekle to improve the deleted-recipe support, so it all happened at around the same time.

src/main/resources/assets/minecolonies/lang/en_us.json
gattsuru: Mostly habit.  Will remove in the next commit.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipe.java
```diff
@@ -426,20 +412,22 @@ public CustomRecipe(final String crafter, final List<ItemStorage> inputs, final
         this.result = primaryOutput;
         this.secondary = secondaryOutput;
         this.altOutputs = altOutputs;
-        this.lootTable = lootTable;
-        if(!researchReq.getPath().isEmpty())
+        if(lootTable != null)
+        {
+            this.lootTable = lootTable;
+        }
```
uecasm: Why the null checks?

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
```diff
@@ -14,25 +21,31 @@
     /**
      * The internal static instance of the singleton
      */
-    private static CustomRecipeManager instance = new CustomRecipeManager();
+    final private static CustomRecipeManager instance = new CustomRecipeManager();
```
uecasm: private static final

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
```diff
@@ -14,25 +21,31 @@
...
-     * The map of loaded recipes
+     * The map of loaded recipes by crafter.
      */
-    private HashMap<String, Map<ResourceLocation, CustomRecipe>> recipeMap = new HashMap<>();
+    final private HashMap<String, Map<ResourceLocation, CustomRecipe>> recipeMap = new HashMap<>();
+
+    /**
+     * The map of all loaded recipes by output.
+     */
+    final private HashMap<Item, List<CustomRecipe>> recipeOutputMap = new HashMap<>();
```
uecasm: private final

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -1211,6 +1211,34 @@ public boolean hasWarehouse()
...
+                    if (building.getBuildingLevel() >= level)
+                    {
+                        return true;
+                    }
+                }
+                else
+                {
+                    sum += building.getBuildingLevel();
+                }
+                if(sum >= level)
```
Raycoms: this can be within the else as well after the sum+=

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeFactory.java
```diff
@@ -0,0 +1,315 @@
...
+        }
+        if(recipe.getLootTable() != null)
+        {
+            compound.putString(RECIPE_LOOTTABLE_PROP, recipe.getLootTable().toString());
+        }
+        compound.putInt(RECIPE_BUILDING_MIN_LEVEL_PROP, recipe.getMinBuildingLevel());
+        compound.putInt(RECIPE_BUILDING_MAX_LEVEL_PROP, recipe.getMaxBuildingLevel());
+        compound.putBoolean(RECIPE_MUST_EXIST, recipe.getMustExist());
+        compound.putBoolean(RECIPE_SHOW_TOOLTIP, recipe.getShowTooltip());
+        // RecipeStorage has its own factory class, but its output is larger than necessary (~1KB per recipe, vs ~800b for below, averaged over all minecolonies recipes)
```
Raycoms: Can't we adjust the serialize to packetBuffer in the recipeStorage factory instead even?

src/main/java/com/minecolonies/coremod/event/ClientEventHandler.java
```diff
@@ -160,6 +172,134 @@ public static void onPlaySoundEvent(final PlaySoundEvent event)
...
+     */
+    @SubscribeEvent
+    public static void onItemTooltipEvent(final ItemTooltipEvent event)
+    {
+        // Vanilla recipe books populate tooltips once before the player exists on remote clients, some other cases.
+        if(event.getPlayer() == null)
+        {
+            return;
+        }
+        IColony colony = IMinecoloniesAPI.getInstance().getColonyManager().getClosestIColony(event.getPlayer().world, event.getPlayer().getPosition());
```
Raycoms: We should be able to ask the chunk the player is in, for the colony id.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeFactory.java
gattsuru: A lot of the differences between this version and the RecipeStorageFactory one are just dropped fields.  The client doesn't care about the IToken or grid size for tooltip drawing, but they may be relevant for the AddRemoveRecipeMessage (or may not be; I don't understand a lot of what RecipeStorage does on the server side).

The byte-based RecipeStorage serializer is a lot more efficient, so it doesn't really matter much; I can switch both, use the byte-based CustomRecipe serialization for the big message, and just warn that the CompoundNBT-based one will be pretty large if anyone ever uses it -- it's genuinely not that big a deal for single recipes, it's only this bulk update that could potentially burst the CompoundNBT limit. 

Using getRecipeStorage _will_ populate all of the custom recipe manager recipe storages on the server, though.  I'm not sure if it's that big a deal, but there's specific code to avoid having that cache being generated until the recipe is called for.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeFactory.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+        if(recipe.getExcludedResearchId() != null)
+        {
+            compound.putString(RECIPE_EXCLUDED_RESEARCHID_PROP, recipe.getExcludedResearchId().toString());
+        }
+        compound.putInt(RECIPE_BUILDING_MIN_LEVEL_PROP, recipe.getMinBuildingLevel());
+        compound.putInt(RECIPE_BUILDING_MAX_LEVEL_PROP, recipe.getMaxBuildingLevel());
+        compound.putBoolean(RECIPE_MUST_EXIST, recipe.getMustExist());
+        compound.putBoolean(RECIPE_SHOW_TOOLTIP, recipe.getShowTooltip());
+        compound.put(RECIPE_TYPE_RECIPE, controller.serialize(recipe.getRecipeStorage()));
```
Mekle001: Why do we need to serialize this at all?

Mekle001: I'd actually go so far as to say you SHOULDN'T serialize this. It should be built/cached when somebody calls getRecipeStorage off the deserialized customRecipe downstream.

Mekle001: Ok, I found why. You're not serializing the core collections, you're depending on recipeStorage to do that. I think that's a bad idea, because it creates the possibility of a de-synced central ID by serializing a cached item which contains an ID that is not included in hashcode/equals.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
```diff
@@ -52,6 +65,20 @@ public void addRecipe(@NotNull final CustomRecipe recipe)
         }
 
         recipeMap.get(recipe.getCrafter()).put(recipe.getRecipeId(), recipe);
+
+        if (!recipeOutputMap.containsKey(recipe.getPrimaryOutput().getItem()))
+        {
+            recipeOutputMap.put(recipe.getPrimaryOutput().getItem(), new ArrayList<>());
```
Mekle001, line 71: How is this going to handle when there are multiple recipes with the same output? We support this today in taught recipes for say charcoal, where you can teach all the logs. If somebody were to add those as a custom recipe, you have a potential problem.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
```diff
@@ -79,13 +115,21 @@ public void removeRecipe(@NotNull final ResourceLocation toRemove)
...
      */
     public Map<String, Map<ResourceLocation, CustomRecipe>> getAllRecipes()
     {
         return recipeMap;
     }
 
+    /**
+     * Get the custom recipes for an item, or null if no matching recipe exists.
+     */
+    public List<CustomRecipe> getRecipeByOutput(final Item item)
```
Mekle001, line 132: This isn't enough, this should support an ItemStorage as the search criteria, so that it can find the recipe with the itemstack that matches.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeFactory.java
gattsuru: I don't know that we do; my original thought process was to serialize the individual components, leaving the RecipeStorages alone; see discussion here (https://github.com/ldtteam/minecolonies/pull/6943#discussion_r615279963). I'm not sure what the consequences of even merely calling `getRecipeStorage` are, and possible desync is a situation (though wasn't sure about the consequences of throwing away old itokens during a data pack reload on a server).

The byte-focused serialization/deserialization operates in that way, and is currently the actually called code; I can switch the nbt-variant back.

Mekle001: Yeah, make the nbt-variant match the byte-focused, and please remove the constructor that takes a recipestorage.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeManager.java
```diff
@@ -124,12 +126,70 @@ public void reset()
...
+            if(recipe.getPrimaryOutput().equals(itemStorage) || recipe.getAltOutputs().contains(itemStorage))
+            {
+                returnList.add(recipe);
+            }
+        }
+        if(returnList.size() > 0)
+        {
+            return returnList;
+        }
+        return null;
```
Mekle001: Why not just return an empty list here? Why force both a null and a isempty check instead of just the isempty check?

gattsuru:
> Why not just return an empty list here? Why force both a null and a isempty check instead of just the isempty check?

The Item variant of the method directly accesses the hashmap, which returns a null if there are no items with matching custom recipes. So it's either a null check in that case, or the isEmpty check here. Can swap if the empty list result is preferred as less likely to cause NPEs.

Mekle001 (approved):

Mekle001:
Ok, I like where this ended up. :shipit:

PR #6953, opened 2021-04-18
Added stone variants to pathblocks

anonymous563:
Added stone, andesite, diorite, and granite, polished versions, and slabs and stairs for each.
https://discord.com/channels/139070364159311872/240841248750043137/833368500754972684

Review please and thanks!


src/main/resources/data/minecolonies/tags/blocks/pathblocks.json
```diff
@@ -6,6 +6,27 @@
     "minecraft:stone_brick_stairs",
     "minecraft:stone_brick_slab",
     "minecraft:mossy_stone_brick_slab",
+    "minecraft:stone",
```
ravenbuilder934: I'd say not stone, in case there's exposed stone as part of worldgen

ravenbuilder934: same for andesite, diorite, and granite (the non-polished variants)

ravenbuilder934 (dismissed):

marchermans (changes requested):
This should be done by including the stone tag

ravenbuilder934:
there is no `stone` tag according to the MC wiki, I don't know what you mean

uecasm:
There's a `forge/stone` block tag.  But you shouldn't use it since it includes the unpolished blocks.

Mekle001 (dismissed):

src/main/resources/data/minecolonies/tags/blocks/pathblocks.json
```diff
@@ -6,6 +6,17 @@
     "minecraft:stone_brick_stairs",
     "minecraft:stone_brick_slab",
     "minecraft:mossy_stone_brick_slab",
+    "minecraft:stone_slab",
+    "minecraft:stone_stairs",
```
someaddons: I think we should only have the stone brick variants, not the plain stone

Raycoms: Yeah, the others are fine, let's remove those here then :shipit:

PR #6985, opened 2021-04-24
Expands turnoffExplosionsInColonies Setting

### LINKED ISSUE #391: Make creepers to still explode inside the colony but doesn't grief
### Prerequisites 

### Context 

- Minecraft Version:
- MineColonies Version:

With the default explosion disable creepers are basically free loots, even the mutant creepers explosion is also destroyed, rendering them helpless.  

Give back creeper's explosion, but the only different is that they don't destroy blocks, or grief terrain.
Therefore making a deadly enemy again.

### Notes or related things
I am expecting this to get a "Good first issue"

---
## Viewers

gattsuru:
Closes https://github.com/ldtteam/minecolonies-features/issues/391

- Makes turnOffExplosionsInColonies into an enum, ~~where zero completely deadens explosions, one allows explosions to damage entities but not the environment, and two allows explosions as they'd normally occur outside of colony borders.  (see discussion [here](https://discord.com/channels/139070364159311872/240841248750043137/834231493076582401) and above) ~~ where DAMAGE_EVERYTHING is vanilla behavior, DAMAGE_ENTITIES is basically a colony-specific no-griefing gamerule, DAMAGE_PLAYERS only allows harm to players and mobs but not citizens or blocks or animals, and DAMAGE_NOTHING stops the explosion from firing.
- Configuration files will automatically reset back to a setting of DAMAGE_NOTHING (ie, completely disabled explosions, ie how things previously were).
- I left the Explosion.Start handler in-place, but as far as I can tell, this almost entirely duplicates what's going on in Explosion.Detonate for totally stopping explosions, but with less fidelity (maybe in exchange for reducing the cost of ray projections?).  It _may make sense to remove it_, as it's why the explosion 'dust' swirl doesn't go off, and its logic chain doesn't match the description of turnOffExplosionsInColonies applying "regardless of colony protection".  But I'm not sure if there are outside cases where it's relevant.
- ~~This is _probably_ going to be murderous to knights.  If that is not desirable, can instead tweak the removeEntities predicate to still remove colonists from the list in mode 1~~
- Added a knight ignore damage that fires for explosive damage, if the knight has and can use a shield.

Review please


Raycoms:
I guess we should add sth to knights then before merging this, such that nights with shields block explosions.

src/api/java/com/minecolonies/api/configuration/ServerConfiguration.java
```diff
@@ -121,7 +121,7 @@
      *  ----------------------------------------------------------------------------- */
 
     public final ForgeConfigSpec.BooleanValue                        enableColonyProtection;
-    public final ForgeConfigSpec.BooleanValue                        turnOffExplosionsInColonies;
+    public final ForgeConfigSpec.IntValue                            turnOffExplosionsInColonies;
```
marchermans: Don't use Integers for this.

As you specified it now, this should be an enum.

marchermans (changes requested):

src/api/java/com/minecolonies/api/configuration/ServerConfiguration.java
```diff
@@ -265,7 +265,7 @@ protected ServerConfiguration(final ForgeConfigSpec.Builder builder)
         swapToCategory(builder, "permissions");
 
         enableColonyProtection = defineBoolean(builder, "enablecolonyprotection", true);
-        turnOffExplosionsInColonies = defineBoolean(builder, "turnoffexplosionsincolonies", true);
+        turnOffExplosionsInColonies = defineInteger(builder, "turnoffexplosionsincolonies", 0, 0, 2);
```
ravenbuilder934: shouldn't the 2nd zero be a one?

gattsuru: defineInteger takes arguments builder, name, default value, minimum value, max value.  The minimum value should always be equal to or lower than the default value.  But OrionDevelopment's right that this is better handled with the defineEnum, so swapping to that.

marchermans (approved):
Yes much better.

This is way more readable and understandable then your first attempt.
Makes the config in one go a lot clearer!

Good job!

PR #7004, opened 2021-04-25
Have custom Recipes delegate reading of ItemStorage inputs to ItemStorage

Mekle001:
- Now, ItemStorage owns the interpretation of the JSON object that encodes an item in customrecipe JSONs
- This allowed me to add a matchType to the object, which informs the ItemStorage what state to set the shouldIgnoreNBT to
- This enables future extension for more complex NBT matches. 

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/crafting/ItemStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/ItemStorage.java
@@ -1,8 +1,12 @@
 package com.minecolonies.api.crafting;
 
+import com.google.gson.JsonObject;
 import com.minecolonies.api.util.ItemStackUtils;
+
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+
 import org.jetbrains.annotations.NotNull;
 
 import java.util.ArrayList;
@@ -40,6 +44,17 @@ public class ItemStorage
      * Amount of the storage.
      */
     private int amount;
+
+    /**
+     * The property name for Count, used both in inputs array and for result
+     */
+    public static final String COUNT_PROP = "count";
+
+    /**
+     * The property name for the item id in the inputs array
+     */
+    public static final String ITEM_PROP = "item";
+
 
     /**
      * Creates an instance of the storage.
@@ -100,6 +115,54 @@ public class ItemStorage
         this.shouldIgnoreDamageValue = false;
         this.shouldIgnoreNBTValue = false;
         this.amount = ItemStackUtils.getSize(stack);
+        this.creativeTabIndex = stack.getItem().getCreativeTabs().stream().filter(Objects::nonNull).map(g -> g.index).collect(Collectors.toList());
+    }
+
+    /**
+     * Creates an instance of the storage from JSON
+     * 
+     * @param jObject the JSON Object to parse
+     */
+    public ItemStorage(@NotNull final JsonObject jObject)
+    {
+        if (jObject.has(ITEM_PROP))
+        {
+            final ItemStack parsedStack = ItemStackUtils.idToItemStack(jObject.get(ITEM_PROP).getAsString());
+            if(jObject.has(COUNT_PROP))
+            {
+                parsedStack.setCount(jObject.get(COUNT_PROP).getAsInt());
+                this.amount = jObject.get(COUNT_PROP).getAsInt();
+            }
+            else
+            {
+                this.amount = parsedStack.getCount();
+            }
+            this.stack = parsedStack;
+            if(jObject.has("matchType"))
+            {
+                String matchType = jObject.get("matchType").getAsString();
+                if(matchType.equals("ignore"))
+                {
+                    this.shouldIgnoreNBTValue = true;
+                }
+                else // includes "exact"
+                {
+                    this.shouldIgnoreNBTValue = false;
+                }
+            }
+            else
+            {
+                this.shouldIgnoreNBTValue = false;
+            }
+
+            this.shouldIgnoreDamageValue= true;
+        } else {
+            this.stack = new ItemStack(Items.AIR);
+            this.amount = 0;
+            this.shouldIgnoreDamageValue = true;
+            this.shouldIgnoreNBTValue = true;
+        }
+    
         this.creativeTabIndex = stack.getItem().getCreativeTabs().stream().filter(Objects::nonNull).map(g -> g.index).collect(Collectors.toList());
     }
 
@@ -272,4 +335,13 @@ public class ItemStorage
         newInstance.setAmount(amount);
         return newInstance;
     }    
+
+    /**
+     * Get an immutable version of this item storage
+     * @return immutable wrapper
+     */
+    public ImmutableItemStorage toImmutable()
+    {
+        return new ImmutableItemStorage(this);
+    }
 }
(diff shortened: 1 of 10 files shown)


src/api/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -103,6 +109,54 @@ public ItemStorage(@NotNull final ItemStack stack)
...
+                    this.shouldIgnoreNBTValue = false;
+                }
+            }
+            else
+            {
+                this.shouldIgnoreNBTValue = false;
+            }
+
+            this.shouldIgnoreDamageValue= true;
+        } else {
```
Raycoms: formatting

src/main/java/com/minecolonies/coremod/colony/crafting/ImmutableItemStorageFactory.java
```diff
@@ -0,0 +1,77 @@
...
+import com.minecolonies.api.crafting.ImmutableItemStorage;
+import com.minecolonies.api.crafting.ItemStorage;
+import com.minecolonies.api.util.constant.TypeConstants;
+
+import org.jetbrains.annotations.NotNull;
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.PacketBuffer;
+
```
Raycoms, line 17: javadoc

Raycoms (dismissed):
Just very small comments

uecasm:
Just as a conflict heads-up, note that #7006 is also trying to mess with ItemStorage serialization...

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -103,6 +109,55 @@ public ItemStorage(@NotNull final ItemStack stack)
...
+            {
+                String matchType = jObject.get(MATCHTYPE_PROP).getAsString();
+                if(matchType.equals(MATCH_NBTIGNORE))
+                {
+                    this.shouldIgnoreNBTValue = true;
+                }
+                else // includes "exact"
+                {
+                    this.shouldIgnoreNBTValue = false;
+                }
```
uecasm, line 141: Could just assign directly.

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -103,6 +109,55 @@ public ItemStorage(@NotNull final ItemStack stack)
...
+            }
+            else
+            {
+                this.shouldIgnoreNBTValue = false;
+            }
+            this.shouldIgnoreDamageValue= true;
+        }
+        else
+        {
+            this.stack = new ItemStack(Items.AIR);
```
uecasm: ItemStack.EMPTY?

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipe.java
```diff
@@ -293,23 +240,14 @@ public static CustomRecipe parse(@NotNull final ResourceLocation recipeId, @NotN
...
-                    {
-                        final ItemStack stack = idToItemStack(ingredient.get(ITEM_PROP).getAsString());
-                        if(ingredient.has(COUNT_PROP))
-                        {
-                            stack.setCount(ingredient.get(COUNT_PROP).getAsInt());
-                        }
-                        recipe.inputs.add(new ItemStorage(stack));
-                    }
-
+                    recipe.inputs.add(new ItemStorage(ingredient));
```
uecasm: This is not equivalent; it can add empty inputs where they would previously be omitted entirely.

src/main/java/com/minecolonies/coremod/colony/requestsystem/management/manager/StandardRecipeManager.java
```diff
@@ -94,7 +94,7 @@ public void read(@NotNull final CompoundNBT compound)
         for (int i = 0; i < list.size(); i++)
         {
             IRecipeStorage recipe = StandardFactoryController.getInstance().deserialize(list.getCompound(i));
-            if (recipe != null && !recipes.containsValue(recipe))
+            if (recipe != null && !recipes.containsValue(recipe) && !recipe.getCleanedInput().isEmpty())
```
uecasm, line 97: ... which I assume is why this was added, but it seems better to solve it at the source instead.

uecasm (commented):
So this whole thing could work for the enchanter's ancient tome (although I don't think it's necessarily _better_ than the NBT-based method previously in #6791 -- it's certainly a lot more code and JSON to accomplish the same thing, albeit with more potential future flexibility).

(Just for reference, since while the code was retained, an example of the JSON didn't survive into the final form of the PR: it looked like [this](https://github.com/ldtteam/minecolonies/blob/e08595f33dde58ead8cec94defffcd2a26096cc2/src/main/generated/resources/data/minecolonies/crafterrecipes/enchanter/tome1.json).)

It doesn't address the clock/compass thing, though, which is what the item tag was intended for.

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
Mekle001: That's what I was trying to remember when I was writing this code. I'll update it.

src/main/java/com/minecolonies/coremod/colony/requestsystem/management/manager/StandardRecipeManager.java
Mekle001: This is added as a safety to deal with recovery from some edge cases already in the wild, where recipes had no inputs.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipe.java
Mekle001: This is what happens when I write this first, and the planned semantics changes and I forget to come back. :)

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
Mekle001: We're using this. for all of the constructor assigns,  and while I could assign the result of the .equals, this is the structure it will become when we add additional options for matchtype.

Mekle001:
For the clock/compass thing, we have several options, including adding recipes that convert to the correct nbt settings, so a clock input that ignores nbt, and an output that has no nbt. 

Alternately, we can do custom recipes for the items that use clock/compass that were 'mustexist' recipes, and had the appropriate input settings.

marchermans (approved):

uecasm:
Ok, so I've tried this out with the enchanter PR, and it doesn't actually work, for a few reasons:
- the `ItemStorageFactory` doesn't serialize/deserialize the `shouldIgnoreNBTValue` field, so it gets lost on world load/save (bad) and on server/client sync (possibly less bad, but might cause surprises).
- even if the above is fixed, the value gets lost by the `StandardRecipeManager` because it considers the recipe-without-ignore to be a match to the recipe-with-ignore, and keeps the former rather than the latter.  I was less sure of the correct way to fix this as `ItemStorage`'s equals/hashCode implementation is a bit peculiar and is likely contributing to the problem.

If you want an example to play with, I can push the (not working) update to the enchanter PR (currently I haven't pushed it, so it's still the working-but-outdated version).

Mekle001:
@uecasm Both of those are actually relatively easy fixes. Let me work on it today, and hopefully have a PR up this evening that you can try.

PR #7006, opened 2021-04-26
ItemStorage fixes, Tooltips, Beekeeper_2 Research

### LINKED ISSUE #285: Apiary Honeybottle and Comb simultaneously
### Prerequisites 

- [x ] I am running the latest **alpha** version of MineColonies for my Minecraft version.
- [x ] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure that my feature does not already exist.
- [ x] I made sure that **this feature request is not a duplicate of any existing feature request.**

### Context 

- Minecraft Version: 1.16.4
- MineColonies Version: minecolonies-0.13.602-ALPHA-universal

One Apiary is enough to make all of the honeybottles and combs you need. 
Problem is that you have to switch it constantly.

## Feature description

A research like the "double trouble" (2 plants for plantation), but for the Apiary, so the beekeeper will use half the hives for bottles and the other half for combs would be nice.

### Notes or related things

---
## Viewers

gattsuru:
Closes https://github.com/ldtteam/minecolonies-features/issues/285

- Fixes serialization issue in ItemStorageFactory causing Research Costs to not display properly.
- Adds tooltips to the Town Hall and Colonist Happiness windows, describing the various Happiness modifiers.
- Adds a maximum amount of stored additionalChildTime, currently set to the amount that will allow one child to be instantly-grown. Handles a case where the colony would store up enough additionalChildTime that every child past that point would be instantly-grown, only really occurring on servers where a player was leaving themselves logged in for long periods of time away from the colony.
- Adds a research allowing Beekeepers to collect both combs and bottles, which alternates between the two.  Various changes to the AI and message structure to support that.  Sorry for stealing the assigned topic, Raven, but owed you given how long the Research system update took me.  I'm not attached to the name or subtitle, but a Winnie the Pooh reference or two seems on-point.

Review please


src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
```diff
@@ -80,6 +80,8 @@ public ItemStorage deserialize(IFactoryController controller, PacketBuffer buffe
     {
         final ItemStack stack = buffer.readItemStack();
         final int size = buffer.readVarInt();
+        // PacketBuffer.readItemStack uses a byte internally for item count, and this is limited (to -128 - 127). Override with the ItemStorage value to keep the two values synced, and support larger stacks.
+        stack.setCount(size);
         return this.getNewInstance(stack, size);
```
Raycoms: Do we need to set this here to the stack? I thought having it in the storage should be good enough?

gattsuru: I don't know.

The short-term problem is that any ItemStorage containing an ItemStack with a number above 127 will be turned into a negative number during serialization sent, and then post-deserialization any ItemStack.get() or ItemStorage.get() will end up returning an Items.AIR object due to how ItemStacks work internally.

The obvious answer is Don't Do That, and never set the count for an ItemStack that's going to get wrapped into an ItemStorage, and that's certainly a viable option for the research loader and packets.  If there's semantic meaning toward ItemStorage.ItemStack.getCount returning 1, it's definitely something I can do instead.

It just seems a landmine.  eg, [this](https://github.com/ldtteam/minecolonies/pull/7004/files#diff-6c39e6841d370b2d36f8846b03058feebe7f2610c35662c0294b490b3e6a69fa) code could potentially trigger the same problem.

uecasm: The Right Way™ is to assume (and perhaps even ensure) that the ItemStack of an ItemStorage always has a count of 1 (or an otherwise-don't-care number between 1 and 64, anyway).  This does require that anything using ItemStorage.getItemStack() is aware of this and knows not to use the count of the stack for any purpose.  (In general, any code that calls that should be treated as suspicious until proven otherwise, rather than the reverse; since at best it represents a lossy conversion.)

But this is the sort of thing that most storage mods do -- the ItemStack is just a representative item(+nbt) for display/comparison/serialization purposes and the actual count is stored elsewhere.

As such, at this end it should probably just not set the stack's count at all -- it's the other end that should ensure it's never set to something silly.

gattsuru: You do have to ensure it: once an ItemStack's got a negative count, pretty much anything that uses getItem() or isEmpty(), including isItemStackEqual (isItemStacksEqual will work, but give false positives) and getDisplayName will start acting up.

At least against accidental behaviors, just changing the ItemStorage constructors will probably be enough for it to never show up, but that seems more Wrong than fixing it here; it's genuinely a serialization-specific issue.

And it's quite possible that the 'setting' side may be outside of our hands, in at least some cases; not only are ItemStorage and ItemStorageFactory an API-side class, a number of its constructors are set to pull the quantity from the input ItemStack.  Or if, worst case, someone decides to make another Increase Max Stack Size mod.

uecasm: It's illegal to set the count of an ItemStack > 64 (or more correctly, > the maxStackSize of the specific item) in the first place.  The real bug is anywhere that sets that.  Assuming that all such bugs are fixed, then there is no need for the serialization code to worry about it, since it's an invalid precondition.

If you really want to write some defensive code, then the right way around is to make the serializer always set the count to 1 before serializing, not to make the deserializer set the stack to a potentially illegal state (which is what this change does).  But (absent other bugs), even that should not be needed.

It is legal for the ItemStorage constructors to pull their counts from the input ItemStack -- the implied precondition is that this is an inventory stack that could therefore never have count > 64 anyway.  Again, for defensive reasons these *could* store only a stack with count 1 (provided they do not alter their arguments directly), but again this should never actually matter to anybody since nobody should be reading that stack count in the first place.

And you don't have to worry about an Increase Stack Size mod.  If someone writes one they'll have to patch the ItemStack serialization anyway (otherwise the mod wouldn't work at all) -- but it doesn't matter since we're ignoring the stack count so we don't care what it does.

gattsuru: If that’s the case, should I revise the WindowListRecipes view, and how should we handle cases where a recipe input has a count above 64?

uecasm: It's not really possible for recipe inputs to have high counts.  Standard vanilla ingredients only support a single item in each crafting grid cell, so with a 3x3 grid it's impossible to have any single item with a count > 9.

We do have a `CountedIngredient` that does permit vanilla recipes with higher counts (currently max is 32 for compost input -- which isn't even a vanilla recipe anyway, so isn't relevant for this), and `CustomRecipe` of course could have higher counts as well, but currently nothing does.  Other mods could do something similar, but again it seems unlikely that they'd do that to a point where it causes an issue, because they still have to work with vanilla crafting mechanics.  So I don't think we need to worry about it.

uecasm: Having said that, `RecipeStorage.getCleanedInput` (which is what is shown in the recipe list) does already return ItemStorage, so anything that reads that should be reading the count from the storage, not the stack.

In an ideal world, yes that should use a custom renderer (and some mods do that, especially storage mods that expect big numbers).  But in practical terms it's probably ok to have it populate an ItemStack from the storage count purely for rendering purposes, since it's just a temporary copy that isn't persisted or sent anywhere, and as noted above it's incredibly unlikely for it to have a high value anyway.  (And itemstacks do render correctly with count>maxStackSize, it's just not a state you should keep one in.)

Similarly for the research rendering, it's not ideal but it's probably ok for you to populate an ItemStack with a high count purely for rendering -- just do it only on client side and only just for the rendering and throw it away.  Only store/transfer the ItemStorage (which contains a stack with 1<=count<=maxStackSize, preferably 1).

gattsuru: I’m not sure there are many places ItemStorage are compared to vanilla recipes.  Maybe the custom recipe must-exist tag?  But that has its own semantic meaning warning that it must follow vanilla rules.  

We don’t use many places where very large ItemStacks make sense, yet (though on spots like the Builder resource tree, that gets closer than you’d think).  But once we open Custom Recipes up for mod pack designer use, it’s both possible and likely to see an outside recipe that wants 1k coal or redstone as an input.  I think such a recipe would even _work_ right now; it just wouldn’t display properly on a remote client.  And short of making a ton on GUI-only copies, I’m not sure how you avoid problems there — does your JEI work handle ItemStorages for such a case, for example?

uecasm: The entire recipe-teaching system revolves around turning vanilla crafting recipes into RecipeStorage, which uses ItemStorage internally.

1k coal in a crafterrecipe?  Possible, perhaps.  Likely, no.  The current implementation of the recipe list will work, but only if the inputs are broken into individual not-more-than-max-stack stacks (such that you could actually fit them into a crafting grid).  It would need some changes if we did want to support recipes with higher counts in a single stack.  But I'm unconvinced that'd be needed in reality.

The current JEI code is outdated, it was written before ItemStorage was used in a lot more places in RecipeStorage, so it's not really representative any more.  Having said that, there's no reason it can't convert an ItemStorage to an ItemStack at the last possible rendering moment, as noted above, especially since it's all client-side anyway.  That's essentially what it does do since (unlike the recipe list) it displays the cleaned and merged inputs, not the original pattern.

ravenbuilder934 (dismissed):
D: for the assignment stealing, :D for doing a great job on it and getting to it faster than I could

src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
gattsuru: I mean, on the output side
image
are valid datapack-style recipes.  They're not a _good idea_, but they both display and craft, just having to be split into normally-sized stacks when getting into the player inventory. 

If we want to stick to the philosophy that ItemStacks should always be valid as inventory ItemStacks, unless the second they're being rendered to the GUI, it's doable (though the way List DataProviders are called makes that a bit more expensive than it seems at first glance).

Raycoms: yes @uecasm is right here. The count of an itemstack in the itemstorage should never be important. We could go as far as always setting the count of a stack in storage to 1, to make sure there is never a high number in there by any chance. At the moment, if someone gives the storage a stack with > 64 that is a problem right there. If you want it to store a big itemstorage then create the itemstorage with a stack and then set the count.

Mekle001: Ok, I'm going to suggest pulling this part of the change. This is really not a serialization problem, this is a problem for ItemStorage itself to solve. 

Talking to Ray, I think we believe the right answer is to update ItemStorage to force the stored stack to 1, and add a new method getStacks() that returns a list of generated/copied stacks that the total equals the amount in the ItemStorage.

Raycoms: I think we should stick with the vanilla limitations of never having more than 64 of anything in a valid inventory. If there is more (let's say 128 items research cost) maybe we should render 2 stacks of 64. In that sense, an itemstorage could have a utility method with getStacks that returns a list of stacks with the correct total size.

gattsuru: I'm not sure we need the getStacks() method, as long as it's well-enough documented that the getItemStack() _isn't_ a representation of the full containing object.  Maybe Ray's got a use case, but for Research it'd end up spitting out 8 stacks of cooked chicken and 32 stacks of redstone at points.  I'll just remove the deserialization change in this PR, and make everything research-related use the `GlobalResearch.hasEnoughResources` method or compare to GetAmount (and possibly add a warning in InventoryUtils.getItemCountInItemHandler), and make the research gui create a copy of the itemstack before adding the amount.

Do you want me to make modifications for cases like the WindowListRecipes et all, or would you rather handle that?

Raycoms: Yes, getStack() is solely to get the "stack data" you could say.

marchermans (dismissed):

src/main/java/com/minecolonies/coremod/research/GlobalResearch.java
```diff
@@ -706,6 +706,10 @@ private ItemStack parseIconItemStacks(final String icon)
             try
             {
                 count = Integer.parseInt(iconParts[2]);
+                if(count > 127)
+                {
+                    count = 127;
+                }
```
Raycoms: Shouldn't we limit that here rather to Stacksize? or maybe even maxStackSize?

gattsuru: If we're being strict on that, I can limit it that heavily.  But these are very explicitly icons, and it seems like it's reasonable to have things like a pick 5 icon.  Only reason I didn't just pass the ItemIcon directly was to avoid potential sidedness issues.  

Or I can even just get rid of the whole behavior, since we're not using DRAW_ICONS at all right now.

Mekle001 (approved):

PR #7029, opened 2021-04-28
Fix ItemStorage serialization, ensure that matching information persists in RecipeStorage

Mekle001:
- Made sure ItemStorage serializes ignoreNBT and ignoreDamage
- Made sure that RecipeStorage properly combines ignoreNBT and ignoreDamage when calculating cleanedInput
- Made sure that RecipeStorage.equals takes ItemStorage comparison information into account

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
+++ b/src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
@@ -44,8 +44,8 @@ public interface IItemStorageFactory extends IFactory<FactoryVoidInput, ItemStor
 
         final ItemStack stack = (ItemStack) context[0];
         final int size = (int) context[1];
-        final boolean ignoreDamage = context[2] != null ? (Boolean) context[2] : false;
-        final boolean ignoreNBT = context[3] != null ? (Boolean) context[3] : false;
+        final boolean ignoreDamage = context.length >= 3 ? (Boolean) context[2] : false;
+        final boolean ignoreNBT = context.length >= 4 ? (Boolean) context[3] : false;
 
         return getNewInstance(stack, size, ignoreDamage, ignoreNBT);
     }
--- a/src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
+++ b/src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
@@ -32,9 +32,22 @@ public interface IItemStorageFactory extends IFactory<FactoryVoidInput, ItemStor
             throw new IllegalArgumentException("Second parameter is supposed to be an Integer!");
         }
 
+        if (context.length >= PARAMS_ITEMSTORAGE + 1 && !(context[2] instanceof Boolean))
+        {
+            throw new IllegalArgumentException("Third parameter is supposed to be an Boolean!");
+        }
+
+        if (context.length >= PARAMS_ITEMSTORAGE + 2 && !(context[3] instanceof Boolean))
+        {
+            throw new IllegalArgumentException("Fourth parameter is supposed to be an Boolean!");
+        }
+
         final ItemStack stack = (ItemStack) context[0];
         final int size = (int) context[1];
-        return getNewInstance(stack, size);
+        final boolean ignoreDamage = context[2] != null ? (Boolean) context[2] : false;
+        final boolean ignoreNBT = context[3] != null ? (Boolean) context[3] : false;
+
+        return getNewInstance(stack, size, ignoreDamage, ignoreNBT);
     }
 
     /**
@@ -45,6 +58,6 @@ public interface IItemStorageFactory extends IFactory<FactoryVoidInput, ItemStor
      * @return a new Instance of ItemStorage.
      */
     @NotNull
-    ItemStorage getNewInstance(@NotNull final ItemStack stack, final int size);
+    ItemStorage getNewInstance(@NotNull final ItemStack stack, final int size, final boolean ignoreDamage, final boolean ignoreNBT);
 }
 
--- a/src/api/java/com/minecolonies/api/crafting/ItemStorage.java
+++ b/src/api/java/com/minecolonies/api/crafting/ItemStorage.java
@@ -269,10 +269,16 @@ public class ItemStorage
 
 
         return stack.isItemEqual(that.getItemStack())
-                 && (this.shouldIgnoreDamageValue || that.getDamageValue() == this.getDamageValue())
+                 && (this.shouldIgnoreDamageValue || that.shouldIgnoreDamageValue || that.getDamageValue() == this.getDamageValue())
                  && (this.shouldIgnoreNBTValue || that.shouldIgnoreNBTValue
                        || (that.getItemStack().getTag() == null && this.getItemStack().getTag() == null)
                        || (that.getItemStack().getTag() != null && that.getItemStack().getTag().equals(this.getItemStack().getTag())));
+    }
+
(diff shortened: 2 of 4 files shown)


src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
```diff
@@ -32,9 +32,22 @@ default ItemStorage getNewInstance(@NotNull final IFactoryController factoryCont
...
+        if (context.length >= PARAMS_ITEMSTORAGE + 2 && !(context[3] instanceof Boolean))
+        {
+            throw new IllegalArgumentException("Fourth parameter is supposed to be an Boolean!");
+        }
+
         final ItemStack stack = (ItemStack) context[0];
         final int size = (int) context[1];
-        return getNewInstance(stack, size);
+        final boolean ignoreDamage = context[2] != null ? (Boolean) context[2] : false;
+        final boolean ignoreNBT = context[3] != null ? (Boolean) context[3] : false;
```
uecasm: This needs to test `context.length` again instead of checking for null.

src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
```diff
@@ -42,9 +52,12 @@
 
     @NotNull
     @Override
-    public ItemStorage getNewInstance(@NotNull final ItemStack stack, final int size)
+    public ItemStorage getNewInstance(@NotNull final ItemStack stack, final int size, final boolean ignoreDamage, final boolean ignoreNBT)
     {
-        return new ItemStorage(stack, size, false);
+        ItemStorage newItem = new ItemStorage(stack,false, false);
```
uecasm: Why does this not use the arguments?

src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
```diff
@@ -65,22 +80,28 @@ public ItemStorage deserialize(@NotNull final IFactoryController controller, @No
     {
         final ItemStack stack = ItemStack.read(nbt.getCompound(TAG_STACK));
         final int size = nbt.getInt(TAG_SIZE);
-        return this.getNewInstance(stack, size);
+        final boolean ignoreNBT = nbt.contains(TAG_SHOULDIGNORENBT) ? nbt.getBoolean(TAG_SHOULDIGNORENBT) : false;
+        final boolean ignoreDamage = nbt.contains(TAG_SHOULDIGNOREDAMAGE) ? nbt.getBoolean(TAG_SHOULDIGNOREDAMAGE) : false;
```
uecasm: `getBoolean` returns false when the tag is missing, so the `contains` checks are not needed.

uecasm:
> Wouldn't it be enough to just write the item itself here (instead of the itemstack) as the count isn't really needed with the amount written afterwards?

No, an ItemStack includes NBT as well.

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -269,12 +269,18 @@ public boolean equals(final Object o)
...
 
         return stack.isItemEqual(that.getItemStack())
-                 && (this.shouldIgnoreDamageValue || that.getDamageValue() == this.getDamageValue())
+                 && (this.shouldIgnoreDamageValue || that.shouldIgnoreDamageValue || that.getDamageValue() == this.getDamageValue())
                  && (this.shouldIgnoreNBTValue || that.shouldIgnoreNBTValue
                        || (that.getItemStack().getTag() == null && this.getItemStack().getTag() == null)
                        || (that.getItemStack().getTag() != null && that.getItemStack().getTag().equals(this.getItemStack().getTag())));
     }
 
+    public boolean matchDefinitionEquals(ItemStorage that)
```
Raycoms, line 283: jdoc

src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
Mekle001: Wow. Clearly distracted yesterday. Thanks.

src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
Mekle001: Another good catch. I'd blame the beverages, but there was none of interest.

src/main/java/com/minecolonies/coremod/colony/crafting/ItemStorageFactory.java
Mekle001: Now, this one I'd forgotten, and just went with the safe pattern. Anyway, it's cleaned up.

src/api/java/com/minecolonies/api/crafting/ItemStorage.java
Mekle001: Again? Oh, right. I did it once, removed it during our discussion yesterday about how to resolve this, and then put it back again, forgetting the Jdoc this time. 

That's my story, and I'm sticking to it.

Raycoms (approved):

src/api/java/com/minecolonies/api/crafting/IItemStorageFactory.java
```diff
@@ -44,8 +44,8 @@ default ItemStorage getNewInstance(@NotNull final IFactoryController factoryCont
 
         final ItemStack stack = (ItemStack) context[0];
         final int size = (int) context[1];
-        final boolean ignoreDamage = context[2] != null ? (Boolean) context[2] : false;
-        final boolean ignoreNBT = context[3] != null ? (Boolean) context[3] : false;
+        final boolean ignoreDamage = context.length >= 3 ? (Boolean) context[2] : false;
+        final boolean ignoreNBT = context.length >= 4 ? (Boolean) context[3] : false;
```
uecasm, line 48: This is better, but it's inconsistent with the comparisons above.  You should make both the same.  (Personally, I'd use the plain number, not the named constant, but that's up to you.)

uecasm (commented):
I haven't had a chance to try this out with the enchanter PR yet; I'll try that later today.

uecasm:
> I haven't had a chance to try this out with the enchanter PR yet; I'll try that later today.

Seems to work.  :shipit:

someaddons (approved):

MeikelLP:
I assume that this PR shall cause all recipes to accept any item (ignoring their nbt tags)?

I'm on 0.14.138-ALPHA and fortress/guardtower still requests a diamond helmet but the builder already has one (with 2 nbt tags)

Is this expected or is this a new bug (if yes, shall I open an issue?)?

Raycoms:
No, you're assuming wrongly, this PR allows us to add recipes that allow the builder to do such things.

MeikelLP:
So an update for all builder recipes will come in the future?

Raycoms:
We won't allow plain recipe conversion for all recipes, I don't think it's in anyones interest for a builder to put a fortune III pickaxe in an itemframe in a building. But we plan it for some limited cases.

MeikelLP:
I have repaired some used diamond armor and stripped their enchantments. Sadly they are useless for builders which is really annoying. Also for clocks and compasses that collide with quark, that would be really really helpful!

Mekle001:
The clocks and compasses are are definitely on the radar.  The armor? That's a harder problem, and while it's something I'd like to solve, it won't be for a while, as it's a much more complex compare.

PR #7049, opened 2021-04-30
Constants cleanup

ravenbuilder934:
Closes #
Closes #
Closes #

- Deletes a barely-used constants class
- Deletes unused constants
- Deletes some unused lang strings

Review please


src/main/java/com/minecolonies/coremod/commands/colonycommands/CommandClaimChunks.java
```diff
@@ -59,7 +58,7 @@ public int onExecute(final CommandContext<CommandSource> context)
             return 0;
         }
 
-        if (chunkManager.getAllChunkStorages().size() > CHUNKS_TO_CLAM_THRESHOLD)
+        if (chunkManager.getAllChunkStorages().size() > 5000)
```
Raycoms: Well, this should still be a constant, right

ravenbuilder934: tbh I don't see the point of a constant if it's only used in one place. But all right, I'll add it into one of the existing constants classes

someaddons (approved):

PR #7068, opened 2021-05-06
Make Zombie Villagers convert to Visitors in colonies with Taverns

ToMe25:
- When a Zombie Villager gets cured in a Village with a Tavern it gets converted to a Visitor, instead of a Villager.
- They have a specific backstory, different from the ones of other visitors.
- This is currently a slightly adapted version of @someaddons suggestion from discord, however if anyone can think of something better I'm not opposed to changing it.

Review please


Aduis1210:
<3 nice work with this - love the addition

someaddons (approved):

PR #7070, opened 2021-05-06
Fix smelter fuel retaining to not be coal greedy

Mekle001:
- change the keep to be only one stack of fuel, and properly compare against allowed fuels

Review please


Raycoms:
Well, maybe furnaces * stacks?

Raycoms:
Or buildinglevel stacks

Mekle001:
Went to a constant 5 stacks retained, as both building level and furnacecount are not available in the constructor.

uecasm:
Can't you just adjust the keepX stack size after the data is loaded / level up?

Raycoms:
> Can't you just adjust the keepX stack size after the data is loaded / level up?

Well, there is the method that queries the keepX where we could adjust it manually.

Mekle001:
Ok, moved all of the retention down to abstractBuildingFurnaceUser, and have it auto-calcing the size to retain based on building level. This actually fixes all of our furnace users, which had varying levels of support for fuel retention.

Raycoms (approved):

PR #7076, opened 2021-05-08
fix Raid npe with no townhall

someaddons:
Closes #
Closes #
Closes #

Fix npe during raids with no townhall, players outside of the colony will no longer hear raid sounds aswell
Review please


marchermans (approved):
Approve of the npe fix.

But I am not sure if reducing the raid sound range is a good idea, i think it would be nice if this is configurable in the colony.

I can think of a lot of situations where it is desirable that i hear these sounds, especially outside of the colony. Alerting me to the fact that i should come back.

someaddons:
hm I do agree that some sound outside the colony is useful, though atm the sound we have is a "Bell" warning which naturally is supposed to resemble a localized warning bell positioned within the colony, so that is kinda odd to be heared super far away. Maybe we add a different sound for a faraway notification instead

PR #7088, opened 2021-05-10
Properly cleanup requests

Raycoms:
- Remove requests from the system once the requester is removed (destroyed)
-
-

Review please


Mekle001 (approved):

marchermans (approved):
Good solution.

Does this only apply here though?

Raycoms:
> Good solution.
> 
> Does this only apply here though?

I think so, yes

PR #7089, opened 2021-05-11
Create reset to default button in the item list module

### LINKED ISSUE #307: Option to (de)select all options in a worker's hut
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: 0.13.636-ALPHA

## Feature description
When I started turning everything on in the composter's hut I noticed that it took a very long time as there are so many options available. It would be so much easier to have an option to (de)select all options in one click.

### Notes or related things

---
## Viewers

chandler05:
Closes ldtteam/minecolonies-features#307
Closes #
Closes #

- Add a reset to default button for resetting all elements in an item list module
-
-

Review please


ravenbuilder934:
Screenshots of the new GUI please

chandler05:
Here's what it looks like.

ravenbuilder934 (dismissed):

Raycoms:
Well, the reason I've not done that yet is because selecting/deselecting all at once comes at a significant negative performance impact (which is why the "inverse" feature exists to start with. I'm still unsure about this. I feel like offering would result in this being used more than necessary resulting in significant performance impact

src/main/java/com/minecolonies/coremod/client/gui/modules/ItemListModuleWindow.java
```diff
@@ -128,6 +136,28 @@ private void switchClicked(@NotNull final Button button)
         resourceList.refreshElementPanes();
     }
 
+    /**
+     * Fired when select or deselect has been clicked.
+     */
+    private void selectAll(@NotNull final boolean enable)
```
MatthiasMann: If this method would be moved into the module the deselecting of all items could at least be optimized a lot (by using listsOfItems.clear() and sending a special message with the same effect).

Raycoms: Well, the problem is slightly different. If we have a cook request with 20000 items to not accept in it. This will blow up the request system by an enormous size.

MatthiasMann: Hmm - maybe the implementation needs to toggle between white list / black list then - eg when you click on "deselect all" it will switch to an empty white list mode, and when you click on "select all" it will switch to an empty black list mode but combine that with the allowed item list of that hut?

Raycoms: Yeah, but that is not simple at all.

chandler05: Still doesn't solve the problem, each item in the list will be sent to the request system causing performance issues with high numbers of items regardless of how the selecting and deselecting is handled.

Raycoms:
I'll close this for now. If we do this, then we need propper handling to not make this perform extremely badly and cause tons and tons of lag and memory cost.

uecasm:
Perhaps it could just be implemented as one-sided for now?  i.e. for regular lists that default to everything off, there is only a "deselect all" button, and for inverse lists that default to everything on, there is only a "select all" button.  In both cases it effectively just clears the filter to the empty/default state.

It's perhaps not as useful as having both, or being able to switch modes, but it's not useless, and it should have less performance issues (if I'm understanding the problem above).

Raycoms:
Yes, we could have a "reset to default" button.

chandler05:
I can refactor my current code to do that.

chandler05:
Done, I have changed the button to "Reset to Default".

Could this be reopened?

Raycoms (approved):

marchermans (approved):

someaddons (approved):

PR #7093, opened 2021-05-11
Fix npe in unReserveGrave

### LINKED ISSUE #7090: Undertaker not doing job, just AI-erroring
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: 0.14.158-ALPHA

### Logs

- latest.log: https://gist.github.com/Lielac/5abc737753d5bcca83ee71a61c9aa7aa

## Steps to reproduce the issue / details

Just reporting the AI issue. 

### Notes

---
## Viewers

chandler05:
Closes #7090
Closes #
Closes #

- Fix an npe caused by the unReserveGrave method
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/managers/GraveManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/GraveManager.java
@@ -197,7 +197,7 @@ public class GraveManager implements IGraveManager
     @Override
     public void unReserveGrave(@NotNull final BlockPos pos)
     {
-        if (graves.get(pos))
+        if (graves.get(pos) != null)
         {
             graves.put(pos, false);
             colony.markDirty();
--- a/src/main/java/com/minecolonies/coremod/colony/managers/GraveManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/GraveManager.java
@@ -197,7 +197,7 @@ public class GraveManager implements IGraveManager
     @Override
     public void unReserveGrave(@NotNull final BlockPos pos)
     {
-        if (graves.get(pos) != null)
+        if (graves.containsKey(pos) && graves.get(pos) != null)
         {
             graves.put(pos, false);
             colony.markDirty();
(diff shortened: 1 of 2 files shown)


Raycoms:
Were you able to reproduce this? I didn't look too deep into this, but, I wonder why this happens in the first place

chandler05:
Couldn't reproduce, not sure why the grave pos would have been null.

chandler05:
Other than that I don't know how it could have happened.

Mekle001 (dismissed):

Raycoms:
I guess this might stem from the player removing the grave before the citizen gets to it, but, you let the chunk unload

someaddons:
I've taken a look and the actual issue of the NPE is not that the blockpositions is null, but that the graves list does not contain the position, which can happen since building might not be up to date on what the graves list  has. So we should add a `graves.contains(pos)` check in addition to the null check.

Also noticed that the Tileentity checks are only tileentity != null, should be instead a `tileentity instanceof GraveType` to have the proper type check too.

src/main/java/com/minecolonies/coremod/colony/managers/GraveManager.java
```diff
@@ -197,7 +197,7 @@ public boolean reserveGrave(@NotNull final BlockPos pos)
     @Override
     public void unReserveGrave(@NotNull final BlockPos pos)
     {
-        if (graves.get(pos))
+        if (graves.containsKey(pos) && graves.get(pos) != null)
```
Raycoms: Doesn't need the null check, contains does this already, so this should be graves.containsKey(pos) && graves.get(pos)
The method that calls this then needs the te instanceof Grave instead of != null too

chandler05: Ok, done.

Raycoms (approved):

someaddons (approved):

src/main/java/com/minecolonies/coremod/colony/managers/GraveManager.java
MatthiasMann: doing a containsKey followed by get will perform the same operation twice on the map. It would be cheaper to just use get and check the result if you never insert null values.

someaddons: you're right that it would be cheaper, although this place isnt called frequent enough to make a difference and is a bit nicer to read this way

PR #7123, opened 2021-05-20
Draft: Modular Citizen Models/Textures

gattsuru:
- Every citizen model is driven from a Resource Pack file, and citizen textures can be added beyond the current limits.  That includes different numbers of textures by style or gender, and new single-colony-style colonist styles.
- This should allow more ready variation between models and textures, including different numbers of textures or models by style or gender, and even new colonist styles eventually bypassing the need for code changes to add those additional styles.  ie, male-lumberjack-default can have 3 textures, while female-lumberjack-default can have 2, and male-lumberjack-medieval could have 4.  Replacements for existing styles will require more textures and files than for new styles, though (ie, both medieval and have 12 student textures, and if you replace behavior for a BASE or CLOTHING layer model, you'll probably need to change all 12 of those students or risk them looking ugly).
- ~~(Should the fallback or default style be overridable on the client side?)~~  Raycoms says not something that needs to be set up for now.
- The system _should_ be able to handle 'partial' styles of citizen models and textures with relatively little expansion, cleanly falling back to the default version.  While not used heavily, it can also handle cases where there's only one texture shared between both genders.

To avoid changelog bloat, this commit does not include the textures or datagen output, nor the minor JavaModel changes.

Costs :
- Performance. Models aren't loaded until needed, and total texture count is lower, but active textures may be higher in some cases and total draws per colonist will be significantly higher.
- File count is higher, and model jsons are not very readable for hand-editing.
- New models must be implemented in data gen before they can be tested in-world, and some traits (mostly physical conflict with other layers, but also the impact on layers by color modifiers) aren't easy to predict from BlockBench, at least from a quick glance.
- Models that vary greatly from a BASE layer model -- the four current ones are Alex, Steve, Kobold, and Dwarf -- may require additional BASE layer models and textures.  Models must be registered to dataGen including their base model attribute, and right now I'm mostly guessing.
- Some models and textures will require additional work from the art team, and there's a _lot_ of general testing to make sure everything looks not-crazy.
- A lot of the resulting textures are dimensionally bigger than they need to be, since they no longer need to represent as much design space (though file size is smaller).
- I don't like the way currently attaching the ClientResourceListener. Not sure if there's a better way to do it; most alternatives were too late to pick up the initial resource pack load.

Some limitations that can be (and in many cases, should be) mitigated, if needed : 
- ~~Does not include a "remove" equivalent~~. Nope.
- Does not handle the worker texture metadata (yet).
- Does not support variety of CitizenRenderSetting settings (yet).  
- Suffix/eye/hair values save/load isn't right (yet).
- Keeping as close compatibility with existing BlockBench java models involved some some compromises.  Some of that's just a minor tweak: isFemale chests have a special logic chain that _might_ be better done as an explicitly set property on the ModelRenderer.  The dataGen access to underlying model data to serialize it is ugly, and handling of Resource Packs to map available files is ugly too (though thankfully off-main-thread).
- The current test version bakes accessory models based on absolute position. I'd like to convert it to relative positions on serialization/deserialization, so they can more easily be adjusted to other models without caring so much about their BASE version, but this and the Animatrix equivalent to have decent overlap where possible, and I'm still trying to draw out a decent plan for (and even start to understand) Animatrix.
- The data structures here are a mess, to understate things. The inheritance diagram needs to be squashed, and beyond seriously, I'm using BipedModel rather than ModelRenderer or LivingEntity as the 'base' for every slot for each model. This should allow us to take as much advantage of the existing animation systems as possible, but it's really inefficient (especially memory-wise) and hacky, and I'm still thinking a bit on how to resolve it.  There's good options, but many involve mixins.
- Performance needs some serious testing.  The layer-based system is naturally and unavoidably going to be a little more expensive, but most of this should only impact situations with a lot of colonists on-screen at once, so some real-ish environments are probably much more meaningful in terms of results, rather than me dropping 400 colonists in a pile on a superflat world.  There's a lot of low-hanging fruit that can be done if impact is too severe, but some will restrict the scope of the project.
- ~~Right now, eyes color, suffix color, and hair color is a hard-coded list (and only has 'genetics' for suffix color). This might be better handled as a config file value, but a) would have to be server-side, and b) _may_ be something we don't want to expose to users for a variety of reasons.~~ Nope.
- Setters are the same as before, entirely based on BipedModelType, selecting by job then house.  This obviously needs to be expanded.   Carlansor's already had some good ideas on that, but need to ponder a bit for how to implement them.
- I've got texture files for the DEFAULT style, but haven't gotten far on the MEDIEVAL styles.  I'm not entirely sold on the current file breakdown, so need to ponder a bit before working them.  Haven't committed either yet, because they swamp the actual code changes.
- Ghost mode isn't ready yet.
- This isn't _too_ smart, in ways it could be. Eg, it could be made to dynamically create models specifically matching their texture sizes.  I'm not sure if this would be valuable at all, though.

Review please


marchermans:
This PR should be pushed to 1.17.x

Mojang added json based entity render modification and specification as far as we can see.
Which would allow this to happen much easier.

Raycoms:
Closed, and as Orion mentioned we should look into the 1.17 changes (1.17 also allows hot reload which is nice)

PR #7139, opened 2021-05-23
Recipe/concrete glass

uecasm:
- Moves concrete mixer and glassblower recipes to data pack.

Review please


uecasm:
As requested, included datagen for glassblower, although it seems a bit too trivial to need it.

Note that as currently implemented these recipes won't replace the ones in existing buildings on upgrade (but they're identical except for having a `recipeSource` now, so it mostly doesn't matter).  This is because the existing code thinks that they might be reduced-input recipes and so prefers the one that's already in the building.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultConcreteMixerCraftingProvider.java
```diff
@@ -0,0 +1,64 @@
...
+
+            CustomRecipeBuilder.create(ModJobs.CONCRETE_ID.getPath(), powder.getRegistryName().getPath())
+                    .inputs(customInput)
+                    .result(new ItemStack(powder, 8))
+                    .build(consumer);
+
+            CustomRecipeBuilder.create(ModJobs.CONCRETE_ID.getPath(), concrete.getRegistryName().getPath())
+                    .inputs(Collections.singletonList(new ItemStorage(new ItemStack(powder))))
+                    .result(new ItemStack(concrete))
+                    //.intermediate(Blocks.WATER)
```
uecasm, line 60: FYI, wasn't sure if this (`intermediate`) should be included or not.  It makes a certain amount of logical sense for it to be there, but the previously existing recipes did not set it.  And currently neither the RS crafting logic nor the JEI rendering support it.

Mekle001: yeah, this is all special case handling. One of these days, as we modularize the crafting code, we should put in a intermediate handler, and allow a broader set of intermediates besides just air and furnace.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingGlassblower.java
```diff
@@ -132,39 +118,6 @@ public void onColonyTick(@NotNull final IColony colony)
...
-    public void checkForWorkerSpecificRecipes()
-    {
-        super.checkForWorkerSpecificRecipes();
-
-        final List<IToken<?>> tokens = new ArrayList<>();
-        for (final Item item : Tags.Items.SAND.getAllElements())
-        {
-            final ItemStack stack = new ItemStack(item);
-            final ItemStack output = FurnaceRecipes.getInstance().getSmeltingResult(stack);
-            if (Tags.Items.GLASS.contains(output.getItem()))
```
Mekle001, line 145: So, it's probably not critical, but this change means that any moded sands that are included in this now no longer get recipes.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultConcreteMixerCraftingProvider.java
Mekle001: Thinking about it, it might make sense to include it in the future, and update the crafter to trigger off the water for the special case handling, rather than use the specific items.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingGlassblower.java
uecasm: Yes, we were discussing that in #discussion on Discord.  The conclusion was that nobody could think of any mods in which "(item tagged as sand) => (item tagged as glass)" were standard furnace smelting recipes.  But if one does come up, a modpack could include a compat recipe datapack.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultConcreteMixerCraftingProvider.java
uecasm: Yeah, that's why I left it commented out, as a reminder for the future.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingGlassblower.java
Mekle001: Ok, I've been away all day, I'll go catch up there.

uecasm: Also, Ray's other idea is that once crafting modules are added, the glassblower could have both a crafting and smelting module and then you'd have to teach it which glass recipe you want to use (based on glass product tag) rather than these being inherent crafter custom recipes.  This avoids most of the potential compat issues, as long as things are tagged appropriately.  (And it wouldn't have to care about sand, just glass.)

The first edition of the JEI PR won't be quite enough to get it there, but Soon™.

Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultConcreteMixerCraftingProvider.java
marchermans: Add a TODO to it then.

marchermans (dismissed):
This looks good.
As discussed in the discord.

Could you mark that as a TODO and then it is mergable.

PR #7148, opened 2021-05-25
specific version of structurize dependency

YoYoK3nny:
structurize dependency versionRange not allowing for newer versions of Structurize

This pull request changes that.


uecasm:
It is entirely deliberate that it requires a specific version, not any later version.

marchermans:
This is deliberate, we test explicitly one version for MC.

PR #7149, opened 2021-05-25
JEI crafting integration, redux

### LINKED ISSUE #282: Teaching recipes - Tell User who can learn it
### Prerequisites 

### Context 

- Minecraft Version:
- MineColonies Version:

When trying to teach a recipe to a worker/building, in the event of an error caused by trying to teach the recipe to the wrong person, it would be nice for the chat message to tell the user which building/worked can be taught that recipe (if any can).

## Viewers

uecasm:
Closes #6540
Probably closes ldtteam/minecolonies-features#282

- Stage one of adding crafting modules to buildings (more work to follow)
- Adds JEI recipe integration

Review please

1. The recipe lookup allows you to see what crafting (or furnace smelting, as appropriate) recipes can be taught to buildings, plus those that they already know via crafterrecipes:
image
    The latter will also indicate if there are level or research restrictions.
2. There's also a slightly different style used for loot-table-based recipes (sifter and enchanter).  These are determined from the recipe itself, so they can be freely mixed in a single building (e.g. enchanter):
image
3. Additionally, clicking the arrow in the "teach recipe" UI will take you to the corresponding page that shows learnable recipes.
image
4. To teach recipes, you can use the (pre-existing) plus button or (new) drag ingredients directly from the JEI GUI (without needing cheat mode):
image


uecasm:
I'm still considering a redesign on the crafting module structure -- rather than having a single interface with policy subclasses, it might be better to have the base `ICraftingBuildingModule` (used directly by custom-only modules), and extend `ICraftingCraftingBuildingModule` and `ISmeltingCraftingBuildingModule` (for each vanilla crafting type) from that.  (Or with better names.)  This would let them have different interfaces, which might be useful once state starts to be added.

On the other hand, the present design is simpler and more easily allows a single ModuleView to serve all cases.

Hard to say which one will be better in the end, it depends how similar the remaining parts of implementation end up being.  Still, it could always be changed later if needed.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingWorker.java
```diff
@@ -314,9 +311,36 @@ public void switchIndex(final int i, final int j)
...
+        return hasSpaceForMoreRecipes() && isRecipeCompatibleWithCraftingModule(token);
+    }
+
+    /**
+     * @param token the recipe token
+     * @return whether the recipe can be added according to the crafting module (or false if there's no module)
+     */
+    protected boolean isRecipeCompatibleWithCraftingModule(final IToken<?> token)
+    {
+        return this.getModules(ICraftingBuildingModule.class).stream()
```
Raycoms: I have a feeling that a simple for each loop here could be much easier and much more efficient.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/smelter/EntityAIWorkSmelter.java
```diff
@@ -399,15 +406,76 @@ private int getRequiredProgressForMakingRawMaterial()
...
         for (final Map.Entry<Enchantment, Integer> entry : enchants.entrySet())
         {
             EnchantedBookItem.addEnchantment(books, new EnchantmentData(entry.getKey(), entry.getValue()));
         }
-        worker.decreaseSaturationForAction();
         return books;
     }
+
+    @NotNull
+    public static List<IGenericRecipe> getAdditionalRecipesForDisplayPurposesOnly()
```
Raycoms: I'm really unsure about all of this, basically, this is the #1 feature that I'm unsure about and wondering about scrapping/modifying it. That's why I'm unsure if I want people to actually see this so obvious in JEI atm.

uecasm: If the feature itself is removed, the JEI display can be removed at the same time (that's an advantage of having them both in the same place rather than in entirely different files).

While it exists, though, it ought to be documented in some way, otherwise nobody would know that it exists.

The other option might be to remove the JEI/crafting module entirely -- which is not unreasonable since the smelter doesn't support crafterrecipes or teachable recipes at all, but it would make the mod less self-documenting.  (I was considering later adding a JEI page for the animal huts too, showing the animal drops as outputs plus some special cases like cow milking.  This would use the same mechanism, although is a bit more complicated due to loot tables.)

someaddons: there is some stuff which is intentionally not shown/displayed as readily available public information such as wiki/jei is

uecasm: Sure, and that makes sense for some things, like specific formulas for raids and XP etc.  It makes less sense to hide crafting mechanics, though.

Raycoms (dismissed):

someaddons (dismissed):

marchermans:
:ship it:

marchermans:
:shipit:

PR #7174, opened 2021-05-30
Add minimum stock to more buildings

dhouck:
This adds minimum stock to every building except for:
 * those where no one uses the inventory (Barracks, Courier, etc.)
 * crafters which doesn't use fuel (Blacksmith, Fletcher, etc.)
 * university and Plantation, because it seemed useless there
 * Postbox, because the UI is different so it wasn't as simple as the
   other buildings

# Notes
Does not close any issues, but touches on ldtteam/minecolonies-features#186 and ldtteam/minecolonies-features#440.

I haven't tested this as much as I want yet.  The "Minimum Stock" tab shows up on all these buildings, seems to work correctly for some of them, and doesn't *visibly* fail for any of them but I haven't fully tested that it works.

I'm getting a display issue where I need to go away from the tab and back for minimum stock to show up, on both the buildings I added it for and the buildings that already have it.  I don't remember seeing this issue before but I bet when I get a chance to test in released minecolonies it'll be there too.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
+++ b/src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
@@ -53,6 +53,7 @@ public final class ModBuildingsInitializer
                                  .setBuildingProducer(BuildingArchery::new)
                                  .setBuildingViewProducer(() -> BuildingArchery.View::new)
                                  .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.ARCHERY_ID))
+                                 .addBuildingModuleProducer(MinimumStockModule::new, MinimumStockModuleView::new)
                                  .addBuildingModuleProducer(BedHandlingModule::new)
                                  .createBuildingEntry();
 
@@ -79,6 +80,7 @@ public final class ModBuildingsInitializer
                                        .setBuildingProducer(BuildingBarracksTower::new)
                                        .setBuildingViewProducer(() -> BuildingBarracksTower.View::new)
                                        .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.BARRACKS_TOWER_ID))
+                                       .addBuildingModuleProducer(MinimumStockModule::new, MinimumStockModuleView::new)
                                        .addBuildingModuleProducer(BedHandlingModule::new)
                                        .createBuildingEntry();
 
@@ -95,6 +97,7 @@ public final class ModBuildingsInitializer
                                  .setBuildingProducer(BuildingBuilder::new)
                                  .setBuildingViewProducer(() -> BuildingBuilder.View::new)
                                  .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.BUILDER_ID))
+                                 .addBuildingModuleProducer(MinimumStockModule::new, MinimumStockModuleView::new)
                                  .addBuildingModuleProducer(() -> new SettingsModule().with(BuildingBuilder.MODE, new StringSetting(BuildingBuilder.AUTO_SETTING, BuildingBuilder.MANUAL_SETTING)), SettingsModuleView::new)
                                  .addBuildingModuleViewProducer(WorkOrderListModuleView::new)
                                  .addBuildingModuleProducer(BuildingResourcesModule::new, BuildingResourcesModuleView::new)
@@ -105,6 +108,7 @@ public final class ModBuildingsInitializer
                                        .setBuildingProducer(BuildingChickenHerder::new)
                                        .setBuildingViewProducer(() -> BuildingChickenHerder.View::new)
                                        .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.CHICKENHERDER_ID))
+                                       .addBuildingModuleProducer(MinimumStockModule::new, MinimumStockModuleView::new)
                                        .addBuildingModuleProducer(() -> new SettingsModule().with(AbstractBuildingWorker.BREEDING, new BoolSetting(true)), SettingsModuleView::new)
                                        .createBuildingEntry();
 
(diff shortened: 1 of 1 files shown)


marchermans:
I don't think this is a good idea on crafters. The RS currently runs the assumption that requests for craftings are changed to prevent things from gumming up. And this will break that assumption. Since if an ingredient is requested that is in stock no further request by the crafter is made. This can however break request execution order since the tree structure might require certain objects to be crafted multiple times using the stock which will break if the stock runs out.

If you want this to have any chance of getting merged the module needs to be mutually exclusive with public crafting at least.if not mutually exclusive with any crafting.

And if you do that, then there are not many buildings left that could benefit from this feature, so I am not sure. Please discuss this in our developer discord if you need more information.

marchermans (changes requested):
The stock module is not compatible with crafting. Please revert.

dhouck:
1. Is there a link to the developer discord separate from the regular discord?  Nobody objected to it when I discussed it there.
2. I don't see anything special in the restaurant or bakery that lets them do minimum stock and crafting, and they already support both.  Did I miss something, or can they run into this issue?
3. If I revert the crafters, I still have:
 * Combat academy and archer training, which can request swords/bows to not have to wait for new ones
 * Animal herders, which can request wheat (or other food), axes, buckets, and shears
 * Composter, which can have a stock of compostables and not need to wait to start a new cycle
 * Fisher, miner, which can stock up on tools (and, for the miner, node construction materials)
 * Library, school, enchanter, which can stock up on books and/or paper
 * Smeltery, which can have a stock of fuel (unlike stone smeltery, the regular smeltery does not handle requests)
 * Florist, which can stock up on compost (or try; usually I don't have enough composters to keep up)
 * Hospital, which can stock up on various cures
 * Beekeeper, which can stock up on glass bottles and shears

Which is quite a lot and still can be useful.

Unfortunately the forester, which inspired me to do this, is a crafter for stripped logs so would be out, but I haven't had an issue stocking up there, maybe because I was only trying to stock saplings.  (It also didn't *help* because of some bugs in the forester AI, but that's a different problem.)

dhouck:
Wait, what I said for 2. isn't quite true, there is something in the cook specifically for minimum stock, and keeping it when the courier is doing a pickup.  But it is for bakery, and I don't think that addresses the issue you were talking about anyway.

OrionDevelopment:
Developer discord: <https://discord.gg/KSWMG9RR>

marchermans:
2) Will indeed have issues when crafting via them is done. And we will need to discuss that then internally since it will cause issues.

dhouck:
Thanks for the link; I joined the server and will ask around more once I have time for a real discussion.

I'm not sure I understand the issue you mention with the crafters; can you give a concrete scenario?  Eg. the restaurant is told to minimum stock golden apples, the hospital and a postbox both ask for golden apples, and some other details happen, causing the system to get stuck.

dhouck:
I seem to have been kicked from the dev discord; presumably someone decided I shouldn't have access to it after all (I rejoined to at least see the publicly-visible sections).

Is there anything you can say in public about the combination of minimum stock and crafter requests?  How bad can it get; how urgent is it to fix the restaurant and bakery?  If it isn't very urgent, might some of the crafters here be worth adding minimum stock to anyway?  If it is urgent, does it make sense for this to just wait until after the fix for those?

dhouck:
I removed four of the crafters based on Discord discussion but kept the forester and farmer awaiting a second opinion.  They're in a separate commit so it's easy enough to move them to a different PR, though.

dhouck:
I think it's unlikely people will try to min stock things that cause issues here, but they might, and based on some discussion in Discord it might be possible for me to add filters to the farmer and forester so that isn't possible.

marchermans:
Use the public discord.

dhouck:
I have been talking about this in the public discord, yes.

dhouck:
This should now only have non-crafters in the PR.  I'll make a new one for just the forester and farmer.  For now it's just the same change for them, because I think it's useful and these issues are unlikely to come up in practice, but I'd like to add a filter to prevent that from coming up if the module supports it.

src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
```diff
@@ -123,6 +128,7 @@ public static void init(final RegistryEvent.Register<BuildingEntry> event)
                                    .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.COMPOSTER_ID))
                                    .addBuildingModuleProducer(() -> new ItemListModule(COMPOSTABLE_LIST), () -> new ItemListModuleView(COMPOSTABLE_LIST, COM_MINECOLONIES_REQUESTS_COMPOSTABLE_UI, false,
                                      (buildingView) -> IColonyManager.getInstance().getCompatibilityManager().getCompostInputs()))
+                                   .addBuildingModuleProducer(MinimumStockModule::new, MinimumStockModuleView::new)
```
Raycoms: this one also interferes with regular working.

dhouck: Removed

Raycoms (approved):

dhouck:
@OrionDevelopment Does this look good to you now?

Also, someone on Discord mentioned that it might be helpful to have min stock for the barracks so you can more easily hire mercenaries, and I think that could make sense.  Similarly, you could min stock some of the things visitors might want in the tavern.  Does that seem reasonable; is there any reason not to add them?

marchermans:
Both barracks and tavern should be compatible. 

The problem is mostly as i said crafting since you easily mess up request trees with that and get citizens stuck.

So if you would not mind and at those then i think this is good to go.

dhouck:
Something really weird is going on with adding a minimum request to the tavern, and the barracks seems to work normally but doesn't actually submit the request to the system (probably because nobody works there?).

Either way both are more complicated than the other buildings and lower priority, and shouldn't hold up this PR.  I'm going to look a bit more but probably not enough to fully fix them, so for now I'll consider this PR to have all the buildings it's going to.

PR #7179, opened 2021-05-30
Guard Module

### LINKED ISSUE #7182: Builder Manual mode no longer has a way to start a task
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: minecolonies-0.14.196-ALPHA-universal.jar
- Structurize Version: structurize-0.13.175-ALPHA-universal.jar

## Expected behavior

In the previous version I was running, minecolonies-0.14.182-ALPHA-universal.jar, when the builder was set to manual mode, there was a page (reached by using the blue arrow in the upper right corner of the builder's hut UI) with work orders.  This page allowed you to select which of the tasks requested of the builder to work on.  For manual mode, this was essential as the builder won't start working on a task without player confirmation.

## Actual behavior

In 0.14.196-ALPHA, there is no blue arrow in the upper right corner of the builder hut UI.  It looks like the equivalent page is a tab on the left hand side of the builder ui titled Work Orders.  The work Orders page lists all of the tasks requested of hte builder, just like the old page, however, it only allows you to cancel tasks, not to start one:

My workaround, for now, is to go to this page, cancel all tasks except for the one I want the builder to work on next, and then set the builder to automatic mode.  Once the builder prints the chat message that they've started work on that task, I then go back to the builder's UI and set it to manual mode again.

1. Set the builder to manual mode
2.  Place a new structure using the build tool
3. In the new structure's Build Options, choose the builder you've set to manual mode and then click the button to build the building.
4. In the builder's screen, click to the tab for Work Orders.
5. See that there's no button to press to tell the builder they can start working on the new building.

### Logs

- latest.log: https://gist.github.com/abadger/f348a8b7ad358277686ccae52e266da1
- crashlog: N/A

### Notes

---
## Viewers

Raycoms:
- This is an extensive rework on how the hostile mobs are handled by the guards.
a) This is now in a select module similar to the item modules, this may be easy extended to other workers if necessary
b) Additional mobs are identified by a special tag
c) Only the disabled monsters are stored in the module (reduces memory overhead by a lot)
- This is also an extensive rework of the guard tower settings putting it a proper module.

Review please


marchermans (dismissed):

src/main/resources/data/minecolonies/tags/entities/hostile.json
```diff
@@ -0,0 +1,6 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:slime"
```
uecasm: The wiki says this needs to be in an `entity_types` dir, not `entities`.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/EntityListModule.java
```diff
@@ -0,0 +1,121 @@
...
+        return this.id;
+    }
+
+    @Override
+    public void serializeToView(@NotNull final PacketBuffer buf)
+    {
+        buf.writeInt(mobsAllowed.size());
+        for (final ResourceLocation entity : mobsAllowed)
+        {
+            buf.writeResourceLocation(entity);
```
uecasm: Writing a list of resource locations is less efficient than writing a list of registry entries.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/EntityListModule.java
```diff
@@ -0,0 +1,121 @@
...
+        this.id = id;
+    }
+
+    @Override
+    public void deserializeNBT(final CompoundNBT compound)
+    {
+        final ListNBT filterableList = compound.getCompound(id).getList(TAG_MOBLIST, Constants.NBT.TAG_COMPOUND);
+        for (int i = 0; i < filterableList.size(); ++i)
+        {
+            mobsAllowed.add(new ResourceLocation(filterableList.getCompound(i).getString(TAG_MOB)));
```
uecasm: This is not compatible with the previous serialization (that used `mobview`, not `mob`), so it's not backwards compatible and all existing guard towers will lose their current settings.

Either that's a bug by itself, or if you're intentionally breaking compatibility then you might as well go all the way and make it a plain list rather than a list of compounds for efficiency.  (Though that also might require some defensive code for the upgrade case, which this doesn't.)

uecasm (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/modules/EntityListModule.java
Raycoms: I have to double check this, but I'm breaking it intentionally to lose the values yes. I can use a string list true. I have to load this from a different compound I believe initially to make sure it doesn't throw an exception though. Forgot about that.

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/EntityListModuleView.java
```diff
@@ -110,7 +111,7 @@ public void deserialize(@NotNull final PacketBuffer buf)
 
         for (int j = 0; j < size; j++)
         {
-            listOfEntities.add(buf.readResourceLocation());
+            listOfEntities.add(buf.readRegistryIdUnsafe(ForgeRegistries.ENTITIES).getRegistryName());
```
uecasm, line 114: You might want to include a null check on this -- it can return null if the ResourceLocation given on the server side wasn't actually in the entities registry (for example, someone saved a mob to the exclude list and then uninstalled the mod that added it).

Ideally you'd check and remove it on the server side as well but it's mostly harmless on that side, and doing it here would also catch weird desyncs.

uecasm (dismissed):

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/EntityListModuleView.java
Raycoms: I thought the same thing this morning, removed it on load on the serverside

uecasm (dismissed):
Sure, that should be fine.  If the registries are desynched between server and client they'll have bigger problems anyway.

Raycoms:
> Sure, that should be fine. If the registries are desynched between server and client they'll have bigger problems anyway.

Exactly xD

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -426,8 +423,47 @@ public void invalidateRecipes(@NotNull final RecipeManager recipeManager)
...
+    /**
+     * Calculate all monsters.
+     */
+    private void discoverMobs()
+    {
+        monsters = new HashSet<>();
+
+        for (final Map.Entry<RegistryKey<EntityType<?>>, EntityType<?>> entry : ForgeRegistries.ENTITIES.getEntries())
+        {
+            if (entry.getValue().getClassification() == EntityClassification.MONSTER)
```
someaddons, line 443: should we check IMob type here?

src/api/java/com/minecolonies/api/configuration/ServerConfiguration.java
```diff
@@ -443,15 +442,11 @@ protected ServerConfiguration(final ForgeConfigSpec.Builder builder)
             "forge:string"),
           s -> s instanceof String);
 
-        guardResourceLocations = defineList(builder, "guardresourcelocations",
```
someaddons, line 446: want to remove the lang entry for `guardresourcelocations` aswell I guess

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
Raycoms: This gives us the EntityType not the Entity class, this is the best way to handle this in 1.16+

src/api/java/com/minecolonies/api/configuration/ServerConfiguration.java
Raycoms: done

src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
someaddons: ah true we dont really have the entity itself here

someaddons (approved):

PR #7192, opened 2021-06-03
fix 7186

### LINKED ISSUE #7186: Forester sapling issues
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: 0.14.197
- Structurize Version: 0.13.175
- Related Mods and their Versions: n/a

## Expected behavior

With replant on, the forester should keep saplings in stock either in their inventory or hut, and use hut saplings if they don't have enough in inventory, or failing that not put saplings in the hut inventory.

## Actual behavior

 * The forester will put many of their saplings in the hut inventory; I'm having trouble tracking down why this happens; maybe this only happens at low hut level level, since it looks like `BuildingLumberjack.getRequiredItemsAndAmount` should let them keep all the saplings at level 3+?  In any case they sometimes run out of various saplings in their inventory, perhaps just from not collecting enough from dark oak trees.
 * Couriers will take saplings from the hut if the forester doesn't have that type on hand, making it not very effective as a backup sapling source.  This is from `AbstractBuilding.buildingRequiresCertainAmountOfItem` calling the aforementioned `getRequiredItemsAndAmount` function, which won't return things the forester isn't holding.
 * The second point here doesn't really matter though, because forester won't use hut saplings.  When they're done chopping and want to replant, if they don't have enough saplings, they will wait a while, then run to the hut to check for saplings.  Unfortunately, this is *just* a check; they don't actually grab any from the hut even if present, because the relevant code in `EntityAIWorkLumberjack.plantSapling` is:
```java
        if (timeWaited >= MAX_WAITING_TIME / 2 && !checkedInHut && !walkToBuilding())
        {
            isInHut(job.getTree().getSapling());
            checkedInHut = true;
        }
```
 * As I was writing this issue, I saw a forester skip the gathering step after chopping a tree.  Not sure what's up there; he might have interrupted himself with eating and forgot to go back to gathering?

The rest (running to the building, running back, etc.) seems to work, although I'm not sure how; I'm probably missing something about how the AI system works or how `timeWaited` is being calculated.

For the first part of the issue about running out of saplings, I don't have better reproduction steps than "have the forester harvest lots of dark oak trees and wait; they probably won't pick up as many saplings as they use in the long run".

For the second part:
1. Put a type of sapling the forester doesn't have on hand in the hut inventory
2. Request pickup
3. Observe courier taking sapling from hut inventory

For the third part:
1. Plant a tree of some type in range of the forester (while they're paused, or at night)
2. Remove all but one leaf, or remove leaves after they start chopping the tree, careful not to let them pick up any relevant saplings
3. Put saplings of that type in hut inventory and remove them from forester inventory (making sure the courier doesn't get it)
4. Observe forester wait around a while, then walk to hut and back and wait more
5. Check hut inventory; saplings are still present
6. Look at base of just-chopped tree; no saplings planted

For the fourth part, I don't know; I just observed it the once.  I'll see if it happens regularly when chopping the tree makes him hungry.

### Logs

- latest.log: None of this shows up
- crashlog: N/A; no crash

### Notes
My recommendations would be:
 * For 1, document on wiki that foresters work best with only 2 types of tree per hut level, and maybe add option to cut all leaves with hoe after harvesting tree, or wait longer before gathering dark oak and jungle
 * For 2, add override for `BuildingLumberjack.buildingRequiresCertainAmountOfItem` to require a stack of each of the saplings present
 * For 3, fix the excerpted code (not actually sure how myself in an efficient and idiomatic way; I'm sure there's a function for grabbing an item from hut inventory but I didn't see it)
 * No suggestion for 4

Additionally, I did some tests before writing this issue, but for most of it I've had a forester with a level-5 hut and most stats manually maxed to 99 harvesting a rotation of trees of five of the six vanilla overworld types running in the background (the spruce was accidentally out of search range).  I started him with a stack of each sapling, and he's down to 52 dark oak, 63 each of birch, oak, and jungle, and 64 acacia (and of course all the spruce).  Then I added in the part about skipping gathering once, and checked again, and saw only 48 dark oak and the same numbers for the rest.  Maybe everything but dark oak is sustainable already, but I wouldn't be surprised if the others can run down too, just slower.

---
## Viewers

Raycoms:
Closes #7186
Closes #
Closes #

- Fix forester correctly picking up saplings
- Fix forester correctly retaining saplings (both inv and hut)
- Minor cleanup and renaming of the involved methods

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
@@ -688,7 +688,8 @@ public abstract class AbstractEntityAIBasic<J extends AbstractJob<?, J>, B exten
     }
 
     /**
-     * Check all chests in the workers hut for a required item.
+     * Check all racks in the workers hut for a required item.
+     * Transfer to worker if found.
      *
      * @param is the type of item requested (amount is ignored)
      * @return true if a stack of that type was found and transferred.
--- a/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
@@ -20,7 +20,6 @@ import com.minecolonies.api.entity.ai.statemachine.AITarget;
 import com.minecolonies.api.entity.ai.statemachine.states.AIBlockingEventType;
 import com.minecolonies.api.entity.ai.statemachine.states.IAIState;
 import com.minecolonies.api.inventory.InventoryCitizen;
-import com.minecolonies.api.tileentities.AbstractTileEntityColonyBuilding;
 import com.minecolonies.api.tileentities.TileEntityRack;
 import com.minecolonies.api.util.*;
 import com.minecolonies.api.util.constant.IToolType;
@@ -34,6 +33,7 @@ import com.minecolonies.coremod.colony.jobs.JobDeliveryman;
 import com.minecolonies.coremod.entity.pathfinding.EntityCitizenWalkToProxy;
 import com.minecolonies.coremod.util.WorkerUtil;
 import net.minecraft.block.Block;
+import net.minecraft.inventory.ItemStackHelper;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.tags.BlockTags;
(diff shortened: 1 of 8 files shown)


dhouck:
This seems to fix the issues I mentioned in #7186, except for the first (which is made mostly irrelevant if the hut block works better) and the last (which is fair because I gave no reproduction).  I’ll actually test it tonight or tomorrow, probably.

Raycoms:
> This seems to fix the issues I mentioned in #7186, except for the first (which is made mostly irrelevant if the hut block works better) and the last (which is fair because I gave no reproduction). I’ll actually test it tonight or tomorrow, probably.

This code should solve both.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -691,34 +691,19 @@ protected final boolean walkToBuilding()
      * Check all chests in the workers hut for a required item.
      *
      * @param is the type of item requested (amount is ignored)
-     * @return true if a stack of that type was found
+     * @return true if a stack of that type was found and transferred.
      */
-    public boolean isInHut(@Nullable final ItemStack is)
+    public boolean checkAndTransferFromHut(@Nullable final ItemStack is)
```
someaddons, line 697: functionality javadoc is outdated now

marchermans (approved):

someaddons (approved):

PR #7193, opened 2021-06-03
Allow hiring from existing huts

mcmanustfj:
related to https://github.com/ldtteam/minecolonies-features/issues/281 but doesn't entirely close it

- makes it so you can hire people employed in other huts
- orders citizens by employment in hire/fire GUI (citizens with jobs are at the bottom)
- Fixes some iffy logic around hiring people with jobs that already existed and anecdotally caused bugs (this PR makes sure to fire people from their existing job before hiring them to the new job)

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
@@ -192,10 +192,31 @@ public class WindowHireWorker extends AbstractWindowSkeleton
         final int id = citizens.get(row).getId();
         @NotNull final ICitizenDataView citizen = citizens.get(row);
 
+        if (citizen.getWorkBuilding() != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -192,12 +192,33 @@ private void doneClicked(@NotNull final Button button)
         final int id = citizens.get(row).getId();
         @NotNull final ICitizenDataView citizen = citizens.get(row);
 
+        if (citizen.getWorkBuilding() != null) {
```
mcmanustfj, line 195: even if you don't take the rest, someone should probably consider doing something like this for hiring from the library.

mcmanustfj:
need to make this work with school too

Raycoms (changes requested):
As I had already said in the discord. I don't think this is a good idea in general to just add them to the back of the list. This will more likely result in confusion of the players especially when no citizens are available.

I would be inclined to accept this if there was a button in that UI to toggle this on/off.

PR #7235, opened 2021-06-11
Integrated buildings support

someaddons:
Adds functionality to support multiple buildings in one schematic, also adds a parent-child relationship to those so we get a hierarchy buildings belonging together.
Allows both inclusion of pre-leveled buildings from schematics, such as scanning a level 3 warehouse schematic with a level 3 courier included aswell as using an included courier with level 0 which needs to be manually leveled. 

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -709,6 +718,8 @@ public void serializeToView(@NotNull final PacketBuffer buf)
         buf.writeBoolean(isDeconstructed());
 
         getModules(IPersistentModule.class).forEach(module -> module.serializeToView(buf));
+
+        buf.writeBlockPos(getParent());
```
uecasm: It would be more logical to group this with the building's own properties (before the module data).

src/main/java/com/minecolonies/coremod/colony/buildings/views/AbstractBuildingView.java
```diff
@@ -403,6 +416,8 @@ public void deserialize(@NotNull final PacketBuffer buf)
         {
             module.deserialize(buf);
         }
+
+        parent = buf.readBlockPos();
```
uecasm: Other half of that...

src/main/java/com/minecolonies/coremod/network/messages/server/DecorationControllerUpdateMessage.java
```diff
@@ -86,7 +86,7 @@ public void onExecute(final NetworkEvent.Context ctxIn, final boolean isLogicalS
         {
             final BlockState state = player.getEntityWorld().getBlockState(pos);
             final Direction basicFacing = state.get(BlockDecorationController.HORIZONTAL_FACING);
-            ((TileEntityDecorationController) tileEntity).setSchematicName(name);
+            ((TileEntityDecorationController) tileEntity).setSchematicName(name + level);
             ((TileEntityDecorationController) tileEntity).setLevel(level);
```
uecasm, line 90: Isn't including the level in the name going to cause issues for upgrades of existing worlds, where that is missing?  (Also the level is stored separately so the combined name can be reconstructed as needed.)

src/main/java/com/minecolonies/coremod/client/gui/WindowDecorationController.java
```diff
@@ -173,7 +173,7 @@ private void doneClicked()
             {
                 final int level = Integer.parseInt(levelString);
                 Network.getNetwork().sendToServer(new DecorationControllerUpdateMessage(controller.getPos(), name, level));
-                controller.setSchematicName(name);
+                controller.setSchematicName(name + level);
```
uecasm, line 176: Another instance of that...

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
```diff
@@ -130,6 +148,8 @@ public CompoundNBT serializeNBT()
 
         compound.putBoolean(TAG_DECONSTRUCTED, isDeconstructed);
 
+        BlockPosUtil.write(compound, TAG_PARENT_SCHEM, parentSchematic);
+        BlockPosUtil.writePosListToNBT(compound, TAG_CHILD_SCHEM, new ArrayList<>(childSchematics));
```
uecasm, line 152: Near here, for example, is some code that you didn't change that does `getSchematicName() + buildingLevel`, which seems like it would be doing the wrong thing now if you're including the level in the name.  Why not just keep them in separate properties?

uecasm (commented):
Doesn't this whole thing already mostly work without this code?  Perhaps less so for creative pasting, but when starting the parent from level 1 as with normal building the inner hutblocks are then able to be built up from level 1 as normal.  Either way the schematic needs a ton of placeholder blocks to cope with it.  And the hutblock itself doesn't get replaced because it's tagged as indestructible.

src/main/java/com/minecolonies/coremod/client/gui/WindowDecorationController.java
someaddons: it is required because scanning does use the exact same format, namely saving schematic name with its level on the end

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
someaddons: these are the positions of the parent/child not the schematic names

src/main/java/com/minecolonies/coremod/network/messages/server/DecorationControllerUpdateMessage.java
someaddons: no its not an upgrade issue, the level here is only stored to have the correct schematic reference used at places outside the controller itself

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
uecasm: I know that.  I'm referring to several lines above that in the same function.  Github won't let me directly comment on lines that aren't in the diff.

uecasm: The point being that there are several places that assume the level number is not appended to the schematic name already, so if you're changing that, you need to change those.

someaddons: no you're confusing buildings with the tileentities now, the schematic data is put into the anchors/tileentities, the buildings have their own hard-coded required name to them, which is different from the tile data and doesnt need to be the same aswell

uecasm: Ok, all right then.  It's hard to tell them apart just from looking at the diff.

someaddons: ye its the same method name, just one is on the building which returns hard-coded hut values like "guardtower" and the other is on the TE which returns the schematics filename(which includes the level)

src/api/java/com/minecolonies/api/tileentities/AbstractTileEntityColonyBuilding.java
```diff
@@ -208,6 +209,24 @@ public void read(final BlockState state, @NotNull final CompoundNBT compound)
         readSchematicDataFromNBT(compound);
     }
 
+    @Override
+    public void readSchematicDataFromNBT(CompoundNBT originalCompound)
```
Raycoms: @NotNull final original?

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -629,7 +638,7 @@ public void removeWorkOrder()
     {
         for (@NotNull final WorkOrderBuild o : colony.getWorkManager().getWorkOrdersOfType(WorkOrderBuild.class))
         {
-            if (o.getBuildingLocation().equals(getID()) && (o instanceof WorkOrderBuildBuilding || o instanceof WorkOrderBuildRemoval))
+            if (o.getSchematicLocation().equals(getID()) && (o instanceof WorkOrderBuildBuilding || o instanceof WorkOrderBuildRemoval))
```
Raycoms, line 645: maybe getSchematicAnchorLocation

src/main/java/com/minecolonies/coremod/network/messages/server/BuildToolPlaceMessage.java
```diff
@@ -380,7 +380,6 @@ private static void setupBuilding(
             if (!(building instanceof IRSComponent))
             {
                 ConstructionTapeHelper.removeConstructionTape(building.getCorners(), world);
-                building.calculateCorners();
```
Raycoms, line 383: If the corner calculation is not needed, the removeConstructionTape shouldn't be either. Needs some good testing though.

someaddons: corner calc happens before now, inbuilt

Raycoms: so the removal can go away.

someaddons: but isnt that to remove construction tape? Shouldnt have anything todo with the corner calc

Raycoms: Yeah I'm not 100% sure tbh. I'm not sure anymore why we removed and re-add here and especially why we did the corner calc inbetween.

uecasm: The remove and re-add is likely _because_ the calc was between.  In case it changed, it needs to move the tape.  So yeah, if you're not recalculating there any more it probably doesn't need to do either thing with the tape.  Unless the tape might not be there at all at this point?  (In which case the add might still be needed.)

someaddons: I guess from the looks of it it was an old hacky fix for removing previously set tape at a different location than normal building corners, possibly from times where we didnt load the schematic to check its size yet. I'll just remove it and we'll see what happens xD

Raycoms (dismissed):

src/main/resources/META-INF/mods.toml
```diff
@@ -53,11 +53,5 @@ MineColonies is a colony simulator within Minecraft! There are numerous types of
...
     ordering="NONE"
     side="BOTH"
-# Here's another dependency
-[[dependencies.minecolonies]]
-    modId="structurize"
-    mandatory=true
-    versionRange="[0.13.177-ALPHA]"
-    ordering="AFTER"
-    side="BOTH"
+
```
ravenbuilder934: ...Structurize is no longer a dep? What?

Raycoms: Debug fail add =D

someaddons: yep, fixed :D

someaddons: raven pls stop auto merging the newest, its annoying!

Raycoms (dismissed):

PR #7253, opened 2021-06-16
Feature/crafting modules

Raycoms:
- Finish crafting modules
-
-

This is a draft. I still need a few things.

Review please


src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -2012,5 +2012,5 @@
   "com.minecolonies.coremod.gui.citizen.family.partner": "Partner",
   "com.minecolonies.coremod.gui.citizen.family.siblings": "Siblings",
   "com.minecolonies.coremod.gui.citizen.family.children": "Children",
-  "com.minecolonies.coremod.gui.citizen.family.unknown": "Unknown"
+  "com.minecolonies.coremod.gui.citizen.family.unknown": "Unknown",
```
ravenbuilder934: wut? are you *trying* to make it crash?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -141,6 +140,14 @@ protected boolean isRecipeCompatibleWithCraftingModule(final IToken<?> token)
     {
         final IGenericRecipe recipe = GenericRecipe.of(token);
         if (recipe == null) return false;
+        for (final CustomRecipe rec : CustomRecipeManager.getInstance().getRecipes(building.getJobName()))
+        {
+            if (rec.getRecipeStorage().equals(IColonyManager.getInstance().getRecipeManager().getRecipes().get(token)))
+            {
+                return true;
+            }
+        }
```
uecasm: This code block needs to be skipped when `building == null`.

But it also seems inefficient, `isRecipeCompatibleWithCraftingModule` gets called a **lot** to establish JEI recipe compatibility.  It would be best to move this code to the caller of this method rather than having it here.  I presume it's only here for the legacy recipe persistence upgrade?

uecasm (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
uecasm: Actually, I was confusing it with another method, the JEI integration doesn't call this particular one directly so it probably shouldn't encounter `building == null`.  But I still think this check probably shouldn't be here, as it could confuse the recipe teaching (you shouldn't be able to teach the custom recipes).

src/main/resources/assets/minecolonies/lang/en_us.json
Raycoms: It's a draft :joy:

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: true, I'll move it into the serialization verification instead, I'm calling this to not remove the pre-taught recipes.

someaddons (dismissed):

src/main/java/com/minecolonies/coremod/client/gui/AbstractWindowModuleBuilding.java
```diff
@@ -24,7 +24,7 @@
 public abstract class AbstractWindowModuleBuilding<B extends IBuildingView> extends AbstractModuleWindow
 {
     /**
-     * Type B is a class that extends {@link com.minecolonies.coremod.colony.buildings.AbstractBuildingWorker.View}.
+     * Type B is a class that extends {@link com.minecolonies.coremod.colony.buildings.AbstractBuildingWorkerView}.
```
dupontct: Isn't this file at com.minecolonies.coremod.colony.buildings.views.AbstractBuildingWorkerView?

src/api/java/com/minecolonies/api/colony/buildings/modules/ICreatesResolversModule.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Interface describing core building stats.
+ * The first core stats module that is found in the building will define the values.
+ */
+public interface ICreatesResolversModule extends IBuildingModule
+{
+    /**
+     * Get the max number of inhabitants this module allows.
+     * @return the modules max number of assigned citizens.
```
uecasm: I don't think either of these javadocs are correct...

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -22,37 +59,153 @@
...
-        //  alternatively, we could keep all recipes in a combined "recipes" tag, as now, with
-        //  the modules smart enough to figure out which one "owns" each recipe for UI purposes
-        //  (which will usually be obvious by looking at the intermediate block).  this would
-        //  be a simpler upgrade path but might be more error-prone or less performant.
+        final CompoundNBT moduleCompound = new CompoundNBT();
+        @NotNull final ListNBT recipesTagList = recipes.stream()
+                                                  .map(iToken -> StandardFactoryController.getInstance().serialize(iToken))
+                                                  .collect(NBTUtils.toListNBT());
+        moduleCompound.put(TAG_NEW_RECIPES + getId(), recipesTagList);
+        compound.put(getId(), moduleCompound);
```
uecasm, line 175: Why the doubled use of `getId()`?  Once should be enough to satisfy anti-collision.
(And in fact if you're using the `getId()` outer compound then there's no need to use `TAG_NEW_RECIPES` at all, you could just use `TAG_RECIPES` again since this is still distinct from the old compound.  See below for why that might be useful.)

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -22,37 +59,153 @@
...
+            final ListNBT recipesTags = compoundNBT.getList(TAG_NEW_RECIPES + getId(), Constants.NBT.TAG_COMPOUND);
+            for (int i = 0; i < recipesTags.size(); i++)
+            {
+                final IToken<?> token = StandardFactoryController.getInstance().deserialize(recipesTags.getCompound(i));
+                if (!recipes.contains(token))
+                {
+                    recipes.add(token);
+                    IColonyManager.getInstance().getRecipeManager().registerUse(token);
+                }
+            }
```
uecasm, line 200: Since the format of `recipesTags` is identical between these two branches, it should be read outside the conditional so that the code isn't repeated.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -22,37 +59,153 @@
...
+    }
+
+    /**
+     * Check if the recipe is a pre-taught recipe through datapack.
+     * @param token the recipe to check.
+     * @return true if so.
+     */
+    private boolean isPreTaughtRecipe(final IToken<?> token)
+    {
+        for (final CustomRecipe rec : CustomRecipeManager.getInstance().getRecipes(getCustomRecipeKey()))
```
uecasm: The only place this is called from already has a list of custom recipes in the `crafterRecipes` variable, which could be passed in here rather than re-queried.

Also, not sure if this is intended, but as written I think this will delete any added-by-datapack recipes that are later "improved".  The place where this is called already handles regular datapack recipes (removals, though not edits) via checking the source, but improved recipes don't have a source.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -67,8 +220,481 @@ public void serializeToView(PacketBuffer buf)
         buf.writeBoolean(this.canLearnCraftingRecipes());
         buf.writeBoolean(this.canLearnFurnaceRecipes());
         buf.writeBoolean(this.canLearnLargeRecipes());
+
+        final List<IRecipeStorage> storages = new ArrayList<>();
+        final Map<ResourceLocation, CustomRecipe> crafterRecipes = CustomRecipeManager.getInstance().getAllRecipes().getOrDefault(building.getJobName() + "_" + getId(), Collections.emptyMap());
```
uecasm: ++Miralderps I think, but this should be using `getCustomRecipeKey()` rather than manually concatenating the job and id.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -67,8 +220,481 @@ public void serializeToView(PacketBuffer buf)
...
+                final List<IToken<?>> assignedTasks = new ArrayList<>(citizen.getJob(AbstractJobCrafter.class).getAssignedTasks());
+                assignedTasks.addAll(citizen.getJob(AbstractJobCrafter.class).getTaskQueue());
+
+                for (final IToken<?> taskToken : assignedTasks)
+                {
+                    final IRequest<? extends PublicCrafting> request = (IRequest<? extends PublicCrafting>) building.getColony().getRequestManager().getRequestForToken(taskToken);
+                    final IRecipeStorage recipeStorage = IColonyManager.getInstance().getRecipeManager().getRecipes().get(request.getRequest().getRecipeStorage());
+                    if (recipeStorage != null)
+                    {
+                        recipes.add(new Tuple<>(recipeStorage, request.getRequest().getCount()));
```
uecasm, line 317: If I'm understanding this correctly, it looks like it will count the recipes twice if there are two modules on the building, and so try to keep twice as many items as it should?  It doesn't look like there's any module-specific filter here.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -67,8 +220,481 @@ public void serializeToView(PacketBuffer buf)
...
+     * @return whether the recipe can bee added based on tokens.
+     */
+    protected Optional<Boolean> canRecipeBeAddedBasedOnTags(final IRecipeStorage storage)
+    {
+        if (storage == null)
+        {
+            return Optional.of(false);
+        }
+
+        final String crafterName = building.getJobName().toLowerCase();
```
uecasm: Given the tag definitions, presumably this needs to sometimes have a `_smelting` suffix (e.g. glassblower), and so probably needs to be calling a virtual method only overridden in those specific cases?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -114,6 +820,26 @@ public boolean isRecipeCompatible(@NotNull final IGenericRecipe recipe)
...
+
+        @Override
+        public boolean canRecipeBeAdded(@NotNull final IToken<?> token)
+        {
+            if (!super.canRecipeBeAdded(token))
+            {
+                return false;
+            }
+            return isRecipeCompatibleWithCraftingModule(token);
+        }
```
uecasm: The super method already checks `isRecipeCompatibleWithCraftingModule`, so this can be removed?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -134,6 +860,26 @@ public boolean isRecipeCompatible(@NotNull final IGenericRecipe recipe)
...
```
uecasm: Same here.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -148,7 +894,28 @@ public boolean isRecipeCompatible(@NotNull final IGenericRecipe recipe)
...
```
uecasm: And here.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -148,7 +894,28 @@ public boolean isRecipeCompatible(@NotNull final IGenericRecipe recipe)
         @Override
         public boolean canLearnLargeRecipes() { return false; }
 
+        //todo need to implement this in the respective workers
```
uecasm: It was not intended that this would ever need to return true on the custom module, because this should only be called when trying to learn recipes, and this module can't be taught, it can only use crafterrecipes.

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/CraftingModuleView.java
```diff
@@ -44,34 +95,63 @@ public JobEntry getJobEntry()
...
-        return false;   // todo this later
+        return isVisible;
     }
 
     @Override
+    @OnlyIn(Dist.CLIENT)
     public Window getWindow()
     {
-        return null;    // todo this later
+        return new WindowListRecipes(buildingView, Constants.MOD_ID + ":gui/layouthuts/layoutlistrecipes.xml", this);
```
uecasm, line 133: Given this string format, shouldn't this be a `ResourceLocation` parameter instead?

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/CraftingModuleView.java
```diff
@@ -44,34 +95,63 @@ public JobEntry getJobEntry()
...
     public String getIcon()
     {
-        return "crafting";
+        return id;
     }
 
     @Override
     public String getDesc()
     {
         return "com.minecolonies.coremod.gui.workerhuts.recipe";
```
uecasm: This might want to be different based on id as well, so you can have a "Crafting" tab and a "Smelting" tab.  (Though I'm still dubious that "custom" should be separate from "crafting", especially in the UI.)

uecasm (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/CraftingModuleView.java
Raycoms: In theory, potentially possible, in practise, that would require changes in all windows and is out of the scope of the PR.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: True on the first, second is difficult, this is backwards compat code to clean-up old recipes, I can't think of a better way we could handle this potentially. Shouldn't the recipeStorage still equal?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: Good catch

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: True, will fix

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: This is quite funny. I think we need input of @Mekle001 here. Most buildings use CraftingUtils.isRecipeCompatibleBasedOnTags which I'm already calling the right way for the right workers.

This method is only called by a) Cook and b) Farmer, by the cook in the getFirstRecipe and by the farmer in the canAddRecipe code. Do we need this code at all?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: @Mekle001 any idea?

uecasm: If it's just for the upgrade path, you could limit the damage by setting a boolean flag when you deserialize with the old NBT tag, and only doing this code when true.  Means you might still lose the improved recipes on upgrade but at least it only happens once.

And if it's only for removing datapack recipes, you could _in theory_ only remove them if recipeSource is not null -- but in practice that's not sufficient, as colonies started before the recipeSource was made more consistent (as of a few months ago) still wouldn't have it set, due to the way the recipes upgraded.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingFarmer.java
```diff
@@ -728,5 +709,23 @@ public boolean isRecipeCompatible(@NotNull final IGenericRecipe recipe)
...
+            Optional<Boolean> isRecipeAllowed;
+
+            if (!super.canRecipeBeAdded(token))
+            {
+                return false;
+            }
+
+            isRecipeAllowed = super.canRecipeBeAddedBasedOnTags(token);
+            return isRecipeAllowed.orElse(false);
+        }
```
uecasm: This looks like a merge derp or something, calling super method twice is silly and calling `canRecipeBeAddedBasedOnTags` happens in the super method anyway.  So it can probably be removed entirely, like most of the other crafters.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
uecasm: Actually all the modules call this, which means that they're checking the tags twice, which is probably wrong.  This is likely an incorrect translation when moving this from the building to the module.

uecasm: Sorry, no, I was searching the wrong method.  You were right, it's just the farmer and cook.  The farmer looks like a merge derp (reviewed separately).  It looks like the usage of the cook should be replaced with `isRecipeCompatibleWithCraftingModule`.  And then this code can go away.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingFarmer.java
Raycoms: True, I'll remove it from here.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: Talked with Mekle, yes, we can remove this.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -520,8 +520,8 @@ public IRecipeStorage getFirstRecipe(final Predicate<ItemStack> stackPredicate)
 
             //If we didn't have a stored recipe, see if there is a smelting recipe that is also a food output, and use it.
             storage = FurnaceRecipes.getInstance().getFirstSmeltingRecipeByResult(stackPredicate);
-            Optional<Boolean> validRecipe =  canRecipeBeAddedBasedOnTags(storage);
-            if(storage != null && ISFOOD.test(storage.getPrimaryOutput().getStack()) && (!validRecipe.isPresent() || validRecipe.get()))
+            boolean validRecipe = isRecipeCompatibleWithCraftingModule(storage.getToken());
```
uecasm: `isRecipeCompatibleWithCraftingModule(storage.getToken())` => `isRecipeCompatible(GenericRecipe.of(storage))`

A bit more efficient, since it's going "forwards" in the data chain and doesn't have to look up the recipe again.  _Technically_ the latter is nullable but it only actually returns null if the parameter is null, so this usage is safe.

uecasm: (Also probably `isRecipeCompatibleWithCraftingModule` should just become another overload of `isRecipeCompatible` since it's actually **on** the crafting module now, instead of on the building.  But that's optional.)

uecasm (dismissed):
I'd still prefer it if the json files just had the crafter name without suffix and the loader figured out which module to send it to, for the sake of tidiness (and datapack simplicity and backwards compatibility).  But that's a style choice, not a blocker.  Otherwise looks good as far as I can see.

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/CraftingModuleView.java
```diff
@@ -44,34 +95,63 @@ public JobEntry getJobEntry()
...
     {
-        return "crafting";
+        return id;
     }
 
     @Override
     public String getDesc()
     {
-        return "com.minecolonies.coremod.gui.workerhuts.recipe";
+        return "com.minecolonies.coremod.gui.workerhuts." + id ;
```
uecasm: Actually wait, no, this should be `"com.minecolonies.coremod.gui.workerhuts.recipe." + id` instead, otherwise the strings are missing.

Raycoms: Uh, happens when you do things distracted xD

uecasm (approved):

someaddons (approved):

marchermans (approved):

PR #7258, opened 2021-06-18
[HotFix] Fix miners using default blockstate tool instead of real blockstate tool.

Asherslab:
Discussed on discord here:
https://discord.com/channels/139070364159311872/741531816422539384/855232112021602334

Mods can change the harvest tool for a blockstate outside the default one.

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
@@ -1148,7 +1148,7 @@ public abstract class AbstractEntityAIBasic<J extends AbstractJob<?, J>, B exten
      */
     private void requestTool(@NotNull final BlockState target, final BlockPos pos)
     {
-        final IToolType toolType = WorkerUtil.getBestToolForBlock(world.getBlockState(pos), target.getBlockHardness(world, pos));
+        final IToolType toolType = WorkerUtil.getBestToolForBlock(target, target.getBlockHardness(world, pos));
         final int required = WorkerUtil.getCorrectHarvestLevelForBlock(target);
         if (getOwnBuilding().getMaxToolLevel() < required && worker.getCitizenData() != null)
         {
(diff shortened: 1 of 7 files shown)


src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -1148,7 +1148,7 @@ public final boolean holdEfficientTool(@NotNull final Block target, final BlockP
      */
     private void requestTool(@NotNull final Block target, final BlockPos pos)
     {
-        final IToolType toolType = WorkerUtil.getBestToolForBlock(target, target.getDefaultState().getBlockHardness(world, pos));
+        final IToolType toolType = WorkerUtil.getBestToolForBlock(world.getBlockState(pos), target.getDefaultState().getBlockHardness(world, pos));
```
Asherslab: kept getBlockHardness here because otherwise it would require passing in the world and pos, no real benefit of doing it that way

uecasm: I don't quite follow this comment, it already has the world and pos.  Shouldn't it be using `world.getBlockState(pos)` for both (or better, as an extra line above)?

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
```diff
@@ -1188,7 +1188,7 @@ private void updateToolFlag(@NotNull final IToolType toolType, final int require
      */
     protected int getMostEfficientTool(@NotNull final Block target, final BlockPos pos)
     {
-        final IToolType toolType = WorkerUtil.getBestToolForBlock(target, target.getDefaultState().getBlockHardness(world, pos));
+        final IToolType toolType = WorkerUtil.getBestToolForBlock(world.getBlockState(pos), target.getDefaultState().getBlockHardness(world, pos));
```
uecasm: Here too.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIBasic.java
uecasm: Or better still, change the parameter from Block to BlockState, since all the callers are already doing the getBlockState anyway.

Raycoms: Agreed

Asherslab:
so, as i said to raycoms during this. most of the code is stuck in tech debt pre-blockstates. i've adapted a few methods very quickly but i don't have time to do a full refactor, only aiming to fix this bug. anyway, a once over refactor of this stuff would be very good

Raycoms (approved):

uecasm (approved):

PR #7300, opened 2021-06-25
Raid starting difficulty adjustment/remove hutblock schems

someaddons:
increase starting raid difficulty a bit, remove no longer needed hutblock schematics which just confuse users occasionally

Review please


uecasm:
Why are the hutblock schems "unneeded"?  Aren't they still required when creating schematics?  Or did I miss a change?

someaddons:
you can place them in creative mod now while sneaking

src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
```diff
@@ -118,7 +118,7 @@
     /**
      * The dynamic difficulty of raids for this colony
      */
-    private int raidDifficulty = MIN_RAID_DIFFICULTY;
+    private int raidDifficulty = 5;
```
Raycoms: why adjust the value instead of the constant?

someaddons: because I don't want to increase the minimum scaling, but the starting level

Raycoms: Should we make it a START_RAID_DIFFICULTY then?

someaddons: no not used elsewhere

Raycoms: I know it's not used elsewhere, but it's to not have it be a magic number but "explained"

ravenbuilder934 (approved):

PR #7315, opened 2021-06-27
Add Rain and Raid interaction

ToMe25:
Closes #
Closes #
Closes #

- Add citizen interactions for when its raining and when the colony is being raided.
- Being raided is more important then rain, so if both events happen at the same time the raid interaction is going to be the one shown.
- The messages might not be perfect yet.

Review please


src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
```diff
@@ -291,5 +291,11 @@ public static void init()
         InteractionValidatorRegistry.registerStandardPredicate(new TranslationTextComponent(COM_MINECOLONIES_COREMOD_BEEKEEPER_NOFLOWERS),
           citizen -> citizen.getWorkBuilding() instanceof BuildingBeekeeper
                        && ((BuildingBeekeeper) citizen.getWorkBuilding()).getModuleMatching(ItemListModule.class, m -> m.getId().equals(BUILDING_FLOWER_LIST)).getList().isEmpty());
+
+        InteractionValidatorRegistry.registerStandardPredicate(new TranslationTextComponent(COM_MINECOLONIES_COREMOD_ENTITY_CITIZEN_RAINING),
+          citizen -> citizen.getEntity().isPresent() && citizen.getEntity().get().getEntityWorld().isRaining());
```
Raycoms: this should also have a !raid check, else shipit

ToMe25: Wouldn't that cancel the interaction of a rain gets started after the citizen already complains about the rain?

Raycoms: Yes, this will cancel the rain interaction of the citizen the moment a raid started.

ToMe25: I will look into always triggering the raid interaction on raid begin and adding this then.

ToMe25: This change had a bigger impact then I expected.
Its done now :)

Raycoms (dismissed):

ToMe25:
Uh btw citizens that already sleep when it starts raining don't get the rain interaction, but since they also don't seem to get the bad weather icon I thought thats fine.

Raycoms:
> Uh btw citizens that already sleep when it starts raining don't get the rain interaction, but since they also don't seem to get the bad weather icon I thought thats fine.

Uhh, that is actually a bug then.

ToMe25:
Well, the method that does both of those things atm doesn't get called on sleeping citizens.
I assume for the icon its fine because they already have the sleeping icon then, so thats probably why no one bothered to change it.
If thats not how it should be I will look into giving them the interaction if it starts raining while they sleep.
Should they also get the rain interaction if they are currently working?
I assume yes.

Raycoms:
> Well, the method that does both of those things atm doesn't get called on sleeping citizens.
> I assume for the icon its fine because they already have the sleeping icon then, so thats probably why no one bothered to change it.
> If thats not how it should be I will look into giving them the interaction if it starts raining while they sleep.
> Should they also get the rain interaction if they are currently working?
> I assume yes.

It's not necessary while they are sleeping, but after they get up, it should show.

ToMe25:
You mean the Icon should show up after, or the interaction?
Or both?
Anyways, as they want to sleep while its raining there shouldn't be an after.

ToMe25:
I just checked, and if the night ends but its still raining they get the rain interaction.
Same goes for if a raid ends and its still raining.

Raycoms:
> I just checked, and if the night ends but its still raining they get the rain interaction.
> Same goes for if a raid ends and its still raining.

great  :shipit:

someaddons (dismissed):

marchermans (approved):

PR #7318, opened 2021-06-28
Fix server-side module views

uecasm:
- Get server builds working again after 200aa5f5716c (couldn't classload `ClientPlayerEntity`)

Review please

This is the minimum fix required to get `runServer` to work.  Probably something more comprehensive is needed to properly isolate the view classloading and prevent this reoccurring in the future, for example to add an extra layer of indirection (e.g. `() -> FloristFlowerListModuleView::new`, similar to how the building views are declared), or something else.  Or finding what it was about that module in particular that couldn't classload on server and making it avoid it.

### DIFF
--- a/src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
+++ b/src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
@@ -48,7 +48,7 @@ public final class ModBuildingsInitializer
                                  .setBuildingProducer(BuildingArchery::new)
                                  .setBuildingViewProducer(() -> BuildingArchery.View::new)
                                  .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.ARCHERY_ID))
-                                 .addBuildingModuleProducer(MinimumStockModule::new, MinimumStockModuleView::new)
+                                 .addBuildingModuleProducer(MinimumStockModule::new, () -> MinimumStockModuleView::new)
                                  .addBuildingModuleProducer(BedHandlingModule::new)
                                  .createBuildingEntry();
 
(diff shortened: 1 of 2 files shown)


marchermans (changes requested):
Then if you know how to fix this, i suggest you fix it fully. Yes a Supplier<Supplier<View>> suffices to prevent the class loader from triggering.

Since this is just a change to the API, as well as some find and replace work. This should not take more then 15 minutes, even when done by hand.

uecasm:
It was a possible way that it could be fixed, but I wasn't sure if it was the right way to fix it.  Hence why I didn't just do it.

marchermans:
It is the way forge fixes this in the cases where it is needed. See things like ISTER registration on the Item.Properties builder (since it happens there as well)

So yeah that is the right way todo it 😉

uecasm:
Alrighty.

marchermans (approved):

src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
```diff
@@ -48,7 +48,7 @@ public static void init(final RegistryEvent.Register<BuildingEntry> event)
                                  .setBuildingProducer(BuildingArchery::new)
                                  .setBuildingViewProducer(() -> BuildingArchery.View::new)
                                  .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.ARCHERY_ID))
-                                 .addBuildingModuleProducer(MinimumStockModule::new, MinimumStockModuleView::new)
+                                 .addBuildingModuleProducer(MinimumStockModule::new, () -> MinimumStockModuleView::new)
```
Raycoms, line 51: Won't those double lambdas create java version issues like it did in structurize?

Nightenom: Symptom for this issue is class being interface

Raycoms: Ah okay, so it should be fine.

marchermans: This is used everywhere by forge to deal with this, and can not really cause an issue.

PR #7329, opened 2021-06-29
Raider spawning

someaddons:
- Improve raider spawning, using pathing to reach close to the pre-chosen spawnpoint
- Simplify guard random pathing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
@@ -276,11 +276,6 @@ public class RaidManager implements IRaiderManager
 
         final int raidLevel = getColonyRaidLevel();
         int amount = calculateRaiderAmount(raidLevel);
-        amount = 5;
-      /*  if (amount <= 0 || raidLevel < MIN_REQUIRED_RAIDLEVEL)
-        {
-            return;
-        }*/
 
         // Splits into multiple raids if too large
         final int raidCount = Math.max(1, amount / BIG_HORDE_SIZE);
--- a/src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
@@ -9,9 +9,9 @@ import com.minecolonies.api.colony.colonyEvents.EventStatus;
 import com.minecolonies.api.colony.colonyEvents.IColonyEvent;
 import com.minecolonies.api.colony.colonyEvents.IColonyRaidEvent;
 import com.minecolonies.api.colony.managers.interfaces.IRaiderManager;
+import com.minecolonies.api.entity.pathfinding.PathResult;
 import com.minecolonies.api.util.BlockPosUtil;
 import com.minecolonies.api.util.Log;
-import com.minecolonies.api.util.Tuple;
 import com.minecolonies.api.util.WorldUtil;
 import com.minecolonies.coremod.MineColonies;
 import com.minecolonies.coremod.colony.Colony;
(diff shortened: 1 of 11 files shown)


src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
```diff
@@ -274,10 +276,11 @@ public void raiderEvent(final String raidType)
 
         final int raidLevel = getColonyRaidLevel();
         int amount = calculateRaiderAmount(raidLevel);
-        if (amount <= 0 || raidLevel < MIN_REQUIRED_RAIDLEVEL)
+        amount = 5;
+      /*  if (amount <= 0 || raidLevel < MIN_REQUIRED_RAIDLEVEL)
```
Raycoms: debug code?

src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -573,7 +574,7 @@ protected Path search()
             totalNodesVisited++;
 
             // Limiting max amount of nodes mapped
-            if (totalNodesVisited > MineColonies.getConfig().getServer().pathfindingMaxNodes.get() || totalNodesVisited > maxRange * maxRange)
```
Raycoms, line 576: ?

src/main/java/com/minecolonies/coremod/colony/managers/RaidManager.java
someaddons: yep^^

src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -1375,6 +1376,11 @@ protected SurfaceType isWalkableSurface(@NotNull final BlockState blockState, fi
             return SurfaceType.NOT_PASSABLE;
         }
 
+        if (blockState.getBlock() == Blocks.WATER || (fluid != null && !fluid.isEmpty() && (fluid.getFluid() == Fluids.WATER || fluid.getFluid() == Fluids.FLOWING_WATER)))
```
Raycoms: don't we have an isWater or so util?

src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
someaddons: the static cap doesnt really fit since depending on the job we have more or less amounts of nodes, esp the spawn pathing is rather big and would unnecessarily capped by a static amount here

src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
someaddons: ye we do, though can't re-use the fluid state queried for the lava above then

Raycoms: maybe break it off into a submethod then that is called, so we don't have different waterchecks all over the place potentially

Raycoms: like a method that gets the fluidstate already,

src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -499,12 +499,29 @@ private static boolean calculateSwimming(@NotNull final IWorldReader world, @Not
...
+    /**
+     * Check if the block at this position is actually some kind of waterly fluid.
+     * @param pos the pos in the world.
+     * @param state existing blockstate or null
+     * @param fluidState existing fluidstate or null
+     * @return true if so.
+     */
+    private static boolean isWater(@NotNull final IWorldReader world, final BlockPos pos, @Nullable BlockState state, @Nullable FluidState fluidState)
+    {
+        if (state == null)
```
Raycoms, line 521: Uh, this is ugly, reusing a param, rather make it a temp var that gets the input from the param =D

PR #7335, opened 2021-06-30
Allow stone smelter to learn more recipes

vroni92:
Closes #
Closes #
Closes #

- *Actually* fixes the problem I had with the stone smelter :D
-
-

Review please


Raycoms (approved):

uecasm:
Unless we explicitly don't want the 2x2 crafters to scale accordingly too (the research description suggests that they should, although it's rarely needed in practice), the condition could just be removed entirely...

PR #7337, opened 2021-06-30
Fix decocontroller not leveling up

someaddons:
-Fix decocontroller not leveling up

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/tileentities/TileEntityDecorationController.java
+++ b/src/main/java/com/minecolonies/coremod/tileentities/TileEntityDecorationController.java
@@ -172,10 +172,30 @@ public class TileEntityDecorationController extends TileEntity implements IBluep
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/tileentities/TileEntityDecorationController.java
```diff
@@ -171,11 +171,31 @@ public void setSchematicCorners(final BlockPos pos1, final BlockPos pos2)
...
+            this.basicFacing = Direction.byHorizontalIndex(compound.getInt(TAG_FACING));
+        }
+    }
+
     @Override
     public void read(final BlockState state, final CompoundNBT compound)
     {
         super.read(state, compound);
-        readSchematicDataFromNBT(compound);
+        IBlueprintDataProvider.super.readSchematicDataFromNBT(compound);
```
uecasm, line 198: Why doesn't this just call the override instead of repeating the code of it?

PR #7344, opened 2021-07-01
Reload TEs with invalid data from default name.

### LINKED ISSUE #7342: Minecraft crashes loading SP saved game world
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: minecolonies-0.14.256-ALPHA-universal
- Structurize Version: structurize-0.13.197-ALPHA-universal
- Related Mods and their Versions: Forge-36.1.32

## Expected behavior

I load my saved SP game world and have it load properly.

## Actual behavior

The saved SP world starts the loading process as expected, then I get a brief view of the HUD and an brief view of sky and then the screen goes black for about 4-5 seconds, then the launcher shows an Exit Code: 0 error on the main screen.

Checking the crash logs, the error seems to indicate that there's an issue with the Mystical Site

1. Create a SP world using a version of Minecolonies, at least 15 versions back.
2. Start a colony and as one of the buildings, build a Mystical site. To properly repro, you'll have to research this.
3. Upgrade to the .256 version of Minecolonies.
4. Start Minecraft and then try to open world.
5. World tries to load and crashes the game.

### Logs

- latest.log: https://gist.github.com/kedamono/826e10b3c37e0cdc9bbd01bf4587cd16
- crashlog: https://gist.github.com/kedamono/b7646763516b10b6c2bacd5f32bcbaf5

### Notes

I created a new world with the same custom modpack and it loaded just fine. Didn't try to start a colony, but otherwise the world was working just fine. 

---
## Viewers

Raycoms:
Closes #7342
Closes #
Closes #

- Reload TEs with invalid data from default name.
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
@@ -403,14 +403,11 @@ public abstract class AbstractSchematicProvider implements ISchematicProvider, I
 
         final LoadOnlyStructureHandler structure = new LoadOnlyStructureHandler(colony.getWorld(), getPosition(), structureName, new PlacementSettings(), true);
         final Blueprint blueprint = structure.getBluePrint();
-        if (blueprint != null)
+        blueprint.rotateWithMirror(BlockPosUtil.getRotationFromRotations(getRotation()), isMirrored() ? Mirror.FRONT_BACK : Mirror.NONE, colony.getWorld());
+        final BlockInfo info = blueprint.getBlockInfoAsMap().getOrDefault(blueprint.getPrimaryBlockOffset(), null);
+        if (info.getTileEntityData() != null)
         {
-            blueprint.rotateWithMirror(BlockPosUtil.getRotationFromRotations(getRotation()), isMirrored() ? Mirror.FRONT_BACK : Mirror.NONE, colony.getWorld());
-            final BlockInfo info = blueprint.getBlockInfoAsMap().getOrDefault(blueprint.getPrimaryBlockOffset(), null);
-            if (info.getTileEntityData() != null)
-            {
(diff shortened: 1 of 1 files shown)


someaddons (approved):

uecasm:
Isn't this still going to crash if the schematic gets renamed/deleted at some point after originally built?  Unless there's something higher level which catches it and deletes the building (which the PR suggested was not the case).

Raycoms:
> Isn't this still going to crash if the schematic gets renamed/deleted at some point after originally built? Unless there's something higher level which catches it and deletes the building (which the PR suggested was not the case).

Yes, if you remove a style this might crash, but this would already cause numerous other problems at other places already. We don't really have recovery code at all for this case.

uecasm:
It has happened before, though, and not that long ago, with people renaming buildings and decos to fix typos.  That simply made the building non-upgradable until deconstruct and rebuild, but that's better than crashing the world.

Raycoms:
> It has happened before, though, and not that long ago, with people renaming buildings and decos to fix typos. That simply made the building non-upgradable until deconstruct and rebuild, but that's better than crashing the world.

Deconstruct is not possible, would have to break the block. And yeah, decos are different.

uecasm:
Now that it's saving the footprint to the TE data properly, it actually would be theoretically possible to deconstruct both buildings and decos (that have a controller) even without a blueprint file...

But that's besides the point.  Crashing the world is bad, we should do everything possible to avoid that.

Raycoms:
In theory it would be possible yeah, in practice we do matching on placeholder blocks etc during deconstruction to avoid destroying things around it etc. Yeah, we should probably add some sort of mechanism that if the building is missing loads the TE in an "invalid" state and reports it to the player and the UI remains not openable.

uecasm:
And oh look, exactly this just happened to someone: https://pastebin.com/nsN4FHm1

PR #7348, opened 2021-07-02
Bump minimum MC/Forge versions to match build.properties

uecasm:
- Bump minimum MC/Forge versions to match build.properties
- Make mods.toml update automatically when bumping versions in the future
- Add optional mod dependencies

Review please

Tested by trying to build mcol for 1.16.4, but that failed due to something related to the zombie villager feature.  So I believe this limit is accurate and the toml shouldn't be over-promising.  But I admit I didn't try taking a prebuilt jar and trying to run it against 1.16.4 outside a dev environment, so it might not be impossible that this was intended.  (Although it looks like the CF upload stopped claiming 1.16.4 compatibility at 0.14.162.)

### DIFF
--- a/src/main/resources/META-INF/mods.toml
+++ b/src/main/resources/META-INF/mods.toml
@@ -60,4 +60,22 @@ MineColonies is a colony simulator within Minecraft! There are numerous types of
     versionRange="[${config.structurize_version}]"
     ordering="AFTER"
     side="BOTH"
-
+# The rest of these are not required, but if you have them then they must be at least this version
+[[dependencies.minecolonies]]
+    modId="tconstruct"
+    mandatory=false
+    versionRange="[${config.tc_version},)"
+    ordering="NONE"
+    side="BOTH"
+[[dependencies.minecolonies]]
+    modId="resourcefulbees"
+    mandatory=false
+    versionRange="[${config.rb_version},)"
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/mods.toml
```diff
@@ -41,7 +41,7 @@ MineColonies is a colony simulator within Minecraft! There are numerous types of
     # Does this dependency have to exist - if not, ordering below must be specified
     mandatory=true #mandatory
     # The version range of the dependency
-    versionRange="[34.1.0,)" #mandatory
+    versionRange="[36.1.0,)" #mandatory
```
uecasm: The build.properties refers to 36.1.24; wasn't sure if I should use that exact version as a minimum or if it should be more lenient.  This seemed like an appropriate compromise.

ravenbuilder934 (dismissed):

uecasm:
While I'm here, should I add the optional minimum for Tinkers as well?  Or should that be a separate PR?

Raycoms:
> While I'm here, should I add the optional minimum for Tinkers as well? Or should that be a separate PR?

You can add this in here

Raycoms (approved):

marchermans (approved):

someaddons (approved):

PR #7352, opened 2021-07-04
Bring back the Build Building button

uecasm:
- Restores the "Build Building" button appearing for level 0 buildings instead of the "Upgrade" button
- When moving a deconstructed building, shows "Build Building" instead of "Repair", to reduce confusion

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/WindowBuildBuilding.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/WindowBuildBuilding.java
@@ -134,19 +134,18 @@ public class WindowBuildBuilding extends AbstractWindowSkeleton
         final Button buttonBuild = findPaneOfTypeByID(BUTTON_BUILD, Button.class);
         final IBuildingView parentBuilding = c.getBuilding(building.getParent());
 
-        if (building.getBuildingLevel() == 0 || parentBuilding != null)
+        if (building.getBuildingLevel() == 0)
         {
-            buttonBuild.setText(LanguageHandler.format("com.minecolonies.coremod.gui.workerhuts.build"));
+            buttonBuild.setText(new TranslationTextComponent("com.minecolonies.coremod.gui.workerhuts.build"));
             findPaneOfTypeByID(BUTTON_DECONSTRUCT_BUILDING, Button.class).hide();
         }
-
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/client/gui/WindowBuildBuilding.java
```diff
@@ -134,19 +134,18 @@ public WindowBuildBuilding(final IColonyView c, final IBuildingView building)
         final Button buttonBuild = findPaneOfTypeByID(BUTTON_BUILD, Button.class);
         final IBuildingView parentBuilding = c.getBuilding(building.getParent());
 
-        if (building.getBuildingLevel() == 0 || parentBuilding != null)
+        if (building.getBuildingLevel() == 0)
```
Raycoms, line 138: Is this remoed on purpose?

uecasm: Yes.  That would make it show "build" and not "upgrade" for any building with a parent, which also seems wrong.

Raycoms: Ye, @someaddons why do we have this here?

uecasm: I suspect the _intent_ of the original change was to make it so that when a parent included a higher-than-0 building in its schematic then the first-build of that building would show a Build button.  But due to the conditional structure it actually didn't do that.  And I don't think it's possible to do that since we can't easily distinguish between the first-build of an embedded building and a subsequent upgrade (the latter of which should show an upgrade button).

uecasm: Unless maybe those first-build buildings are still in a deconstructed state?  That could possibly be checked for... I'm not sure if I have any embedded schematics to test that with.

uecasm: Ok, this later change is probably a simpler fix for the original intent.  It turns the "repair" button into a "build building" button when the building is in a deconstructed state (which makes more sense when moving buildings too, while retaining the ability to move-and-upgrade if desired).  This will also affect the embedded child buildings on first build, if I'm understanding how they work correctly, so they'll get a "build" button too (and they won't get an upgrade button if they're already at their local max level).

someaddons (dismissed):

PR #7362, opened 2021-07-06
Mojmaps

Raycoms:
-
- Changes mappings to mojmaps.
-

Some small internal changes:

1) I renamed getName() in our hutBlocks to getHutName()
2) I changed level in decoration controller to "tiers" as getLevel() is the world. We might want to change more of these things in the future to not create confusion with level.

Review please


marchermans (dismissed):
I only cursorly scanned this PR.

It looks good, and yeah I agree there are probably other places that need renaming but those are a good start.

src/api/java/com/minecolonies/api/blocks/AbstractBlockMinecoloniesNamedGrave.java
```diff
@@ -16,11 +16,11 @@
     /**
      * The direction the block is facing.
      */
-    public static final DirectionProperty FACING = HorizontalBlock.HORIZONTAL_FACING;
+    public static final DirectionProperty FACING = HorizontalBlock.FACING;
 
     public AbstractBlockMinecoloniesNamedGrave(final Properties properties)
     {
-        super(properties.notSolid());
+        super(properties.noOcclusion());
```
uecasm, line 23: I wonder if `AbstractBlockHut` needs this too.  Most of the new hut block models are not solid.  Possibly explains why the previews show x-rays around them.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBarracks.java
```diff
@@ -124,14 +124,14 @@ public void registerBlockPosition(@NotNull final BlockState block, @NotNull fina
         {
             if (world.getBlockState(pos).getBlock() != ModBlocks.blockHutBarracksTower)
             {
-                world.setBlockState(pos, ModBlocks.blockHutBarracksTower.getDefaultState().with(BlockBarracksTowerSubstitution.FACING, block.get(AbstractBlockHut.FACING)));
-                final TileEntity tile = world.getTileEntity(pos);
+                world.setBlockAndUpdate(pos, ModBlocks.blockHutBarracksTower.defaultBlockState().setValue(BlockBarracksTowerSubstitution.FACING, block.getValue(AbstractBlockHut.FACING)));
```
uecasm: While it doesn't actually _matter_, since `BlockBarracksTowerSubstitution.FACING` and `AbstractBlockHut.FACING` are equal references, technically this logic is backwards and these two symbols should be reversed.

src/api/java/com/minecolonies/api/blocks/AbstractBlockMinecoloniesNamedGrave.java
Raycoms: AbstractBlockHut has this

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingBarracks.java
Raycoms: agreed, will adjust this in this PR anyway, while it doesn't really matter, it's small enough to adjust in this PR

PR #7365, opened 2021-07-06
Changes Related to https://github.com/ldtteam/Structurize/pull/406

Justen-N:
DO NOT MERGE UNLESS ldtteam/Structurize#406 is accepted. 
- Abstract BlockHut Now implements  IBlockHut, a new marker interface. 
-
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
+++ b/src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
@@ -77,7 +77,7 @@ public abstract class AbstractBlockHut<B extends AbstractBlockHut<B>> extends Ab
     private static final VoxelShape SHAPE = VoxelShapes.create(0.1, 0.1, 0.1, 0.9, 0.9, 0.9);
 
     /**
-     * The hut's lower-case building-registry-compatible name.s
+     * The hut's lower-case building-registry-compatible name.
      */
     private final String name;
 
--- a/src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
+++ b/src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
@@ -49,7 +49,7 @@ import org.jetbrains.annotations.Nullable;
  * All AbstractBlockHut[something] should extend this class.
  */
 @SuppressWarnings("PMD.ExcessiveImports")
-public abstract class AbstractBlockHut<B extends AbstractBlockHut<B>> extends AbstractBlockMinecolonies<B> implements IBuilderUndestroyable, IAnchorBlock
+public abstract class AbstractBlockHut<B extends AbstractBlockHut<B>> extends AbstractBlockMinecolonies<B> implements IBuilderUndestroyable, IAnchorBlock, IBlockHut
 {
     /**
      * Hardness factor of the pvp mode.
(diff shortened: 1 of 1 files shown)


src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
```diff
@@ -77,7 +77,7 @@
     private static final VoxelShape SHAPE = VoxelShapes.create(0.1, 0.1, 0.1, 0.9, 0.9, 0.9);
 
     /**
-     * The hut's lower-case building-registry-compatible name.
+     * The hut's lower-case building-registry-compatible name.s
```
ravenbuilder934: is this change intentional?

Justen-N: No. Also, damn you're fast.

ravenbuilder934: heh. fastest triage in the west, and all that

Justen-N: In fact, I didn't even touch that line, my fork just has weird crap happening. I no longer have direct access to the repo, so that's a thing.

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
```diff
@@ -49,7 +49,7 @@
  * All AbstractBlockHut[something] should extend this class.
  */
 @SuppressWarnings("PMD.ExcessiveImports")
-public abstract class AbstractBlockHut<B extends AbstractBlockHut<B>> extends AbstractBlockMinecolonies<B> implements IBuilderUndestroyable, IAnchorBlock
+public abstract class AbstractBlockHut<B extends AbstractBlockHut<B>> extends AbstractBlockMinecolonies<B> implements IBuilderUndestroyable, IAnchorBlock, IBlockHut
```
Raycoms, line 52: Can't we move IHutBlock to MineColonies, and in MineColonies extend structurename and use our own version which does the hut check or so?

Justen-N:
By the way, when this is reviewed, I would like to ask: Do we want to do away with the blockhut naming convention for blockhuts? Currently, this would require an overhaul of multiple sections of code and translation, and I don't want to start in on it if we don't want to do it

ravenbuilder934:
my two cents: get rid of it, but in 1.17, so we don't need to worry about datafixeruppers/etc

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
Raycoms: If we go as far as Improving this, then we should remove structurize having to know about huts

Justen-N: If I were to write this on MC's side, I would just check for membership in the AbstractHutBlock class, and not worry about the interface at all.  However, this check only matters for schematics, which seems to be why we would want it in Structurize.

Raycoms: Well, it's only for schematics, but who defines special treatment of schematics should be the mInecolonies side, and not the structurize side

Justen-N: Alright. I'll see what I can on that front

Justen-N: Hold on. I'm gonna cancel this PR and we just need to convince Orion to wait on these changes. Section/style/level should all be in the Metadata changes. This is adding a 'fix' for something that really will be fixed at a later date in a better way.

uecasm: It would be nice if the levelled decorations could have a level selector similar to huts (would save a lot of "why is it building level 1 instead of level 3" questions).  Having said that, that's tricky with the current UI arrangement and you probably wouldn't want to treated the levelled decos as "huts" as far as appearing individually in the first selector, so it's not as easy as that sounds.

PR #7366, opened 2021-07-07
fix 7225

### LINKED ISSUE #7212: Mirroring of barracks tower placeholder doesn't work
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: 0.14.210-alpha
- Structurize Version: 0.13.177-alpha

## Expected behavior

When you place a structure with a barracks tower placeholder (eg. medieval birch alternative barracks) mirrored, the barracks tower hut blocks will be in the correct mirrored state, and the barracks tower will be built mirrored.

## Actual behavior

Rotation works, but preview shows the barracks tower placeholder in the wrong orientation, and actually building the barracks tower puts it in unmirrored.

1. Grab a barracks hut block and begin placing it in a style that has barracks towers inside
2. Find a barracks tower in the preview and observe its orientation.  Mirror the barracks.
3. Find the corresponding barracks tower hut block and observe that it is in a different orientation relative to the rest of the structure
4. Build the barracks mirrored
5. Attempt to build the barracks tower.  Notice the construction tape is in the wrong location.

### Logs

- latest.log: Unlike the other issues I'm uploading now, this does show up a bit in latest.log, but I think not usefully.  https://gist.github.com/dhouck/ae2c7f039644daad6452fb300da7b28d, relevant bit starting around 1:54:36

### Notes

I'm not fully sure what causes this.  The `mirror` function of BlockBarracksTowerSubstitution.java looks reasonable so I'd at least expect this to show up correctly in the previews.  I notice that `AbstractBlockHut.java` doesn't have a `mirror` function even though it does have `rotate`, and a bit more work might need to be done to pass the mirror bit down to the hut block's building, though.

---
## Viewers

### LINKED ISSUE #7225: Teleport Scrolls lost registeration information (UI only) after player going into Nether.
### Prerequisites 

- [ ] I am running the latest **alpha** version of MineColonies and Structurize for my Minecraft version.

### Context 

- Minecraft Version: 0.14.189-beta-universal
- MineColonies Version: 1.16.5
- Structurize Version: 0.13.174

## Expected behavior

After I registered the Teleport Scrolls, they should always have the registered information even I carry them and go into other dimensions and return.

## Actual behavior

Once I go into Nether, the scroll lost all registered information. the target becomes **none**, but another player can see the colony name I registered.

Though it shows None in Colony target, I can still use the scroll to teleport.

1.  Register an Ultrasafe Teleport Scroll (or several as a stack), 
2. go into Nether
3.  the registered information will be lost.
4. return to Overworld
5. It will show None in the Colony target, but it actually can be used.
6. if you logout and back, it may got back the information

### Logs
No logs are specified with this bug. it's not a crash.

- latest.log: 
- crashlog: 

### Notes

I don't know this is intended or not, If you guys don't want the player use the teleport scrolls in other dimensions, you can disable the usage in other dimension not let the scrolls lost its registered information.

---
## Viewers

Raycoms:
Closes #7225
Closes #

- Fix scrolls displaying none message even though registered
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/blocks/BlockBarracksTowerSubstitution.java
+++ b/src/main/java/com/minecolonies/coremod/blocks/BlockBarracksTowerSubstitution.java
@@ -64,18 +64,6 @@ public class BlockBarracksTowerSubstitution extends AbstractBlockMinecolonies<Bl
         return state.setValue(FACING, rot.rotate(state.getValue(FACING)));
     }
 
-    /**
-     * @deprecated (Remove this as soon as minecraft offers anything better).
-     */
-    @SuppressWarnings(DEPRECATION)
-    @NotNull
-    @Override
-    @Deprecated
-    public BlockState mirror(@NotNull final BlockState state, final Mirror mirrorIn)
-    {
-        return state.rotate(mirrorIn.getRotation(state.getValue(FACING)));
-    }
-
     @Nullable
     @Override
     public BlockState getStateForPlacement(final BlockItemUseContext context)
--- a/src/main/java/com/minecolonies/coremod/blocks/BlockBarracksTowerSubstitution.java
+++ b/src/main/java/com/minecolonies/coremod/blocks/BlockBarracksTowerSubstitution.java
@@ -64,6 +64,18 @@ public class BlockBarracksTowerSubstitution extends AbstractBlockMinecolonies<Bl
         return state.setValue(FACING, rot.rotate(state.getValue(FACING)));
     }
 
+    /**
+     * @deprecated (Remove this as soon as minecraft offers anything better).
+     */
+    @SuppressWarnings(DEPRECATION)
+    @NotNull
+    @Override
+    @Deprecated
+    public BlockState mirror(@NotNull final BlockState state, final Mirror mirrorIn)
+    {
+        return state.rotate(mirrorIn.getRotation(state.getValue(FACING)));
+    }
+
     @Nullable
     @Override
     public BlockState getStateForPlacement(final BlockItemUseContext context)
--- a/src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
+++ b/src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
@@ -26,10 +26,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.state.DirectionProperty;
 import net.minecraft.state.StateContainer;
 import net.minecraft.tileentity.TileEntity;
-import net.minecraft.util.ActionResultType;
-import net.minecraft.util.Direction;
-import net.minecraft.util.Hand;
-import net.minecraft.util.Rotation;
+import net.minecraft.util.*;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.BlockRayTraceResult;
 import net.minecraft.util.math.shapes.ISelectionContext;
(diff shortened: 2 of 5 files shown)


src/main/java/com/minecolonies/coremod/blocks/BlockBarracksTowerSubstitution.java
```diff
@@ -64,18 +64,6 @@ public BlockState rotate(@NotNull final BlockState state, final Rotation rot)
...
-     * @deprecated (Remove this as soon as minecraft offers anything better).
-     */
-    @SuppressWarnings(DEPRECATION)
-    @NotNull
-    @Override
-    @Deprecated
-    public BlockState mirror(@NotNull final BlockState state, final Mirror mirrorIn)
-    {
-        return state.rotate(mirrorIn.getRotation(state.getValue(FACING)));
-    }
```
uecasm: Possibly of note is that most other MineColonies blocks include similar code, either directly or via inheritance from `HorizontalBlock`...

Raycoms: public abstract class AbstractBlockHut<B extends AbstractBlockHut<B>> extends AbstractBlockMinecolonies<B> implements IBuilderUndestroyable, IAnchorBlock

uecasm: Yes, AbstractBlockHut doesn't.  I was referring to things like barrels, racks, graves, flags, timber frames, etc.  (And sort of deco controllers, but they override `mirror` differently anyway so don't really count.)

Shingles also seem to get messed up (as noted in discord), although those don't have a custom mirror, they get it from vanilla stairs.  So it seems like there's something more systemic going wrong.

Raycoms: This here is for the schematic misalignment, not of blocks weirdly running into issues when mirrored.

uecasm: Ok, but I just tested on this branch and the mirroring is still wrong.  Asian barracks, level 5, no rotation.  With this code change it fixes the incorrect mirroring of the front two towers (because these should not change facing) but not that of the back two towers (because these should change facing).  Again, it mostly just looks like the deleted code was correct but structurize is mirroring the schematic as a whole on the wrong axis, such that it doesn't agree with what Minecraft itself does to blocks.

Raycoms: Okay I checked the code and I know what's up. The barrackstowerplaceholders do not store the mirror anywhere. All other buildings store the mirror in the tileEntity and use the rotation of the block. So this code doesn't actually help because the barrackstowersubs don't have a tileEntity. So, the real fix is removing the barrackstowesubstitution and use the new integrated system for this properly.

uecasm: That may also be true, but it's still getting the block facing incorrect.  With this change reverted, it now changes the facing of the front two and still doesn't change the facing of the back two, which is worse.  (And again, suggests that the code is correct but it's getting the wrong mirror axis being passed in.)

Raycoms: Getting someone to rescan the barracks with barrackstower instead of placeholder to fix this for good.

someaddons (approved):

PR #7382, opened 2021-07-10
Bump JEI and cache recipes to improve startup perf

uecasm:
- Bump JEI to slightly newer build and use newer methods for better translation support
- Generate generic vanilla recipes once in advance instead of per module to improve startup speed

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/compatibility/jei/JEIPlugin.java
+++ b/src/main/java/com/minecolonies/coremod/compatibility/jei/JEIPlugin.java
@@ -109,7 +109,7 @@ public class JEIPlugin implements IModPlugin
             return;
         }
 
-        populateRecipes(cacheVanillaRecipes(), registration::addRecipes);
+        populateRecipes(buildVanillaRecipesMap(), registration::addRecipes);
         recipesLoaded = true;
     }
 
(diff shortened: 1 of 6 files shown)


Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/compatibility/jei/JEIPlugin.java
```diff
@@ -155,6 +145,66 @@ public void onRuntimeAvailable(@NotNull final IJeiRuntime jeiRuntime)
         this.weakRuntime = new WeakReference<>(jeiRuntime);
     }
 
+    private Map<IRecipeType<?>, List<IGenericRecipe>> cacheVanillaRecipes()
```
marchermans: Unless I am missing something, but I think you are missing the cache here.

You loop the recipe manager and build a map with the results.

That map is never stored anywhere for reuse and the map is rebuild every time...

Can you point out to me how this is a caching function?
If so then please document this since the functionality is not obvious on first sight.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/compatibility/jei/JEIPlugin.java
uecasm: It needs to be built once each time this method is called because this is called on reload when the input recipes have all changed.  As such there is no point in storing it anywhere.

As was already stated in the description, how this acts as a cache is that this list of generic recipes is then used repeatedly when scanning the individual crafting modules.  Prior to this change it would rebuild the list from scratch for each and every module.

Which means this code is now run ~22x less often.

marchermans: Ahhhh okey.
I would maybe rename this then to buildRecipes.
Since this does not actually do caching, but instead just builds them and then other code reuses the result.

marchermans (approved):

PR #7401, opened 2021-07-14
Add sleeping and mourning interactions

ToMe25:
Closes #
Closes #
Closes #

- Sleeping citizens will now have an interaction when you click on them asking you why you woke them up.
- The sleeping interaction text is `zZzz... I want to sleep... Why are you bothering me?`.
- This interaction does not cause the chat icon to appear above them.
- Mourning Citizens now have an interaction telling you why they aren't working.
- This interactions is currently `I'm still processing %ss death, I can't work today.` I am aware this isn't great, but wasn't able to come up with anything better.
- The rain interaction also no longer shows the icon.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenMournHandler.java
+++ b/src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenMournHandler.java
@@ -67,6 +67,11 @@ public class CitizenMournHandler implements ICitizenMournHandler
     }
 
     @Override
+    public Set<String> getDeceasedCitizens() {
+    	return deceasedCitizens;
+    }
+
+    @Override
     public void removeDeceasedCitizen(final String name)
     {
         deceasedCitizens.remove(name);
--- a/src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenMournHandler.java
+++ b/src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenMournHandler.java
@@ -67,7 +67,8 @@ public class CitizenMournHandler implements ICitizenMournHandler
     }
 
     @Override
-    public Set<String> getDeceasedCitizens() {
+    public Set<String> getDeceasedCitizens()
+    {
     	return deceasedCitizens;
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenMournHandler.java
```diff
@@ -66,6 +66,11 @@ public void addDeceasedCitizen(final String name)
         deceasedCitizens.add(name);
     }
 
+    @Override
+    public Set<String> getDeceasedCitizens() {
```
Raycoms: formatting is off here

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -1651,6 +1651,8 @@
   "com.minecolonies.coremod.entity.citizen.sleep": "All colonists are tucked into bed.",
   "com.minecolonies.coremod.entity.citizen.raining": "I don't like staying out in the rain, my clothes are gonna get wet.",
   "com.minecolonies.coremod.entity.citizen.raid": "Help! The colony is being raided. I am running for shelter.",
+  "com.minecolonies.coremod.entity.citizen.sleeping": "zZzz... I'm sleeping... Why did you wake me up?",
```
Raycoms: What is triggerd when they are on the way to bed?

src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenMournHandler.java
ToMe25: oops

src/main/resources/assets/minecolonies/lang/en_us.json
ToMe25: Its triggered at the same time the icon in the gui is, for consistency.

Raycoms: So maybe something more generic like "zZzz... I want to sleep... Why are you bothering me?"

ToMe25: Yeah makes sense, I'll change that.

ToMe25: Wait no I remember now, It gets triggered here because that is the best place I could find, but the validator only allows it while the citizen actually sleeps.
So it only shows up while the citizen actually sleeps, is that something I should change?

Raycoms: Yeah, probably better

ToMe25: which?
having it show when they sleep or when they want to sleep?

Raycoms: when they want to sleep

ToMe25: ok I'll change that then, I thought this was the intended way

ToMe25: done.

Raycoms (approved):

marchermans (approved):

someaddons (approved):

PR #7427, opened 2021-07-21
Crafting improvements

Raycoms:
Closes #
Closes #
Closes #

- Improve crafting speed
- Improve issue with wrong storage matching
-

Review please


Mekle001 (approved):

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -623,9 +623,13 @@ public IRecipeStorage getFirstFulfillableRecipe(final Predicate<ItemStack> stack
             final IRecipeStorage storage = IColonyManager.getInstance().getRecipeManager().getRecipes().get(token);
             if (storage != null && (stackPredicate.test(storage.getPrimaryOutput()) || storage.getAlternateOutputs().stream().anyMatch(i -> stackPredicate.test(i))))
             {
-                final List<IItemHandler> handlers = building.getHandlers();
+                final Set<IItemHandler> handlers = new HashSet<>();
+                for (final ICitizenData workerEntity : building.getAssignedCitizen())
+                {
+                    handlers.add(workerEntity.getInventory());
```
someaddons, line 629: not sure if this really works, the caller of this function might not have access to the resources of all worker inventories

uecasm: Yeah, should probably only be the specific worker that will be doing the crafting.  Looks like all call sites do know the specific worker, except for `AbstractCraftingProductionResolver`.  The module should also know which specific worker it relates to anyway, similar to how it picks the associated crafting job.

Having said that, assuming that the actual fulfilment does the same thing as this then it means workers at the same building would cooperate to fulfil recipes (without needing to explicitly dump items into the hut inventories), which might not be a bad thing.  Currently this only affects the restaurant.

Raycoms: Yeah, atm this part of the code is very meh (this is only a copy of the old code, so this is nothing new).
This will change properly when I add the worker employment modules which allows us to handle this propperly.

Raycoms: The old code path "building.getHandlers()" returns the citizen inventory handlers.

someaddons (approved):

PR #7435, opened 2021-07-22
Fix cobble/ladder deserialization

### LINKED ISSUE #7386: Level 1 Miner gets "Rock Bottom" advancement and won't new shafts
### Prerequisites 

- [X] I am running the latest **alpha** version of MineColonies and Structurize for my Minecraft version.
- [X] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure my issue is not covered there.
- [X] I made sure that **this issue is not a duplicate of any existing issue.**

### Context 

- Minecraft Version:
- MineColonies Version: Alpha 0.14.279
- Structurize Version: 0.13.198

## Expected behavior
A newly hired miner does not achieve the "Rock Bottom" advancement and begins digging the shaft lower.

## Actual behavior
After the completion of the miner's hut, I hired the miner and immediately received the rock bottom advancement.

The Unconstructed lower level/Cave Entrance

Relevant Advancement

Mine Levels

Miner's Stats

1. Create a colony
2. Find a location with a cave close to the surface.
3. Place the Miner's hut on top of the cave (Use the medieval oak style mine)
4. Construct the Mine at Y=64
5. Hire a Miner
6. If all goes wrong, you should get the bug

### Logs
Latest Log: https://github.com/Graepix/minecolonies-logs/blob/main/2021-07-10-1.log.tar
No crash occurred

### Notes
My miner's strength is 13 and his stamina is 15. Also I have not built any guard towers yet and my miner sleeps in the tavern which is 130 blocks away. I'm also playing on peaceful. 
My colony has a constructed Tavern lv 1, Woodcutter (MOalt) lv 1, Courier (MOalt) lv 1, Warehouse (MOalt)  lv 1, Builder (MOalt) lv 2, Farmer (MO) lv 1, Restaurant (MOalt) lv 1, and two houses (1 MO)(1MOalt) lv 1. It also has placed schematics for the townhall (MO), the library (MO), and the university MO). MO= Medieval Oak, MOalt = Medieval Oak Alternative

I relogged into the world to check the Y level the block was built at, all sound was missing and the miner briefly split into two entities the Miner and a regular citizen with a technical information name. Here's the log for that: https://github.com/Graepix/minecolonies-logs/blob/main/latest.log

I constructed my mine on top of a cave, and during construction, my builder kept teleporting into the cave and path finding her way out.

I'm playing with a custom modpack, the only mods I have that affect the world generation are Serene Seasons (version: 1.16.5-4.0.1.119) and Farmer's Delight version (1.16.5-0.4.5). Serene Seasons seems to have no affect on my other colonists, but it does change the temperature and grass color of a biome. My miner was constructed in late autumn. 
---
## Viewers

uecasm:
Closes #7386

- Teach miners hired to a previously unused miner building which direction is down, and to not claim early victory.

Review please

This fixes the miners refusing to dig the shaft and immediately earning the Rock Bottom advancement despite not digging the shaft; caused by deserializing a `BlockPos.ZERO` instead of a `null` for these positions.  (The positions aren't calculated until a worker starts working for the first time.)

Simplest way to reproduce this happening is to set hiring to manual (so no miner is assigned), build a mine, then save and reload the world, then finally hire someone (who will then refuse to work).

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingMiner.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingMiner.java
@@ -71,11 +71,6 @@ public class BuildingMiner extends AbstractBuildingStructureBuilder
     private static final String MINER = "miner";
 
     /**
-     * True if shaft is at bottom limit.
-     */
-    private boolean clearedShaft = false;
-
-    /**
      * The location of the topmost cobblestone the ladder starts at.
      */
     private BlockPos cobbleLocation;
@@ -180,7 +175,6 @@ public class BuildingMiner extends AbstractBuildingStructureBuilder
     {
         super.deserializeNBT(compound);
 
-        clearedShaft = compound.getBoolean(TAG_CLEARED);
         ladderLocation = BlockPosUtil.readOrNull(compound, TAG_LLOCATION);
         cobbleLocation = BlockPosUtil.readOrNull(compound, TAG_CLOCATION);
     }
@@ -189,7 +183,6 @@ public class BuildingMiner extends AbstractBuildingStructureBuilder
     public CompoundNBT serializeNBT()
     {
         final CompoundNBT compound = super.serializeNBT();
-        compound.putBoolean(TAG_CLEARED, clearedShaft);
 
         BlockPosUtil.writeOptional(compound, TAG_CLOCATION, cobbleLocation);
         BlockPosUtil.writeOptional(compound, TAG_LLOCATION, ladderLocation);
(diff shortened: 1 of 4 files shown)


uecasm:
Incidentally, it looks like nothing is actually calling `hasClearedShaft()`.  Not sure if that's a new thing due to modularization, but it looks like it might have been like that for a while.  Does this still need to exist or should it be removed?  (Or should the view pay attention to it, so it can indicate that the shaft is still being dug?)

Raycoms:
> Incidentally, it looks like nothing is actually calling `hasClearedShaft()`. Not sure if that's a new thing due to modularization, but it looks like it might have been like that for a while. Does this still need to exist or should it be removed? (Or should the view pay attention to it, so it can indicate that the shaft is still being dug?)

The view shouldn't need it.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingMiner.java
```diff
@@ -186,16 +187,28 @@ public CompoundNBT serializeNBT()
...
-        }
+        writeOptionalBlockPos(compound, TAG_CLOCATION, cobbleLocation);
+        writeOptionalBlockPos(compound, TAG_LLOCATION, ladderLocation);
+
+        return compound;
+    }
 
-        if (ladderLocation != null)
+    @Nullable
+    private static BlockPos readOptionalBlockPos(@NotNull final CompoundNBT compound,
```
Raycoms: Those should maybe go to BlockPosUtil.

uecasm: Maybe, although the implementation of `readOptionalBlockPos` is deliberately "incorrect" for general purposes (it treats ZERO == null, in order to fix already-broken miners).  Still, there's other places (parent-child buildings, for example) that treat ZERO == null, so I guess it'd be in good company.

Raycoms (approved):

PR #7460, opened 2021-08-02
Vanilla-Style Huts

WenXin20:
- New vanilla-style models and textures for all the huts, blocks, and items.
- I can also work on improving the hitboxes of huts if this is merged.
- Huts are also temporary given a bigger hitbox to fix cullfacing working improperly with the hitboxes they previously had.
  - For example, if there was a carpet next to a hut, the side of the hut would be invisible.
- Before: 
blockhutmysticalsite_old
- After: 
blockhutmysticalsite_new

Review please


Imrielle:
Fixing the hit boxes is great, but otherwise - theres a large problem with this PR.

The entire point of the previous model update a few months ago was to make Minecolonies resource pack compatible, without needing the person making the resource pack to have to go through and make 100+ textures for the huts. This undoes all of that, making the previous work pointless.

If this was just a fix for the cullfaces and the hitboxes, I'd be all for it. Instead, I'm going to leave it up to Ray and the rest to approve, because I very much do not approve of adding 314 textures back in.

Raycoms:
First of all, I do agree with @Imrielle comment, we don't want to have non vanilla textures where we can avoid them to guarantee resource pack compat. Especially considering vanilla has so many different textures available for us to use.
On a second note, I do like that our blocks don't look "so vanilla-like". It makes them seem more "purposeful" as they are. Vanilla villager work blocks have almost no options/abilities/etc.

WenXin20:
I liked how it was previous to the model update a few months ago, mainly because the current look doesn't feel very minecrafty to me and a lot of the current models have either squished pixels, stretched pixels, or are using a larger than 16x16 pixel grid, which don't look good to me and I don't think it would look very nice with other resource packs either.

But if this isn't preferred, I can just move it out to my own resource pack and just push in the larger hitboxes. Cullfaces probably won't work very well with the current models though.

Edit: There are also some item and tab icon retextures, if any of those are wanted.

carlansor:
I like the idea of a more "minecrafty" look for models and such, at least to keep proportions coherent. The problem for me is the textures, if they arent linked to vanilla ones. Might check a thing about this, but surely it would be a good option to have it as a resource pack.

WenXin20:
Have you tried it in game to see how it looks? Might look better than just the image above. 
I also have it set up already for a resource pack that I can post on curseforge if that ends up being preferred.

Raycoms:
Can we get some ingame screenshots of this?

WenXin20:
Something I've also noticed is that the scarecrow sometimes switches textures.
New (proposed):


Old:

ravenbuilder934:
> Something I've also noticed is that the scarecrow sometimes switches textures.

intentional iirc

WenXin20:
> > Something I've also noticed is that the scarecrow sometimes switches textures.
> 
> intentional iirc

I mean it switches between game loads and without breaking/moving it. In the image above, one shows both scarecrows, while the other only has one.

WenXin20:
Any update on the status of this pr?

Raycoms:
> Any update on the status of this pr?

We won't want all of our models adjusted like this. But we believe some of the models do indeed look more fitting than our current ones. But, for that we have to go with a comb through it.

WenXin20:
Alright, sounds good. If any changes are needed, I can do that as well. I'm in the discord chat if that's easier to communicate with.

marchermans:
We talked about this internally, we will publish this as a resourcepack for patreons first, and get their feedback, things to fix etc.
From there we will see what people think of it, and how to incorporate this properly into our existing models.

Thanks for your help improving the models, if you want to help us with the models in general contact us on discord to see how we can work together further on this. 

Greets,
Orion.

WenXin20:
Alright, sounds good! Glad to help :)

WenXin20:
Can you let me know which channel in the minecolonies discord?

marchermans:
You can just send a message in the general chat, and ping me. I will see it then, and see how to get you into a dedicated area for it

PR #7528, opened 2021-08-23
Fix colony permissions with enabled pvp

HubertBalcerzak:
- Fix colony permissions when pvp is enabled. It was possible to break blocks regardless of permission settings.

Review please


someaddons (approved):

Kuboczoch (approved):
Great! Can't wait!

PR #7534, opened 2021-08-24
Patch cobblestone mine missing cover issue 5006 903131781

leagris:
Fixes a schem issue (https://github.com/ldtteam/minecolonies/issues/5006#issuecomment-903131781)

- Add the missing solid block place-holder roof layer

Review please

edit raven: remove language making github think this closes the entire schematic issue


Talyda:
Can I comment here an opinion? I'm gonna anyway =P
I don't think this is a good idea (to copy paste my thoughts from discord =P)
because more and more packs and even the world itself is making bigger caves that the miner needs to work through, and they simply aren't mining enough cobble half the time to provide for themselves, let alone the rest of the colony =)
I've had to go off mining myself to supply cobble for the miner to finish making tunnels because of the cave mods that get added to Minecolonies packs
*edits* or get rid of cobble tunnels altogether and just make them all wood =P

Raycoms:
Can I get a "before after" screenshot of the schem?

leagris:
> Can I get a "before after" screenshot of the schem?

Maybe put this on hold because all other cobblestone nodes are missing the top layer of placeholder blocks, have insufficient lighting with dark spots letting hostile spawns and already consume excessive cobblestone.

So this patch is only providing the missing roof on the straight node. It is a band-aid partial fix.

I have been chatting in discord `#shem-discussion` with Shindria and Motionlesstrain. I am finally working on a totally different cobblestone node design with center pillar and torches on it. Will consume less cobblestone but more torches and some oak signs indicating `EXIT ==>` on 3-way and 4-way nodes, with no dark spot and a roof of placeholder blocks exactly like the default oak wood mines has already.

BlackWarrior2004:
> Can I comment here an opinion? I'm gonna anyway =P
> I don't think this is a good idea (to copy paste my thoughts from discord =P)
> because more and more packs and even the world itself is making bigger caves that the miner needs to work through, and they simply aren't mining enough cobble half the time to provide for themselves, let alone the rest of the colony =)
> I've had to go off mining myself to supply cobble for the miner to finish making tunnels because of the cave mods that get added to Minecolonies packs
> _edits_ or get rid of cobble tunnels altogether and just make them all wood =P

I guess you are Shindria.. Btw it will only consume cobblestone when theres a blank over the mineshaft, and it aint the case all time plus having to cover everything is way more beneficial than letting mobs come in mineshafts and kill the citizens, like for example when it leads to a ravine, or some huge caves for example from quark, not to mention there are many mobs(modded) that are way stronger for guards to handle
I dont know whether this issue is fixed or not, (I am a noob on Github), but if its not pls fix it

Talyda:
> Btw it will only consume cobblestone when theres a blank over the mineshaft

Yeah, I am, and yeah, I know that it will only do it then. That is my problem with it - there is too many big caves and the miner keeps needing more cobble than they mine up.  I don't think it's a good idea, and I don't the the benefits of it outweigh the need for me to go strip mining for cobblestone to fill in the mine that my miner is making =)

MotionlessTrain:
The problem of big caves is one with all the mine node styles. For instance, the default oak one (e.g. used with truedwarven) already has a roof above it, and will have those problems as well, that is the miner needing cobblestone instead of generating it while being in a large cave. (As I understand it,) this pull request would make it at least equal across the different mining node styles

Talyda:
Again, I think it is a problem with ANY of the cobblestone styles. Change all the mines to wooden, because wood is a renewable resource, and it wouldn't be a problem.   Or create a way for the miner to pull cobble out their arse. Let them make a cobblegen or something. As it is right now it isn't sustainable half the time

uecasm:
Thus speaks someone who has probably never had their mines burn down from lava...

Talyda:
Of course I have had them burn down - but even burned down, it is a renewable resource. I can fix that easily.  I can't be the only person that has had to go strip mining for cobblestone because the miner just doesn't produce enough.

MotionlessTrain:
As far as I'm aware, a miner fills the walls, floors (and ceilings in some mine node styles) with cobblestone, regardless of styles. I agree that it gives extra problems with stone and fortress (and cave?) specifically, because they use cobblestone as well for the structures inside the node, so the miners in those styles use cobblestone for two things.

For fortress, a solution could be to use dark oak mining nodes. But that wouldn't work for stone

uecasm:
> I can't be the only person that has had to go strip mining for cobblestone because the miner just doesn't produce enough.

I just feed a cobblegen into a barrel, and if the warehouse is getting low I dump some in.  This could even be automated, with the help of some additional tech.

Raycoms:
All schematics should have a roof. Some not having is a bug. Now there are two details:
a) We will add a way to select the solid block of the mine (can be any type of stone variant) which allows you to set dirt, andesite, diorite, other modded stones.
b) The miner provides ores to the colony. We need a quarry to provide cobble

BlackWarrior2004:
> All schematics should have a roof. Some not having is a bug. Now there are two details:
> a) We will add a way to select the solid block of the mine (can be any type of stone variant) which allows you to set dirt, andesite, diorite, other modded stones.
> b) The miner provides ores to the colony. We need a quarry to provide cobble

agree with you either pls fix that cobblestone mineshaft roof or atleast make it so that it uses wooden one, else its very difficult to protect miner from various mobs, and not to mention skeletons can snipe from the gap above and the guards cant do anything to them, plus guards dont really follow the miners they patrol on that level on their own
And I have no problem with wooden ones cuz I disable fire tick

PR #7458, opened 2021-08-01
Some fixes

someaddons:
Closes #7456
Closes #7442
Closes #7428

# Changes proposed in this pull request:
Fix guard building near not refreshing
Fix fisher having a hard time catching the fish items
Fix fisher spamming xp orb messages in the log, they now wait 5 ticks before those can be picked up
Fix nearby guard state never refreshing
Add fisher a chance to not make it in time for retrieving their rod 
Fix decocontroller loosing path

Review please

PR #7338, opened 2021-06-30
Improve nearby guard interactions

someaddons:
# Changes proposed in this pull request:
Improve nearby guard interactions

Review please

PR #7542, opened 2021-08-28
Fix weird behaviour when building a guard tower or barracks

uecasm:
- When building or upgrading a guard tower or barracks, now correctly calculates the "am I near a guard tower" state on other buildings.
- Reverts to a chunk-aligned cubic range around the guard tower, but does still limit Y distance.

Review please

Previously, this would force an entire quadrant of the world to think it was guarded, even when it wasn't.  It would be fixed automatically after reloading the world or upgrading/repairing the building in question, however, thanks to a recent change by someaddons (#7458).

~~This does a recalculation on all buildings, which is an O(n²) evaluation (though it's deferred until the next time the building wants to know, which should smear out the time a bit).  It might be possible to optimize that a bit more (which is what the previous code was attempting) but doing it properly requires the old level of the guard building, which is not available at this point.  (And this is the same thing that occurs when loading the world anyway.)~~

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
@@ -41,6 +41,7 @@ import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.nbt.ListNBT;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.MutableBoundingBox;
 import net.minecraft.world.World;
 import net.minecraft.world.chunk.Chunk;
 import net.minecraftforge.common.util.Constants;
@@ -615,10 +616,13 @@ public class BuildingManager implements IBuildingManager
 
         for (final IBuilding colonyBuilding : getBuildings().values())
         {
-            if (colonyBuilding instanceof IGuardBuilding
-                  && (colonyBuilding.getClaimRadius(colonyBuilding.getBuildingLevel()) * 16) > building.getID().distManhattan(colonyBuilding.getID()))
+            if (colonyBuilding instanceof IGuardBuilding)
             {
-                return true;
+                final MutableBoundingBox guardedRegion = BlockPosUtil.getChunkAlignedBB(colonyBuilding.getPosition(), colonyBuilding.getClaimRadius(colonyBuilding.getBuildingLevel()));
+                if (guardedRegion.isInside(colonyBuilding.getPosition()))
+                {
+                    return true;
+                }
             }
         }
 
@@ -628,10 +632,11 @@ public class BuildingManager implements IBuildingManager
     @Override
     public void guardBuildingChangedAt(final IBuilding guardBuilding, final int newLevel)
     {
-        final int claimRadius = guardBuilding.getClaimRadius(Math.max(guardBuilding.getBuildingLevel(), newLevel)) * 16;
+        final int claimRadius = guardBuilding.getClaimRadius(Math.max(guardBuilding.getBuildingLevel(), newLevel));
+        final MutableBoundingBox guardedRegion = BlockPosUtil.getChunkAlignedBB(guardBuilding.getPosition(), claimRadius);
         for (final IBuilding building : getBuildings().values())
         {
-            if (claimRadius > building.getID().distManhattan(guardBuilding.getID()))
+            if (guardedRegion.isInside(building.getPosition()))
             {
                 building.resetGuardBuildingNear();
             }
--- a/src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
@@ -628,20 +628,9 @@ public class BuildingManager implements IBuildingManager
     @Override
     public void guardBuildingChangedAt(final IBuilding guardBuilding, final int newLevel)
     {
-        for (final IBuilding building : colony.getBuildingManager().getBuildings().values())
+        for (final IBuilding building : getBuildings().values())
         {
-            if (building.getPosition().getX() <= guardBuilding.getPosition().getX() + 16 * guardBuilding.getClaimRadius(newLevel)
-                  && building.getPosition().getZ() <= guardBuilding.getPosition().getZ() + 16 * guardBuilding.getClaimRadius(newLevel))
-            {
-                if (newLevel > 0)
-                {
-                    building.setGuardBuildingNear(true);
-                }
-                else
-                {
-                    building.setGuardBuildingNear(hasGuardBuildingNear(building));
-                }
-            }
+            building.resetGuardBuildingNear();
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
```diff
@@ -628,20 +628,9 @@ public boolean hasGuardBuildingNear(final IBuilding building)
     @Override
     public void guardBuildingChangedAt(final IBuilding guardBuilding, final int newLevel)
     {
-        for (final IBuilding building : colony.getBuildingManager().getBuildings().values())
+        for (final IBuilding building : getBuildings().values())
         {
-            if (building.getPosition().getX() <= guardBuilding.getPosition().getX() + 16 * guardBuilding.getClaimRadius(newLevel)
-                  && building.getPosition().getZ() <= guardBuilding.getPosition().getZ() + 16 * guardBuilding.getClaimRadius(newLevel))
```
someaddons, line 634: If we just fix the distance calculation here we dont need to mark all buildings, just those in the vincinity. It just needs to be a distance calc(2d) of current building compared to the guard building, as the current doesnt work with negative coords

uecasm: As I said, it's not possible to fix the distance calculation without knowing the old level of the building, which is no longer available.  (This is called after the upgrade, so it has access to two copies of the new level and zero copies of the old level.  And it can't assume `newLevel - 1` due to deconstruction and creative paste.)

This is because the range it needs to update is the maximum of the old range and the new range.

someaddons: the new range is never lower than the old range

uecasm: It is on deconstruct.

uecasm: For the deconstruct case, the implementation in the PR is more efficient than before, since it defers the calculations until something actually checks the building, instead of recalculating immediately.

someaddons: you can just change the level the guard towers(and barrakcs) pass on deconstruct to this, then we can just mark all in the vincinity for recalc easily

uecasm:
I've made the reset conditional based on distance.  (Aside: it's weird that the same interface provides both `getID()` and `getPosition()` and they're always equal.  This usage seems like it ought to be using `getPosition` but I made it use `getID` for consistency with `hasGuardBuildingNear`.)

I think this will work in most cases, but I'm less confident that it will work in some corner cases such as build tool pasting.  Still, I guess it would still only be temporarily wrong.

src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
```diff
@@ -628,19 +628,12 @@ public boolean hasGuardBuildingNear(final IBuilding building)
     @Override
     public void guardBuildingChangedAt(final IBuilding guardBuilding, final int newLevel)
     {
-        for (final IBuilding building : colony.getBuildingManager().getBuildings().values())
+        final int claimRadius = guardBuilding.getClaimRadius(Math.max(guardBuilding.getBuildingLevel(), newLevel)) * 16;
+        for (final IBuilding building : getBuildings().values())
         {
-            if (building.getPosition().getX() <= guardBuilding.getPosition().getX() + 16 * guardBuilding.getClaimRadius(newLevel)
-                  && building.getPosition().getZ() <= guardBuilding.getPosition().getZ() + 16 * guardBuilding.getClaimRadius(newLevel))
+            if (claimRadius > building.getID().distManhattan(guardBuilding.getID()))
```
someaddons: looks good now, just a small thing: you likely want to use a 2d manhattan distance so y levels dont affect it

uecasm: You might be right, but this is the same check being done in `hasGuardBuildingNear`.  So it's consistent as is.

And it would be a bit weird for a guard building at y = 100 to be successfully defending a building at y = 30...

MotionlessTrain:
Sorry to interfere, but isn't the manhattan distance a different metric than is used for claiming areas? That could mean that a building in the corner of the claimed area is in the 2 chunk radius of guard tower expansion, but the manhattan distance would be close to 32+32 = 64 blocks, twice the radius
The 5x5 chunks claimed by a level 1 guard tower (centered on the guard tower, so you get a "radius" of 2 chunks) would be a distance metric of max(|x1-x2|, |z1-z2|) < 2*16

uecasm:
Correct, the guard's "this building is guarded" radius is not aligned to chunk boundaries.  It *used* to be, but it was changed in #7338.  I assume that was intentional.

(The tower's actual colony chunk claiming is separate, and is chunk-aligned.  Though it does use the same chunk range as this.)

someaddons:
ye you're right its not really manhattan distance, more a box. Might be easiest to inflate an AABB box around the guard building pos and then check if it contains other buildings. lets us limit y too

MotionlessTrain:
Ah, you're right. That would at least explain why one of my builders is not happy with the guard towers being just a couple of chunks away on a diagonal. 😅

uecasm:
So you do want me to change the implementation to an AABB instead of Manhattan distance?  (This is different again from both before and after #7338.)

someaddons:
ye please make it an AABB, that way we can properly define the area too :D

on both checks^^

src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
```diff
@@ -615,10 +616,13 @@ public boolean hasGuardBuildingNear(final IBuilding building)
 
         for (final IBuilding colonyBuilding : getBuildings().values())
         {
-            if (colonyBuilding instanceof IGuardBuilding
-                  && (colonyBuilding.getClaimRadius(colonyBuilding.getBuildingLevel()) * 16) > building.getID().distManhattan(colonyBuilding.getID()))
+            if (colonyBuilding instanceof IGuardBuilding)
             {
-                return true;
+                final MutableBoundingBox guardedRegion = BlockPosUtil.getChunkAlignedBB(colonyBuilding.getPosition(), colonyBuilding.getClaimRadius(colonyBuilding.getBuildingLevel()));
+                if (guardedRegion.isInside(colonyBuilding.getPosition()))
```
MotionlessTrain: I think you meant to use `building` here. colonyBuilding is the guard tower and will always be in its own box

uecasm: No, it's the region of the guard building that matters.

MotionlessTrain: In line 621, you make a bounding box around colonyBuilding.getPosition(), and in line 622 you check if colonyBuilding.getPosition() is inside the bounding box. That will always be true. The `building` itself doesn't get checked here at all, where previously the distance between `colonyBuilding` and `building` was used

uecasm: Sorry, you're right; I misread and thought you were talking about the other line.

someaddons (approved):

PR #7558, opened 2021-09-03
Adds two general utility schematics

uecasm:
Closes ldtteam/Structurize#429

- Adds `fill`, an 8x8x8 of solid placeholders (to fill an area with solid blocks).
- Adds `dig`, an 8x8x8 of air (to dig out or clear an area).

Review please

Inspired by a conversation with Shindria on Discord.


Aduis1210:
Could we add additional sizes? 
2x2 3x3 4x4 ect?
Or even make a tweakable option?

Raycoms:
> Could we add additional sizes?
> 2x2 3x3 4x4 ect?
> Or even make a tweakable option?

That would be the shape-tool integration then.

Aduis1210:
> > Could we add additional sizes?
> > 2x2 3x3 4x4 ect?
> > Or even make a tweakable option?
> 
> That would be the shape-tool integration then.

If possible :) It would be an awesome addition to the builder
I'd imagine it would be very sought after in the community.

uecasm:
Yeah, I'm planning to have a look at shape tool integration today.  But I still think these should be merged, they're faster to use, and 8x8x8 is a reasonably useful size.

PR #7566, opened 2021-09-04
Integrate shape tool into survival/builder

uecasm:
Depends on ldtteam/Structurize#435

- Allows the Shape Tool to be used in MineColonies survival (it triggers a deco build)
    - If you're in creative, then you can use the new paste button to still do an instant free paste of the shape
    - Otherwise the tick will request that a builder do the build instead of doing a paste as in Structurize alone (regardless of your mode)
    - If you're not in creative, then the undo/paste buttons are hidden.
- Fixes the schematic name shown in builder chat and in the town hall (now uses a compact name instead of the full path)

Review please

- TODO: there is ~not yet~ a crafting recipe for the Shape Tool, and there probably should be one now.  Any suggestions?
    - I'm thinking that this should be a recipe added by MineColonies alone, not one in Structurize itself, since it remains a creative-only item over there.
    - The obvious recipe would probably be an emerald on a pair of sticks, similar to the build/scan tool recipes.
- Caveat: if you set the shape tool to RANDOM from a MineColonies dev environment then it crashes with the same error as ldtteam/Structurize#213.  This does not affect a Structurize dev environment nor normal in-game use.  Not sure what's behind that other than "weird gradle things", but I don't think it's a new issue.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -10,7 +10,8 @@ fml_range=[36,)
 mappings_channel=official
 mappings_version=1.16.5
 # Dependencies
-structurize_version=0.13.210-ALPHA
+#structurize_version=0.13.210-ALPHA
+structurize_version=unspecified
 datagen_version=0.1.48-ALPHA
 jei_mcversion=1.16.5
 jei_version=7.6.4.90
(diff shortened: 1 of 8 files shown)


Aduis1210:
Same as Build wand but change the tip to andesite?

someaddons (approved):

PR #7588, opened 2021-09-10
Add groundlevel support to supply camps

someaddons:
-Add groundlevel support to supply camps

Review please


Raycoms (dismissed):

uecasm:
Ideally it should be supported (as optional) for all buildings and decorations, as an initial placement offset.  This would help to reduce accidental basements.

That should probably happen on the structurize side instead though.

Raycoms:
> Ideally it should be supported (as optional) for all buildings and decorations, as an initial placement offset. This would help to reduce accidental basements.
> 
> That should probably happen on the structurize side instead though.

Yeah, I want to do some unpacking on MineColonies side that "moves the structure up" to groundlevel if exists.

uecasm:
Well, you _could_ do it that way, but I think it makes more sense just to do it in structurize when first loading the build gui, so that it can apply to everything.  Given a value saved in the blueprint, it'd be fairly straightforward.

I don't think this would work as-is unless you put a decoration controller in your camp, which seems a bit odd.

My proposal for doing this is ldtteam/Structurize#436 instead.

someaddons:
we'll just do both, this one is for basing the supply camp's placing restriction code on the groundlevel instead of beeing hardcoded, while your PR is for defaulting the preview to the right Y level + vanishing tag holders. The only actual overlap those two now have is the util class

uecasm:
I guess, although you'd be able to read the ground offset directly from the `Settings` class now, so you wouldn't need to use the tag util method for the camp at all.  (And it's more debatable if the "large flat ground area" requirement makes as much sense if the ground offset is large because you're doing a flying ship/skyblock style camp that probably doesn't go near the ground in most places... but that's another topic.)

Something would still be needed for the raid ship placement, but that could use the method from Structurize without repeating it here.  (Though note that my implementation is slightly different from yours -- as noted there, it looks like you refactored the code correctly but the original code wasn't actually reading the correct data.  Something that probably needs to be double-checked on the mcol side after that PR is merged.)

src/main/java/com/minecolonies/coremod/util/BlueprintTagUtils.java
```diff
@@ -0,0 +1,44 @@
...
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.util.math.BlockPos;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Utilities for locating tags in a blueprint
+ */
+public class BlueprintTagUtils
```
Raycoms: You wanted to replace that with a call yet.

src/main/java/com/minecolonies/coremod/items/ItemSupplyCampDeployer.java
```diff
@@ -155,13 +156,20 @@ public static boolean canCampBePlaced(
         final int sizeX = Settings.instance.getActiveStructure().getSizeX();
         final int sizeZ = Settings.instance.getActiveStructure().getSizeZ();
 
+        int groundLevel = zeroPos.getY();
+        final BlockPos groundLevelPos = BlueprintTagUtils.getFirstPosForTag(Settings.instance.getActiveStructure(), "groundlevel");
+        if (groundLevelPos != null)
+        {
+            groundLevel = groundLevelPos.getY();
+        }
+
```
uecasm, line 165: As previously noted, you could just call `Settings.instance.getGroundOffset()` instead of repeating this search.

PR #7591, opened 2021-09-12
Productive bees compatibility

LetterShapedGlyphs:
- Adds compatibility for Beekeepers to harvest unique combs from Productive Bees hives
- Allows Hive Tool and Apiary Hut to count Advanced Hives from Productive Bees as valid hive blocks.

Review please

Note - Currently using the curse.maven for testing purposes while I try to get an alternative maven repository that works.


Raycoms:
Looks good, just awaiting the other maven link then =)

JaisDK:
There's no maven link other than curse maven.

I see a problem with the logic here though. The worker will only go to hives with a HONEY_LEVEL of 5, but an advanced hive will produce combs at any honey level. On top of that, this harvest code only pulls items from inventory, it does not harvest the honey level once it does hit 5.
As it is now it will work just fine, but the beekeeper will only start getting combs once honey level is 5 and then it will keep going (even if there are no combs) since the honey level wont drop.

I don't know if it's a viable solution, but to do a full harvest of the hive you need to:
- In `EntityAIWorkBeekeeper::harvestHoney()` check the inventory and the honey level and decide based on that
- In `ProductiveBeesCompat` also check the honey level and do a shears harvest OR
- since Advanced Hives can harvest the honey level automatically, insert empty glass bottles into the inventory

LetterShapedGlyphs:
> I don't know if it's a viable solution, but to do a full harvest of the hive you need to:
> 
> * In `EntityAIWorkBeekeeper::harvestHoney()` check the inventory and the honey level and decide based on that
> * In `ProductiveBeesCompat` also check the honey level and do a shears harvest OR
> * since Advanced Hives can harvest the honey level automatically, insert empty glass bottles into the inventory

The beekeeper AI already sets the honey level to 0 when harvesting comb, that just isn't in any of the files I modified.
`world.setBlockAndUpdate(hive, world.getBlockState(hive).setValue(BlockStateProperties.LEVEL_HONEY, 0));`

I'm still trying to think of the cleanest way to tie in the auto-bottling of honey (using the input slot) when that setting is enabled in the Apiary hut but this works.

JaisDK:
:+1:

marchermans:
No curse maven for production dependencies.
If the mod does not provide a maven of their own then we can not provide it.

marchermans (changes requested):
No cursemaven for production dependencies.

LetterShapedGlyphs:
> No curse maven for production dependencies.
> If the mod does not provide a maven of their own then we can not provide it.

You might have noticed I made mention of this already in my initial PR description. I'm currently seeing if I can work with the mod dev to get an alternative repo up, but I was told explicitly that this would be fine for testing purposes in the mean-time.

marchermans:
Since this PR is only made for testing, I converted it into a draft untill the Dependency issue is resolved.

Raycoms:
Any update on a maven here?

JaisDK:
I won't be setting up a maven anytime soon

Raycoms:
Optimally that would become an addon.

PR #7592, opened 2021-09-12
DRAFT: Initial refactor of ItemStorage into two objects

Mekle001:
- Create ItemStackHandling for comparison purposes
- Create ItemStackStorage for storage purposes
- Both extend the now-abstract ItemStorage
- Much of the functionality is still in the base, we can work to simplify and move unique functionality to the specific classes in the future. 

This is not yet ready to merge, much testing is still required. 

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Food.java
+++ b/src/api/java/com/minecolonies/api/colony/requestsystem/requestable/Food.java
@@ -3,6 +3,8 @@ package com.minecolonies.api.colony.requestsystem.requestable;
 import com.google.common.reflect.TypeToken;
 import com.minecolonies.api.MinecoloniesAPIProxy;
 import com.minecolonies.api.colony.requestsystem.factory.IFactoryController;
+import com.minecolonies.api.crafting.ItemStackHandling;
+import com.minecolonies.api.crafting.ItemStackStorage;
 import com.minecolonies.api.crafting.ItemStorage;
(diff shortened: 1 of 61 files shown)


marchermans:
I am going to be honest, I don't see a reason todo this.

From your code, and from the PR description this is completely unclear, increases the clutter and maintenance burden, and has no advantages, over just once class that can do both.

Could you make it more clear why we should take the hits in both complexity as well as burden of maintenance vs the older solution?

Raycoms:
> I am going to be honest, I don't see a reason todo this.
> 
> From your code, and from the PR description this is completely unclear, increases the clutter and maintenance burden, and has no advantages, over just once class that can do both.
> 
> Could you make it more clear why we should take the hits in both complexity as well as burden of maintenance vs the older solution?

I had been talking with him about this. This has one big reason. We want to be able to do partial matching.
I.e: We look for an item with nbt x, we could allow fulfillment if the item in storage has nbt x+y. Now, if we do this, we need to know which one is the stored the stack (shouldn't ever have a ignorenbt etc) and the queried stack.

marchermans:
I see no reason Todo this.
Your search will still be in a linear factor. So no performance improvement is to be expected.

Raycoms:
> I see no reason Todo this.
> Your search will still be in a linear factor. So no performance improvement is to be expected.

It's not for performance. It's so that we can do one sided matches.

Mekle001:
this is old enough, it's probably faster to recreate it post modularization.

PR #7595, opened 2021-09-14
Bump structurize to get the new build tool ground level behaviour

uecasm:
- ldtteam/structurize#424
- ldtteam/structurize#436

Review please

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -10,7 +10,7 @@ fml_range=[36,)
 mappings_channel=official
 mappings_version=1.16.5
 # Dependencies
-structurize_version=0.13.219-ALPHA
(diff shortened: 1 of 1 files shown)


Raycoms (approved):

marchermans (approved):

uecasm:
FYI the list of buildings that currently provides a groundlevel tag is a little disjointed -- most styles, only the miner (but medievaloak, only citizen and not miner), then wooden has it in most buildings, except sometimes not the L1 or L5 of the building. I haven't checked which ones actually needed it (it's redundant if it's on the y-level immediately below the hut).

PR #7582, opened 2021-09-07
optimize images

ravenbuilder934:
Closes #
Closes #
Closes #

# Changes proposed in this pull request:
- optimizes images, reduces file size by about 0.2MB total
-
-

Review please

PR #7596, opened 2021-09-15
optimize images

ravenbuilder934:
#7582


Raycoms:
I'm gonna be honest. I don't think this is worth it actually. For 0.2mb risking that textures are harder to edit afterwards or risking weird texture bugs.

ravenbuilder934:
I mean..I've optimized them before, and haven't heard of anything like that. Did you have glitches?

PR #7611, opened 2021-09-23
Mob_Attack_Blacklist tag

Justen-N:
- Mob_attack_blacklist tag added
-EventHandler.onEntityAdded modified to check for mob_attack_blacklist membership before the attack citizens goal is injected. 

Three possible use cases: 
- Developer only: modify the mobs that will attack citizens in code. 
- Granular Config: Possible changes to the mobAttackCitizens config to allow for an admin to limit specific mobs from attacking citizens.
-In-Game menu: allow aggressive mod settings to be changed via a dropdown, in the same vein as the 'guards attack mobs' gaurd tower menu. 

Review please


src/main/resources/data/minecolonies/tags/entity_types/mob_attack_blacklist.json
```diff
@@ -0,0 +1,6 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:slime"
+  ]
```
Raycoms, line 6: Looks mostly fine, here I'd remove the slime, and add Llama and Enderman (so that you can remove them in the code above.

Justen-N: that actually makes way more sense than using slime as a placeholder... I don't know why that didn't translate in my head.

Raycoms (approved):

someaddons (approved):

PR #7616, opened 2021-09-28
Fix groundlevel again

uecasm:
Depends on ldtteam/Structurize#443

- Makes the supply camp/ship tools consistent with build/shape tool behaviour (selection of initial anchor and groundlevel)
  - ~~Caveat: they will now appear one block above ground by default unless someone rescans them with a tag anchor on top of the ground (and optionally a groundlevel tag, but just the anchor is sufficient if placed correctly --~~ best position is close to the centre and directly sitting on the "ground", but can be anywhere on the centre line if also using a groundlevel tag [and doesn't strictly need to be centred, but that helps rotation])

Review please

(Just waiting on a structurize version bump for the linked PR before mergeable)

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -10,7 +10,7 @@ fml_range=[36,)
 mappings_channel=official
 mappings_version=1.16.5
 # Dependencies
-structurize_version=0.13.230-ALPHA
+structurize_version=0.13.242-ALPHA
 datagen_version=0.1.48-ALPHA
 jei_mcversion=1.16.5
 jei_version=7.6.4.90
--- a/src/main/java/com/minecolonies/coremod/client/gui/WindowMinecoloniesBuildTool.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/WindowMinecoloniesBuildTool.java
@@ -41,20 +41,22 @@ public class WindowMinecoloniesBuildTool extends WindowBuildTool
      * @param pos           the position.
      * @param structureName the structure name.
      * @param rotation      the rotation.
+     * @param groundstyle   one of the GROUNDSTYLE_ values.
      */
-    public WindowMinecoloniesBuildTool(@Nullable final BlockPos pos, final String structureName, final int rotation)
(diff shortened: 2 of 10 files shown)


uecasm:
For maximum compatibility with the existing schematics, it now uses a different set of rules to decide the untagged ground level for each building "source":
* The build tool uses "one block below the anchor"
* The supply camp uses "one layer of ground at the bottom of the schematic"
* The supply ship uses "three layers of water at the bottom of the schematic"
* The raid ships use "four layers of water at the bottom of the schematic" (but they're all tagged anyway, so it doesn't matter)

This does mean that if you use the build tool to build a supply camp/ship then it will not be aligned correctly to the ground by default, although you can still adjust it to the correct position before building.  This can easily be fixed by rescanning with a tag anchor block or decoration controller placed anywhere and an explicit groundlevel tag (contrary to before, both are required now).

marchermans (dismissed):
So I don't like the special casing here.
Because I am not sure this covers all legacy cases.

Would it not be better to default back to the old placement behaviour if there is no groundtag?
Because this special casing here makes me feel iffy, since if you missed one old schematic which relied on that placing logic, you killed that schematic.

uecasm:
Unless there is a way to place schematics other than those four, then this covers all legacy cases.

Granted not all existing schematics will perfectly fit these rules, but they do fit the majority of untagged schematics, and for the ones that do not fit, the worst that happens is that the ground level of the initial preview will appear above ground instead of flush with the ground, and the player will have to click minus a few times.

Which is strictly an improvement on previous behaviour, since most schematics will line up correctly, and the few that don't will be very obviously lined up wrong -- vs. prior behaviour where the wrong ones tended to have "ground" underground and the player didn't notice and the poor builder would carve out huge areas.

I'm not really a fan of the special-casing either, but I did it at Ray's request to reduce the impact on untagged schems and make them act as much like previous behaviour as possible.  It can be removed again after the special schems are all tagged appropriately.

Raycoms (approved):

PR #7624, opened 2021-10-03
Rendering port

Nightenom:
**Needs ldtteam/structurize#439**

- Refactor world rendering (to match new structurize) into standalone classes connected with rendering context holder (thus removing any world rendering from ClientEventHandler)
- Fix MutableChunkPos to be up to date with vanilla version (comments should be used to find classes which depend on vanilla code so we can update them easily)
- Moved client method from unsided event handler to client one
- Made multipiston optional dependency in mods.toml (not sure if actually applicable, didn't check code dependencies)
- **Didn't test** everything but most works

Review please


src/main/resources/META-INF/mods.toml
```diff
@@ -73,7 +73,7 @@ MineColonies is a colony simulator within Minecraft! There are numerous types of
     side="BOTH"
 [[dependencies.minecolonies]]
     modId="multipiston"
-    mandatory=true
+    mandatory=false
```
Raycoms: Like to have this mandatory for people still

Raycoms (dismissed):

src/main/resources/META-INF/mods.toml
marchermans: Why is this even an issue.
We are publishing this properly...

Nightenom: Cuz it's not hard dep (no code ref), not sure if it's used in schematics tho

Raycoms: I had it hard dep because its used in schematics

src/main/java/com/minecolonies/coremod/entity/pathfinding/Pathfinding.java
```diff
@@ -92,191 +77,4 @@ public static void enqueue(@NotNull final AbstractPathJob job)
...
     }
-
-    /**
-     * Render debugging information for the pathfinding system.
-     *
-     * @param frame       entity movement weight.
-     * @param matrixStack the matrix stack to apply to.
-     */
-    @OnlyIn(Dist.CLIENT)
-    public static void debugDraw(final double frame, final PoseStack matrixStack)
```
someaddons, line 103: is this just removed? or moved elsewhere

src/main/resources/META-INF/mods.toml
uecasm: That seems like a good reason to keep it mandatory.  Unless we want to rescan them to populate the `required_mods` and have the build tool warn you if a schematic needs a mod you don't have, but then that'd be a different PR anyway...

src/main/java/com/minecolonies/coremod/entity/pathfinding/Pathfinding.java
Raycoms: He moved it into a separate class

someaddons: ah didnt see that one then, kk

PR #7634, opened 2021-10-07
Stuckhandler/Raider Block break fixes

someaddons:
Closes #
Closes #
Closes #

Fix threatlist not dropping invalid marked targets on death or timeout
Fix stuck handler global timeout happening before other options could be tried, global timeout is now the last fallback if unstuck could not be detected in normal ways.
This also fixes raiders not breaking blocks as intended.
Fixed facing issues where the util inverted one of the axis to determine facing, leading to issues on raider trying to break blocks in the wrong directions.
The facing change also affected racks, updated their models to now have proper facing and not require the one axis inverted behaviour we had before.
Fixed Targeting AI to only consider visible targets on area searches

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/combat/threat/ThreatTable.java
+++ b/src/api/java/com/minecolonies/api/entity/combat/threat/ThreatTable.java
@@ -150,8 +150,8 @@ public class ThreatTable<T extends LivingEntity & IThreatTableEntity>
             return null;
         }
 
-        /// Add ignore logic
-        final ThreatTableEntry current = threatList.get(currentTargetIndex);
+        // Threat value target change thresholds
+        ThreatTableEntry current = threatList.get(currentTargetIndex);
         final ThreatTableEntry top = threatList.get(0);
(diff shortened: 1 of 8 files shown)


src/api/java/com/minecolonies/api/entity/combat/threat/ThreatTable.java
```diff
@@ -170,27 +170,20 @@ public ThreatTableEntry getTarget()
...
                 {
                     currentTargetIndex = 0;
-                    return top;
+                    current = top;
                 }
             }
         }
 
-        if (current.getThreat() < 0)
+        if ((owner.level.getGameTime() - current.getLastSeen()) > MAX_TRACKING_TICKS || !current.getEntity().isAlive())
```
marchermans: Make the time check use a Math.abs call to guarantee that rejoins into the server don't cause things to mess up.

marchermans (changes requested):

Raycoms (dismissed):

Issue #541, opened 2017-02-02
[b2464] placing supply ship too close to claimed area ...


kreezxil:
In build 2464, when placing a supply ship, if you place it close to another players claimed area, instead of telling you that it is too close, it tries to build the ship and fails only rendering as much of it as it can and because it fails to build the ship, you don't get the starter chest with the town hall placer in it and further you can't place another supply ship obviously.

![2017-02-02_17 28 46](https://cloud.githubusercontent.com/assets/3880072/22573146/2bdc11d4-e96d-11e6-8f54-023571c0bb05.png)

PR #7635, opened 2021-10-07
Feature #541: Update smelting logic to use fortune chances for the multiplier

UserModel:
This will now use the same logic as fortune does to determine the smelting multiplier.

Building Level 1: Fortune 0
Building Level 2: Fortune 0, with a chance for Fortune 1
Building Level 3: Fortune 1, with a chance for Fortune 2
Building Level 4: Fortune 2, with a chance for Fortune 3
Building Level 5: Fortune 3, with a chance for Fortune 4

The chances for increasing to the next tier is still the same as it was previously.

Closes feature [#541](https://github.com/ldtteam/minecolonies-features/issues/541)

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSmeltery.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSmeltery.java
@@ -122,21 +122,22 @@ public class BuildingSmeltery extends AbstractBuildingFurnaceUser
     @SuppressWarnings(MAGIC_NUMBERS_SHOULD_NOT_BE_USED)
     public int ingotMultiplier(final int skillLevel, final Random random)
     {
+        int fortuneLevel;
         switch (getBuildingLevel())
         {
-            case 1:
-                return random.nextInt(ONE_HUNDRED_PERCENT - skillLevel / 2) == 0 ? DOUBLE : 1;
-            case 2:
-                return random.nextInt(ONE_HUNDRED_PERCENT - skillLevel) == 0 ? DOUBLE : 1;
-            case 3:
-                return 2;
-            case 4:
-                return random.nextInt(ONE_HUNDRED_PERCENT - skillLevel / 2) == 0 ? TRIPLE : DOUBLE;
-            case 5:
-                return random.nextInt(ONE_HUNDRED_PERCENT - skillLevel) == 0 ? TRIPLE : DOUBLE;
-            default:
-                return 1;
+            case 2: // Building Level 2 - Default: Fortune 0, Chance for Fortune 1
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSmeltery.java
```diff
@@ -122,21 +122,22 @@ public Skill getSecondarySkill()
...
+                fortuneLevel = random.nextInt(ONE_HUNDRED_PERCENT - skillLevel) == 0 ? 2 : 1; break;
+            case 4: // Building Level 4 - Default: Fortune 2, Chance for Fortune 3
+                fortuneLevel = random.nextInt(ONE_HUNDRED_PERCENT - skillLevel / 2) == 0 ? 3 : 2; break;
+            case 5: // Building Level 5 - Default: Fortune 3, Chance for Fortune 4
+                fortuneLevel = random.nextInt(ONE_HUNDRED_PERCENT - skillLevel) == 0 ? 4 : 3; break;
+            default: // Building Level 1 - Default: Fortune 0
+                fortuneLevel = 0; break;
         }
+        int multiplier = random.nextInt(fortuneLevel + 2) - 1;
+        return multiplier > 0 ? 1 + multiplier : 1;
```
Raycoms, line 140: But that is again not properly like vanilla fortune loottable, right?

UserModel: This is the same formula that the vanilla fortune loottable uses.

Raycoms: Yeah, but it's item agnostic, so it doesn't speedup lapis better than iron

UserModel: Yeah, this doesn't speed up or make multiples for things like redstone or lapis. This returns a multiplier, like before, that states whether the items need to be doubled, tripled, or quadrupled. Do I need to add extra logic for ores that can drop multiple? From my understanding, those special ores only smelt into one with the current smeltery.

Raycoms: Yeah, the idea is to use some logic that allows to replicate the way vanillla does ore doubling through fortune for each block (and inherently supporting modded too)

UserModel: Okay, I'll take a look, but that would be changes to the method that calls the ore doubling, not the ore doubling itself. The multipliers for iron and lapis are the same, as we are just doubling, tripling, or quadrupling.

Raycoms: Yes, will need both

PR #7659, opened 2021-10-12
Remove the restrictive BipedModelType enum. Use the Model Type registry instead.

TagnumElite:
This pull request allows for the use of model types outside of BipedModelType.

The enum BipedModelType ordinal is used to store the citizens model type. This restricts model types to just the enum.
My goal here was to make the IModelTypeRegistry more useful. A new class ModelTypeEntry along withs its Builder allows for
registering custom model types.

Citizens now store there model types instead as a resource location string. This allows the citizen to fetch its model from the model type registry instead of trying to find it in the BipedModelType enum.

Since I 'improved' the IModelTypeRegistry. There is now validation for model types creation.

- ~Uses my proposed changes in ldtteam/Structurize#452 to register huts using the HUTS parameter.~
- Change IModeTypeRegistry to support registering ResourceLocations and more to allow for mod model types.
- Renames getModel to getModelType in IJob.

Review please


marchermans:
I don't understand this pr

marchermans:
This change is completely not needed.
The api gives access to this registry.

Since this is client side only a forge registry is not needed, since their main advantage is the ability to sync between client and server using it's automatic syncing.

TagnumElite:
I am currently busy with the change.

Also, the problem with having custom models is them saving them to entity data, it expects the BipedModelType enum.
If I want to have my own models from my own mod, it doesn't save properly.

TagnumElite:
I have my course to go to and there is no power so I can only work on this later. I did have an idea for model textures. Instead of specify model texture count and name in code, instead there could be a data generator that generates a file containing a list of textures for both genders for each style. This would allow for validation that all textures exist and allow the modification of texture variations per model, so that it can be changed.

TagnumElite:
Sorry, I made the mistake of using the GitHub Client. I had to force overwrite this pull request. To see the old changes look at [this branch](https://github.com/Tagnum-Mods/minecolonies/tree/dedepend_backup).

Justen-N:
Tagnum, were you attempting to change the enum at runtime? i.e., to add another type or remove one? If that is the case, then Orion is absolutely right, and you should be able to use the API to add what you were using this for. IN other words, I don't see this as necessary either.

marchermans:
The change away from the enum is needed since those are not expandable at runtime. However it suffice to offer a concurrent map via the api and not a full-blown registry

TagnumElite:
I am not gonna touch that can of worms, extending enums at runtime. No, the problem I have was just [this](https://github.com/ldtteam/minecolonies/blob/version/main/src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenJobHandler.java#L88) or more specifically the fact that Citizen EntityData Model was set to enum name and fetched via enum name. All I really did was just change it to a 'ResouceLocation' and fetched everything from the model registry.

Raycoms:
> I am not gonna touch that can of worms, extending enums at runtime. No, the problem I have was just [this](https://github.com/ldtteam/minecolonies/blob/version/main/src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenJobHandler.java#L88) or more specifically the fact that Citizen EntityData Model was set to enum name and fetched via enum name. All I really did was just change it to a 'ResouceLocation' and fetched everything from the model registry.

Okay, I gave this another look. I understand the problem, but I think the solution is an overkill. Instead of changing everything around it, we could just make the BipedModelType a class instead of an enum and add a way to register additional ones.

TagnumElite:
I can simplify this much further, infact. [Here, I just did exactly that](https://github.com/Tagnum-Mods/minecolonies/commit/41bca7b79a5ac30a6f7fe2aa70315e18c33ab00d).
Is this more in line with what you want?

Raycoms:
> I can simplify this much further, infact. [Here, I just did exactly that](https://github.com/Tagnum-Mods/minecolonies/commit/41bca7b79a5ac30a6f7fe2aa70315e18c33ab00d). Is this more in line with what you want?

That looks much better already.

TagnumElite:
Well, I can push this branch to that commit or close and make a new pull request if you prefer that commit more.

Raycoms:
> Well, I can push this branch to that commit or close and make a new pull request if you prefer that commit more.

Yes please

TagnumElite:
Sorry about the above close and re-open. GitHub automatically closed the commit becuase I push before I pulled by mistake.

src/api/java/com/minecolonies/api/client/render/modeltype/ModModelTypes.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final ResourceLocation GLASSBLOWER_ID = new ResourceLocation(Constants.MOD_ID, "glassblower");
+    public static final ResourceLocation DYER_ID = new ResourceLocation(Constants.MOD_ID, "dyer");
+    public static final ResourceLocation MECHANIST_ID = new ResourceLocation(Constants.MOD_ID, "mechanist");
+    public static final ResourceLocation FLETCHER_ID = new ResourceLocation(Constants.MOD_ID, "fletcher");
+    public static final ResourceLocation CONCRETE_MIXER_ID = new ResourceLocation(Constants.MOD_ID, "concretemixer");
+    public static final ResourceLocation RABBIT_HERDER_ID = new ResourceLocation(Constants.MOD_ID, "rabbitherder");
+    public static final ResourceLocation PLANTER_ID = new ResourceLocation(Constants.MOD_ID, "planter");
+    public static final ResourceLocation BEEKEEPER_ID = new ResourceLocation(Constants.MOD_ID, "beekeeper");
+
+    public static BipedModelType settler;
```
Raycoms: all upper case, constants

src/main/java/com/minecolonies/apiimp/initializer/ModModelTypeInitializer.java
```diff
@@ -0,0 +1,124 @@
+package com.minecolonies.apiimp.initializer;
+
+import com.minecolonies.api.client.render.modeltype.BipedModelType;
+import com.minecolonies.api.client.render.modeltype.CitizenModel;
+import com.minecolonies.api.client.render.modeltype.ModModelTypes;
+import com.minecolonies.api.client.render.modeltype.registry.IModelTypeRegistry;
+import com.minecolonies.coremod.client.model.*;
+
+public class ModModelTypeInitializer
```
Raycoms, line 9: wrong method formatting in file

src/api/java/com/minecolonies/api/client/render/modeltype/ModModelTypes.java
TagnumElite: I copied the format from [ModBuildings.java](https://github.com/ldtteam/minecolonies/blob/version/main/src/api/java/com/minecolonies/api/colony/buildings/ModBuildings.java).

src/main/java/com/minecolonies/apiimp/initializer/ModModelTypeInitializer.java
TagnumElite: Do you mean you want everything instantiated first, then registered?

Raycoms: No
{

}

TagnumElite: Oh, completely flew over my head. Yeah I will fix that now

src/api/java/com/minecolonies/api/client/render/modeltype/IModelType.java
```diff
@@ -23,4 +23,18 @@
...
+     * @return The male model
+     */
+    CitizenModel<AbstractEntityCitizen> getMaleModel();
+
+    /**
+     * Get the female model for this model type
+     *
+     * @return The female model
+     */
+    CitizenModel<AbstractEntityCitizen> getFemaleMap();
```
marchermans: Wrong method name.

src/api/java/com/minecolonies/api/client/render/modeltype/BipedModelType.java
```diff
@@ -1,78 +1,72 @@
 package com.minecolonies.api.client.render.modeltype;
 
+import com.minecolonies.api.entity.citizen.AbstractEntityCitizen;
+import net.minecraft.util.ResourceLocation;
+
 /**
  * Enum with possible citizens.
  */
-public enum BipedModelType implements ISimpleModelType
+public class BipedModelType implements ISimpleModelType
```
marchermans: This should stay an enum.

You should never need this in the API anyway.... everwhere where it is in use in the API or internally it should use the ISimpleModelType interface.

src/api/java/com/minecolonies/api/client/render/modeltype/BipedModelType.java
```diff
@@ -1,78 +1,72 @@
 package com.minecolonies.api.client.render.modeltype;
```
marchermans: Who put this in the API? Did I? Because this is an internal implementation detail, and should never be here in the first palce.

src/api/java/com/minecolonies/api/client/render/modeltype/ModModelTypes.java
```diff
@@ -0,0 +1,84 @@
...
+    public static BipedModelType CONCRETE_MIXER;
+    public static BipedModelType RABBIT_HERDER;
+    public static BipedModelType PLANTER;
+    public static BipedModelType BEEKEEPER;
+
+    private ModModelTypes()
+    {
+        throw new IllegalStateException("Tried to initialize: ModModelTypes but this is a Utility class.");
+    }
+}
```
marchermans: Again not the way this should be implemented.
You are exposing that BipedModelType which is completely not needed at all.

If this needs to be extendable, make a registry (any will do, even a `ConcurrentHashMap` and have everybody use that) That way you only need to expose the IDs as constants and not the underlying objects, reducing the chance for binary compatibility breaking bugs later on.

src/api/java/com/minecolonies/api/client/render/modeltype/registry/IModelTypeRegistry.java
```diff
@@ -1,23 +1,32 @@
...
 
-    IModelTypeRegistry register(IModelType type, CitizenModel<AbstractEntityCitizen> maleModel, CitizenModel<AbstractEntityCitizen> femaleModel);
-
-    Map<IModelType, CitizenModel<AbstractEntityCitizen>> getMaleMap();
+    /**
+     * @param name
+     * @param modelType
+     * @return
+     */
+    void register(IModelType modelType);
```
marchermans, line 28: See this is way better.
If you have a registry, why are you exposing the implemented object then?

marchermans (changes requested):
This now both exposes the API and the underlying implementing class.
That last thing has to go.

Reduce the exposure of the BipedModelType and you are good.

src/api/java/com/minecolonies/api/client/render/modeltype/ModModelTypes.java
TagnumElite: Hmm, I never thought of that. Will make changes shortly.

TagnumElite:
So for `IJob#getModel()` do you want me to return a `ResourceLocation` instead of a `IModelType`.

Raycoms:
@TagnumElite is this ready or anything missing?

TagnumElite:
Mostly, going through the changes I made, I saw that `IModelTypeRegistry` had empty doc strings.

TagnumElite:
That should be me done unless there are more changes that need to be done. Also, `gradle\configuration.gradle` has incorrect variables `structurize_version` -> `structurizeVersion` and `jei_version` -> `jeiVersion`.
Which will probably be the reason for failed builds.

marchermans (approved):
Sorry missed that this one is complete 👍 
I had reviewed it already but somehow it got not recorded, blame my phone.

PR #7684, opened 2021-10-17
Guard hire bed message

someaddons:
Displays a chat message when a guard is assigned to a guard building about the citizen home/tavern now having a free bed

Review please

### DIFF
--- a/src/main/resources/assets/minecolonies/lang/en_us.json
+++ b/src/main/resources/assets/minecolonies/lang/en_us.json
@@ -659,6 +659,9 @@
 
   "com.minecolonies.coremod.gui.workerhuts.level0": "You must construct this hut before you can hire a worker!",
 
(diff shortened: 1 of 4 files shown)


src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -659,6 +659,9 @@
 
   "com.minecolonies.coremod.gui.workerhuts.level0": "You must construct this hut before you can hire a worker!",
 
+  "com.minecolonies.coremod.gui.workerhuts.guardassignbed": "%s became a guard. The %s at %s now has an open bed slot for a new citizen.",
```
Raycoms, line 662: Not better to have just one, and then %s became %. The %.....

someaddons: easier to adjust individual messages this way, so depends on what raven changes

ravenbuilder934: i'd say having one message for all three is fine, as I don't foresee a situation where I'd change only one of them

PR #7404, opened 2021-07-15
Feature/farm+unimodule

Raycoms:
Closes #
Closes #
Closes #

# Changes proposed in this pull request:
- Move farmer to modules
- Move University to modules
- More Graveyard to modules
- Move Enchanter to modules

Review please

PR #7695, opened 2021-10-20
Fixes crash if creative-fancy-pasting a tag-anchored schematic

uecasm:
- Fixes a server crash if someone fancy-pastes a schematic containing a tag anchor
    - Could also happen if pasting a schem containing a hut that no longer exists (e.g. missing addon mod)

Review please

This was also found in 1.17 with some versions of the Frontier supply camp (not that you should need to be fancy-pasting that, but still...).

### DIFF
--- a/src/api/java/com/minecolonies/api/util/CreativeBuildingStructureHandler.java
+++ b/src/api/java/com/minecolonies/api/util/CreativeBuildingStructureHandler.java
@@ -80,9 +80,12 @@ public final class CreativeBuildingStructureHandler extends CreativeStructureHan
         if (teData != null && teData.contains(TAG_BLUEPRINTDATA))
         {
             final TileEntity te = getWorld().getBlockEntity(worldPos);
-            ((IBlueprintDataProvider) te).readSchematicDataFromNBT(teData);
-            ((ServerWorld) getWorld()).getChunkSource().blockChanged(worldPos);
-            te.setChanged();
+            if (te != null)
+            {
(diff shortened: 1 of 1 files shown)


src/api/java/com/minecolonies/api/util/CreativeBuildingStructureHandler.java
```diff
@@ -80,9 +80,12 @@ public void triggerSuccess(final BlockPos pos, final List<ItemStack> list, final
         if (teData != null && teData.contains(TAG_BLUEPRINTDATA))
         {
             final TileEntity te = getWorld().getBlockEntity(worldPos);
-            ((IBlueprintDataProvider) te).readSchematicDataFromNBT(teData);
-            ((ServerWorld) getWorld()).getChunkSource().blockChanged(worldPos);
-            te.setChanged();
+            if (te != null)
```
uecasm, line 83: I was debating whether this should be an `instanceof IBlueprintDataProvider` check instead.  It might be a little safer, but there shouldn't really be any other cases where this happens unless the build itself failed.

uecasm:
FWIW, this code was introduced by #7235 and #7404... not entirely sure why though.  I haven't explicitly checked, but this doesn't look like it would be doing anything different from what structurize would have done when building it anyway.  So why is this needed just for mcol creative paste?

Raycoms:
I'm not 100% sure anymore, I think it was to update a building when pasting over it.

uecasm:
This is already merged to 1.17, so.....

PR #7718, opened 2021-10-26
Replace furnace building type with module

Raycoms:
- This is one of the last building modules. No more extends AbstractBuildingFurnace but instead a proper module
-
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/util/constant/BuildingConstants.java
+++ b/src/api/java/com/minecolonies/api/util/constant/BuildingConstants.java
@@ -93,6 +93,11 @@ public final class BuildingConstants
     public static final String BUILDING_FLOWER_LIST = "flowers";
 
     /**
+     * The list of fuel.
+     */
+    public static final String FUEL_LIST = "fuel";
+
+    /**
      * Private constructor to hide implicit public one.
      */
     private BuildingConstants()
--- a/src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
+++ b/src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
@@ -14,11 +14,7 @@ import com.minecolonies.api.items.ModTags;
 import com.minecolonies.api.util.InventoryUtils;
 import com.minecolonies.api.util.ItemStackUtils;
 import com.minecolonies.coremod.colony.buildings.AbstractBuilding;
-import com.minecolonies.coremod.colony.buildings.AbstractBuildingFurnaceUser;
-import com.minecolonies.coremod.colony.buildings.modules.EnchanterStationsModule;
-import com.minecolonies.coremod.colony.buildings.modules.FarmerFieldModule;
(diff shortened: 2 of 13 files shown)


someaddons (approved):

uecasm:
Shouldn't this be merged into the `AbstractCraftingBuildingModule.Smelting` module rather than being a separate one?

Is there ever any reason to have a building be able to use furnaces but not know furnace recipes, or vice versa?  (You can still disable teaching furnace recipes in the building's smelting module if you like.)

Of the buildings in this PR, it looks like the only one that lacks a smelting module is the Baker ... which might possibly be why the baker has been having trouble turning dough into bread?  (Although their crafting module does contain the recipe for it...)

Raycoms:
> Shouldn't this be merged into the `AbstractCraftingBuildingModule.Smelting` module rather than being a separate one?
> 
> Is there ever any reason to have a building be able to use furnaces but not know furnace recipes, or vice versa? (You can still disable teaching furnace recipes in the building's smelting module if you like.)
> 
> Of the buildings in this PR, it looks like the only one that lacks a smelting module is the Baker ... which might possibly be why the baker has been having trouble turning dough into bread? (Although their crafting module does contain the recipe for it...)

Not necessarily, technically, the cook AI uses it, without doing recipes.

But yes, that might be the problem of the Baker

uecasm:
The cook does have a smelting module.

Raycoms:
> The cook does have a smelting module.

But that one is only used for requests, not for regular cooking, right?

uecasm:
The crafting modules themselves don't really **do** anything, they're just storage for recipes and helpers for JEI integration, and related recipe queries.  It's always the AI that drives actually doing work one way or the other.

The point being that the cook still has one so it doesn't break the pattern where the two modules could be merged.  The only one that does break the pattern is the baker, which might itself be a bug.

Raycoms:
> The crafting modules themselves don't really **do** anything, they're just storage for recipes and helpers for JEI integration, and related recipe queries. It's always the AI that drives actually doing work one way or the other.
> 
> The point being that the cook still has one so it doesn't break the pattern where the two modules could be merged. The only one that does break the pattern is the baker, which might itself be a bug.

Yeah, the baker is probably the one where it's a bug because it's actually a request smelter too. 
But I don't see why we have to mix them, not necessarily they have to be together and having modules be pretty self contained is a big advantage in terms of maintenance.

uecasm:
They don't _have_ to be mixed, but it does seem kinda silly for them to be separate if, as we've determined, when you want one you always want the other too.  And it just makes sense to me for the furnace crafting module to also contain the list of furnaces that can be crafted at.

Imagine adding some third type of crafting that also needed work blocks (e.g. something like Domum but where the schem did have to contain an Architect's Cutter, or some other mod integration, or even something like the concrete mixer's water mixing) -- it makes sense to have one module that wraps all of that up into a neat package, including both the list of work blocks and the recipes that can be made using them, and helpers for the building and AI to actually use them.

PR #7739, opened 2021-11-03
Solve two small issues

Raycoms:
Closes #
Closes #
Closes #

- This should stop some of the select res crashes
- Also will solve some of the other crashes too
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/event/ClientEventHandler.java
+++ b/src/main/java/com/minecolonies/coremod/event/ClientEventHandler.java
@@ -424,12 +424,13 @@ public class ClientEventHandler
                       true);
                     if (!wrapper.hasBluePrint() || !wrapper.isCorrectMD5(md5))
                     {
-                        Log.getLogger().debug("Blueprint error, requesting" + structureName + " from server.");
                         if (ServerLifecycleHooks.getCurrentServer() == null)
                         {
+                            Log.getLogger().debug("Blueprint error, requesting" + structureName + " from server.");
                             Network.getNetwork().sendToServer(new SchematicRequestMessage(structureName));
                             continue;
                         }
+                        continue;
                     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/minecolonies/coremod/event/ClientEventHandler.java
```diff
@@ -424,12 +424,12 @@ private static void handleRenderBuildTool(@NotNull final RenderWorldLastEvent ev
...
                     if (!wrapper.hasBluePrint() || !wrapper.isCorrectMD5(md5))
                     {
-                        Log.getLogger().debug("Blueprint error, requesting" + structureName + " from server.");
                         if (ServerLifecycleHooks.getCurrentServer() == null)
                         {
+                            Log.getLogger().debug("Blueprint error, requesting" + structureName + " from server.");
                             Network.getNetwork().sendToServer(new SchematicRequestMessage(structureName));
-                            continue;
                         }
+                        continue;
```
uecasm, line 447: This will make it not log anything in the case where it's single-player and the blueprint can't be found, which seems unhelpful.  Although log spam on render is also unhelpful... but on a related note, wouldn't this still cause message spam on each render tick on multiplayer?  I don't see anything obvious that will limit it...

Raycoms: Not very helpful anyway. removed log

uecasm: Ok, but it will still spam packets, unless I'm missing something.

Raycoms: Good point, will have to think of sth there. Potentially make it a static map<string, boolean> for which ones were requested already during a session. then I can readd the log too

marchermans (approved):

PR #7744, opened 2021-11-05
Feature: fishing bonus loot tables

uecasm:
- Refactors the existing loot table datagen to use vanilla datagen (+ LootTableAnalyzer where needed) instead of LDT datagenerators library.
    - Now the only thing remaining which uses that library is the sounds datagen.  I have not investigated how easy it would be to replace that too and remove the dependency entirely.
    - A bunch of datagen files have "changed" as a result of this, but it's just reformatting; unless I've overlooked something there's no actual behavioural change.
- Moves the fisherman bonus loot generation from config file to datapack loot table.  This makes it easier to tweak for modpack authors, among other benefits.
    - Intended change: instead of spontaneously appearing in the fisherman's inventory the loot now spawns in the world at the fishing bobber and is pulled towards the worker, exactly as the fish does.
    - Not-completely-intended change: while the base chance (at zero worker skill) for the extra drops is the same as before, the overall chance of bonus loot is slightly lower (e.g. at skill = 100, chance of a prismarine shard is 9.6% instead of 12.5%), due to the way that loot table scaling works.  Not sure if this is still ok or if we want to tweak it further.
    - ~~Not changed at all, but noticed in passing and it seems weird: since the fisherman is using a custom fishing bobber (presumbly since the vanilla one is hard-coded for players), it automatically fails the loot check for FISHING_TREASURE and this table is never used.  (Consequently, enchanting a fishing rod with Luck of the Sea is a net **loss** (since it decreases the drop weights of non-treasure, but then doesn't add anything else to compensate; this also happens for player-used enchanted rods if not fishing "in open water", whatever that means).)~~
    - ~~Also not changed, and not sure if intended: the config comment specified that sponges would not drop before Fisherman L4, but in actual fact everything starts dropping at L3.  But there's separate loot tables now per hut level so it's easy to customise if we want to (or a modpack author wants to).~~  The sponge will no longer drop from L3.
    - Also not changed, but seems probably intended: the fish/junk loot table is affected solely by the Luck of the Sea enchantment, while the bonus loot chance is affected solely by the worker's primary skill level.
- Switches the standard fish loot to use custom tables (but these still delegate to the vanilla tables for the most part, so should get similar loot automatically).
    - This is primarily to allow the FISHING_TREASURE loot to drop, which was otherwise blocked by use of a custom bobber.
    - The condition for treasure loot to drop is different from vanilla -- here we simply check whether the bobber itself is in a biome classed as ocean, rather than the AABB block check performed for vanilla fishing.  Note that this means it is very unlikely that fishing in a hut-integrated fishing lake will ever produce treasure loot, but that's better than never.  (This also means that using Luck of the Sea is still counterproductive unless they are actually fishing in an ocean biome.)
- Added JEI integration for the fisherman:
image
(It flashes between items of equal probability, so there are some additional drops not pictured here, but you'll see when playing.)

Review please (I've split the commits so that some of the runData noise is easier to filter out)
(I'll make a 1.17 port of this later.)


src/main/java/com/minecolonies/coremod/compatibility/jei/FishermanRecipeCategory.java
```diff
@@ -0,0 +1,137 @@
...
+        super(ModJobs.fisherman.getHandlerProducer().apply(null), ModJobs.FISHERMAN_ID,
+                new ItemStack(ModBuildings.fisherman.getBuildingBlock()), guiHelper);
+    }
+
+    private static final int LOOT_SLOTS_X = CITIZEN_X + CITIZEN_W + 4;
+    private static final int LOOT_SLOTS_W = WIDTH - LOOT_SLOTS_X;
+
+    @NotNull
+    @Override
+    public Class<? extends FishingRecipe> getRecipeClass() {
```
Raycoms: minor formatting thing

Raycoms (changes requested):
I like the move to datapacks and the JEI compat, as well as the in world spawn.
I'm fine with the slight chance decrease.
Yes the custom fishing bobber is a thing because of the hard-coded vanilla one.

In this regard, since you're touching this, can you get us access to FISHING_TREASURE if it's in an ocean?

Was not intended for sponges to drop at level 3.
The last thing is indeed intended.

uecasm:
> In this regard, since you're touching this, can you get us access to FISHING_TREASURE if it's in an ocean?

I can have a look at it, although since most fisher huts include a fishing lake (though not all) it might not help.

Having said that, as best I can tell (but I haven't thoroughly investigated) the vanilla definition of "open water" in this case is actually just a 5x3x5 area containing nothing but water, air, and optionally lily pads (and nothing else).  So a fishing lake might meet the definition sometimes, albeit probably only when the line is cast to the exact centre of the lake.  Or we could use a different definition, such as checking the biome.

Raycoms:
> > In this regard, since you're touching this, can you get us access to FISHING_TREASURE if it's in an ocean?
> 
> I can have a look at it, although since most fisher huts include a fishing lake (though not all) it might not help.
> 
> Having said that, as best I can tell (but I haven't thoroughly investigated) the vanilla definition of "open water" in this case is actually just a 5x3x5 area containing nothing but water, air, and optionally lily pads (and nothing else). So a fishing lake might meet the definition sometimes, albeit probably only when the line is cast to the exact centre of the lake. Or we could use a different definition, such as checking the biome.

Yeah, I'd check the biome

uecasm:
Ugh.  Forge continues to be bad at loot tables.

So the fundamental issue disabling the treasure fishing loot is that the `FISHING` loot table conditions the lookup into the `FISHING_TREASURE` on the `FishingPredicate`, which in turn requires `instanceof FishingBobberEntity`.

Looking into this, there are a few options, none of which I'm entirely happy with.  I'm open to better ones:

1. We could change our `NewBobberEntity` to extend `FishingBobberEntity`.  This is probably the cleanest option (from a loot perspective), but I'm not sure if that will have any undesirable side effects.  The main hassle is that we'd need to pass it a non-null `PlayerEntity` at least for the constructor, even if we then override everything so it gets ignored after that.  This would have to be a fake player since we can't rely on any particular player existing otherwise.

2. We could use `LootTableLoadEvent` to modify the existing fishing loot table and add a new entry for the `FISHING_TREASURE` loot table that's conditioned on our own custom condition that recognises `NewBobberEntity` instead.  This *would* have been the cleanest option, except that Forge doesn't provide any API to modify an existing loot pool (any more).  (There is a method to add an extra pool, but that will not correctly affect probabilities.)  So at minimum this will need an AT or reflection.  This also risks being lost if a datapack replaces the fishing loot table entirely without taking this change into account.  And I think this doesn't get triggered if a datapack replaces the table (but I haven't confirmed that), and it's more invisible to modpack makers.

3. We could replace the fishing loot table ourselves in json, which is a bit like the second option except it's more visible to modpack/datapack authors.  But it seems less compatible with other mods.

4. We could modify `FishingPredicate` itself to also recognise `NewBobberEntity`.  This requires a mixin.  (Which is easy, but I know people are philosophically opposed to.)

5. We could use Forge's Global Loot Modifiers system, except afaict this behaves like an extra pool so it won't correctly alter the probabilities of the other fishing loot entries.

Raycoms:
> Ugh. Forge continues to be bad at loot tables.
> 
> So the fundamental issue disabling the treasure fishing loot is that the `FISHING` loot table conditions the lookup into the `FISHING_TREASURE` on the `FishingPredicate`, which in turn requires `instanceof FishingBobberEntity`.
> 
> Looking into this, there are a few options, none of which I'm entirely happy with. I'm open to better ones:
> 
> 1. We could change our `NewBobberEntity` to extend `FishingBobberEntity`.  This is probably the cleanest option (from a loot perspective), but I'm not sure if that will have any undesirable side effects.  The main hassle is that we'd need to pass it a non-null `PlayerEntity` at least for the constructor, even if we then override everything so it gets ignored after that.  This would have to be a fake player since we can't rely on any particular player existing otherwise.
> 2. We could use `LootTableLoadEvent` to modify the existing fishing loot table and add a new entry for the `FISHING_TREASURE` loot table that's conditioned on our own custom condition that recognises `NewBobberEntity` instead.  This _would_ have been the cleanest option, except that Forge doesn't provide any API to modify an existing loot pool (any more).  (There is a method to add an extra pool, but that will not correctly affect probabilities.)  So at minimum this will need an AT or reflection.  This also risks being lost if a datapack replaces the fishing loot table entirely without taking this change into account.  And I think this doesn't get triggered if a datapack replaces the table (but I haven't confirmed that), and it's more invisible to modpack makers.
> 3. We could replace the fishing loot table ourselves in json, which is a bit like the second option except it's more visible to modpack/datapack authors.  But it seems less compatible with other mods.
> 4. We could modify `FishingPredicate` itself to also recognise `NewBobberEntity`.  This requires a mixin.  (Which is easy, but I know people are philosophically opposed to.)
> 5. We could use Forge's Global Loot Modifiers system, except afaict this behaves like an extra pool so it won't correctly alter the probabilities of the other fishing loot entries.

I guess 1. with a fakeplayer is probably the cleanest way then.

uecasm:
> I guess 1. with a fakeplayer is probably the cleanest way then.

Oh, actually it's not that easy.  There's no way to pass through the right `EntityType` that way.  I _think_ to get that to work we'd have to get rid of `NewBobberEntity` entirely and make the fisherman themselves have a fake player they use to use the vanilla bobber.  Which might be a thing that could be done, but it's a much larger change than I wanted to include in this PR.

Raycoms:
> > I guess 1. with a fakeplayer is probably the cleanest way then.
> 
> Oh, actually it's not that easy. There's no way to pass through the right `EntityType` that way. I _think_ to get that to work we'd have to get rid of `NewBobberEntity` entirely and make the fisherman themselves have a fake player they use to use the vanilla bobber. Which might be a thing that could be done, but it's a much larger change than I wanted to include in this PR.

Ohh, it does an entitytype check. darn it.

uecasm:
> Ohh, it does an entitytype check. darn it.

Not as such, but I assume it's bad to create an entity that thinks it's the wrong type.  I suppose `getType()` could be overridden too, but that seems really fugly.

src/main/java/com/minecolonies/coremod/compatibility/jei/FishermanRecipeCategory.java
uecasm: Fixed.  Dangers of auto-collapsing single-line methods.

uecasm:
The cleanest and safest and smallest _code delta_ would be option 4.  But I suspect that isn't going to be the preferred option. :)

Using the vanilla bobber would be the _nicest_ code (and seems more future-proof).  Not sure if there's downsides to making a fake player for the fisherman for that integration though, or other complications that'd mean that might not work.  But if we want to go this way I think it should be a separate PR anyway.

Raycoms:
> The cleanest and safest and smallest _code delta_ would be option 4. But I suspect that isn't going to be the preferred option. :)
> 
> Using the vanilla bobber would be the _nicest_ code (and seems more future-proof). Not sure if there's downsides to making a fake player for the fisherman for that integration though, or other complications that'd mean that might not work. But if we want to go this way I think it should be a separate PR anyway.

Yeah, I'd like to avoid mixins if it's not 100% necessary. I mean, just using the vanilla bobber with a fakeplayer would be pretty easy to test. Just needs to create a vanilla bobber instead, should be a 3 line change.

Raycoms:
> > The cleanest and safest and smallest _code delta_ would be option 4. But I suspect that isn't going to be the preferred option. :)
> > Using the vanilla bobber would be the _nicest_ code (and seems more future-proof). Not sure if there's downsides to making a fake player for the fisherman for that integration though, or other complications that'd mean that might not work. But if we want to go this way I think it should be a separate PR anyway.
> 
> Yeah, I'd like to avoid mixins if it's not 100% necessary. I mean, just using the vanilla bobber with a fakeplayer would be pretty easy to test. Just needs to create a vanilla bobber instead, should be a 3 line change.

Nvm, I just remembered, we use the bobber callback for a lot of stuff.

uecasm:
> should be a 3 line change

Pretty sure it'd be more complicated than that.  It'd need a custom fake player class for a start (`FakePlayer` doesn't work because it doesn't have a world position, which the bobber requires).

someaddons:
How about we simply define our own custom fishing loottable for our fisher, as modded items added to the default table often enoguh are just unusable/hindering citizens work and pollute the storage + it may disturb the other mods balance aswell. I'd just make our own custom fishing loot table which is easy to add aswell since we have our custom bobber which does the looting and if needed a modpack can always provide a datapack for the fisher's custom loottable and add more stuff

Raycoms:
> How about we simply define our own custom fishing loottable for our fisher, as modded items added to the default table often enoguh are just unusable/hindering citizens work and pollute the storage + it may disturb the other mods balance aswell. I'd just make our own custom fishing loot table which is easy to add aswell since we have our custom bobber which does the looting and if needed a modpack can always provide a datapack for the fisher's custom loottable and add more stuff

yeah, that's one of the options. I also think that's probably the best thing to do. Take vanilla loot table and alter it a bit to our requirements. In the end there is already a fisher loottable now.

uecasm:
True, that's probably a lot easier and tidier than the other alternatives above.  We sacrifice a bit of compatibility but _most_ people will probably be adding things to the child tables instead of directly to the fishing root table.  (Or using the global loot modifiers, which are utterly incompatible with JEI previews anyway.)  I'll get on that then.

(To be clear, I was thinking only of making a new top-level table that still uses the same child tables, so that we still do get modded fish etc when they're added to the "right" vanilla table, so we still retain some automatic compatibility.  Which is I think not quite what someaddons was suggesting.  But modpack authors will have the option to completely override it with something entirely custom if they don't like what some mod is adding.)

uecasm:
> (To be clear, I was thinking only of making a new top-level table that still uses the same child tables, so that we still do get modded fish etc when they're added to the "right" vanilla table, so we still retain some automatic compatibility. Which is I think not quite what someaddons was suggesting. But modpack authors will have the option to completely override it with something entirely custom if they don't like what some mod is adding.)

Or I guess the other option would be to merge the two such that the current "bonus loot" table becomes the only loot table, and then you can completely customise the drops per building level, which is perhaps closer to what someaddons was suggesting.  But you'd definitely lose all mod-fish compatibility that way, so I'm less enthusiastic about it.  (You'd also lose the "Luck of the Sea" effect behaviour.)

Raycoms:
> > (To be clear, I was thinking only of making a new top-level table that still uses the same child tables, so that we still do get modded fish etc when they're added to the "right" vanilla table, so we still retain some automatic compatibility. Which is I think not quite what someaddons was suggesting. But modpack authors will have the option to completely override it with something entirely custom if they don't like what some mod is adding.)
> 
> Or I guess the other option would be to merge the two such that the current "bonus loot" table becomes the only loot table, and then you can completely customise the drops per building level, which is perhaps closer to what someaddons was suggesting. But you'd definitely lose all mod-fish compatibility that way, so I'm less enthusiastic about it. (You'd also lose the "Luck of the Sea" effect behaviour.)

let's go with the first then.

uecasm:
Updated: now removes the sponge from L3 and sets up some side tables for the regular fish loot, but with a different treasure condition that will actually work for our custom bobber.  (They mostly delegate to the standard loot tables, but can easily be made unique in a modpack if someone wants to.)  Also fixed up some loot display so that it properly renders drops with counts, damage, and enchants.

Note that there's only one set of regular fishing loot tables (not per-hut-level) and they are still keyed off the Luck of the Sea enchant, while the bonus loot tables are split per hut level and are keyed off the worker skill.  Both tables are rolled on every cast (and if they produce an item will also produce XP orbs).

Raycoms (approved):

someaddons (approved):

PR #7749, opened 2021-11-07
Assignment rework

Raycoms:
Assignment rework:
- Clean up job translation keys
- Give jobs and views their entry
- Connect the buildings and crafting modules with the right job
- Create assignment module for each job creation type.
- Rework hiring window to work for each module type.
- Allow to mix archers/knights in a barracks tower

Review please


src/main/java/com/minecolonies/coremod/client/gui/modules/WindowHutCrafterTaskModule.java
```diff
@@ -48,7 +48,7 @@ public void onOpened()
         super.onOpened();
         final List<IToken<?>> tasks = new ArrayList<>();
 
-        for (final int citizenId : ((IBuildingWorkerView) buildingView).getWorkerId())
+        for (final int citizenId : buildingView.getModuleView(WorkerBuildingModuleView.class).getWorkerId())
```
uecasm: There's some other places where you use `getModuleViews`, implying that there may be zero or more than one, neither of which this would cope with.  Is that intended?  (There may be other places that also do this.)

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -642,16 +662,15 @@ public IRecipeStorage getFirstFulfillableRecipe(final Predicate<ItemStack> stack
     public boolean fullFillRecipe(final IRecipeStorage storage)
     {
         final List<IItemHandler> handlers = building.getHandlers();
+        final ICitizenData data = building.getModuleMatching(WorkerBuildingModule.class, m -> m.getJobEntry().equals(jobEntry)).getFirstCitizen();
```
uecasm: I feel like this expression should be in a helper method, since it's used in a few places (`getModuleForJob` or similar).

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -664,38 +683,28 @@ public boolean fullFillRecipe(final IRecipeStorage storage)
         return storage.fullfillRecipe(builder.create(RecipeStorage.recipeLootParameters), handlers);
     }
 
+    @Nullable
     @Override
-    public void updateWorkerAvailableForRecipes()
+    public IJob<?> getCraftingJob()
     {
-        for (final IToken<?> token : recipes)
+        if (jobEntry == null)
```
uecasm, line 690: Unlike the former `getMainBuildingJob`, this will always return a null-citizen Job, never the actual job with the real citizen data.  That doesn't matter for anything JEI-related (as it would always be null anyway) but the theory was that there might be some cases where it would be useful -- though I don't know if anything did rely on it so far, so perhaps this is ok.

src/main/resources/assets/minecolonies/lang/en_us.json
```diff
@@ -2032,6 +2004,46 @@
...
+  "com.minecolonies.job.teacher": "Teacher",
+  "com.minecolonies.job.dyer": "Dyer",
+  "com.minecolonies.job.glassblower": "Glassblower",
+  "com.minecolonies.job.mechanic": "Mechanic",
+  "com.minecolonies.job.fletcher": "Fletcher",
+  "com.minecolonies.job.planter": "Planter",
+  "com.minecolonies.job.rabbitherder": "Rabbitherder",
+  "com.minecolonies.job.concretemixer": "Concrete Mixer",
+  "com.minecolonies.job.beekeeper": "Beekeeper",
+  "com.minecolonies.job.cookassistant": "Assistant Cook"
```
uecasm: Is there some reason the `coremod` part was removed?  Granted it seems a little silly but changing this will break any existing translations...

Raycoms: Yes, on one hand, I want to slowly get rid of the "coremod" stuff. And on the other hand are all of those now consistent. I wanted to be able to properly differentiate between new old and if sth breaks where we are still expecting the old it will be easier to find.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: Yeah I was debating that too, I added such a helper in the AI, but I wasn't sure if I should add one into the building actually.

src/main/java/com/minecolonies/coremod/client/gui/modules/WindowHutCrafterTaskModule.java
Raycoms: Not sure I understood what you meant here.

uecasm: There's a mixture of calls to `getModuleView`, which can only return exactly one module, and `getModuleViews`, which can return zero or more modules.  It's suspicious to use both on the same module type, because they're contradictory.     It's not _necessarily_ a bug, especially if there can only be zero or one (and not more than one), but it's still potentially inconsistent.

Raycoms: True, let me make this more consistent.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
Raycoms: true, adjusted

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
```diff
@@ -0,0 +1,279 @@
...
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+
+import static com.minecolonies.api.util.constant.NbtTagConstants.*;
+
+/**
+ * The living module for citizen to call their home.
```
someaddons: docs

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
```diff
@@ -0,0 +1,279 @@
...
+        this.canWorkingDuringRain = canWorkingDuringRain;
+        this.sizeLimit = sizeLimit;
+    }
+
+    @Override
+    public boolean assignCitizen(final ICitizenData citizen)
+    {
+        if (citizen.getWorkBuilding() != null)
+        {
+            citizen.getWorkBuilding().getFirstOptionalModuleOccurance(WorkerBuildingModule.class).ifPresent(m -> m.removeCitizen(citizen));
```
someaddons: should probably lookup all assignment modules here since we can have more than one on a building right

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
```diff
@@ -0,0 +1,279 @@
...
+            Log.getLogger().warn("Unable to assign citizen:" + citizen.getName() + " to building:" + building.getSchematicName() + " jobname:" + getJobDisplayName());
+            return false;
+        }
+        return true;
+    }
+
+    @Override
+    public void deserializeNBT(final CompoundNBT compound)
+    {
+        if (compound.getAllKeys().contains(TAG_WORKER))
```
someaddons, line 106: no super deserialize?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
```diff
@@ -0,0 +1,279 @@
...
+        {
+            final ListNBT workersTagList = compound.getList(TAG_WORKER, Constants.NBT.TAG_COMPOUND);
+            for (int i = 0; i < workersTagList.size(); ++i)
+            {
+                final ICitizenData data = building.getColony().getCitizenManager().getCivilian(workersTagList.getCompound(i).getInt(TAG_WORKER_ID));
+                if (data != null && data.getJob() != null && data.getJob().getJobRegistryEntry().equals(jobEntry))
```
someaddons: this looks a bit weird, is it not this very class creating the job instance? and shouldnt it do so on loading?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
```diff
@@ -0,0 +1,279 @@
...
+    {
+        this.jobEntry = entry;
+        this.primary = primary;
+        this.secondary = secondary;
+        this.canWorkingDuringRain = canWorkingDuringRain;
+        this.sizeLimit = sizeLimit;
+    }
+
+    @Override
+    public boolean assignCitizen(final ICitizenData citizen)
```
someaddons, line 82: shouldnt this create the job on assign? the colony tick call seems to indicate so

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
```diff
@@ -0,0 +1,279 @@
...
+            final ICitizenData joblessCitizen = colony.getCitizenManager().getJoblessCitizen();
+            if (joblessCitizen != null)
+            {
+                assignCitizen(joblessCitizen);
+            }
+        }
+    }
+
+    @Override
+    public void serializeNBT(final CompoundNBT compound)
```
someaddons, line 151: no super call aswell?

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/WorkerBuildingModuleView.java
```diff
@@ -165,9 +156,37 @@ public int getMaxInhabitants()
...
+        return jobEntry;
+    }
+
+    /**
+     * Check if the module is full.
+     * @return true if so.
+     */
+    public boolean isFull()
+    {
+        return getWorkerId().size() >= getMaxInhabitants();
```
someaddons: should probably refactor it to getWorkerIDList()

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -176,14 +170,13 @@ public BlockPos getNextSittingPosition()
...
 
-
     /**
      * Get the status of the assistant processing requests
      * @return true if currently crafting
      */
     public boolean getIsCooking()
     {
-        ICitizenData citizen = getAssistant();
+        final ICitizenData citizen = this.getModuleMatching(CraftingWorkerBuildingModule.class, m -> m.getJobEntry().equals(ModJobs.cookassistant)).getFirstCitizen();
```
someaddons: can always compare job entries with ==

src/main/java/com/minecolonies/coremod/colony/jobs/views/DefaultJobView.java
```diff
@@ -26,9 +27,9 @@
     private final IColonyView colonyView;
 
     /**
-     * The synched job name.
+     * The synched job registry entry.
      */
-    private String name;
+    private JobEntry entry;
```
someaddons, line 32: We probably also should give the non-view jobs their entry object instead of hardcoding the return, to at least be consistent with job/jobview

someaddons (commented):
Looks overall fine, left some comments

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
Raycoms: No, there is none in any super class.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
Raycoms: No, on loading the citizen at this point is loaded with its job already.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
someaddons: ah should probably move the saving of the citizen list into the assignment module then and serialize/deserialize it there

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
Raycoms: No, the onAssignment sets the workerBuilding

    void onAssignment(final ICitizenData citizen)
    {
        citizen.setWorkBuilding(building)

And in here the job is created

Raycoms: I didn't change this, this could potentially moved up to "onAssignment" but that might change the current workflow and could be risky.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/WorkerBuildingModule.java
Raycoms: No, on purpose I didn't because we want to properly divide the "housing assignment" from the "living assignment". That is especially important for backwards compat.

src/main/java/com/minecolonies/coremod/colony/jobs/views/DefaultJobView.java
Raycoms: Can do in the next PR, feel like there is even more things we can do justifying more PRs =D

someaddons: well it'd be part of your description of "Give jobs and views their entry" :D

PR #7770, opened 2021-11-14
Some fixes

someaddons:
Fix block search Util not advancing Y levels correctly
Fix racks to update to their newer rotations

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
+++ b/src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
@@ -47,6 +47,16 @@ import static com.minecolonies.api.util.constant.TranslationConstants.RACK;
 public class TileEntityRack extends AbstractTileEntityRack
 {
     /**
+     * All Racks current version id
+     */
+    private static final byte VERSION = 2;
+
+    /**
+     * The racks version
+     */
+    private byte version = 0;
+
+    /**
      * The content of the chest.
      */
     private final Map<ItemStorage, Integer> content = new HashMap<>();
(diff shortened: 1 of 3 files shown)


src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
```diff
@@ -474,6 +486,11 @@ public void rotate(final Rotation rotationIn)
     @Override
     public <T> LazyOptional<T> getCapability(@Nonnull final Capability<T> capability, final Direction dir)
     {
+        if (version != VERSION)
+        {
+            updateBlockState();
```
Mekle001, line 491: Shouldn't we be updating the version here? It seems like it will always be zero and do this every time.

Raycoms (approved):

Mekle001 (changes requested):

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
Raycoms: yep

Raycoms (changes requested):

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
someaddons: indeed, good catch

Mekle001 (approved):

src/api/java/com/minecolonies/api/tileentities/TileEntityRack.java
uecasm: It'll also do it again on version 2->3, although I suppose the code can be changed when that actually happens to fix it then...

PR #7771, opened 2021-11-15
better permission handling

Raycoms:
Closes #
Closes #
Closes #

- Add permission for hurt_citizens and hurt_visitor for damage > 1
- Allow editing owner permission for owner only
- Some cleanup

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowPermissionsPage.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowPermissionsPage.java
@@ -438,16 +438,14 @@ public class WindowPermissionsPage extends AbstractWindowTownHall
 
         final IPermissions permissions = building.getColony().getPermissions();
         final PlayerEntity playerEntity = Minecraft.getInstance().player;
-        final Rank owner = building.getColony().getPermissions().getRankOwner();
-        if (!permissions.hasPermission(playerEntity, Action.EDIT_PERMISSIONS)
-              || (actionsRank == owner && (building.getColony().getPermissions().getRank(playerEntity) != owner || action == Action.EDIT_PERMISSIONS)))
+        if (!permissions.hasPermission(playerEntity, Action.EDIT_PERMISSIONS) || !permissions.canAlterPermission(permissions.getRank(playerEntity), actionsRank, action))
         {
             return;
         }
 
         final boolean trigger = LanguageHandler.format(COM_MINECOLONIES_COREMOD_GUI_WORKERHUTS_RETRIEVE_ON).equals(button.getTextAsString());
         Network.getNetwork().sendToServer(new PermissionsMessage.Permission(building.getColony(), PermissionsMessage.MessageType.TOGGLE_PERMISSION, actionsRank, action));
-        building.getColony().getPermissions().togglePermission(actionsRank, action);
+        building.getColony().getPermissions().togglePermission(permissions.getRank(playerEntity), actionsRank, action);
 
         if (trigger)
         {
--- a/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowPermissionsPage.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowPermissionsPage.java
@@ -440,7 +440,7 @@ public class WindowPermissionsPage extends AbstractWindowTownHall
         final PlayerEntity playerEntity = Minecraft.getInstance().player;
         final Rank owner = building.getColony().getPermissions().getRankOwner();
         if (!permissions.hasPermission(playerEntity, Action.EDIT_PERMISSIONS)
-              || (actionsRank == owner && building.getColony().getPermissions().getRank(playerEntity) != owner))
+              || (actionsRank == owner && (building.getColony().getPermissions().getRank(playerEntity) != owner || action == Action.EDIT_PERMISSIONS)))
         {
             return;
         }
--- a/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowPermissionsPage.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowPermissionsPage.java
@@ -67,6 +67,7 @@ public class WindowPermissionsPage extends AbstractWindowTownHall
      * A list of ranks (excluding owner)
      */
     private final List<Rank> rankList = new LinkedList<>();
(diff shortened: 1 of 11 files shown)


someaddons (dismissed):

src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowPermissionsPage.java
```diff
@@ -430,10 +433,19 @@ private void addBlock()
...
+        final IPermissions permissions = building.getColony().getPermissions();
+        final PlayerEntity playerEntity = Minecraft.getInstance().player;
+        final Rank owner = building.getColony().getPermissions().getRankOwner();
+        if (!permissions.hasPermission(playerEntity, Action.EDIT_PERMISSIONS)
+              || (actionsRank == owner && (building.getColony().getPermissions().getRank(playerEntity) != owner || action == Action.EDIT_PERMISSIONS)))
+        {
+            return;
+        }
+
+        final boolean trigger = LanguageHandler.format(COM_MINECOLONIES_COREMOD_GUI_WORKERHUTS_RETRIEVE_ON).equals(button.getTextAsString());
```
uecasm, line 446: If this is only checked on client-side, it means that technically someone could hack the client to let them still mess with the owner's permissions...

src/main/java/com/minecolonies/coremod/colony/permissions/Permissions.java
```diff
@@ -571,8 +573,7 @@ public void savePermissions(@NotNull final CompoundNBT compound)
     @Override
     public boolean hasPermission(final Rank rank, @NotNull final Action action)
     {
-        return (rank == ranks.get(OWNER_RANK_ID) && action != Action.GUARDS_ATTACK)
-                 || Utils.testFlag(permissionMap.get(rank), action.getFlag())
+        return Utils.testFlag(permissionMap.get(rank), action.getFlag())
```
uecasm, line 584: ... while this means that owners don't automatically get su-style "ignore permission settings" for everything.  Which should mostly be fine, but they could lose permissions if they're set weird currently.  (Notably it also means that the owners who currently deliberately perform "lethal selection" on their colonists will suddenly not be able to any more until they change this setting...)

This is not necessarily a bad thing, and is probably intended.  But it seems worth pointing out.

src/main/java/com/minecolonies/coremod/colony/permissions/PermissionsView.java
```diff
@@ -125,8 +125,7 @@ public boolean hasPermission(final UUID id, @NotNull final Action action)
      */
     public boolean hasPermission(final Rank rank, @NotNull final Action action)
     {
-        return (rank.getId() == OWNER_RANK_ID && action != Action.GUARDS_ATTACK)
-                 || (permissions != null && action != null && permissions.containsKey(rank) && Utils.testFlag(permissions.get(rank), action.getFlag()));
+        return permissions != null && action != null && permissions.containsKey(rank) && Utils.testFlag(permissions.get(rank), action.getFlag());
```
uecasm, line 128: Same as above.

src/main/java/com/minecolonies/coremod/colony/permissions/Permissions.java
Raycoms: That probably applies to the above, serverside enforcement. To not allow the server owner from locking themselves out.

Raycoms: The lethal selection "guards attack" actually means "guards will attack this player" which by default is false for owners.

uecasm: No, by "lethal selection" I meant players intentionally killing citizens/visitors because they wanted different ones.  Just pointing out that this change will mean that owners will no longer be able to do that unless they explicitly enable the permission.

The other part is that prior to this change, the code would just grant all permissions (except `GUARDS_ATTACK`) regardless of whatever settings were in the town hall, and after this change the settings in the town hall will actually matter to owners.  This will be fine if the town hall settings are left at defaults, but if they've toggled things around to a silly configuration then it may affect them poorly.

As I said, these are both probably intended changes, but are worth noting and may surprise people and generate questions.

What _might_ be still worthwhile actually changing though is that while `togglePermission` protects against disabling the permissions that owners should always have, if an existing colony has some of these required owner permissions already missing then they're going to now take effect and the owner won't be able to fix their colony permissions manually.  (Unless there's already somewhere outside of this PR that fixes these permissions if missing; I haven't checked the rest of the code.)

Raycoms: "No, by "lethal selection" I meant players intentionally killing citizens/visitors because they wanted different ones. Just pointing out that this change will mean that owners will no longer be able to do that unless they explicitly enable the permission."

Yes, gotta be either lava, or alter the permission. That's okay.

"The other part is that prior to this change, the code would just grant all permissions (except GUARDS_ATTACK) regardless of whatever settings were in the town hall, and after this change the settings in the town hall will actually matter to owners. This will be fine if the town hall settings are left at defaults, but if they've toggled things around to a silly configuration then it may affect them poorly."

Yes, that's why I tried to make sure they can't change any setting that could lock them out.

"What might be still worthwhile actually changing though is that while togglePermission protects against disabling the permissions that owners should always have, if an existing colony has some of these required owner permissions already missing then they're going to now take effect and the owner won't be able to fix their colony permissions manually. (Unless there's already somewhere outside of this PR that fixes these permissions if missing; I haven't checked the rest of the code.)"

Someone has to test this, from my tests, for the owner, everything defaults to on. But might be worth a test.

uecasm: > Someone has to test this, from my tests, for the owner, everything defaults to on. But might be worth a test.

For new colonies, sure.  But if you make a colony in an older version, can you toggle off those permissions there (which has no effect in that version) and then upgrade to this version and then oops, you're locked out?

Or even if the old version UI wouldn't let you do that either, if there's some corrupt data in the colony permissions and those are absent then nobody would have noticed before (since it would have worked anyway) but now their colony is unusable after upgrade.

(And we know we've had some weird colony permissions data saved in the past, such as the null custom ranks issues.)

Raycoms: Old UI didn't have owner toggles.

someaddons (approved):

PR #7786, opened 2021-11-19
Fix raiders & guards attacking while dying (#7785)

### LINKED ISSUE #7785: Raiders can attack during their death animation
### Prerequisites 

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: 1.0.147-ALPHA
- Structurize Version: 0.13.252-ALPHA

## Expected behavior

When a mob's health reaches zero, there's a short period of time where their death animation is played (they turn red, flop sideways and disappear). Raiders should not be able to attack during this death animation.

## Actual behavior

Raiders continue to attack while dying.

1. (Optional) Minimize the attack delay for raiders to make the bug more obvious. Imo, the easiest way to do this is by temporarily changing [this](https://github.com/ldtteam/minecolonies/blob/61f47ed35fad461cac06406ea65ca47c15ba7bd4/src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/RaiderRangedAI.java#L138) and [this](https://github.com/ldtteam/minecolonies/blob/61f47ed35fad461cac06406ea65ca47c15ba7bd4/src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/RaiderMeleeAI.java#L69) to return 1.
2. Start a raid at a colony using `/minecolonies raid-All now`
3. Use `/tp @e[type=minecolonies:barbarian,limit=1] @p` to isolate one of the raiders from the others.
4. Allow the raider to hit you, then kill it.
5. The raider should land one or two hits on you while it's dying.

### Logs

- latest.log: 
- crashlog: 

### Notes

---
## Viewers

Vortrox:
Closes #7785

- Stop the entity using AttackMoveAI from attacking if it is dead or dying. This affects raiders and guards.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/mobs/AbstractEntityMinecoloniesMob.java
+++ b/src/api/java/com/minecolonies/api/entity/mobs/AbstractEntityMinecoloniesMob.java
@@ -378,6 +378,11 @@ public abstract class AbstractEntityMinecoloniesMob extends MobEntity implements
     @Override
     public void aiStep()
     {
+        if (!this.isAlive())
+        {
+            return;
+        }
+
         updateSwingTime();
 
         if (invulTime > 0)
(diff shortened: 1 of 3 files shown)


someaddons (commented):
Hi and thanks for wanting to contribute to minecolonies, for this fix I'd prefer if we instead of messing with the case by case logic here would add an isAlive() check into the entities aiStep() call to prevent any AI from running when the entity is dead instead.

Raycoms:
> Hi and thanks for wanting to contribute to minecolonies, for this fix I'd prefer if we instead of messing with the case by case logic here would add an isAlive() check into the entities aiStep() call to prevent any AI from running when the entity is dead instead.

I second that, good comment

Raycoms (approved):

someaddons (approved):

PR #7825, opened 2021-12-01
Fix following guards idling on player death

Vortrox:
Prior to this, guards that were set to follow the player from their building's hut block would stop following the player and instead idle around the player's spawn if the player died.

The problem was that the PlayerEntity object of the player to follow before and after dying was not the same and it wasn't null. It seems like the old PlayerEntity is sent to the player's spawn instead of being deleted while the player is given a new PlayerEntity to control. This makes the building guards refer to the wrong entity.

This fix works by keeping track of the player's UUID (which doesn't change after death) and resolving that into the correct PlayerEntity object when needed.

Closes #
Closes #
Closes #

- Make AbstractBuildingGuards store the UUID of the player to follow instead of PlayerEntity, then get the correct PlayerEntity object when needed
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
@@ -531,7 +531,7 @@ public abstract class AbstractBuildingGuards extends AbstractBuilding implements
     public BlockPos getPositionToFollow()
     {
         PlayerEntity followPlayer = getPlayerFromUUID(followPlayerUUID, this.colony.getWorld());
-        if (getSetting(GUARD_TASK).getValue().equals(GuardTaskSetting.FOLLOW) && followPlayer != null)
+        if (getSetting(GUARD_TASK).getValue().equals(GuardTaskSetting.FOLLOW) && followPlayer != null && followPlayer.level.dimension() == this.colony.getDimension())
         {
             return new BlockPos(followPlayer.position());
         }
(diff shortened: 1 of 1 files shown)


Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingGuards.java
```diff
@@ -529,6 +530,7 @@ public void setGuardPos(final BlockPos guardPos)
     @Override
     public BlockPos getPositionToFollow()
     {
+        PlayerEntity followPlayer = getPlayerFromUUID(followPlayerUUID, this.colony.getWorld());
         if (getSetting(GUARD_TASK).getValue().equals(GuardTaskSetting.FOLLOW) && followPlayer != null)
```
Raycoms: In general it's fine. But we should have another check here to make sure the player is in the same dimension.

Vortrox: While working on this I found out that following guards can't use portals so they wouldn't be able to follow players across dimensions anyway, is this intended?

Raycoms: Yes fully, citizens don't work well outside of their own dimension.

Raycoms (approved):

someaddons (approved):

PR #7833, opened 2021-12-03
Better guard arrows

someaddons:
-Make guards only consume arrows for damage when actually hitting entities

Review please


src/main/java/com/minecolonies/coremod/entity/CustomArrowEntity.java
```diff
@@ -83,6 +90,20 @@ protected void onHitEntity(EntityRayTraceResult traceResult)
 
         // Set the old actual damage value back
         setBaseDamage(prevDamage);
+        if (onHitCallback != null && onHitCallback.test(traceResult))
```
Raycoms, line 93: Kinda find it weird if a predicate has a side-effect, maybe make it a producer or a function?

someaddons: its the same as a function, it just doesnt need unboxing

PR #7842, opened 2021-12-07
Update smelter to be 1.17 aware

Mekle001:
- Add tags for breakable ores and for raw ores
- Enable smelter to break ores into raw ores with building-level appropriate fortune
- Enable smelter to also break gems/etc. 

Review please


src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
```diff
@@ -352,6 +352,7 @@ public static void init(final RegistryEvent.Register<BuildingEntry> event)
                                   .addBuildingModuleProducer(() -> new CraftingWorkerBuildingModule(ModJobs.smelter, Skill.Athletics, Skill.Strength, false, (b) -> 1), () -> WorkerBuildingModuleView::new)
                                   .addBuildingModuleProducer(() -> new BuildingSmeltery.SmeltingModule(ModJobs.smelter), () -> CraftingModuleView::new)
                                   .addBuildingModuleProducer(FurnaceUserModule::new)
+                                  .addBuildingModuleProducer(() -> new BuildingSmeltery.CraftingModule(ModJobs.smelter), () -> CraftingModuleView::new)
```
uecasm: ~~Should probably list the crafting module before the smelting module, for consistency (this affects the tab order, I think).~~  Realised afterwards that this is a custom module rather than a crafting module, so the order is actually consistent -- but the class is poorly named.

Or use the variant that doesn't show a tab (omit creating the view), since it's a custom module and I don't think there'd be any need to re-order the "recipes".

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSmeltery.java
```diff
@@ -159,4 +172,69 @@ private static IGenericRecipe createSmeltingRecipe(final ItemStorage input, fina
...
+        {
+            super.checkForWorkerSpecificRecipes();
+
+            for(Item input: ModTags.breakable_ore.getValues())
+            {
+                Block b = Block.byItem(input);
+                List<ItemStack> drops = Block.getDrops(b.defaultBlockState(), (ServerLevel) building.getColony().getWorld(), building.getID(), null);
+                for(ItemStack drop: drops)
+                {
+                    if(!drop.isEmpty())
```
uecasm: Parenthesis spacing.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/smelter/EntityAIWorkSmelter.java
```diff
@@ -27,7 +44,7 @@
 import static com.minecolonies.api.util.constant.Constants.RESULT_SLOT;
 import static com.minecolonies.api.util.constant.Constants.STACKSIZE;
 import static com.minecolonies.api.util.constant.TranslationConstants.*;
-
+import static com.minecolonies.api.entity.ai.statemachine.states.AIWorkerState.*;
 /**
```
uecasm, line 43: Missing blank line.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/smelter/EntityAIWorkSmelter.java
```diff
@@ -71,7 +87,7 @@ protected void extractFromFurnace(final FurnaceBlockEntity furnace)
     {
         final ItemStack ingots = new InvWrapper(furnace).extractItem(RESULT_SLOT, STACKSIZE, false);
         final int multiplier = getOwnBuilding().ingotMultiplier((int) (getSecondarySkillLevel()), worker.getRandom());
-        int amount = ingots.getCount() * multiplier;
+        int amount = ingots.getCount(); //* multiplier;
```
uecasm: Multiplier calculation should be removed if it's not going to be used.

src/main/resources/data/minecolonies/tags/items/breakable_ore.json
```diff
@@ -0,0 +1,23 @@
...
+      "minecraft:redstone_ore",
+      "minecraft:deepslate_redstone_ore",
+      "minecraft:emerald_ore",
+      "minecraft:deepslate_emerald_ore",
+      "minecraft:lapis_ore",
+      "minecraft:deepslate_lapis_ore",
+      "minecraft:diamond_ore",
+      "minecraft:deepslate_diamond_ore",
+      "minecraft:nether_gold_ore",
+      "minecraft:nether_quartz_ore"
```
uecasm, line 21: These should use the minecraft ore tags (e.g. `#minecraft:coal_ores` includes both regular and deepslate already).

And preferably, there should be a datagen for this rather than being hand-written (although this is not without precedent).

But also if this is going to be a tag for ore blocks then it should be used _instead of_ `Tags.Items.ORES`, not in addition to it.  Or rather, this tag should just include the standard ores tag plus (temporarily) the vanilla ores that Forge hasn't added yet, such that additional modded ores will likely already be included in the list without need for further datapack adjustment.  Either way, the code should not be checking both.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/smelter/EntityAIWorkSmelter.java
```diff
@@ -71,7 +87,7 @@ protected void extractFromFurnace(final FurnaceBlockEntity furnace)
     {
         final ItemStack ingots = new InvWrapper(furnace).extractItem(RESULT_SLOT, STACKSIZE, false);
         final int multiplier = getOwnBuilding().ingotMultiplier((int) (getSecondarySkillLevel()), worker.getRandom());
```
Raycoms: I'd cleanup the old ingotMultiplier code here entirely

src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
Raycoms: I'd show it, like in the forester to be informative (gotta make sure teaching is disabled though).
But yes, should be consistent order

src/main/resources/data/minecolonies/tags/items/breakable_ore.json
Raycoms: Yeah, joining the vanilla ore tag + the missing here would be better

uecasm: Might need a safety check in the code such that if the result of breaking a `breakable_ore` is still a `breakable_ore` then it doesn't get stuck (just ignores it for breaking, and either smelts if smeltable or ignores entirely until the couriers remove), since some modded ores may still drop as themselves instead of as raw ores.  But I don't think the tag itself is the right way to distinguish that.

Mekle001: Not using the ores tag was intentional, because of modded ores.

Raycoms: Ohh, these here are the breakable ores, yes, we shouldn't just have a generic ore tag in here. We could include though things like #minecraft:coal_ores because we know those are broken down properly.

Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSmeltery.java
Mekle001: I'm not sure what you're getting at here, I don't see anything wrong with the spacing

src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
Mekle001: Renamed

src/main/resources/data/minecolonies/tags/items/breakable_ore.json
Mekle001: I wasn't sure which of those were right, and which weren't, and was having troubles viewing them so opted for consistency.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingSmeltery.java
uecasm: You don't have a space between the `for` and `(` or before the `:`, both of which are standard in the codebase.

This also applies elsewhere (such as with `if`) and in other files.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -670,19 +670,36 @@ public boolean fullFillRecipe(final IRecipeStorage storage)
...
-        final int luck = (int)(((primarySkill + 1) * 2) - Math.pow((primarySkill + 1 ) / 10.0, 2));
 
         LootContext.Builder builder =  (new LootContext.Builder((ServerLevel) building.getColony().getWorld())
                                           .withParameter(LootContextParams.ORIGIN, worker.position())
                                           .withParameter(LootContextParams.THIS_ENTITY, worker)
-                                          .withParameter(LootContextParams.TOOL, worker.getMainHandItem())
+                                          .withParameter(LootContextParams.TOOL, getCraftingTool(worker))
                                           .withRandom(worker.getRandom())
-                                          .withLuck((float) luck));
+                                          .withLuck(getCraftingLuck(worker)));
```
Raycoms, line 679: Don't we have the fortune already, not sure what the luck is for then.

Mekle001: We don't use luck in this case. Luck is for other loottable scenarios, where there isn't a tool. The future netherworker for example.

someaddons (approved):

PR #7845, opened 2021-12-08
Hut block as anchor in decoration

MotionlessTrain:
- Let a builder place a hut block on the anchor position of a decoration.
- This fixes the netherguard decoration of fortress, and allows custom building schematics be used as parent buildings in a cluster.
- Solve the javadoc issues, that lets the build server think the build failed the last week(s).

At the moment, it seems like the hut that gets placed try to find their schematics in the "wooden" style.
I tested it with Decorations -> fortressalternative -> netherguard1, and it was searching for schematics/wooden/netherguard2.
I haven't found yet why it defaults to wooden there and what can be done to correct that.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
@@ -98,7 +98,8 @@ public abstract class AbstractEntityAIStructure<J extends AbstractJobStructure<?
 
         return worldState.getBlock() instanceof IBuilderUndestroyable
                  || worldState.getBlock() == Blocks.BEDROCK
-                 || (info.getBlockInfo().getState().getBlock() instanceof AbstractBlockHut && handler.getWorldPos().equals(worldPos));
+                 || (info.getBlockInfo().getState().getBlock() instanceof AbstractBlockHut && handler.getWorldPos().equals(worldPos)
+                       && worldState.getBlock() instanceof AbstractBlockHut);
     };
 
     /**
--- a/src/main/java/com/minecolonies/coremod/client/gui/huts/WindowHutCrusherModule.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/huts/WindowHutCrusherModule.java
@@ -45,7 +45,7 @@ public class WindowHutCrusherModule extends AbstractWindowWorkerModuleBuilding<B
     /**
      * Constructor for the window of the miner hut.
      *
-     * @param building {@link BuildingMiner.View}.
+     * @param building {@link BuildingCrusher.View}.
      */
     public WindowHutCrusherModule(final BuildingCrusher.View building)
     {
--- a/src/main/java/com/minecolonies/coremod/client/gui/modules/FarmerFieldsModuleWindow.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/modules/FarmerFieldsModuleWindow.java
@@ -113,7 +113,7 @@ public class FarmerFieldsModuleWindow extends AbstractModuleWindow
     /**
      * Constructor for the window of the farmer.
      *
-     * @param building {@link BuildingFarmer.View}.
+     * @param moduleView {@link FarmerFieldModuleView}.
      */
     public FarmerFieldsModuleWindow(final IBuildingView building, final FarmerFieldModuleView moduleView)
     {
--- a/src/main/java/com/minecolonies/coremod/client/gui/modules/GraveyardManagementWindow.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/modules/GraveyardManagementWindow.java
@@ -72,7 +72,7 @@ public class GraveyardManagementWindow extends AbstractModuleWindow
     /**
      * Constructor for the window of the graveyard.
      *
-     * @param building {@link BuildingGraveyard.View}.
+     * @param moduleView {@link GraveyardManagementModuleView}.
      */
     public GraveyardManagementWindow(final IBuildingView building, GraveyardManagementModuleView moduleView)
     {
--- a/src/main/java/com/minecolonies/coremod/client/gui/modules/WindowHutMinerModule.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/modules/WindowHutMinerModule.java
@@ -41,7 +41,7 @@ public class WindowHutMinerModule extends AbstractModuleWindow
     /**
      * Constructor for the window of the miner hut.
      *
-     * @param building {@link BuildingMiner.View}.
+     * @param moduleView {@link MinerLevelManagementModuleView}.
      */
     public WindowHutMinerModule(final IBuildingView building, final MinerLevelManagementModuleView moduleView)
     {
--- a/src/main/java/com/minecolonies/coremod/client/gui/modules/WindowMineGuardModule.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/modules/WindowMineGuardModule.java
@@ -36,7 +36,7 @@ public class WindowMineGuardModule  extends AbstractModuleWindow
     /**
      * Constructor for the window of the miner hut.
      *
-     * @param building {@link BuildingMiner.View}.
+     * @param building {@link IBuildingView}.
      */
     public WindowMineGuardModule(final IBuildingView building)
     {


src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAIStructure.java
```diff
@@ -98,7 +98,8 @@
 
         return worldState.getBlock() instanceof IBuilderUndestroyable
                  || worldState.getBlock() == Blocks.BEDROCK
-                 || (info.getBlockInfo().getState().getBlock() instanceof AbstractBlockHut && handler.getWorldPos().equals(worldPos));
+                 || (info.getBlockInfo().getState().getBlock() instanceof AbstractBlockHut && handler.getWorldPos().equals(worldPos)
+                       && worldState.getBlock() instanceof AbstractBlockHut);
     };
```
Raycoms, line 103: What exactly does this change?

MotionlessTrain: It makes sure that the anchor block with the AbstractBlockHut in the schematic is only ignored when there is an AbstractBuilding in the world. E.g. for a building request, it will ignore the position as before, but in a decoration request, if there needs to go a hut block there, the builder will place it there.

Raycoms (approved):

marchermans:
I do not find this a good idea....

We explicitly defined the concept here to not allow this..

But alright

uecasm:
Why do you want a decoration with a hut block as anchor?

The usual rules (as far as I understand them):

- if you want it to be a building, the anchor must be the hut block for that building
- otherwise if you want it to be an upgradable decoration, the anchor must be a Decoration Controller
- otherwise if you want it to have tags (currently just `groundlevel`), the anchor must be a Tag Anchor
- otherwise the anchor must be a Tag Anchor or a regular block that is none of the above

I don't see the point of having a hut block as anchor if it's not a building, other than being lazy when scanning and not overriding the default anchor search when one isn't set.

marchermans:
> Why do you want a decoration with a hut block as anchor?
> 
> The usual rules (as far as I understand them):
> 
> * if you want it to be a building, the anchor must be the hut block for that building
> * otherwise if you want it to be an upgradable decoration, the anchor must be a Decoration Controller
> * otherwise if you want it to have tags (currently just `groundlevel`), the anchor must be a Tag Anchor
> * otherwise the anchor must be a Tag Anchor or a regular block that is none of the above
> 
> I don't see the point of having a hut block as anchor if it's not a building, other than being lazy when scanning and not overriding the default anchor search when one isn't set.

I totally agree, at the moment I see no reason to change this logic so drastically.

uecasm:
What might perhaps be useful would be to have the scan tool display an error and refuse to scan when the anchor isn't set (or isn't within the scan bounds) and the search finds more than one possible anchor block.  Won't fix all cases (e.g. a deco containing only one hut and no controller/tag anchor) but should act as a reminder for forgetful people for the common cases.  Though that would be a Structurize change rather than here.

MotionlessTrain:
> Why do you want a decoration with a hut block as anchor?
> 
> The usual rules (as far as I understand them):
> 
> * if you want it to be a building, the anchor must be the hut block for that building
> * otherwise if you want it to be an upgradable decoration, the anchor must be a Decoration Controller
> * otherwise if you want it to have tags (currently just `groundlevel`), the anchor must be a Tag Anchor
> * otherwise the anchor must be a Tag Anchor or a regular block that is none of the above
> 
> I don't see the point of having a hut block as anchor if it's not a building, other than being lazy when scanning and not overriding the default anchor search when one isn't set.

The idea was to be able to make cluster buildings with a custom style. E.g. a guard tower which can be used to create an entire district with multiple herders inside of it. If you want to do multiple of those, the <style>/guardtower and <stylealternative>/guardtower are not enough (especially if you want to design another “regular” guard tower as well). The solution for that is to make a decoration, containing only the guard tower, which can then be built and upgraded to get the herders.
Out of the discussion of discord, the question arose if the level 1 of the custom style of the guard tower could serve as that decoration itself. That turned out not to be the case, (as found with schematics/frontieralternative/netherguard1), as everything but the hut block gets placed.
According to SomeAddons, a hut block should also get placed in decorations with that hut block as anchor block, hence this pull request.

uecasm:
> The idea was to be able to make cluster buildings with a custom style. E.g. a guard tower which can be used to create an entire district with multiple herders inside of it. If you want to do multiple of those, the <style>/guardtower and /guardtower are not enough (especially if you want to design another “regular” guard tower as well). The solution for that is to make a decoration, containing only the guard tower, which can then be built and upgraded to get the herders.
Out of the discussion of discord, the question arose if the level 1 of the custom style of the guard tower could serve as that decoration itself. That turned out not to be the case, (as found with schematics/frontieralternative/netherguard1), as everything but the hut block gets placed.

It wouldn't really be in the correct directory to be a decoration that way anyway.  Just make a `schematics/decorations/frontier/netherguard` decoration (with a regular non-hut anchor), that in turn contains the `schematics/frontieralternative/netherguard0` hut block.  It could also have additional huts, or you could wait and put the extra huts in the `netherguard1` or higher level schematic instead if you want them to progressively show up.  (Although the longer you leave it, the more likely the player would have already built the regular version elsewhere.)

Or perhaps rather than `decorations` we could standardize on another name (like the `roads`/`rails` etc) for this sort of thing, e.g. `districts`, especially if more styles do it.

MotionlessTrain:
> What might perhaps be useful would be to have the scan tool display an error and refuse to scan when the anchor isn't set (or isn't within the scan bounds) and the search finds more than one possible anchor block. Won't fix all cases (e.g. a deco containing only one hut and no controller/tag anchor) but should act as a reminder for forgetful people for the common cases. Though that would be a Structurize change rather than here.

If it turns out this is not allowed at all, building a custom building as decoration, I think it is better that it is disallowed in some way. They way it is now is non intuitive: the preview and material list both show a guard tower is going to be placed, but it never will be.
The scan tool change would be a good step forwards to disallow it I think, to prevent scanning errors causing this. Perhaps a change for the build tool is also necessary, so that custom buildings like the netherguard can never be placed as decoration request in the first place

someaddons (approved):
I think the key point here is that if the schematic has a hut block at a place, and its missing in the world the builder should place it, regardless of anchor pos or not

Raycoms:
> I think the key point here is that if the schematic has a hut block at a place, and its missing in the world the builder should place it, regardless of anchor pos or not

I agree with SomeAddons here. That is a fine condition.

uecasm:
> I think the key point here is that if the schematic has a hut block at a place, and its missing in the world the builder should place it, regardless of anchor pos or not

I don't disagree with that, but it just seems like this whole thing started from a poor premise.

marchermans:
> I think the key point here is that if the schematic has a hut block at a place, and its missing in the world the builder should place it, regardless of anchor pos or not

This is not the default? Why?

someaddons:
> > I think the key point here is that if the schematic has a hut block at a place, and its missing in the world the builder should place it, regardless of anchor pos or not
> 
> This is not the default? Why?

Its an oversight, a bug xD

marchermans:
> > > I think the key point here is that if the schematic has a hut block at a place, and its missing in the world the builder should place it, regardless of anchor pos or not
> > 
> > 
> > This is not the default? Why?
> 
> Its an oversight, a bug xD

Then that should be fixed, and merged, but this PR does not do that.

marchermans:
Or at least does not do that targeted at that.

Raycoms:
> Or at least does not do that targeted at that.

It fixes it, although they encountered this bug in a different context.

LadyShatana:
Adding to this, I propose a new hut block, called something like "District" or something as an anchor block. Works like a merger of the Deco Controller and the Barracks, where the main Building can be upgraded and targets specific "child" blocks. Then those standard child hut block in the build will build according to the instructions of the new anchor block. 

Specific example: Herder Hall - the new District controller indicates that this is level one of the hall in the target folder. It targets the cavern/alternative child blocks. 4 child blocks are added to the schematic (sheep, cow, chicken, pic) and will build in place the exact same way the BarracksGuard towers do.

My first time adding anything to the github, I hope it makes sense and I did it right!

MotionlessTrain:
> > The idea was to be able to make cluster buildings with a custom style. E.g. a guard tower which can be used to create an entire district with multiple herders inside of it. If you want to do multiple of those, the <style>/guardtower and /guardtower are not enough (especially if you want to design another “regular” guard tower as well). The solution for that is to make a decoration, containing only the guard tower, which can then be built and upgraded to get the herders.
> > Out of the discussion of discord, the question arose if the level 1 of the custom style of the guard tower could serve as that decoration itself. That turned out not to be the case, (as found with schematics/frontieralternative/netherguard1), as everything but the hut block gets placed.
> 
> It wouldn't really be in the correct directory to be a decoration that way anyway. Just make a `schematics/decorations/frontier/netherguard` decoration (with a regular non-hut anchor), that in turn contains the `schematics/frontieralternative/netherguard0` hut block. It could also have additional huts, or you could wait and put the extra huts in the `netherguard1` or higher level schematic instead if you want them to progressively show up. (Although the longer you leave it, the more likely the player would have already built the regular version elsewhere.)
> 
> Or perhaps rather than `decorations` we could standardize on another name (like the `roads`/`rails` etc) for this sort of thing, e.g. `districts`, especially if more styles do it.

I can at least fix the netherguard decoration so that it can be build in the correct way (assuming I get Moobien's consent for it). I think that will be a separate pull request though.

uecasm:
> Adding to this, I propose a new hut block, called something like "District" or something as an anchor block. Works like a merger of the Deco Controller and the Barracks, where the main Building can be upgraded and targets specific "child" blocks. Then those standard child hut block in the build will build according to the instructions of the new anchor block.
> 
> Specific example: Herder Hall - the new District controller indicates that this is level one of the hall in the target folder. It targets the cavern/alternative child blocks. 4 child blocks are added to the schematic (sheep, cow, chicken, pic) and will build in place the exact same way the BarracksGuard towers do.

If I'm understanding this correctly, the only effect of this different from using a decoration as parent is that it forces you to build five levels of the parent and to upgrade the parent before you can upgrade any of the children.  Which sounds potentially annoying to players... and you can already do it by picking an existing building type (e.g. farmer or mystical site) to be the parent (which is what Frontier does, incidentally).

LadyShatana:
> 
> If I'm understanding this correctly, the only effect of this different from using a decoration as parent is that it forces you to build five levels of the parent and to upgrade the parent before you can upgrade any of the children. Which sounds potentially annoying to players... and you can already do it by picking an existing building type (e.g. farmer or mystical site) to be the parent (which is what Frontier does, incidentally).

These are alternate builds, if the player doesn't want a full Hall they have the regular huts to choose from. In this particular case the herder hut footprint was a small part of the entire build (like the barracks towers) and the required materials at each level were in both the base building and the hut (like the barracks). 

By allowing a new anchor (Dirstrict/Hall) block it would also be less confusing than telling folks "OK build the Alternate Cowboy and it will be this giant Hall and then you build the other 3 herders in that"  All this said....I can work around what is allowable now.....would just be nice to have this other setup option.

uecasm:
> These are alternate builds, if the player doesn't want a full Hall they have the regular huts to choose from. In this particular case the herder hut footprint was a small part of the entire build (like the barracks towers) and the required materials at each level were in both the base building and the hut (like the barracks).
> 
> By allowing a new anchor (Dirstrict/Hall) block it would also be less confusing than telling folks "OK build the Alternate Cowboy and it will be this giant Hall and then you build the other 3 herders in that" All this said....I can work around what is allowable now.....would just be nice to have this other setup option.

Ok, but I don't see how that's any different from doing one of:
* Make both `stylename/hutname1` and `stylename/alt/hutname1` (and `stylename/hall/hutname1` etc).
    * With the hutblock in hand, the player can easily choose the type from the middle combo.
    * This is more visible when the player is already trying to build that specific building (so it helps when the relationship is logical, such as warehouse to courier).
    * Downside: they need to have the hutblock in hand (which usually means someone needs to tell them that the alternative hall styles exist, when the connection might be less logical, such as cowboy vs other herders).
* Make a `stylename/decorations/xyz-hall` decoration that contains the special unconstructed children.
    * This is more visible when the player is looking around the menu without a specific building in mind.
    * You can still limit the children to the parent version if you want by only having one hut in the decoration and having the others show up when that hut is built.
        * But I still wouldn't recommend using the cowboy as the parent for that case; there's no logical reason why players should be prevented from upgrading their shepherd before their cowboy, or whatever other order they like.
    * Mystical Site or Guard Tower are probably reasonably good do-mostly-nothing candidates to use as parent buildings, when nothing more logical (like warehouse/courier or university/library/school) presents itself.

LadyShatana:
I don't want the player to have to/be able to choose the style. I would also love to be able to lock the huts to the main hall.

Huts within the hall are built on a much smaller, specific footprint. If a player used it outside the hall they could "cheat the system" a bit by using this much smaller version.

If the new block is a merge with the Deco Controller, we can add a command to point to the specific style of the child blocks.

So a person building, for example, the Cavern Herder Hall would:

1) make the Hall/District block

2) Select the style CavernHerder

Place the build, builder builds the Hall build. They now have inside the Hall the 4 child herder huts that are already in the correct place, and will only build and upgrade to CavernAlternative. (Or CavernHerder if we can take that next step)

As they upgrade the Hall, they can upgrade the huts. In total the Hall + Huts equal to the same amount of supplies, difficulty, rarity of the 4 huts put together. But in a lovely harmonized build.

The player still has the regular huts as an option, so if they don't want to go this route they don't have to.

MotionlessTrain:
This thing you describe is already possible, by abusing a guard tower, as @uecasm already mentioned.
Make a guard tower which would encompass the entire hall, and save it under schematics/cavernherder/guardtower1 (to 5)
That guard tower would do exactly the same as that district building, plus expanding the boundary by two chunks (instead of one), giving the workers the feeling they are save, and you get an extra guard of course.
In this guard tower, you can include the herders with e.g. cavernherder/shepherd style, or if you want to forbid the players to build it loose, save the buildings with another name (cavernherder/s1 for the level 1 shepherd for instance). They need to be saved in cavernherder if you place that "district" guard tower in caverherder as well, otherwise they won't work

uecasm:
Or better, save it under `cavern/herder` rather than `cavernherder`.  Hierarchical style organisation is better than polluting the top level, existing examples to the contrary notwithstanding.

Which reminds me: @Raycoms: while granted, it might annoy patreons a bit, any chance we can slip in a rename of all the `Xalternative` styles to `X/alt` or something before 1.18 goes public and it becomes entrenched forever again?

It's unfortunately a bit harder now than it used to be since the schematic data needs to be updated as well, it's not a pure file rename.  But that can be done with a script (or updating the schem server).

LadyShatana:
I get what you are saying about this being _possible_ by abusing existing huts. I would simply like a proper system with a block for this purpose, which is easier for new players to use and understand. Want to know what Halls/Districts are available? Grab the District block and look in the menu, see (for example) Cavern/Herder, Cavern/Crafter, Frontier/Herder, Frontier/Supply etc etc.  Also the increased utility of the block possible by what I have suggested already, which would make builders happy.

That said, I will table this dream for now. I know the dev team is very busy and doing an amazing job.

uecasm:
Building types should usually **do** something, though.  Even the mystical site has a function.

Building type aside, It sounds like you're mostly asking for two things:

* A way to make it easier to find districts/halls/neighbourhoods/whatever

    This could be handled by coming up with some reasonable common grouping directory for decorations, like the way that fields/roads/rails are split off from "general" decorations (as I said before).  And this way, you don't need to craft an extra hut block to get them, they're always available.

* An easier way to get a decoration parent to dictate the style of its child huts.

    This sounds like a good feature request ;)

PR #7858, opened 2021-12-16
Journeymap integration

uecasm:
- Changes permission map to 64-bit, since we now have 33 permissions.
- Adds permission upgrade code for existing colonies, and fixes upgrade of "hurt citizen/visitors" permissions.
- Syncs colony grave locations to client as part of the general colony sync data.
    - Does not update the GraveyardManagementModule to use this instead of separately passing the info
- Adds two kinds of integration with Journeymap:
    - Colony borders (uses colony team colour)
    - Colonist deathpoints (auto-removed when grave decays or is harvested)
image image
image

- Each has two ways to disable display:
    - A particular player can disable display globally by changing a setting in their client config.
    - The owner/officer of a colony can toggle whether players with particular roles can see them.  This also allows them to turn them off for themselves temporarily during gameplay.
- By default, borders are visible to all players, and deathpoints are only visible to owners/officers.

Review please

Caveats:
- ~~Not yet mergeable; the colony border doesn't work in the current build of Journeymap (due to a JM bug) and requires an unreleased prerelease.~~  I'm also hoping that some of the utility code might get upstreamed.
- ~~Not yet tested in multiplayer.  Don't see any obvious footguns but I've been wrong about that before.~~
- Due to the way that colony chunk data is loaded, colony borders will not appear until you've moved about 2 chunks from your initial login position.  (This is the same as the build tool border render.)
- Colony borders and deathpoints are non-persistent and will disappear if you log out or switch dimensions.  They will not reappear again until you re-enter the colony (so you're not going to see persistent borders for multiple colonies unless you've been there recently).  So _mostly_ it's just for seeing your own borders without the build tool up, not for keeping an eye on neighbours.
- A custom icon for the deathpoints is included, but JM currently does not actually support displaying a custom icon (despite it being in their API).  It's there as a future enhancement.


uecasm:
New Journeymap has released now, so bumped versions.

marchermans (dismissed):

someaddons (dismissed):

PR #7865, opened 2021-12-20
Clean compat dependencies in test environment

uecasm:
- Omits compat dependencies from development/test environment by default; they're opt-in instead.

Review please

(This reduces potentially confusing log messages generated from mods outside of our control.)

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -26,16 +26,19 @@ dependencies {
     library fg.deobf("com.ldtteam:structurize:${project.structurize_version}")
     library "com.ldtteam:datagenerators:${project.dataGeneratorsVersion}"
     library fg.deobf("mezz.jei:jei-${project.minecraftVersion}:${project.jei_version}")
-    library fg.deobf("com.resourcefulbees:ResourcefulBees:${project.minecraftVersion}-${project.resourceFullBeesVersion}")
-    library fg.deobf("com.ferreusveritas.dynamictrees:DynamicTrees-${project.minecraftVersion}:${project.dynamicTreesVersion}")
-    library fg.deobf("slimeknights.tconstruct:TConstruct:${project.minecraftVersion}-${project.tinkersConstructVersion}")
-    library fg.deobf("slimeknights.mantle:Mantle:${project.minecraftVersion}-${project.mantleVersion}")
 
     apiLibrary fg.deobf("com.ldtteam:structurize:${project.structurize_version}")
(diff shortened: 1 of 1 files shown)


marchermans (changes requested):
Testing with compat dependencies is important, this achieves only the fact that people will forget to test with these mods.

uecasm:
They're important when you're testing compat -- otherwise they can just get in the way of what you're actually trying to test.

For example, with this set of mods loaded, there is a large JEI warning spam on startup about runtime ingredients (and I have no idea which ones, because there's no attribution).  This makes it less clear which warnings are generated by Minecolonies and which externally, which makes it harder to tell which ones can actually be fixed.

They also potentially cause undesirable changes of behaviour -- e.g. imagine if Quark were added to the compat list later on, with its massive list of vanilla feature changes.  This makes it harder to test cases where the mod is not included.

Even in the current mod list, Dynamic Trees is another such problematic default include -- generate a world with dynamic trees loaded and now you can't test vanilla tree behaviour any more.  (Well, it's not completely impossible, but you have a lot more hoops to jump through -- when Dynamic Trees is not in many modpacks and so it's a very non-default configuration to test in.)

It's always easier to add a mod for testing to an existing world than it is to remove it again (missing blocks/items and other errors, plus forever a warning on startup for the missing mod), which is another reason why they should be opt-in, not opt-out.

Raycoms:
> They're important when you're testing compat -- otherwise they can just get in the way of what you're actually trying to test.
> 
> For example, with this set of mods loaded, there is a large JEI warning spam on startup about runtime ingredients (and I have no idea which ones, because there's no attribution). This makes it less clear which warnings are generated by Minecolonies and which externally, which makes it harder to tell which ones can actually be fixed.
> 
> They also potentially cause undesirable changes of behaviour -- e.g. imagine if Quark were added to the compat list later on, with its massive list of vanilla feature changes. This makes it harder to test cases where the mod is not included.
> 
> Even in the current mod list, Dynamic Trees is another such problematic default include -- generate a world with dynamic trees loaded and now you can't test vanilla tree behaviour any more. (Well, it's not completely impossible, but you have a lot more hoops to jump through -- when Dynamic Trees is not in many modpacks and so it's a very non-default configuration to test in.)
> 
> It's always easier to add a mod for testing to an existing world than it is to remove it again (missing blocks/items and other errors, plus forever a warning on startup for the missing mod), which is another reason why they should be opt-in, not opt-out.

I agree on this here, pre OPC we didn't have all those deps loaded by default either.

PR #7869, opened 2021-12-22
Fix tree search

someaddons:
Closes #
Closes #
Closes #

Fix tree pathfinding accessing world chunks from pathing thread

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
@@ -1,5 +1,6 @@
 package com.minecolonies.coremod.colony.managers;
 
+import com.google.common.collect.ImmutableMap;
 import com.ldtteam.structurize.util.LanguageHandler;
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.IColony;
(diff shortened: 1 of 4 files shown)


Raycoms (approved):

src/main/java/com/minecolonies/coremod/colony/managers/BuildingManager.java
```diff
@@ -506,8 +507,19 @@ public IBuilding addNewBuilding(@NotNull final AbstractTileEntityColonyBuilding
     @Override
     public void removeBuilding(@NotNull final IBuilding building, final Set<ServerPlayerEntity> subscribers)
     {
-        if (buildings.remove(building.getID()) != null)
+        if (buildings.containsKey(building.getID()))
         {
+            final ImmutableMap.Builder<BlockPos, IBuilding> builder = new ImmutableMap.Builder<>();
+            for (final IBuilding tbuilding : buildings.values())
+            {
+                if (tbuilding != building)
```
uecasm, line 515: Shouldn't this be comparing the building ids instead?  It might be possible to not have reference equality...

someaddons: if we do not have reference equality here then we have a huge error elsewhere, as this very map should be the one holding all existing references

marchermans (approved):

PR #7870, opened 2021-12-22
Fixes rendering of rack window

uecasm:
- Makes rendering of rack window use the same constants and conditions as the rack container slots, for consistency.
    - There's still a few magic numbers left behind, but I'm not sure of the derivation of those.
- Simplifies rendering of large racks.
- Corrects calculation of rack window size.
    - This makes JEI correctly keep out of the way of the larger windows.

Review please

(click me)
[preview](https://youtu.be/Ga1IuOfJr_I)


someaddons (approved):

Raycoms (approved):

marchermans (approved):
Very nice

PR #7878, opened 2021-12-28
Port Journeymap and racks etc to 1.18

uecasm:
- Port #7844
- Port #7870
- Port #7858
    - Minimap doesn't show colony border for some reason, but it seems like a JM bug since it works in 1.16
- Clean up dependency version properties and automate mods.toml
- Bump min forge version both for JM compatibility and to ensure log4j-safe version

Review please


uecasm:
(And now actually use the correct base branch, derp.)

PR #7884, opened 2021-12-31
Netherminer

Mekle001:
- First off, we have a new building/worker pair, in the Nether Miner and Nether Mine
- Wooden themed Nether Mine buildings level 1-5
- Skins and models for the NetherMine hut block, and the Nether Miner
- Research to unlock the building
- Tweaks all over the place to make things work. 
  - Added transparency layer support to hut blocks
  - Added support for invisible workers, to suppress many 'normal' behaviors while invisible. 
  - Added support to recipe fulfillment to return the list of generated items without populating the worker's inventory. 
  - Built a rudimentary 'simulation' capability based off of items generated from loottables. 
    - Loot table generates tokens with information about the encounter
    - a battle is simulated, with ~50% chance to hit on either side.
    - For the duration of the simulation, the worker wears the armor/weapon, so damage is taken, and reduced. 
    - Rounds continue to the death
    - Loot table also generates loot
    - Blocks are 'mined' with the appropriate tool, applying damage to the tool, and converting the blocks to the blocks loottable.
    - Number of items returned affects how long the overall simulation lasts. 

Review please


Mekle001:
This is a draft, because it's waiting on a few more textures (_d, _b, _a) to complete the set. The ones currently in the PR are placeholders.

uecasm:
I thought the idea was to have a "nether explorer/gatherer", not a miner?

Mekle001:
That's effectively what this is. It's been through 3 or 4 sets of names, and I'm open to others. But Nether Miner seems to resonate. From an AI perspective, it's a souped up variant of the sifter. The worker is modelled as a cross between a guard and a miner. It uses armor, and takes damage while it's away, with a significant chance of death. It returns with nether blocks and possibly a few mob drops if it survives. 

Currently, the blocks and mob drops are limited to open-nether spaces, and don't include bastions or fortresses. Some of the drops require you to have a higher level building, you don't start seeing ancient debris until building level 5 for instance.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
```diff
@@ -0,0 +1,675 @@
...
+                        
+                        damage -= damage * (getSecondarySkillLevel() * SECONDARY_DAMAGE_REDUCTION);
+
+                        int slotOfStack = InventoryUtils.findFirstSlotInItemHandlerNotEmptyWith(worker.getItemHandlerCitizen(), 
+                            itemStack -> !ItemStackUtils.isEmpty(itemStack) && itemStack.getItem() instanceof SwordItem );
+                        
+                        if(slotOfStack != -1)
+                        {
+                            int swordDamage = (int) Math.floor(mob.getHealth() / 3f);
+                            worker.getInventoryCitizen().damageInventoryItem(slotOfStack, swordDamage, worker, entity -> {});
```
uecasm: Shouldn't the strength of the sword (stone vs diamond vs extra enchants) affect the durability damage (as in, a higher damage output means less swings to kill, means less durability damage)?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
```diff
@@ -0,0 +1,675 @@
...
+    /**
+     * Close the nether portal while idle around the building
+     */
+    protected IAIState closePortal()
+    {
+        final BlockState block = world.getBlockState(getOwnBuilding().getPortalLocation());
+
+        if(block.is(Blocks.NETHER_PORTAL))
+        {
+            useFlintAndSteel();
```
uecasm, line 675: Closing the portal shouldn't require flint & steel.  Unless the portal is "off" while they're away, and they're actually opening and closing it again quickly to return, perhaps?

src/main/resources/data/minecolonies/loot_tables/recipes/netherworker/trip.json
```diff
@@ -0,0 +1,656 @@
+{
+    "name": "minecolonies:recipes/trip5",
```
uecasm: This name doesn't match the filename -- I assume this was because you had separate loot tables per hut level originally before adding the scoreboard stuff.  TBH, it'd be nicer if you went back to doing that, the JEI and recipe integration already supports separate loot tables per hut level but doesn't currently support scoreboard conditions.  While not impossible to support, it'd be quite complicated and entirely unnecessary if it was just done the easy way instead. 😀 

You can write a datagen for the loot table and custom recipes instead of hand-crafting them, to reduce the duplication of specification.  See the existing sifter/enchanter/fisher datagens.

Mekle001: The separate loot tables never existed, they were planned but I got the scoreboard conditions working, and never went there. 

Datagen was broken for me all week, so I've been working around it, and now I'm running out of holiday time, so I'd like to take that as a potential follow-on change.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: Yeah, it's open when they get back. I can remove it, but I didn't figure the extra damage was that onerous, and my goal was to keep costs of running this beast high.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: It should, if this were a full calculation. If you look closely, the worker actually only ever takes one hit of damage from any mob today, and the amount of damage was a rough estimate for multiple swings. this is a case where I was erroring on the side of greater damage, and higher risk for death of the worker. I'd rather back it down than ramp it up later.

uecasm: `damage` is the damage dealt to the worker by the mob, yes.  But this calculation isn't related to that, it's the durability damage to the sword.  Which is currently based on the mob health (as it should be), but _not_ based on the sword's strength (which it should be).  A stone sword will still break faster than a diamond sword (due to lower total durability) but it's still very different from how an actual fight would work.

Mekle001: Ok, I basically rewrote the simulation, so damage taken to the sword depends on how many times it hits the mob.

src/main/java/com/minecolonies/coremod/client/model/ModelEntityNetherworkerMale.java
```diff
@@ -0,0 +1,119 @@
...
+		leftArm.texOffs(32, 48).addBox(-1.0F, -2.0F, -2.0F, 4.0F, 12.0F, 4.0F, 0.0F, false);
+		leftArm.texOffs(48, 48).addBox(-1.0F, -2.0F, -2.0F, 4.0F, 12.0F, 4.0F, 0.25F, false);
+
+		rightLeg = new ModelRenderer(this);
+		rightLeg.setPos(-1.9F, 12.0F, 0.0F);
+		rightLeg.texOffs(0, 16).addBox(-2.0F, 0.0F, -2.0F, 4.0F, 12.0F, 4.0F, 0.0F, false);
+		rightLeg.texOffs(0, 32).addBox(-2.0F, 0.0F, -2.0F, 4.0F, 12.0F, 4.0F, 0.25F, false);
+
+		leftLeg = new ModelRenderer(this);
+		leftLeg.setPos(1.9F, 12.0F, 0.0F);
```
Raycoms: Please run a formatter on those two

src/main/java/com/minecolonies/coremod/colony/jobs/JobNetherWorker.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    public boolean isInNether()
+    {
+        return citizenInNether;
+    }
+
+    /**
+     * Get the queue of CraftedResults
+     */
+    public Queue<ItemStack> getCraftedResults()
```
Raycoms, line 163: A bit unclean this way. Instead of getting and inserting into the list. Rather have a method here that adds to the respective lists.

Mekle001: That's a lot of methods to surface, with add, addAll, isEmpty, poll and clear.

Raycoms: No, general poll and isEmpty are fine with getList (i.e. assume that the list is a copy). Just clear and add ItemStack ... stack is needed then.

Mekle001: poll on a queue is both a read and a remove. It can't operate on a copy.

Raycoms: Ah thought you meant with poll just a get and not pop.

Raycoms: Hmm, then add a comment to the return method to make clear that the content of the queue is manipulated on return. And only the add a list of stacks method for add.

src/main/java/com/minecolonies/coremod/client/model/ModelEntityNetherworkerMale.java
Mekle001: Done

src/main/java/com/minecolonies/coremod/colony/jobs/JobNetherWorker.java
Mekle001: done

src/main/resources/data/minecolonies/loot_tables/recipes/netherworker/trip.json
Mekle001: Thank you for your help with this conversion.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
```diff
@@ -0,0 +1,958 @@
...
+        boolean haveShovel = checkForToolOrWeapon(ToolType.SHOVEL);
+        boolean haveSword = checkForToolOrWeapon(ToolType.SWORD);
+        boolean haveLighter = checkForToolOrWeapon(ToolType.FLINT_N_STEEL);
+        if(haveAxe || havePick || haveShovel || haveSword || haveLighter)
+        {
+            worker.getCitizenData().setIdleAtJob(true);
+            return IDLE;
+        }
+
+        if (currentRecipeStorage == null)
```
Raycoms, line 217: when can this happen?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
```diff
@@ -0,0 +1,958 @@
...
+                }
+            }
+
+            return getState();
+        }
+        else 
+        {
+            if (!getOwnBuilding().canDoTrip())
+            {
+                worker.getCitizenData().setIdleAtJob(false);
```
Raycoms, line 242: Isn't that idle true then?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
```diff
@@ -0,0 +1,958 @@
...
+            {
+                int itemDelay = 0;
+                if(currStack.getItem() instanceof BlockItem)
+                {
+                    final BlockItem bi = (BlockItem) currStack.getItem();
+                    final Block block = bi.getBlock();
+                    final net.minecraftforge.common.ToolType toolType;
+
+                    if(block.getHarvestTool(block.defaultBlockState()) == null)
+                    {
```
Raycoms, line 497: formatting

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
```diff
@@ -0,0 +1,958 @@
...
+        }
+
+        return NETHER_RETURN;
+    }
+
+    // calculate the XP coming from certain ores
+    private int xpOnDrop(Block block)
+    {
+        Random rnd = worker.getRandom();
+        if (block == Blocks.COAL_ORE) {
```
Raycoms: formatting

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
```diff
@@ -0,0 +1,958 @@
...
+            InventoryUtils.transferFoodUpToSaturation((IBuilding) getOwnBuilding(), worker.getInventoryCitizen(), amountToTransfer, stack -> edible.matches(stack));
+        }
+    }
+
+    /**
+     * Attempt to eat to restore some saturation
+     */
+    protected void attemptToEat()
+    {
+        final IDeliverable edible = new StackList(getEdiblesList(), "Edible Food", 1);
```
Raycoms, line 811: A few of the methods down here seem like we got very similar utilities already (i.e. in the eating AI) we should potentially opt into attempting to reuse those

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: This is the start of a trip. When we have all of the basic equipment, and nothing else is blocking us is when we hit this point.  Further along, if we don't have the actual trip recipe inputs, we'll check to see if the portal is supposed to be closed, and isn't.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: No, this is the case when we *can't* do the job, potentially for days, because we've done one too recently. That's not idle, that's recovering. Otherwise we hit the idle state way too fast.

Raycoms: so setIdleAtJob = true -> we're not idle and false -> we're idle?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: Some of them were lifted from there with minor modifications. I tried to reuse them, but that attempt made quite a mess out of the eating AI, because of the "during simulation" vs "during statemachine" differences.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: That's not what I said. when !canDoTrip, we're not idle, because we're recovering from our last trip. If we set period to a multi day period, say 3 days, and our idle timeout before the gear is 3 days, then we're at high risk of the gear, even though we can't do the job because we're counting down the timeout. This way, we're not "idle" until we could go to the nether and haven't.

Raycoms: Ahhhh, understood. Maybe we rename canDoTrip then to sth "isTakingABreak" or so.

Mekle001: Let me think about that. How about "noRecentTrip"? So !noRecentTrip is logically consistent with !canDoTrip, but reads better.

Mekle001: Hmm.. I don't know that I like that either. It's structurally designed to support 2 trips every 3 days.

Raycoms: isReadyForTrip that implies more "waiting" rather than "lack of sth"

Mekle001: Done and done.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: fixed

src/main/java/com/minecolonies/coremod/entity/ai/citizen/netherworker/EntityAIWorkNether.java
Mekle001: fixed

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
```diff
@@ -0,0 +1,334 @@
...
+import java.util.Map;
+
+
+public class BuildingNetherWorker extends AbstractBuilding
+{
+
+    /**
+     * Settings
+     */
+    public static final ISettingKey<BoolSetting>       CLOSE_PORTAL      =
```
Raycoms: The formatting of this message seems weird.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
```diff
@@ -0,0 +1,334 @@
...
+
+    @Override
+    public int buildingRequiresCertainAmountOfItem(final ItemStack stack, final List<ItemStorage> localAlreadyKept, final boolean inventory)
+    {
+        if (stack.isEmpty())
+        {
+            return 0;
+        }
+
+        if (inventory && getFirstModuleOccurance(MinimumStockModule.class).isStocked(stack))
```
Raycoms, line 194: Shouldn't this be in the super already? If not we should probably do that then.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
```diff
@@ -0,0 +1,334 @@
...
+        {
+            return stack.getCount();
+        }
+
+        if (isAllowedFood(stack) && (localAlreadyKept.stream().filter(storage -> ISFOOD.test(storage.getItemStack())).mapToInt(ItemStorage::getAmount).sum() < STACKSIZE || !inventory))
```
Raycoms: I've been thinking a bit over this food business and I think we're doing a bit of an overkill here. Instead, we might just do food through:
a) We use the food selection list for the food request as is. This should be at most a oneliner in the AI then. (If not this should be easy to generify from the cook.
b) The worker itself is content with ANY food they have then. If food gets delivered to them, they eat it, no need to further check it. (Restaurant needs more complex things there because this one also hands out food).

So here, we just need an "is food check" which we can also simply have in keep x and avoid this entirely.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
```diff
@@ -0,0 +1,334 @@
...
+     * This should be a 'air block' in the portal that can directly be checked to see if the portal is open
+     * @return the block above the tag, null if not available
+     */
+    public BlockPos getPortalLocation()
+    {
+        if (tileEntity != null && !tileEntity.getPositionedTags().isEmpty())
+        {
+            for (final Map.Entry<BlockPos, List<String>> entry : tileEntity.getPositionedTags().entrySet())
+            {
+                if(entry.getValue().contains("portal"))
```
Raycoms: Don't we have any util to get blockpos for a tag, maybe make a getPosListForTag or so and just do get(0), this way we avoid having  a lot of duplicate code.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    /**
+     * On initial construction or reset request, excludes the tagged food by default.
+     *
+     * @param listModule The food exclusion module.
+     */
+    public static void onResetFoodExclusionList(final ItemListModule listModule)
+    {
+        listModule.clearItems();
```
Raycoms, line 307: Where does this one come from?

Raycoms (changes requested):
I made another partial review, the only thing missing to review fully now is the AI.

Did you implement already the "invisible downstairs room" ?

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
uecasm: Well, the caveat with that is that you don't want farmers eating raw carrots they've just farmed, or guards eating zombie flesh, or rabbithands eating the carrots or the raw rabbit ... unless the restaurant is configured to hand out carrots, in which case maybe you do.

To repeat an idea that's been suggested before, it might be simplest if colonists had a special "food" inventory slot that only the player and restaurant can touch, and they'll only eat food from there.  It's slightly more silly in that the farmer with carrots would refuse to eat them until they dump them in the warehouse, go to the restaurant, and then are given carrots to eat, but that's probably only a short-term weirdness that will go away once the colony has better food supplies anyway.

A general food overhaul is probably out of scope for the nether miner PR anyway though 😁

Raycoms: This code here applies only to the netherworker though. For the farmer stuff we got special job based checks.

uecasm: I know, but you were making general statements.  Though note that the nether miner can collect raw pork chops (and I think zombie flesh too)... (theoretically even cooked ones if they had a Fire Aspect sword, though I don't think that's considered in the simulated loot at present).  So "just eat any food" is not necessarily safe for them either.

Raycoms: "any edibles"

Raycoms: And no, I was trying to make general statements about the netherminer

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
Mekle001: Should be covered by the toKeep definition above. Removed

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
Mekle001: There isn't any util function like that, nor is there a lot of duplicate code The other places this is used is structured very differently.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingNetherWorker.java
Mekle001: From the cook.

someaddons (dismissed):

TagnumElite:
I feel like it would be nice to abstract this out so that there could be other "Explorers/Adventurers" later.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
```diff
@@ -190,6 +190,27 @@ else if (block instanceof BlockMinecoloniesRack)
         }
     }
 
+    /**
+     * Gets the list of tags, and finds the first location registered there. 
+     * @param tagName the name of the tag to query
+     * @return the BlockPos, or null if not found
+     */
+    protected BlockPos getLocationFromTag(String tagName)
```
Raycoms: getFirstLocationFromTag

someaddons (approved):

PR #7916, opened 2022-01-07
Netherminer: switch to per-level datagen

uecasm:
- Changes from static scoreboard-based recipe & loot table to datagen per-level recipes & loot tables.
    - This makes it mostly compatible with the existing JEI integration.  It also makes (the code) much easier to read.
- Fixes the adventure token loot table specification, which as far as I could tell was invalid before.
    - The quality value is an integer, so almost all of the values previously specified were being ignored.  I've multiplied the weights/qualities by 100 so that the qualities can actually take effect approximately how you had them.
    - The "group" wrapper was completely pointless (since each group only contained one child).
    - The "group" wrapper was counterproductive (since "weight"/"quality" on group is ignored; only "conditions" takes effect) -- so all your adventure tokens actually had weight=1 quality=0 before, despite what they looked like they were doing.

Review please

I have not tested the behaviour of the actual netherminer building, but in theory it should Just Work™ due to the default custom recipe management.  At least for this purpose the `HutLevel` scoreboard is no longer required; I haven't checked if it is used elsewhere (but I haven't removed it anyway).

The display of Adventure Tokens in JEI is probably not optimal (especially given that they don't have a language string, although that might be a bug anyway).  I'm open to suggestions for alternative displays; some possible options include:

1. Just hide them (not preferred)
2. Expand them into typical loot from the mob (probably preferred, since this is similar to the animal herders)
3. Leave them as is but give them a better tooltip or something

You can either merge this as is and I'll do another PR for JEI token updates or leave it for a bit and I'll add that to this PR later.


uecasm:
Also, it looks like previously the adventure tokens were not conditional on hut level at all.  Perhaps they should be; that'd be easy enough to change with the current code structure.

For example, Endermen mostly only appear in the Nether in the Warped biome (although this is not strictly exclusive), and Warped blocks are only gathered starting at level 3, so it might make sense to only start encountering them at that level too.  I haven't thought much about other possible changes.

Mekle001:
The groups were intended to go away, since early on they had both the mob loot and the adventure token, and the loot got removed with the last simulator rewrite.  I didn't realize that groups didn't support weighting. As for the floating points vs integers, that's my bad. for some reason I had it stuck in my head that floating points worked there. 

As for what to do about Adventure tokens, they are intended to never be seen by the player. They are an indicator from the loot table that an encounter took place, but they never make it into an inventory. 

This is likely why I forgot the language string, since they weren't visible anywhere I was looking. I think the easiest answer is to go with option 3

Overall, this looks really good, thank you @uecasm!

PR #7917, opened 2022-01-08
Expands adventure tokens into corresponding mob loot

uecasm:
- Expands adventure tokens in loot analysis to the corresponding mob loot instead.

Review please

For example, this allows you to press R (or click) on "pork chops" in JEI to see that you can get them both from the nether miner and the pig herder.


Mekle001 (approved):
Now I understand what you mean by "expand", I had something else in mind. I love this!

Issue #7899, opened 2022-01-04
Blacksmith doesn't proceed with requests from other citizen from the colony


Tellurnirn:
Minecolonies

Version: Minecolonies-1.18.1-1.0.278 on Server

Expected behavior

    When workers request a tool, the blacksmith should start make the tool. We also placed postboxes and stash ones, in order to place request or let citizen do it too, as we read it was a possible solution amongst fixed issues threads on here: https://github.com/ldtteam/minecolonies/issues/3534
  

Actual behaviour

    The blacksmith does nothing even though she has all the needed recipes and the material in her shop. It is only possible to type words in a box on the post boxes menu, it does not lead anywhere and nothing changes in the display doing so.

Addendum:

Don't know if it is related: but the list of items, to be selected in the minimum stock page after the player clicks on ADD, does not display anything but air in all the buildings. Hence nothing can be added to the list.

Thanks in advance

Raycoms:
Any startup logs?

Tellurnirn:
Is there anything that should be deleted from the logs so what I share here can't be used to access the server/mess with it?

Raycoms:
Not that I am aware of.

Tellurnirn:
Ok actually, the issue evolved. Somehow restarting the server every single time the display fails to show recipes works.
It's really time consuming but does the job.

So now the new problem is about the miner's hut block that simply can't find the framed cobblestone or the frame spruce plank items for the player to add to the minimal stock list. Typing their name in the search gives no results even though the miner says they can't work any further if they don't get these blocks. Blocks which totally exist as they're easy to find in creative mode...

Any ideas how to fix this?

PR #7918, opened 2022-01-08
Allows requesting items from the player's inventory as well

uecasm:
Related to #7899
Closes [discord request](https://discord.com/channels/139070364159311872/372217492547829762/928869704859279441)

- Adds unique items from the player inventory to the postbox and minimum stock requestables list (this allows things like dyed armor and DO blocks)

Review please

This was relatively straightforward and is consistent with how the scan tool replace feature supports DO blocks, although it's perhaps not the ideal solution -- it would be better to read the list of possible recipe outputs currently taught to any colony building and use that to augment the default item list **instead** of the player inventory.  However this is client side and the colony recipe data is all server side AFAIK (and the recipe manager is global, not per-colony, so there isn't any existing list of "all local craftables"), so that does not seem straightforward.

### DIFF
--- a/src/api/java/com/minecolonies/api/util/ItemStackUtils.java
+++ b/src/api/java/com/minecolonies/api/util/ItemStackUtils.java
@@ -1016,7 +1016,7 @@ public final class ItemStackUtils
             if (stack.isDamageableItem() && stack.isDamaged())
             {
                 pristine = stack.copy();
-                stack.setDamageValue(0);
+                pristine.setDamageValue(0);
                 // in case the item wasn't already in the set, we want to only store a pristine one!
             }
             allItems.add(new ItemStorage(pristine, true));
--- a/src/api/java/com/minecolonies/api/util/ItemStackUtils.java
+++ b/src/api/java/com/minecolonies/api/util/ItemStackUtils.java
@@ -998,14 +998,8 @@ public final class ItemStackUtils
      */
     public static Set<ItemStack> allItemsPlusInventory(@NotNull final PlayerEntity player)
     {
-        // using ItemStorage for better hashing
-        final Set<ItemStorage> allItems = new HashSet<>();
-
-        // all known items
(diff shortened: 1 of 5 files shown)


src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -991,5 +988,41 @@ public static ItemStack idToItemStack(final String itemData)
...
+        for (final ItemStack stack : player.inventory.items)
+        {
+            if (stack.isEmpty()) continue;
+
+            ItemStack pristine = stack;
+            if (stack.isDamageableItem() && stack.isDamaged())
+            {
+                pristine = stack.copy();
+                stack.setDamageValue(0);
+                // in case the item wasn't already in the set, we want to only store a pristine one!
```
MotionlessTrain, line 1014: Shouldn’t this be the `pristine` ItemStack? Now the item stack in the players inventory gets modified, and the damaged one gets stored in `allItems`

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -991,5 +988,41 @@ public static ItemStack idToItemStack(final String itemData)
...
+     * @param player The player whose inventory to check.
+     * @return The set of items.
+     */
+    public static Set<ItemStack> allItemsPlusInventory(@NotNull final PlayerEntity player)
+    {
+        // using ItemStorage for better hashing
+        final Set<ItemStorage> allItems = new HashSet<>();
+
+        // all known items
+        for (final ItemStack stack : IColonyManager.getInstance().getCompatibilityManager().getListOfAllItems())
```
Raycoms: I feel like we should partially cache this or so. So it's not rebuilt on every search

uecasm: Done.

someaddons (approved):

PR #7948, opened 2022-01-16
Improvements to feeding

uecasm:
- Allow chicken herder and shepherd to force-feed babies to increase growth rate too.
- Prevent butchering adults that have been recently fed (avoid wasting food).
- Add "feeding" setting to allow disabling force-feeding baby animals to reduce food usage.

Review please

Regarding the first, I can sort of see why disabling this for sheep might have been deliberate since they will also self-feed via eating grass blocks to improve their own growth rate -- but it's still possible for the player to hand-feed baby sheep, so it should be possible for the shepherd too.  I can't think of any reason why chickens might have been deliberate.

Having said that, there doesn't really seem to be any sort of rate limit on it -- once over the necessary (low) skill threshold they'll just keep using food to age them up until they're all fully grown or they run out of food.  While there's no limit in vanilla either, a player is probably less likely to always age up the babies.  Perhaps this needs a limit or a setting or both (~~not included here~~ setting now added, but perhaps still could use rate).

Regarding the second, I never directly observed this happening (and I'm not sure how likely it is based on timing) but the scenario to imagine is this:

1. A level 2 herder with 4 adult animals (at max limit).
2. They feed all the animals; this is two breeding pairs and so should result in two babies being born.
3. As soon as the first baby is born, there's now 5 animals in the pen and so the herder will start looking for animals to slaughter.
4. Previously, they could have selected one of the breeding pair that had not yet produced a baby (because they were still trying to path to each other, but the herder got there first).
    * The net result of this is that there will now be three adults (two just-bred, one just-fed and looking for a partner), and just one baby.
    * This is a waste of two units of food -- for the one killed while fed and also the one fed without being able to find a partner, as the "fed" state will wear off over time.  (And the herder will only ever feed two adults at a time, so even if another adult is introduced before it wears off, they won't feed the new adult until the old one is no longer fed.)

The change will refuse to slaughter just-fed animals (or babies, as before); the herder will still start looking for an animal to slaughter as soon as the first baby is born but will always choose one not just-fed (i.e. usually one of the parents of the new baby).

This should cause a clean sweep in the breeding cycle -- they will feed all the adults, kill all the parents as soon as they produce babies, feed the babies to grow them up faster, then repeat.  (In practice they'll probably get out of sync due to pathfinding delays and other reasons.)

It's possibly also worth mentioning that the beekeeper doesn't force-feed baby bees (which is why they lack that setting), even though it's possible for the player to do that, because they're using different AI from the other herders.  This hasn't been changed here.

### DIFF
--- a/src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
+++ b/src/main/java/com/minecolonies/apiimp/initializer/ModBuildingsInitializer.java
@@ -138,7 +138,8 @@ public final class ModBuildingsInitializer
                                        .setRegistryName(new ResourceLocation(Constants.MOD_ID, ModBuildings.CHICKENHERDER_ID))
                                        .addBuildingModuleProducer(() -> new WorkerBuildingModule(ModJobs.chickenHerder, Skill.Adaptability, Skill.Agility, false, (b) -> 1), () -> WorkerBuildingModuleView::new)
                                        .addBuildingModuleProducer(MinimumStockModule::new, () -> MinimumStockModuleView::new)
-                                       .addBuildingModuleProducer(() -> new SettingsModule().with(AbstractBuilding.BREEDING, new BoolSetting(true)), () -> SettingsModuleView::new)
+                                       .addBuildingModuleProducer(() -> new SettingsModule().with(AbstractBuilding.BREEDING, new BoolSetting(true))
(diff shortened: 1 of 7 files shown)


Raycoms (commented):
As this was previously capped by worker level, we might want this setting only to be visible if citizen hut level > x or worker level > x

uecasm:
The worker skill requirement is still there (and also still affects how much food is required to age up the babies).  The setting is in addition to that, so you can set it before you hire someone.

someaddons (approved):

PR #7982, opened 2022-01-24
improve danger check

Raycoms:
- Improve pathing passable danger check
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
+++ b/src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
@@ -1373,6 +1373,7 @@ public abstract class AbstractPathJob implements Callable<Path>
                         return true;
                     }
                 }
+                return false;
             }
         }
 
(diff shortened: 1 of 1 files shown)


marchermans (dismissed):

src/main/java/com/minecolonies/coremod/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -1353,17 +1354,25 @@ protected boolean isPassable(@NotNull final BlockState block, final BlockPos pos
...
+                }
+
+                if (shape.isEmpty() || shape.max(Direction.Axis.Y) <= 0.1  && !isLiquid((block)) && (block.getBlock() != Blocks.SNOW || block.getValue(SnowBlock.LAYERS) == 1))
+                {
+                    final PathNodeType pathType = block.getAiPathNodeType(world, pos);
+                    if (pathType == null || pathType.getDanger() == null)
+                    {
+                        return true;
+                    }
+                }
```
MotionlessTrain, line 1375: Is the possibility to `return false` in case the condition of the `if` is `false` removed intentionally?
Now, if the condition is false, it falls through to the `return true` at line 1379 instead

Raycoms: good sighting

someaddons (approved):

PR #7998, opened 2022-01-27
Added the capture outside animals feature to Herders

ImEr1c:
Closes #
Closes #
Closes #

-Added a feature to make Herders able to capture animals outside their work building
-
-

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/ai/statemachine/states/AIWorkerState.java
+++ b/src/api/java/com/minecolonies/api/entity/ai/statemachine/states/AIWorkerState.java
@@ -356,6 +356,10 @@ public enum AIWorkerState implements IAIState
      */
     HERDER_FEED(false),
 
+    HERDER_FIND(true),
+
+    HERDER_GET(true),
+
     /*
 ### Cowboy ###
      */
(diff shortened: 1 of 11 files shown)


Raycoms (changes requested):
First of all, thanks a lot for your contribution. We highly appreciate contributions from the community. 

First some general comments:
- Please try to contact either through a discussion or on discord before implementing a feature so that we can provide feedback on this beforehand and guarantee that this feature is actually something we "want".
- Second, we have a specific format that comes with the project, please apply this to all your changes before making a PR.

Some more specific to the implementation:
- The extra herder building is not necessary. If we provided a pen location it would be with the help of the tag tool (i.e. some center position in the pen). This avoids a lot of unnecessary complexity and a potential lag source.
- This brings me to point two, MineCraft is extremely sensitive in terms of lag and, as we expect many many herders on a server, small details can bring down an entire server (There were instances in the past where herders already did that).
So always make sure:
-- You double check a chunk is loaded before accessing it
-- Don't scan a large area, rather have the citizens walk around randomly in the colony and scan a rather small area (could re-use guard code)
-- Keep the "decide" method in the AI clean of heavy code. The herder should probably only attempt to find animals if they got nothing else to do.
- Immersion also plays a big role for us, so we'd want the herder to bring the animal on the leash back to the pasture and not teleport it.


The detail points aside, your implementation seems to have a particular focus to put animals back in the pen that escaped the herder. My plan to solve this was requiring large pastures for the animals that fulfill certain requirements (up to discussion) and may be built as a decoration (including tags that show the corners) or allow player built pastures even, and provide a tool in the building where a player can "set the bounding box".

PR #8055, opened 2022-02-11
attempt to fix #7963

### LINKED ISSUE #7963: Occasional mine node connection error
### Prerequisites 

- [x ] I am running the latest **alpha** version of MineColonies and Structurize for my Minecraft version.
- [x ] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure my issue is not covered there.
- [x ] I made sure that **this issue is not a duplicate of any existing issue.**

### Context 

- Minecraft Version: 1.18.1
- MineColonies Version: 1.18.1-1.0.328-ALPHA
- Structurize Version: 1.18.1-1.0.319-BETA

## Expected behavior
all mine nodes would connect to each other as intended

## Actual behavior
The first node from the mineshaft is always built as a straight node. I think the name is minerx2top. Occasionally the miner will build a node off the side of this node, even though it may not physically connect (depending on the schematic). So the actual root node is a minerx2top node, but the internal logic of how the mine interconnects must think it is a minerx3topleft or minerx3topright or minerx3leftright. (Sorry, the naming convention and how the nodes actually connect make no sense to me.) I think it is usually, perhaps always, the version that has a node connecting to the left as you are facing away from the mineshaft.

So the miner will create another node that, depending on the schematic, can't even be reached. I have manually fixed the connection in my game, but I shouldn't have to. Maybe the node physically built in-game is hard-coded to be minerx2top, but connection-logic thinks it could be something else?

This could result in a section of mine that isn't accessible to colonists.

0. Create a mine that uses minerx**** schematics that are all unique so you can identify which schematic was built. Also, it is best if you use a schematic that has a mechanism that points towards the root (mineshaft). For example, torches always placed away from the root/mineshaft.
1. Instruct your miner to mine on a fresh floor.
2. Let you miner mine for a while. Enough that the nodes diagonal from the mineshaft are mined out.
3. Check to see if one of the diagonal nodes from the mineshaft "thinks" it is connected to the side of one of the minerx2top nodes directly connected to the mineshaft.
4. If the bug didn't occur, start over at step 1 above.

### Logs

No logs...

### Notes
Could someone explain the naming logic of the minerx**** schematics? Maybe explain how to orient, modify, and save them in creative mode? This would be nice to have in the wiki. I got a set to work but it was by trial and error and took a long time. Since there are no anchor blocks, it's difficult to determine how to orient and save them.

---
## Viewers

Raycoms:
Closes #7963
Closes #
Closes #

- Should fix the mineshaft issues, still needs a bit of testing
-
-

Review please


MotionlessTrain:
The nodes in the new orientation and the code changes work correctly!

I did notice a few minor schematic issues though.

someaddons (approved):

PR #8056, opened 2022-02-11
Wall mode

uecasm:
Depends on ldtteam/structurize#486

- fix translation string in decoration controller "cancel repair/upgrade"
- fix translation string in rally banner gui
- decoration controller repair/upgrade now shows you the list of expected materials and allow custom builder assignment, just like the initial deco build
- make work orders for decoration controller repair/upgrade have a short name instead of the full schematic path
- support for "~wall~ extendable mode" (decorations only) -- see linked Structurize PR for usage
  - an upgradable wall build only gets one decoration controller, which indicates that it's ~a wall~ extended in the gui
- support for PlacementSettings api changes
- adds top-level sections for Roads, Rail, Walls, and Fields (this removes them from the Decorations menu)
- all `/walls/` will automatically get the Extendable Mode UI.  other schematics (including roads/rails) can only use it if a tag is added (but walls should be updated to use the tag too, so eventually this backwards-compatibility check can be removed)

Review please

image
Now: image

image
image

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -28,7 +28,8 @@ mappingsVersion=1.18
 
 dataGeneratorsVersion=0.1.48-ALPHA
 blockUI_version=1.18-0.0.42-ALPHA
-structurize_version=1.18.1-1.0.342-RELEASE
+#structurize_version=1.18.1-1.0.342-RELEASE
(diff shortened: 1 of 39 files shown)


uecasm:
Rebranding to Extendable Mode based on feedback; moving some mcol-specific compat code to here instead of struct.

PR #8079, opened 2022-02-19
Tags and Recipes datagen

### LINKED ISSUE #8067: Stonesmelter doesn't recognize the smooth nether bricks recipe
### Prerequisites 

- [X] I am running the latest **alpha** version of MineColonies and Structurize for my Minecraft version.
- [X] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure my issue is not covered there.
- [X] I made sure that **this issue is not a duplicate of any existing issue.**

### Context 

- Minecraft Version:
- MineColonies Version:
- Structurize Version:

## Actual behavior

Only the smooth nether bricks have this issue, other smooth blocks is recognized as normal

1. 
2. 
3. 

### Logs

- latest.log: 
- crashlog: 

### Notes

---
## Viewers

uecasm:
Closes #8067

- Moves tags to datagen
- Moves non-crafter recipes to datagen
- Makes tags use standard loading mechanism to reduce complexity
    - This does reintroduce some glitches in the tag contents when used on a server and enabling/disabling a datapack while live, but a `/reload` fixes it (may require a double reload on edit).  And presumably other mods suffer similar fates, so this shouldn't be too surprising.  Works as expected in SP and when first connecting to an MP server, as far as I can tell (but not extensively tested)
- Recipes now have unlock conditions, so will work in the vanilla recipe book
    - Most recipes are unlocked by having a build tool and sometimes another item in the inventory.  Some of these might be a bit too harsh (e.g. farm won't unlock until you craft and hold a wooden hoe, which you might never do if you've already advanced beyond that) so possibly could be wound back a bit.  But I did want to try to avoid unlocking absolutely everything on getting the build tool.  (And of course the unlocks only matter if you're playing with `doLimitedCrafting` enabled, which is not the default.)
- Some small tweaks to the tag contents
- Enables smoking and campfire cooking recipes for mcol foods (from dough) ... not too likely to be used, since the baker will usually do the full round using the smelting recipe, but seemed useful for consistency.

Review please

I did originally want to just do the recipes, but it turns out they don't play nice unless the tags are converted first.
As before, the commits are separated to make the diffs a bit more manageable.
I've avoided renaming the crafting tags at present to retain backwards compatibility, but there are a few that are a bit weird.  And ideally we'd use the same names for the tags as we use for the `crafter` in the custom recipes, which is not quite the case at present.
I'll call out some of the specific intended changes to the tag contents shortly.

The tags can be a bit harder to read in code than in JSON, especially for non-coders, and because things are smooshed into one file rather than many.  But at least for reading it's still possible to look at the generated JSON files; just makes non-coder PRs possibly a bit trickier.  This does provide a bit of future-proofing when Mojang change the JSON formats between versions, though.


src/datagen/generated/minecolonies/data/minecolonies/tags/items/reduceable_ingredient.json
```diff
@@ -11,8 +11,8 @@
     "#forge:cobblestone",
     "#forge:gravel",
     "#forge:sand",
-    "#forge:dusts/redstone",
-    "#forge:gems/diamond",
+    "#forge:dusts",
+    "#forge:gems",
```
uecasm, line 15: Here's one change -- it seemed reasonable to allow other dusts and gems to be reduced in improved recipes.

src/datagen/generated/minecolonies/data/minecolonies/tags/items/reduceable_ingredient.json
```diff
@@ -29,9 +29,10 @@
     "#minecraft:stone_bricks",
     "minecraft:stick",
     "minecraft:leather",
+    "minecraft:rabbit_hide",
```
uecasm, line 32: And adding rabbit hide along with leather.

src/datagen/generated/minecolonies/data/minecolonies/tags/items/reduceable_product_excluded.json
```diff
@@ -9,11 +9,13 @@
     "#forge:storage_blocks",
     "minecraft:glowstone",
     "minecraft:dried_kelp_block",
+    "minecraft:honey_block",
+    "minecraft:honeycomb_block",
```
uecasm: A couple of other storage blocks that should be excluded from reduction ... although that reminds me that I wanted to abstract this further.

src/datagen/generated/minecolonies/data/minecolonies/tags/items/stonemason_ingredient.json
```diff
@@ -15,7 +15,8 @@
     "minecraft:polished_blackstone_slab",
     "minecraft:polished_blackstone_bricks",
     "minecraft:basalt",
+    "minecraft:polished_basalt",
```
uecasm, line 18: Seemed reasonable.

src/main/java/com/minecolonies/coremod/event/EventHandler.java
```diff
@@ -138,10 +134,9 @@ public static void onCommandsRegister(final RegisterCommandsEvent event)
     @SubscribeEvent
     public static void onEntityAdded(@NotNull final EntityJoinWorldEvent event)
     {
-        ITag<EntityType<?>> mobBlacklist = EntityTypeTags.getAllTags().getTag(new ResourceLocation(Constants.MOD_ID,"mob_attack_blacklist" ));
         if (!event.getWorld().isClientSide())
         {
-            if (MineColonies.getConfig().getServer().mobAttackCitizens.get() && (event.getEntity() instanceof IMob)  && !(mobBlacklist.contains(event.getEntity().getType())))
+            if (MineColonies.getConfig().getServer().mobAttackCitizens.get() && (event.getEntity() instanceof IMob)  && !(ModTags.mobAttackBlacklist.contains(event.getEntity().getType())))
```
uecasm, line 139: I didn't test this, but it seemed better to use a tag object here.

Raycoms (dismissed):

src/datagen/generated/minecolonies/data/minecolonies/tags/items/reduceable_product_excluded.json
uecasm: Now further abstracted.

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultRecipeProvider.java
```diff
@@ -0,0 +1,517 @@
...
+
+        ShapedRecipeBuilder.shaped(ModBlocks.blockHutCrusher)
+                .pattern("XTX")
+                .pattern("CBC")
+                .pattern("XXX")
+                .define('X', ItemTags.PLANKS)
+                .define('C', Items.COBBLESTONE)
+                .define('B', Items.IRON_INGOT)
+                .define('T', buildTool.get())
+                .unlockedBy("has_items", InventoryChangeTrigger.Instance.hasItems(buildTool.get(), Items.IRON_INGOT))
```
someaddons: what are these kinda arbitrary cost unlocked by things for? Do we add an advancement for each recipe?

uecasm: Yes.  That's what unlocks the vanilla recipe book recipes.  You can't actually create a vanilla recipe datagen without supplying one.

PR #8081, opened 2022-02-20
Add copper ore to the default lucky ore list

4wincode:
It's a little more common than iron to make up for the fact that you only get ore, not raw copper.

- Add copper ore to the default lucky ore list with a weight of 48

Review please


uecasm:
> It's a little more common than iron to make up for the fact that you only get ore, not raw copper.

That shouldn't be a consideration; the smelter will turn the ore into the raw metal item as normal (just as if you'd mined it yourself), sometimes with fortune.  As such even were the values equal you would likely end up with a lot more copper than iron.

PR #8119, opened 2022-03-08
Fix Compost Barrel logic to not void excess when extracting compost

DaemonUmbra:
- Fix Compost Barrel logic to not void excess when extracting compost
- Also changed the method of dispensing finished compost:
  - If a side is given (the clicked side of the barrel), finished compost is spawned as an item entity moving in the direction of the indicated side.
  - If null side is given, fixed old logic is invoked inserting results directly into the player's inventory with overflow being spawned on top of the barrel with no velocity.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/tileentities/TileEntityBarrel.java
+++ b/src/main/java/com/minecolonies/coremod/tileentities/TileEntityBarrel.java
@@ -113,20 +113,6 @@ public class TileEntityBarrel extends AbstractTileEntityBarrel implements ITicka
     /**
      * Method called when a player uses the block. Takes the needed items from the player.
      *
-     * @deprecated Use side-sensitive version
-     * @param playerIn  the player
-     * @param itemstack the itemStack on the hand of the player
-     * @return if the barrel took any item
-     */
-    @Deprecated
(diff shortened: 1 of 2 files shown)


marchermans (dismissed):

src/main/java/com/minecolonies/coremod/tileentities/TileEntityBarrel.java
```diff
@@ -106,15 +113,44 @@ private void doBarrelCompostTick(final Level worldIn, final BlockPos pos, final
     /**
      * Method called when a player uses the block. Takes the needed items from the player.
      *
+     * @deprecated Use side-sensitive version
      * @param playerIn  the player
      * @param itemstack the itemStack on the hand of the player
      * @return if the barrel took any item
      */
+    @Deprecated
     public boolean useBarrel(final Player playerIn, final ItemStack itemstack)
```
uecasm: While deprecating the old method when changing parameters like this is in general a good practice, there isn't anywhere else that calls the old method, so it's not really needed.

marchermans: It is also not part of the API so it should indeed be good to go

marchermans (approved):

PR #8122, opened 2022-03-10
Allow worker huts to hire from other jobs

mcmanustfj:
See [old PR](https://github.com/ldtteam/minecolonies/pull/7193) for some context.

This feature will allow huts hiring for "normal" (`WorkerBuildingModuleView`) jobs to hire workers that are already employed, firing them from their current job.  
Special huts (quarry, warehouse, school hiring students) will not be able to do this.

This feature is disabled by default, with a button in the hire window screen to allow it.  This button is disabled for huts that can't use this feature.

This is what the UI looks like when showing employed:
image

## Improvements that need to be made before this is ready
 - Right now,

- [ ]  there's supposed to be a Y or N in the 'show employed' button, but it's not showing up and I don't know why.  Please take a look and advise, because I can't figure this out.
- [x]  The button is still appearing for pupils when it shouldn't be (and lets you hire adults as pupils). (fixed)
- [ ]  I moved the "Turn the hiring mode of this hut to manual to fire this worker" warning above the building hiring mode, which slightly infringes on the worker list view.  I think this is fine in most circumstances (because you wouldn't be doing this on automatic hiring mode) but it's still a slight eyesore.
image

Review please


mcmanustfj:
wrong branch

PR #8123, opened 2022-03-10
 Allow hiring of employed workers

mcmanustfj:
See [old PR](https://github.com/ldtteam/minecolonies/pull/7193) for some context.  In short, this PR makes it so you don't have to fire a citizen before hiring them somewhere else.  This was rejected a year ago with the feedback that I should add a toggle, so I did.

- This feature will allow huts hiring for "normal" (`WorkerBuildingModuleView`) jobs to hire workers that are already employed, firing them from their current job.  Special huts (quarry, warehouse, school hiring students) will not be able to do this.* 
- Workers with jobs are listed after unemployed workers and workers who would normally be able to be hired (i.e. from the library).
- This feature is disabled by default, with a button in the hire window screen to allow it.  This button is disabled for huts that can't use this feature.

This is what the UI looks like when showing employed:
image

## Improvements that should be made before this is ready
 - Right now, there's supposed to be a Y or N in the 'show employed' button, but it's not showing up and I don't know why.  Please take a look and advise, because I can't figure this out.
 - ~~The button is still appearing for pupils when it shouldn't be (and lets you hire adults as pupils).~~ (fixed)
 - I moved the "Turn the hiring mode of this hut to manual to fire this worker" warning above the building hiring mode, which slightly infringes on the worker list view.  I think this is fine in most circumstances (because you wouldn't be doing this on automatic hiring mode) but it's still a slight eyesore.  I don't know how to do this better, so I'd love feedback.
image

Review please


src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -207,45 +221,89 @@ private void doneClicked(@NotNull final Button button)
         final int row = citizenList.getListElementIndexByPane(button);
         @NotNull final ICitizenDataView citizen = citizens.get(row);
```
Raycoms, line 223: Formatting is off. we do:

{

}

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -207,45 +221,89 @@ private void doneClicked(@NotNull final Button button)
...
         final int row = jobList.getListElementIndexByPane(button);
         selectedModule = moduleViews.get(row);
+        setupShowEmployed();
         setupSettings(findPaneOfTypeByID(BUTTON_MODE, Button.class));
         updateCitizens();
         citizenList.refreshElementPanes();
         jobList.refreshElementPanes();
     }
 
+    protected void showEmployedToggled(@NotNull final Button button)
```
Raycoms, line 277: javadoc

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -207,45 +221,89 @@ private void doneClicked(@NotNull final Button button)
...
+    {
+
+        button.setText(showEmployed ? "N" : "Y");
+        showEmployed = !showEmployed;
+
+        onOpened();
+    }
+
+    private void setupShowEmployed()
```
Raycoms, line 291: javadoc

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -207,45 +221,89 @@ private void doneClicked(@NotNull final Button button)
...
+    {
+        Button button = findPaneOfTypeByID(TOGGLE_SHOW_EMPLOYED, Button.class);
+        button.setEnabled(selectedModule instanceof WorkerBuildingModuleView
+                && !(selectedModule instanceof PupilBuildingModuleView));
+        button.setText("N");
+        showEmployed = false;
+    }
+
+    private boolean canAssign(ICitizenDataView citizen)
```
Raycoms, line 305: javadoc

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
```diff
@@ -207,45 +221,89 @@ private void doneClicked(@NotNull final Button button)
         final int row = citizenList.getListElementIndexByPane(button);
         @NotNull final ICitizenDataView citizen = citizens.get(row);
 
+        // Fire citizen if they already have a job
+        if (citizen.getWorkBuilding() != null && selectedModule instanceof WorkerBuildingModuleView) {
+            IBuildingView oldJob =  colony.getBuilding(citizen.getWorkBuilding());
+            oldJob.getModuleViewMatching(IAssignmentModuleView.class,
```
Raycoms, line 228: Does this trigger it on th eserverside too?

Raycoms (changes requested):
Several comments

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
mcmanustfj: It should? It uses the same `.removeCitizen` function that `fireClicked()` does.

mcmanustfj:
Added the researcher XP nerfs as discussed in discord.  Picked 5 out of a hat as an XP number - should be the same XP for the primary stat (which was 0.7) at like 14 int (without housing/building modifiers) but totally removed the absurd mana xp they were getting

src/main/java/com/minecolonies/coremod/client/gui/WindowHireWorker.java
Raycoms: kk

Raycoms (dismissed):

mcmanustfj:
~~whoops was giving 5 xp instead of .5 xp~~
giving 5 xp before int reductions was intentional, but when playtesting it was still way too high - gained like 7 levels in one day with 30 int and a 60% building bonus.  Will bump back up a bit.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/research/EntityAIWorkResearcher.java
```diff
@@ -23,9 +23,9 @@
     public static final int STUDY_DELAY = 60;
 
     /**
-     * XP gained per study position
+     * base XP gained per study position
      */
-    private static final double XP_PER_STUDYPOS = 0.7;
+    private static final double XP_PER_STUDYPOS = 3.125;
```
someaddons: I think XP and studies/day are still off, currently we have a study delay of just 60 ticks(3seconds) + walking time inbetween, which is likely more than just ~3 studies a day you need for getting to the average 10xp base.

A minecraft day is 10 minutes, ~3 of those are spent at getting to or from work, so lets say 7mins effective working time. If they now study for 3s + ~20s walking time you get a total of 18 study actions a day.
Following the baseline of 10Xp a day a worker should get, you'll end up giving them 10/18 = ~0.55 xp per study action.

I think we should do two things here: 
First increase the study delay, 60 ticks is very fast I think they should be looking at a "book" near a bookcase for about 60 seconds instead which would be 60 * 20 ticks.

When we study every 60 seconds and estimate another 20seconds of walking time we get 80 seconds per study, with the estimated minecraft work day of 7minutes that is ~ 5 Study actions per day. 

At 10 Xp Base and 5 study actions a day you'd give it 2 XP per study action.

mcmanustfj: Will do!  I misread the XP calculation, I thought that

`localXp += localXp * (intelligenceLevel / 100.0);`

was actually 

`localXp = localXp * (intelligenceLevel / 100.0);`

(in which case I should have given them 31.25 xp per action).  
I'll do the things you suggested!

someaddons (approved):

PR #8152, opened 2022-03-16
Fixed miner depth

mcmanustfj:
Fixes depth limit
Review please


uecasm (approved):
This is consistent with the original code, though I wonder if it should support level 0 and level 5 better than it currently does...

marchermans (approved):

PR #8165, opened 2022-03-20
fix 8103

### LINKED ISSUE #8103: [BUG] [1.16.5] Assistant cook keeps what he makes

### Minecraft Version

1.16

### MineColonies Version

minecolonies-1.16.5-1.0.462-ALPHA

### Structurize Version

structurize-1.16.5-1.0.343-RELEASE

Complete mod list: https://gist.github.com/Sandriell/0688b4e8e61c46e1bd44c36ad9ff6fc0

### Current Behavior

The assistant cook is very greedy and is not putting what he makes into the inventory of the restaurant and is just keeping it in his own inventory. With a minimum stock set, it results in an endless loop.

I taught the restaurant the recipe for carrot bread from Pam's Harvestcraft and set a minimum stock of one stack. The assistant cook requested all the ingredients, gets them delivered, makes the stack of carrot bread and then just keeps it. He never deposits the finished food into the restaurants inventory so another stack is immediately requested to try and fulfill the minimum stock order.

Tested to see if this was a Harvestcraft issue in particular and it also occurs with beetroot soup.

Additionally, two other related issues I noticed during this is that the assistant cook crafts entire stacks instantly once he has the ingredients. Seems like the should be smacking a crafting table or the hut block for awhile like other crafters do. The second issue is that the assistant cook struggles to recognize that he has the Harvestcraft cooking utensils (such as the bakeware) and still makes a request for them even if they are in his own inventory.

### Expected Behavior

After finishing the request, the food is deposited into the hut's inventory.

### Reproduction Steps

1. Teach the assistant cook a recipe such as beetroot soup.
2. Set a minimum stock for that item.
3. Wait for/supply all the ingredients.
4. Assistant cook makes the request (instantly as pointed out)
5. Immediately requests to make another stock to try and satisfy the minimum stock requirement

### Logs

N/A

### Anything else?

Raycoms:
Closes #8103
Closes #
Closes #

- Make assistant cooks dump everything.
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
@@ -243,6 +243,8 @@ public class BuildingCook extends AbstractBuilding
             return stack.getCount();
         }
 
+        // Make the assistant cook drop everything. We don't want them to keep food.
+        // Neither like the cook does here, nor how the average worker does in the super call.
         if (jobEntry == ModJobs.cookassistant)
         {
             return stack.getCount();
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
@@ -231,7 +231,7 @@ public class BuildingCook extends AbstractBuilding
     }
 
     @Override
-    public int buildingRequiresCertainAmountOfItem(final ItemStack stack, final List<ItemStorage> localAlreadyKept, final boolean inventory)
+    public int buildingRequiresCertainAmountOfItem(final ItemStack stack, final List<ItemStorage> localAlreadyKept, final boolean inventory, final JobEntry jobEntry)
     {
         if (stack.isEmpty())
         {
@@ -242,6 +242,12 @@ public class BuildingCook extends AbstractBuilding
         {
             return stack.getCount();
         }
+
+        if (jobEntry == ModJobs.cookassistant)
+        {
+            return stack.getCount();
+        }
+
         if (isAllowedFood(stack) && (localAlreadyKept.stream().filter(storage -> ISFOOD.test(storage.getItemStack())).mapToInt(ItemStorage::getAmount).sum() < STACKSIZE || !inventory))
         {
             final ItemStorage kept = new ItemStorage(stack);
@@ -266,7 +272,7 @@ public class BuildingCook extends AbstractBuilding
             return 0;
         }
 
-        return super.buildingRequiresCertainAmountOfItem(stack, localAlreadyKept, inventory);
+        return super.buildingRequiresCertainAmountOfItem(stack, localAlreadyKept, inventory, jobEntry);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -242,6 +242,12 @@ public int buildingRequiresCertainAmountOfItem(final ItemStack stack, final List
         {
             return stack.getCount();
         }
+
+        if (jobEntry == ModJobs.cookassistant)
+        {
+            return stack.getCount();
+        }
+
```
marchermans, line 252: Don't hardcode this.
Make an interface for this and check if the job has that interface implemented.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
someaddons: I think the proper way when this stuff is job dependent, not building, is to forward the call to each assigned citizen instead

MotionlessTrain: Or to the affected citizen at least, as this method is called from a context where a citizen is known (that is where jobEntry is coming from)

marchermans: I don't care particularly how this is solved, as long as it is not hardcoded.

someaddons: if it is only going from citizen to building to citizen, then it may aswell just stay within the citizen instead, indeed

Raycoms: I can't really forward this call to the citizen or job. This is the only instance. And the cook building decides how the two cooks behave. This is okay, this is not in a generic high level class.

someaddons: Ofc you could forward the call to the citizens job. 
`This is the only instance` what do you mean with that what exactly is the "only instance"?
Why is it the cook building required to decide this behaviour when apparently you need job-selective behaivour?

Raycoms: I'm not gonna move a shitton of code around into the jobs to fix this little bug that only affects the cook assistant.

Raycoms: Especially because this code is also touched by the courier pickup which does not know the job, then it's null, and then the behaviour is different (which is fine, this is only 1 small corner case for the assistant dump)

Raycoms: I understand not wanting to have this in AbstractBuilding, but this is in the BuildingCook

someaddons: ah thats why your answer didnt make a lot of sense, it was "mimimi" dont wanna change it xD

Raycoms: okay change it yourself.

marchermans (approved):

PR #8180, opened 2022-03-24
Allow Farmers to work sloped Fields

SaintWacko:
- Modify the AI for the Farmer to be capable of working on a slope
- Add a new research project to prevent Farmers from breaking cropland when they land on it (as they tend to when working on a slope)

Review please


src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,63 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
      * @param position the position to check.
      * @param field    the field close to this position.
-     * @return true if should be hoed.
+     * @return position of hoeable surface or null if there is not one
      */
-    private boolean shouldHoe(@NotNull final BlockPos position, @NotNull final ScarecrowTileEntity field)
+    private BlockPos findHoeableSurface(@NotNull BlockPos position, @NotNull final ScarecrowTileEntity field)
     {
+        position = getSurfacePos(position);
+        if ((position == null)
```
Raycoms: Unnecessary brackets around the first null check

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,63 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
+            || field.isNoPartOfField(world, position) 
+            || (world.getBlockState(position.above()).getBlock() instanceof CropsBlock)
+            || (world.getBlockState(position.above()).getBlock() instanceof BlockScarecrow)
+            || !world.getBlockState(position).getBlock().is(Tags.Blocks.DIRT)
+        )
+            {
```
Raycoms: bracket formatting seems a bit off, if you're using IDEA, you should be able to import the formatted rules that are in the repo

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,63 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
+
 
-        return !field.isNoPartOfField(world, position) && !(world.getBlockState(position.above()).getBlock() instanceof CropsBlock)
-                 && !(world.getBlockState(position.above()).getBlock() instanceof BlockScarecrow)
-                 && (world.getBlockState(position).getBlock().is(Tags.Blocks.DIRT) || world.getBlockState(position).getBlock() instanceof GrassBlock);
+    private BlockPos getSurfacePos(final BlockPos position)
+    {
+        return getSurfacePos(position, 0);
+    }
+    /**
```
Raycoms, line 428: space

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,63 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
+                return null;
+            }
+        return position;
+    }
+
 
-        return !field.isNoPartOfField(world, position) && !(world.getBlockState(position.above()).getBlock() instanceof CropsBlock)
-                 && !(world.getBlockState(position.above()).getBlock() instanceof BlockScarecrow)
-                 && (world.getBlockState(position).getBlock().is(Tags.Blocks.DIRT) || world.getBlockState(position).getBlock() instanceof GrassBlock);
+    private BlockPos getSurfacePos(final BlockPos position)
```
Raycoms, line 423: jdoc

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,63 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
+    * Finds the position of the surface near the specified position
+    *
+    * @param position the location to begin the search
+    * @return the position of the surface block or null if it can't be found
+    */
+    private BlockPos getSurfacePos(final BlockPos position, final Integer depth)
+    {
+        if (depth > MAX_DEPTH)
+        {
+            return position.below();
```
Raycoms: shouldn't this be null too?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,63 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
+        }
+        else if (depth < (MAX_DEPTH * -1))
```
Raycoms: why "times -1" ?

Raycoms: and just -MAX_DEPTH

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,63 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
+        {
+            return null;
+        }
+        final BlockState curBlockState = world.getBlockState(position);
+        @Nullable final Block curBlock = curBlockState.getBlock();
+        if (curBlock.is(Tags.Blocks.DIRT) || curBlock.is(Tags.Blocks.STONE) || curBlockState.getMaterial().isLiquid() || curBlock instanceof FarmlandBlock)
```
Raycoms: I thought a bunch about this. We should start at position MAX_DEPTH and then move down to -MAX_DEPTH until we reach a "solid block". If this solid block is not farmland, we can't plant here. (Avoid planting in a little cave opening).

Raycoms (changes requested):
Some thinks require a little bit of changes yet.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
SaintWacko: Actually... The intention here was that once it reached the max height the farmer would level the ground there to work, but you're right, I think it would be better if they just stopped. That would allow farmland to butt up against a cliff, too

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
SaintWacko: It *was* because it had different behaviour if it reached the end searching up than it did when searching down, but now that they're both null I can just take the absolute value and compare it.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
SaintWacko: The problem with that is what if you wanted your farmland in a cave? Or in a stone-ceilinged room? If we started at the top and worked down, it would prevent you from planting if the (stone or dirt) ceiling was lower than five blocks above

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
SaintWacko: Ah, yeah, I was using IDEA for testing, but still doing my editing in VSCode.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
Raycoms: Okay. So essentially we want to move in both direction in the same speed actually.
Like "until reaching MAX_DEPTH do i++
check pos.above(i) and pos.below(i) if it's farmland.

When down from the field, we want to stop once we hit a solid non field block
When going up, we want to stop once we reach an air block.

SaintWacko: I'm not sure I understand. Why would we want to check both directions? We're only looking for one surface block.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
Nightenom: Reach me out (discord) if you want formatter for eclipse/vscode that is close to idea's in repo

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
SaintWacko: To clarify: At the moment it still does check both up and down, but it figures out which direction the surface should be and only goes in that direction, rather than scanning both ways.

Raycoms: We want to check if the farmland block is downwards or upwards. But we probably want to search in both directions in the same speed (as we return the first we find).

Raycoms: The block might be solid but the next farmland block is actually down (because of the cave stuff)

SaintWacko: Yeah, it could be an issue if the land sloped down into a cave where the roof of the cave is at the same level as the block the scarecrow is sitting on. Turning down the MAX_DEPTH would make that much less likely, but either way it's an edge case that I can't see a good solution for. The only thing I could think of would be to track where the surface is, and start the search from the surface of an adjacent block. And that just seems messy.

SaintWacko: Also, I realized you mentioned looking for the farmland block, but this code is also used when the Farmer is determining where to hoe, before any farmland blocks exist.

Raycoms: yeah, "farmland or block that can be turned into farmland" essentially.

SaintWacko: Okay, yeah. So the way the code works, there are some pretty rare situations where it could behave oddly, but I can't think of a way to easily fix it that wouldn't lead to it behaving oddly in more common situations.

SaintWacko: So are you good with this as it stands?

marchermans (dismissed):

src/main/java/com/minecolonies/coremod/entity/ai/citizen/farmer/EntityAIWorkFarmer.java
```diff
@@ -391,14 +397,66 @@ private IAIState canGoPlanting(@NotNull final ScarecrowTileEntity currentField,
...
+    */
+    private BlockPos getSurfacePos(final BlockPos position, final Integer depth)
+    {
+        if (Math.abs(depth) > MAX_DEPTH)
+        {
+            return null;
+        }
+        final BlockState curBlockState = world.getBlockState(position);
+        @Nullable final Block curBlock = curBlockState.getBlock();
+        if (curBlock.is(Tags.Blocks.DIRT) || curBlock.is(Tags.Blocks.STONE) || curBlockState.getMaterial().isLiquid() || ((curBlock instanceof IWaterLoggable) && curBlockState.getValue(BlockStateProperties.WATERLOGGED)) || curBlock instanceof FarmlandBlock)
```
Raycoms: Still, stone tag is not any useful, check for solid material rather then.

SaintWacko: I was trying to emulate the current behavior as closely as possible. Currently, if you have wood or crafted stone on top of the field, the farmer will break it in order to till the field. I'm happy to update it to not do that anymore, though. That would allow farms to butt up against buildings without damaging them, too.

SaintWacko: I'm out right now, but when I get home I'll update it to look for solid blocks, liquids, or waterlogged blocks

Raycoms (approved):
Did you test that in a few different scenarios?

SaintWacko:
Yep, I have a fairly robust testing field setup that I've been using.

> Did you test that in a few different scenarios?

Raycoms (approved):

someaddons (approved):

PR #8183, opened 2022-03-24
Make archery prefer "work" tag to decide standing locations

uecasm:
- Allows glowstone to be used decoratively in tagged buildings
- Falls back on using glowstone block list if no tags

Review please

Also updates the cook to use the new convenience method.  Didn't update the other tag queries because they check for more than one tag per building, so using the convenience method would be a very slight perf regression from redundant recalcs (but probably nobody cares).

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
@@ -194,7 +194,8 @@ public abstract class AbstractBuildingContainer extends AbstractSchematicProvide
     @Nullable
     protected BlockPos getFirstLocationFromTag(@NotNull final String tagName)
     {
-        return getLocationsFromTag(tagName).stream().findFirst().orElse(null);
+        final List<BlockPos> locations = getLocationsFromTag(tagName);
+        return locations.isEmpty() ? null : locations.get(0);
     }
 
     /**
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
@@ -28,10 +28,7 @@ import org.jetbrains.annotations.NotNull;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.function.Predicate;
 
 import static com.minecolonies.api.colony.requestsystem.requestable.deliveryman.AbstractDeliverymanRequestable.getMaxBuildingPriority;
@@ -194,20 +191,25 @@ public abstract class AbstractBuildingContainer extends AbstractSchematicProvide
      * @param tagName the name of the tag to query
      * @return the BlockPos, or null if not found
      */
-    protected BlockPos getFirstLocationFromTag(String tagName)
+    @Nullable
+    protected BlockPos getFirstLocationFromTag(@NotNull final String tagName)
     {
-        if (tileEntity != null && !tileEntity.getPositionedTags().isEmpty())
+        return getLocationsFromTag(tagName).stream().findFirst().orElse(null);
+    }
+
+    /**
+     * Gets the list of tags, and finds all locations registered there.
+     * @param tagName the name of the tag to query
+     * @return all the matching BlockPos, or an empty list if not found
+     */
+    @NotNull
+    protected List<BlockPos> getLocationsFromTag(@NotNull final String tagName)
+    {
+        if (tileEntity != null)
         {
-            for (final Map.Entry<BlockPos, List<String>> entry : tileEntity.getPositionedTags().entrySet())
-            {
-                if(entry.getValue().contains(tagName))
-                {
-                    //Tagged block for a portal location should be one of the obsidian block at the bottom of the opening
-                    return getPosition().offset(entry.getKey());
-                }
-            }
+            return new ArrayList<>(tileEntity.getWorldTagNamePosMap().getOrDefault(tagName, Collections.emptySet()));
         }
-        return null;
+        return Collections.emptyList();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
```diff
@@ -123,19 +122,8 @@ public void initTagPositions()
...
-            for (final Map.Entry<BlockPos, List<String>> entry : te.getWorldTagPosMap().entrySet())
-            {
-                if (entry.getValue().contains(TAG_SITTING))
-                {
-                    sitPositions.add(entry.getKey());
-                }
-            }
-        }
+        sitPositions = getLocationsFromTag(TAG_SITTING);
+        initTags = !sitPositions.isEmpty();
```
uecasm, line 126: This is a slight semantic change, but it didn't seem unreasonable.  Alternative is to just set `initTags = true` unconditionally (which is also slightly different).

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
```diff
@@ -194,20 +191,25 @@ else if (block instanceof BlockMinecoloniesRack)
      * @param tagName the name of the tag to query
      * @return the BlockPos, or null if not found
      */
-    protected BlockPos getFirstLocationFromTag(String tagName)
+    @Nullable
+    protected BlockPos getFirstLocationFromTag(@NotNull final String tagName)
     {
-        if (tileEntity != null && !tileEntity.getPositionedTags().isEmpty())
+        return getLocationsFromTag(tagName).stream().findFirst().orElse(null);
```
Raycoms: stream, findFirst, seems like isEmpty ? null : get(0)

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
Raycoms, line 126: Actually, initTags is always true if the TE was not null

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
```diff
@@ -194,20 +191,25 @@ else if (block instanceof BlockMinecoloniesRack)
...
+    protected List<BlockPos> getLocationsFromTag(@NotNull final String tagName)
+    {
+        if (tileEntity != null)
         {
-            for (final Map.Entry<BlockPos, List<String>> entry : tileEntity.getPositionedTags().entrySet())
-            {
-                if(entry.getValue().contains(tagName))
-                {
-                    //Tagged block for a portal location should be one of the obsidian block at the bottom of the opening
-                    return getPosition().offset(entry.getKey());
```
MotionlessTrain, line 206: It seems like the nether mine used an offset to get the portal from the "portal" tag.
That seems to be gone during the rewrite

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
uecasm: Yes, that's the semantic change I mentioned above.  `te != null` is not visible here any more, so this is the next best thing since `getLocationsFromTag` will return an empty list if the te is null.

Since the te should never be null here anyway unless something is broken, setting unconditionally to true may be reasonable behaviour too.  Unless we're expecting the building to exist before the te does?

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
uecasm: That's converting the relative tag offset to world coords.  `getLocationsFromTag` uses `getWorldTagNamePosMap`, which has already applied that offset.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
uecasm: That would be equivalent, but not a one-liner.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
Raycoms: The TE is null if the hut is not loaded.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
MotionlessTrain: Thanks for clarification. I got confused by the comment

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
Raycoms: Yeah, but performs much better because it doesn't require to create a stream

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
uecasm: The differences in actual behaviour between these options:
* With previous implementation, it will keep re-checking as long as te==null, but afterwards stop even if the list is empty.
* With current implementation, it will keep re-checking the tags each time this is called until at least one tag exists.  This allows the player to add the tag themselves after building the structure, if it was missing for some reason.  But possibly has a slight performance hit to keep re-checking if this is called too often.
* With "just set to true", it will give up immediately if te == null or tags is empty on first call after build/repair/upgrade.

Raycoms: technically, you coudl return null fromt he getLocations method if the TE is null.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuildingContainer.java
uecasm: k

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingCook.java
uecasm: True, and I could change it to that if the distinction is meaningful.  Since it's a convenience method, I wanted to be convenient and not return null though.

Raycoms: It should be mostly fine though. I added most of this code before most of the restaurants were tagged.

someaddons (approved):

PR #8190, opened 2022-03-27
Make build mode setting respect current Structurize setting before researched

uecasm:
- Before researching Builder Modes, uses the current Structurize config interator, not whatever (invisible) setting is saved in the builder hut.

Review please

As before, the current Structurize setting gets saved in the builder hut as of the moment when it is first built, regardless of whether the research is unlocked or not.  (So when the research is unlocked, the iterator may change, if the config was edited after the hut was first built.)

Also as before, work orders save the iterator when building begins, and so will ignore further changes after this unless the order is cancelled and restarted.

This introduces `building.getOptionalSetting`, which should be the preferred way to access building settings (and most, if not all, existing usages should probably switch to it, rather than accessing the module directly and/or using the non-optional variant).  Currently however this PR does not change any existing usages other than to fix this specific issue.

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
@@ -510,9 +510,18 @@ public interface IBuilding extends IBuildingContainer, IRequestResolverProvider,
      * Get setting for key. Utility function.
      * @param key the key.
      * @param <T> the key type.
-     * @return the optional wrapping the value.
+     * @return the setting.
      */
     <T extends ISetting> T getSetting(@NotNull final ISettingKey<T> key);
+
+    /**
+     * Get setting for key. Utility function.
+     * @param key the key.
+     * @param <T> the key type.
+     * @return the optional wrapping the value.
+     */
+    @NotNull
+    <T extends ISetting> Optional<T> getOptionalSetting(@NotNull final ISettingKey<T> key);
 
     /**
      * Check if the assigned citizens are allowed to eat the following stack.
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/modules/settings/BuilderModeSetting.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/modules/settings/BuilderModeSetting.java
@@ -2,7 +2,11 @@ package com.minecolonies.coremod.colony.buildings.modules.settings;
 
 import com.ldtteam.structurize.Structurize;
 import com.ldtteam.structurize.placement.StructureIterators;
+import com.minecolonies.api.colony.buildings.IBuilding;
+import com.minecolonies.api.colony.buildings.modules.ISettingsModule;
 import com.minecolonies.api.colony.buildings.modules.settings.ISettingsModuleView;
+import com.minecolonies.coremod.colony.buildings.workerbuildings.BuildingBuilder;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.List;
 
@@ -34,8 +38,22 @@ public class BuilderModeSetting extends StringSetting
     }
 
     @Override
+    public boolean isActive(final ISettingsModule module)
+    {
+        return module.getBuilding().getColony().getResearchManager().getResearchEffects().getEffectStrength(BUILDER_MODE) > 0;
+    }
+
+    @Override
     public boolean isActive(final ISettingsModuleView module)
     {
         return module.getColony().getResearchManager().getResearchEffects().getEffectStrength(BUILDER_MODE) > 0;
+    }
+
+    @NotNull
+    public static String getActualValue(@NotNull final IBuilding building)
+    {
+        return building.getOptionalSetting(BuildingBuilder.BUILDING_MODE)
+                .map(StringSetting::getValue)
+                .orElse(Structurize.getConfig().getServer().iteratorType.get());
     }
 }
--- a/src/api/java/com/minecolonies/api/colony/buildings/modules/AbstractBuildingModule.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/modules/AbstractBuildingModule.java
@@ -40,6 +40,12 @@ public abstract class AbstractBuildingModule implements IBuildingModule
     }
 
     @Override
+    public IBuilding getBuilding()
+    {
+        return this.building;
+    }
+
+    @Override
     public IBuildingModule setBuilding(final IBuilding building)
     {
         this.building = building;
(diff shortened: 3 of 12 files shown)


src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -510,10 +510,19 @@ else if (getBuildingLevel() <= WOOD_HUT_LEVEL)
...
 
+    /**
+     * Get setting for key. Utility function.
+     * @param key the key.
+     * @param <T> the key type.
+     * @return the optional wrapping the value.
+     */
+    @NotNull
+    <T extends ISetting> Optional<T> getOptionalSetting(@NotNull final ISettingKey<T> key);
+
```
marchermans, line 525: This was proposed to @Raycoms by me, but he hates this kind of functional API.
Not sure this as such will get accepted ....

src/main/java/com/minecolonies/coremod/colony/buildings/modules/settings/BuilderModeSetting.java
```diff
@@ -33,9 +37,23 @@ public BuilderModeSetting(final List<String> value, final int curr)
...
+    public boolean isActive(final ISettingsModule module)
+    {
+        return module.getBuilding().getColony().getResearchManager().getResearchEffects().getEffectStrength(BUILDER_MODE) > 0;
+    }
+
     @Override
     public boolean isActive(final ISettingsModuleView module)
     {
         return module.getColony().getResearchManager().getResearchEffects().getEffectStrength(BUILDER_MODE) > 0;
     }
```
marchermans, line 50: Reduce the duplicate code here, neither methods rely on the passed in module or view, so redirect both please to an internal private method.

marchermans (changes requested):

src/main/java/com/minecolonies/coremod/colony/buildings/modules/settings/BuilderModeSetting.java
uecasm: While the code looks similar, the types in the method chain are not the same.  It would only be possible to move `.getResearchEffects()` and on to a separate method, which didn't seem sufficiently worthwhile to me.  But I can change that if preferred.

marchermans: Nah leave it then..
I hate our view concept....

I checked it out a while ago, we don't really save that much on network bandwidth and it brings us double the hassle and code to maintain, but @Raycoms wants to keep it. So who am I to judge.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
uecasm: It's not appropriate in all contexts, but using `Optional`'s `ifPresent`/`map`/`flatMap`/`orElse` interface seems safer to me than surprise nulls, or using a setting that the user can't actually edit (so might not be appropriate)...

There's a bunch of places that have a defensive `building.hasModule(ISettingsModule.class) && building.getSetting(xxx).getValue()`, which is a leaked abstraction (and _still_ potentially vulnerable to nulls).  Would be better (and safer) with `building.getOptionalSetting(xxx).map(StringSetting::getValue).orElse("default")` or similar.

Raycoms: I'm just against "getOptionalSetting" in all context. In limited contexts I'm fine with it, it has it's uses, but I don't want to have to rely on it 100%

src/main/java/com/minecolonies/coremod/colony/buildings/modules/settings/BuilderModeSetting.java
Raycoms: It's simply because it isn't possible to serialize everything to both sides =D

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
uecasm: Well, it should probably be used in all contexts where the setting might not exist.  So, any setting with an `isActive` check, or any case where you have a generic building that may or may not have a specific setting.

It's less important in other cases, where you're fairly certain you can't get a null.  But it seems simpler to have a uniform way to query settings even in those cases.

Raycoms: i.e. in some cases we do rely on getOptionalModule as well

Raycoms: I think we should make this not a getOptionalSetting, but directly a getSettingOrDefault, so we don't require to unpack that on the other side always

uecasm: Well, that could be useful too (particularly in this case), but then you couldn't call other methods on `ISetting`, such as `.ifPresent(setting -> setting.set(foo))`

uecasm: Hrm.  I don't think Java provides any good way to actually write a generic `getSettingOrDefault`.  `ISetting` doesn't provide `getValue`, for a start, nor is there any simple way to extract the value type from the `ISettingKey`.  This probably isn't impossible, but it would require a pretty hefty rewrite.

Raycoms: Ah, darn, then :shipit:

Raycoms (approved):

src/main/java/com/minecolonies/coremod/colony/buildings/modules/settings/BuilderModeSetting.java
marchermans: What?
Yes it is -> NBT We already store the entire state in it, write the entire colony to NBT dump it in a packet and send it...... Done.

someaddons: The question is how often is that send, if we send a 1mb colony every tick then it'll require 20mb/sec

Raycoms: Not all state is persisted too, there is a lot of temporary state too, without views we don't know which state we can use and which not easily.

marchermans: Regardless of something being temporary or not, we could still use the same implementations / interfaces everywhere...

marchermans: > The question is how often is that send, if we send a 1mb colony every tick then it'll require 20mb/sec

I agree that it should not be a single packet doing the update, but we could write a set of packets which sync the managers individually or when dirty etc.
I was just making a point about the fact that we technically have all data.

My point is not the sync, my point is the dual interfaces and classes for them.

someaddons (approved):

PR #8212, opened 2022-04-01
Plantation can select one and multiple crops when "Double Trouble" is researched

### LINKED ISSUE #494: Plantation should be allowed to grow one crop despite the research "Double trouble"
### Prerequisites 

- [ ] I am running the latest **alpha** version of MineColonies for my Minecraft version.

### Context 

- Minecraft Version: 1.16.5
- MineColonies Version: minecolonies-0.14.293-ALPHA-universal.jar

After researching "Double trouble" you are allowed to grow 2 crops in a plantation at once, however the options to grow only 1 crop are removed for whatever reason, this is unlike the beekeeper which still retains the option to either collect honey, beecombs or both.

After having researched "Double trouble" still retain the option to only choose one crop to plant

### Notes or related things
n/a

---
## Viewers

Thodor12:
Closes ldtteam/minecolonies-features#494

- Previously when "Double Trouble" got researched the options to only plant a single item became unavailable, this has been fixed
- Planter will attempt to clear out blocks which do not belong on the sand patches
- Planter will continue to farm plants he no longer has selected in the plant options, this is to prevent letting these plants go to waste (note that if a plant is currently not in the planting cycle, the planter will remove it despite being 3 blocks tall, he'll immediately clear out the plants, this also prevents players from manually planting plants)

Review please


src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingPlantation.java
```diff
@@ -43,7 +47,20 @@
...
      */
-    public static final ISettingKey<PlantationSetting> MODE = new SettingKey<>(PlantationSetting.class, new ResourceLocation(com.minecolonies.api.util.constant.Constants.MOD_ID, "mode"));
+    public static final ISettingKey<PlantationSetting> MODE =
+      new SettingKey<>(PlantationSetting.class, new ResourceLocation(com.minecolonies.api.util.constant.Constants.MOD_ID, "mode"));
+
+    /**
+     * The combinations of items/blocks/tags.
+     */
+    public static final Map<Item, ItemCombination> COMBINATIONS = Collections.unmodifiableMap(new HashMap<Item, ItemCombination>()
+    {
```
Raycoms: does this need the double parenthesis?

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingPlantation.java
```diff
@@ -214,16 +281,125 @@ public CraftingModule(final JobEntry jobEntry)
...
+         *
+         * @return the item combination data.
+         */
+        public ItemCombination getCombination()
+        {
+            return combination;
+        }
+    }
+
+    public static class ItemCombination
```
Raycoms: Confusing name imo

Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/entity/ai/citizen/planter/EntityAIWorkPlanter.java
```diff
@@ -106,23 +142,29 @@ private IAIState farm()
             return START_WORKING;
         }
 
-        if (walkToBlock(workPos.above()))
+        if (isItemPositionAir(workPos))
+        {
+            return START_WORKING;
+        }
+
+        if (walkToBlock(workPos.getPosition().above()))
```
someaddons: I think we should rename the workPos, since its always the ground block and we keep having to do above() its kinda confusing. Maybe plantableSoilPos? or so

src/main/java/com/minecolonies/coremod/entity/ai/citizen/planter/EntityAIWorkPlanter.java
```diff
@@ -164,38 +206,51 @@ protected IAIState decide()
...
         final BuildingPlantation plantation = getOwnBuilding();
 
-        final List<BlockPos> list = plantation.getPosForPhase();
-        for (final BlockPos pos : list)
+        final Item currentItem = plantation.getNextPhase();
+        final List<Item> availablePlants = plantation.getAvailablePlants();
+        final List<BuildingPlantation.PlantationItemPosition> list = plantation.getAllWorkPositions();
+
+        // Iterate all positions to check if there's a valid block or if there's anything to harvest
+        for (final BuildingPlantation.PlantationItemPosition plantationItemPosition : list)
```
someaddons: while we're on reworking a bit, I know this is additional work but it probably would be nice if the worker would actively walk and "inspect" the fields. This gives a much more lively/active impression to the players watching. Would probably be another new AI State "INSPECT" or so

Thodor12: This would probably be the lowest priority action right? So that state would occur all the way at the end instead of `START_WORKING` state?
Also it'd probably be best if this selects a random location of the allowed plants right?
How would I prevent him constantly walking around but giving it some delay?

Also how would this work alongside the new leisure system and inactive workers?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/planter/EntityAIWorkPlanter.java
```diff
@@ -217,16 +272,44 @@ private void requestPlantable(final Item current)
...
-        return !(world.getBlockState(pos.above(1)).getBlock() instanceof AirBlock)
-                 && !(world.getBlockState(pos.above(2)).getBlock() instanceof AirBlock)
-                 && !(world.getBlockState(pos.above(3)).getBlock() instanceof AirBlock);
+        BlockPos position = plantationItemPosition.getPosition();
+        Block firstBlock = world.getBlockState(position.above(1)).getBlock();
+        if (plantationItemPosition.getCombination().getBlock() != firstBlock)
+        {
+            return false;
+        }
+        return firstBlock == world.getBlockState(position.above(2)).getBlock() && firstBlock == world.getBlockState(position.above(3)).getBlock();
```
someaddons: How about you move the height dependence into the PlantableBlockItem class thing aswell, so that one also sets the max height

src/main/java/com/minecolonies/coremod/entity/ai/citizen/planter/EntityAIWorkPlanter.java
someaddons: I think you could have a % chance to go into the INSPECT state, which would solve the doing it constantly part. 
The locations could be in-order if we sort the list beforehand. Maybe sorting by blockpos and then having the "walk to pos" have a bigger distance, like 5 blocks so he  seems to check out patches/areas at once. Probably want to save a current list index then.

Thodor12: What if I randomly select any position of any plant in the current planting cycle and move to that one?
A minimum distance walk check could potentially have problems in certain schematics.

Also the % chance is a good idea, I'll add that

someaddons: I'd do it two-fold, you move to the blockpos in question with a bigger minimum distance of e.g. 5 to "look" if something needs to harvested or w/e, and then if you see sth that needs to then you go closer and do Actions at the block.

someaddons (approved):

PR #8227, opened 2022-04-06
Resource scroll displays warehouse contents

Thodor12:
No issue present, discussed with Ray over Discord

- When shift right-clicking on a warehouse you can create a "snapshot"
- As long as there is a snapshot loaded on the resource scroll it'll display amounts (just like with deliveries) next to the requested items


src/main/java/com/minecolonies/coremod/client/gui/WindowResourceList.java
```diff
@@ -54,15 +71,45 @@
...
+     * The position of the warehouse clicked on.
+     */
+    @Nullable
+    private final BlockPos warehousePos;
+
+    /**
+     * The compound data to store the warehouse snapshot to.
+     */
+    @Nullable
+    private final CompoundNBT compound;
```
Raycoms, line 84: Why do we need this field?

src/main/java/com/minecolonies/coremod/client/gui/WindowResourceList.java
```diff
@@ -77,6 +124,43 @@ public WindowResourceList(final int colonyId, final BlockPos buildingPos)
...
+    private void pullResourcesFromWarehouse()
+    {
+        if (warehousePos == null)
+        {
+            return;
+        }
+
+        Map<Item, BuildingBuilderResource> resourceMap = resources.stream().collect(Collectors.toMap(ItemStorage::getItem, v -> v));
+
+        warehouseSnapshot = new HashMap<>();
```
Raycoms: can't we always just initiate this with an empty map?

Raycoms (changes requested):
Some small questions

src/main/java/com/minecolonies/coremod/client/gui/WindowResourceList.java
Thodor12: In the constructor it attempts to load the snapshot if any NBT data is present, else it's an empty hashmap, however this resets the snapshot in order to clear out a potential existing snapshot to then proceed to populate it with new data.

src/main/java/com/minecolonies/coremod/client/gui/WindowResourceList.java
Thodor12: The snapshot is calculated inside of this class, so we have to pass along the compound in order to save the snapshot.
It was either passing along the compound or the resource scroll itself.

Raycoms: yeah, we pass it along, but do we need it as a field. couldn't see where we needed the field where we didn't already pass it as a param

src/main/java/com/minecolonies/coremod/client/gui/WindowResourceList.java
Raycoms: ah kk

src/main/java/com/minecolonies/coremod/client/gui/WindowResourceList.java
Thodor12: I don't think so as this saving is not directly called through the constructor.
It first goes through `onOpened` which then proceeds to call `pullResourcesFromWarehouse`, that is where the saving happens.

Raycoms: I double checked the code, the compound is only used in saving, but do we need that data there at all? Shouldn't we write completely fresh data anyway?

Raycoms: technically, we could give the message only the warehouse snapshot data, colony id and building id (which are needed extra anyway to identify)

Raycoms (dismissed):

someaddons (dismissed):

marchermans (dismissed):

Raycoms (approved):

marchermans (approved):

PR #8241, opened 2022-04-08
Add citizen direct item interactions

someaddons:
Added citizen interactions with certain items, e.g. right click food to feed or golden apple to cure sickness, using a book on a child gives it some intelligence xp.
Citizens now behave more responsive to players when they bump into one.
Position searches, including spawn point ones, now also consider the initial pos.
Improved standing up from sitting, the standing pos now depends on the originally targeted block for sitting.
Child pupils no longer teleport from far to their sitting position.
Fix guards doing wandering movement when they should not
Citizen levelup now plays the levelup sound.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/jobs/IJob.java
+++ b/src/api/java/com/minecolonies/api/colony/jobs/IJob.java
@@ -5,20 +5,22 @@ import com.minecolonies.api.colony.IColony;
 import com.minecolonies.api.colony.jobs.registry.JobEntry;
 import com.minecolonies.api.colony.requestsystem.token.IToken;
 import com.minecolonies.api.entity.citizen.AbstractEntityCitizen;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.network.FriendlyByteBuf;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.damagesource.DamageSource;
 import net.minecraft.world.entity.ai.goal.Goal;
 import net.minecraft.world.entity.ai.goal.GoalSelector;
 import net.minecraft.world.item.ItemStack;
-import net.minecraft.nbt.CompoundTag;
-import net.minecraft.network.FriendlyByteBuf;
-import net.minecraft.world.damagesource.DamageSource;
 import net.minecraftforge.common.util.INBTSerializable;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.Set;
 
 import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_COMPLAINS_DAYS;
-import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_DEMANDS_DAYS;;
+import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_DEMANDS_DAYS;
+
+;
 
 public interface IJob<AI extends Goal> extends INBTSerializable<CompoundTag>
 {
@@ -274,7 +276,18 @@ public interface IJob<AI extends Goal> extends INBTSerializable<CompoundTag>
 
     /**
      * Set the registry entry of the job.
+     *
      * @param jobEntry the job entry belonging to it.
      */
     void setRegistryEntry(JobEntry jobEntry);
+
+    /**
+     * Whether the job is a guard
+     *
+     * @return
+     */
+    default boolean isGuard()
+    {
+        return false;
+    }
 }
--- a/src/api/java/com/minecolonies/api/colony/jobs/IJob.java
+++ b/src/api/java/com/minecolonies/api/colony/jobs/IJob.java
@@ -20,8 +20,6 @@ import java.util.Set;
 import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_COMPLAINS_DAYS;
 import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_DEMANDS_DAYS;
 
-;
-
 public interface IJob<AI extends Goal> extends INBTSerializable<CompoundTag>
 {
     /**
--- a/src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
+++ b/src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
@@ -6,7 +6,6 @@ import com.minecolonies.api.client.render.modeltype.registry.IModelTypeRegistry;
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.ICitizenDataView;
 import com.minecolonies.api.colony.jobs.IJob;
-import com.minecolonies.api.colony.jobs.ModJobs;
 import com.minecolonies.api.colony.requestsystem.location.ILocation;
 import com.minecolonies.api.entity.MinecoloniesMinecart;
 import com.minecolonies.api.entity.ai.DesiredActivity;
@@ -368,8 +367,8 @@ public abstract class AbstractEntityCitizen extends AbstractCivilianEntity imple
     @Override
     public void onPlayerCollide(final Player player)
     {
-        final IJob job = getCitizenData().getJob();
-        if (job == null || job.getJobRegistryEntry() != ModJobs.archer && job.getJobRegistryEntry() != ModJobs.knight && job.getJobRegistryEntry() != ModJobs.druid)
+        final IJob<?> job = getCitizenData().getJob();
+        if (job == null || !job.isGuard())
         {
             super.onPlayerCollide(player);
         }
--- a/src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
+++ b/src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
@@ -5,6 +5,8 @@ import com.minecolonies.api.client.render.modeltype.ModModelTypes;
 import com.minecolonies.api.client.render.modeltype.registry.IModelTypeRegistry;
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.ICitizenDataView;
+import com.minecolonies.api.colony.jobs.IJob;
+import com.minecolonies.api.colony.jobs.ModJobs;
 import com.minecolonies.api.colony.requestsystem.location.ILocation;
 import com.minecolonies.api.entity.MinecoloniesMinecart;
 import com.minecolonies.api.entity.ai.DesiredActivity;
@@ -17,7 +19,6 @@ import com.minecolonies.api.inventory.InventoryCitizen;
 import com.minecolonies.api.sounds.EventType;
 import com.minecolonies.api.util.CompatibilityUtils;
 import com.minecolonies.api.util.ItemStackUtils;
-import com.minecolonies.api.util.Log;
 import com.minecolonies.api.util.SoundUtils;
 import com.mojang.datafixers.util.Pair;
 import net.minecraft.core.BlockPos;
@@ -134,14 +135,15 @@ public abstract class AbstractEntityCitizen extends AbstractCivilianEntity imple
 
     /**
      * Get the default attributes with their values.
+     *
      * @return the attribute modifier map.
      */
     public static AttributeSupplier.Builder getDefaultAttributes()
     {
         return LivingEntity.createLivingAttributes()
-                 .add(Attributes.MAX_HEALTH, BASE_MAX_HEALTH)
-                 .add(Attributes.MOVEMENT_SPEED, BASE_MOVEMENT_SPEED)
-                 .add(Attributes.FOLLOW_RANGE, BASE_PATHFINDING_RANGE);
+          .add(Attributes.MAX_HEALTH, BASE_MAX_HEALTH)
+          .add(Attributes.MOVEMENT_SPEED, BASE_MOVEMENT_SPEED)
+          .add(Attributes.FOLLOW_RANGE, BASE_PATHFINDING_RANGE);
     }
 
     public GoalSelector getTasks()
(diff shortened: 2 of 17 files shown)


Thodor12:
I see you implemented golden apples, books, cacti and glowstone dust.
Maybe milk is a good idea too so you can clear any effects from a citizen aswell?

someaddons:
hm effects are not really used by us at the moment for citizens

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -399,6 +414,195 @@ public InteractionResult checkAndHandleImportantInteractions(final Player player
...
+                    ChatFormatting.RED)),
+                  player.getUUID());
+            }
+            return null;
+        }
+
+        if (usedStack.getItem() == Items.GOLDEN_APPLE && getCitizenDiseaseHandler().isSick())
+        {
+            usedStack.setCount(usedStack.getCount() - 1);
+            player.setItemInHand(hand, usedStack);
```
Raycoms, line 443: where is the healing chance here?

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -399,6 +414,195 @@ public InteractionResult checkAndHandleImportantInteractions(final Player player
...
+
+        if (usedStack.getItem() == Items.CACTUS)
+        {
+            usedStack.setCount(usedStack.getCount() - 1);
+            player.setItemInHand(hand, usedStack);
+
+            if (!level.isClientSide())
+            {
+                player.sendMessage(new TranslatableComponent("com.minecolonies.coremod.interaction.ouch", getCitizenData().getName()), player.getUUID());
+                getNavigation().moveAwayFromLivingEntity(player, 5, 1);
```
Raycoms, line 501: disappointed, no jump up (move up)

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -399,6 +414,195 @@ public InteractionResult checkAndHandleImportantInteractions(final Player player
...
+    /**
+     * Interaction with children for offering food
+     *
+     * @param usedStack
+     * @param player
+     * @param hand
+     */
+    private void childFoodInteraction(final ItemStack usedStack, final Player player, final InteractionHand hand)
+    {
+        if (usedStack.getDisplayName().getString().toLowerCase(Locale.ENGLISH).contains("cookie"))
```
Raycoms, line 547: for mod support generic? maybe we make a sweets tag and put chocolate in too?

someaddons: no by design they're hard set on some form of cookie! If you name an item cookie thats ok too

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -399,6 +414,195 @@ public InteractionResult checkAndHandleImportantInteractions(final Player player
...
+        {
+            usedStack.setCount(usedStack.getCount() - 1);
+            player.setItemInHand(hand, usedStack);
+            interactionCooldown = 100;
+
+            if (!level.isClientSide())
+            {
+                playSound(SoundEvents.GENERIC_EAT, 1.5f, (float) SoundUtils.getRandomPitch(getRandom()));
```
MotionlessTrain: Should the child actually gain anything from the cookie, e.g. saturation?

someaddons: oh true I didnt add a gain yet

src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
```diff
@@ -362,6 +365,21 @@ public void push(@NotNull final Entity entityIn)
         super.push(entityIn);
     }
 
+    @Override
+    public void onPlayerCollide(final Player player)
+    {
+        final IJob job = getCitizenData().getJob();
+        if (job == null || job.getJobRegistryEntry() != ModJobs.archer && job.getJobRegistryEntry() != ModJobs.knight && job.getJobRegistryEntry() != ModJobs.druid)
```
uecasm: Perhaps a good time for `job.isGuard()` or sth?  Elsewhere an instanceof Guard check is used.

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -399,6 +414,209 @@ public InteractionResult checkAndHandleImportantInteractions(final Player player
...
+            {
+                playSound(SoundEvents.VILLAGER_NO, 0.5f, (float) SoundUtils.getRandomPitch(getRandom()));
+                player.sendMessage(new TranslatableComponent("com.minecolonies.coremod.interaction.notnow", this.getCitizenData().getName()).setStyle(Style.EMPTY.withColor(
+                    ChatFormatting.RED)),
+                  player.getUUID());
+            }
+            return null;
+        }
+
+        if (usedStack.getItem() == Items.GOLDEN_APPLE && getCitizenDiseaseHandler().isSick())
```
uecasm, line 440: Why is this specifically a golden apple, when that's only half of the items needed to cure one disease (that might not be the one they have)?  Shouldn't it just check if the item is one needed for their current disease's cure, then just add it to inventory (with no cooldown) and use the standard code for curing once they have all of them?

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -399,6 +414,209 @@ public InteractionResult checkAndHandleImportantInteractions(final Player player
...
+            else
+            {
+                eatFoodInteraction(usedStack, player, hand);
+            }
+            return InteractionResult.CONSUME;
+        }
+
+        if (usedStack.getItem() == Items.BOOK && isBaby())
+        {
+            usedStack.setCount(usedStack.getCount() - 1);
```
uecasm: All of these will actually consume the item fully in creative, which is not how creative is usually supposed to work.  I'm away from my PC right now so I can't check the actual method name, but I'm pretty sure there's a `decrement()` method or sth that handles this without emptying the slot in creative.

src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenSkillHandler.java
```diff
@@ -180,7 +177,7 @@ public void addXpToSkill(final Skill skill, final double xp, final ICitizenData
 
         final IBuilding home = data.getHomeBuilding();
 
-        final double citizenHutLevel = home == null ? 0 : home.getBuildingLevel();
+        final double citizenHutLevel = home == null ? 1 : home.getBuildingLevel();
```
uecasm: Why treat homeless people as if they have a level 1 house?

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
Raycoms: this has a cooldown, and is supposed to be a small chance

Raycoms: I'd rather make this a more of 1/5th though

uecasm: Yeah but if one of the diseases is bad enough to require a golden apple *and* something else, why can just the apple cure it?  Why would someone try using the apple when the citizen is specifically asking for different items?  Why, since most of those are food items, should the citizen just eat them if given instead of holding on to them to cure?

Perhaps this could use the enchanted golden apple instead, or a healing potion of some sort (then perhaps a high chance of success is ok), but I still think it should also accept the specific healing items too.

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
someaddons: it doesnt consume it in creative, tested with it

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
Raycoms: Atm there is already a small chance of a "miracle healing" in general. The idea was that independent of the disease there is a small chance that this helps enough (kinda realistic compared to reality).

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
uecasm: Ok, I guess that's handled elsewhere then.  But still, why not just use `usedStack.shrink(1);` instead?

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
uecasm: Hmm.  Ok, I see why not "enchanted golden apple" now.  I remember those having a crafting recipe, but apparently that was a ~modpack~ 1.7 thing and they don't by default.

I still think it should use a potion rather than golden apple though, and regardless should also detect and accept the proper cure items.  Someone's just going to report it as a bug later otherwise.

MotionlessTrain: I can't remember how often I have right clicked a citizen right after I harvested some potatoes and carrots, and still have either of those in my hand.
It would be inconvenient at the very least if the citizen decided to eat the potato or carrot upon interacting, where I actually meant to open their interface to give both items to them.

src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
someaddons: ye if we need this again I'll generalise it

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
someaddons: I'll up the chance to just 1 in 3, but the important distinction here is discoverability. There is and should not be any info about this future spread around publically, it is partly a reward of discovering the interaction possibility. Cost wise we're fine with 1.5Apples per cure on avg.
Some annoyance with interact-feeding is expected/factored in, but I'll turn it off for sick people, those realistically aren't that interactive

src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenSkillHandler.java
someaddons: I've raised the base here, because before homeless people never could gain any XP. However now that there is a chance to give them some I think having 1 be the lowest makes sense

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
someaddons: fixed^^

src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenSkillHandler.java
uecasm: Then why not just +1 the level (and maxlevel), such that people in a L1 house do still get a skillcap boost over L0?  Pushing L4 to a cap of 50 (from 40) doesn't seem like a problematic change, there's still lots of headroom before L5.

On an unrelated note (existing prior to this PR), it looks like a citizen living in an L3 tavern will have no skill cap, just like one living in an L5 house.  Is that intended?

src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
uecasm: Again though, it's already checked as `job instanceof AbstractJobGuard` elsewhere, so this ought to be consistent.  And that's more future-proof/addon-proof for new guard types.

someaddons: abstractjobguard is not in the api

MotionlessTrain: Though IJob could have an isGuard method, which is overridden in AbstractJobGuard to return true instead of false. I don't think that would cause problems with that class not being in the API

Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/entity/citizen/citizenhandlers/CitizenSkillHandler.java
someaddons: ye added +1, disallowed all buildings below max 5 from going up higher

src/api/java/com/minecolonies/api/colony/jobs/IJob.java
```diff
@@ -5,20 +5,22 @@
...
 import net.minecraftforge.common.util.INBTSerializable;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.Set;
 
 import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_COMPLAINS_DAYS;
-import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_DEMANDS_DAYS;;
+import static com.minecolonies.api.util.constant.HappinessConstants.IDLE_AT_JOB_DEMANDS_DAYS;
+
+;
```
MotionlessTrain: Interesting lonely semicolon

PR #8282, opened 2022-04-21
Feature/domum paper

uecasm:
- `runData` added carpets to spawn blocks, apparently.
- Moves crafting of paper extra from stonemason to plantation.
    - Doesn't make the plantation responsible for DO-crafting paper wall or other DO blocks that include paper extra (since it's not currently a DO crafter), but that's a thing that could be done if we want to.
- Ports #8288 
- Makes DO crafting use a new crafting type.
    - There should be no actual change in behaviour from this.
- Adds recipe auditing for DO outputs and inputs.
- Since it shares code with the above, JEI will now also hint which blocks are acceptable for DO crafting at each crafter.

Review please

See [the spreadsheet](https://docs.google.com/spreadsheets/d/1fPdA5BDmnW3r2BXfRik8bmVt0nxwI3tOrrpc1vYXP_I/edit?usp=sharing):
- The "After" tab is before this PR
- The "WithDomum" tab is after this PR
- The "DomumInput" tab is a new visualisation of the acceptable input blocks.
    - For each output block type, it shows the possible input "skin" blocks and which crafter will accept them.
    - As before, only one of the input blocks needs to be acceptable to the crafter; the other blocks can be any valid skin, crafter-acceptable or not.

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/buildings/modules/ICraftingBuildingModule.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/modules/ICraftingBuildingModule.java
@@ -9,9 +9,9 @@ import com.minecolonies.api.crafting.IRecipeStorage;
 import com.minecolonies.api.entity.citizen.AbstractEntityCitizen;
 import com.minecolonies.api.crafting.registry.CraftingType;
 import com.minecolonies.api.util.OptionalPredicate;
-
-import net.minecraft.world.item.ItemStack;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.crafting.RecipeType;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
--- a/src/api/java/com/minecolonies/api/colony/buildings/modules/ICraftingBuildingModule.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/modules/ICraftingBuildingModule.java
@@ -7,6 +7,7 @@ import com.minecolonies.api.colony.requestsystem.token.IToken;
 import com.minecolonies.api.crafting.IGenericRecipe;
 import com.minecolonies.api.crafting.IRecipeStorage;
 import com.minecolonies.api.entity.citizen.AbstractEntityCitizen;
+import com.minecolonies.api.crafting.registry.CraftingType;
 import com.minecolonies.api.util.OptionalPredicate;
 
 import net.minecraft.world.item.ItemStack;
@@ -15,6 +16,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
+import java.util.Set;
 import java.util.function.Predicate;
 
 /**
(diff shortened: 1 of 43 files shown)


src/api/java/com/minecolonies/api/colony/buildings/modules/ICraftingBuildingModule.java
```diff
@@ -80,20 +81,23 @@ public interface ICraftingBuildingModule extends IBuildingModule
...
 
     /**
      * Check if this building type can learn (or otherwise process)
-     * vanilla smelting recipes of some kind.
+     * vanilla crafting recipes of some kind.
      *
      * @return True if so; false otherwise.
      */
-    boolean canLearnFurnaceRecipes();
+    default boolean canLearnRecipeType(@NotNull RecipeType<?> type)
```
Raycoms: that's pretty similar to what I did with the alchemist, but we need a custom type to include brewing.

uecasm: Well, that's annoying.  Why wouldn't minecraft/forge use their standard crafting recipe system for potion crafting?

Raycoms: Because they're consistently inconsistent

uecasm: Hmm.  Reading up on potions it looks like they're not implemented in a datapack-friendly format at all.  So quite possibly you should be using an entirely separate kind of module for them, because whatever they are, it's not crafting.

Raycoms: Nah, it fits very will in there.

uecasm: Welp, now I've seen the alchemist code, and this is definitely the better way to do it.  I could perhaps add a brewing recipe type to this to make it easier to port your PR to use this method instead?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/SimpleCraftingModule.java
```diff
@@ -39,6 +39,12 @@ public List<IRequestResolver<?>> createResolvers()
...
     }
 
+    @Override
+    public boolean canLearnManyRecipes()
+    {
+        return false;
+    }
+
     @Override
     public boolean canLearnLargeRecipes()
```
Raycoms: but you kept the large recipe type here as a separate method, the idea was to get rid of this extra thing, which is "only good for distinguishing 2x2 vs 3x3 crafting"

uecasm: Getting rid of the extra thing is not a useful goal, especially when that conflates two unrelated things.

Raycoms: There are different things, there are MODULE_CRAFTING_TYPES which includes LARGE, SMALL, SMELTING, BREWING (everything that we consider crafting to some sorts) and there is the Mojang RECIPE_TYPE which are two different things imo.

uecasm: Oh, I see the confusion.  You're coming at this from the UI side top-down, and 2x2 and 3x3 crafting have different UIs.  I'm coming at it from the recipe side, bottom-up, and they have the same recipe type.  So you see them as different and I see them as the same.

uecasm: But regardless, I think basing it on RecipeType is better than inventing our own type, since it allows for extension to new crafting types by addons or future features without modifying this type (OCP).  And that sees 2x2 and 3x3 crafting as the same, so this should too.

If you really want to get rid of the large crafting check, it could be replaced with an instanceof, since there's only one type that ever sets it to false.  But I don't really think it needs to be removed at all.  Or potentially it could be pulled higher into the `Crafting` policy class rather than the base interface, since it's not relevant to anything else.  But that makes it less of just a thin policy class, which was the intent.

Raycoms: Yes, in the end, our workers don't really rely on or care about the vanilla recipe system at all (outside from the UI side of things) that's why I wanted to avoid using the vanilla RecipeType (I was about to use it as well but then decided otherwise).

I do understand the concern of extendability though. Instead of enum we can make this an object with some pre-defined static types as well. This way anyone could create a new static object for their type when required.

I just want to avoid mixing it up with the Recipe Type because our workers don't mess with vanilla recipe handling outside of the teacvhing

uecasm: Well, it's also extremely convenient for JEI and recipe auditing that this does use RecipeType.  Though that doesn't preclude using our own type provided that it's convertable to RecipeType.... but then I don't really see the point of not using RecipeType, other than the annoyance of brewing not already being one.

src/api/java/com/minecolonies/api/colony/buildings/modules/ICraftingBuildingModule.java
uecasm: Though I wasn't originally planning to backport this to 1.16 since much of it is DO-focused, but I guess I should if you're going to want to build alchemist things on top of it.

uecasm:
Just tagging this as a draft until it gets rebased on the crafting types update

uecasm:
Now rebased on crafting types.

Raycoms (approved):

someaddons (approved):

PR #8286, opened 2022-04-22
Feature/alchemist

Raycoms:
Closes #8324
Closes #
Closes #

# Changes proposed in this pull request:
- Adjusts slightly the recipe checks and creates a proper minecolonies recipe type
- Add alchemist that brews potions (potion teaching window)
- Alchemist also creates magic potion for druid
- Druid now uses magic potions instead of mistletoe + water bottle
- Alchemist gathers mistletoes + nether wart

Review please

PR #8288, opened 2022-04-23
Refactor crafting modules to use crafting types

uecasm:
- Adds a registry for CraftingTypes (to allow future expansion via addons)
- Refactors crafting modules, recipe auditor, and JEI to use CraftingTypes
- Includes `CraftingType.BREWING` in anticipation of #8286

Review please

This extracts and expands on common ideas between #8286 and #8282 and is intended to be merged before those, so that they can be updated to use this method instead of their own.

I've also added recipe enumeration for brewing recipes; in theory, as soon as #8286 is updated to have a crafting module for `CraftingTypes.BREWING` then it should automatically get recipe auditing and JEI support as well, but I haven't tested this.

There is possibly some other almost-common logic that could be moved to CraftingType -- perhaps some of the logic to decide what recipe GUI to show?

### DIFF
--- a/src/main/java/com/minecolonies/apiimp/initializer/ModCraftingTypesInitializer.java
+++ b/src/main/java/com/minecolonies/apiimp/initializer/ModCraftingTypesInitializer.java
@@ -1,0 +1,38 @@
+package com.minecolonies.apiimp.initializer;
+
+import com.minecolonies.api.crafting.ModCraftingTypes;
+import com.minecolonies.api.crafting.RecipeCraftingType;
+import com.minecolonies.api.crafting.registry.CraftingType;
+import com.minecolonies.coremod.recipes.BrewingCraftingType;
+import net.minecraft.item.crafting.IRecipeType;
+import net.minecraftforge.event.RegistryEvent;
+import net.minecraftforge.registries.IForgeRegistry;
+
+public final class ModCraftingTypesInitializer
+{
+    private ModCraftingTypesInitializer()
+    {
+        throw new IllegalStateException("Tried to initialize: ModCraftingTypesInitializer but this is a Utility class.");
+    }
+
+    public static void init(final RegistryEvent.Register<CraftingType> event)
+    {
+        final IForgeRegistry<CraftingType> reg = event.getRegistry();
+
+        ModCraftingTypes.SMALL_CRAFTING = new RecipeCraftingType<>(ModCraftingTypes.SMALL_CRAFTING_ID,
+                IRecipeType.CRAFTING, r -> r.canCraftInDimensions(2, 2));
+        reg.register(ModCraftingTypes.SMALL_CRAFTING);
+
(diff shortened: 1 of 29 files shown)


marchermans (approved):
Nice work

src/main/java/com/minecolonies/apiimp/initializer/ModCraftingTypesInitializer.java
```diff
@@ -0,0 +1,38 @@
...
+    {
+        final IForgeRegistry<CraftingType> reg = event.getRegistry();
+
+        ModCraftingTypes.SMALL_CRAFTING = new RecipeCraftingType<>(ModCraftingTypes.SMALL_CRAFTING_ID,
+                IRecipeType.CRAFTING, r -> r.canCraftInDimensions(2, 2));
+        reg.register(ModCraftingTypes.SMALL_CRAFTING);
+
+        ModCraftingTypes.LARGE_CRAFTING = new RecipeCraftingType<>(ModCraftingTypes.LARGE_CRAFTING_ID,
+                IRecipeType.CRAFTING, r -> r.canCraftInDimensions(3, 3)
+                    && !r.canCraftInDimensions(2, 2));
```
Raycoms, line 28: this seems weird, this one can also do 2x2

uecasm: Yes, but that's already covered by the `SMALL_CRAFTING`.  The large crafting module always does both types, but each type has an exclusive subset of recipes with no overlaps, otherwise it would count them more than once.

Raycoms: Couldn't we make it then so that the code that checks has either LARGE or SMALL but not together? That seems like it would be cleaner than this

uecasm: We could, but there's a few places that check for only `SMALL_CRAFTING` because they assume that anything that has large crafting does both, and they don't care what crafting size, only that it's vanilla crafting.  If we were to make them mutually exclusive then they'd have to explicitly check for both, which is more wordy.  This just makes everything simpler.  (This is the same thing you did in your own PR, incidentally...)

Raycoms: Yeah, I know that I had done that in mine too, but now I think it makes more sense to separate that better. There is also a few calls where we check both in some way too, so overall the complexity should stay the same. Technically in some of those places we can even check if IRecipeType.Crafting instead of both (as they got that in common)

uecasm: I disagree.  And there are no places where we check more than one type at once.  It is much much simpler and more sensible the way it is now.

The only reason I can think of where it might make sense to split up the checks is if the 2x2 and 3x3 crafting did have completely separate GUI classes.  But they don't -- the "can do 3x3" is just a modifier flag to the same window for both.

PR #8344, opened 2022-05-06
Fix/request name

uecasm:
- Fixes `@NotNullable` method that returns null, plus associated javadoc
- Allows request `getRequesterDisplayName` to work again if called server-side
- Fixes NPE if calling `building.getRequesterDisplayName` with invalid request (e.g. one owned by a module rather than the building itself).

Review please

Most of these are not things directly run into in MineColonies code itself, they were encountered by trying to use the API from my mod.

The building/requestresolver code still strikes me as a bit weird.  I followed the pattern of the existing code but I do wonder whether it would be better to make the requestresolver just call the building, and have the building be responsible for checking its modules if it doesn't know about the request itself.  It seems a bit weird for external code to be reaching directly into modules.

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
@@ -78,9 +78,10 @@ public interface IBuilding extends IBuildingContainer, IRequestResolverProvider,
      * @param clazz the class of the module.
      * @param modulePredicate the predicate to match.
      * @param <T> the module type.
-     * @return the first matching module (could be null).
+     * @return the first matching module.
+     * @throws IllegalArgumentException if your condition does not match any modules
      */
-    @Nullable
+    @NotNull
     <T extends IBuildingModule> T getModuleMatching(Class<T> clazz, Predicate<? super T> modulePredicate);
 
     /**
--- a/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
@@ -77,10 +77,10 @@ public interface IBuilding extends IBuildingContainer, IRequestResolverProvider,
      * Get a module matching a certain predicate.
      * @param clazz the class of the module.
      * @param modulePredicate the predicate to match.
-     * @param <T> the optional type.
-     * @return optional of the matching predicate (could be empty).
+     * @param <T> the module type.
+     * @return the first matching module (could be null).
      */
-    @NotNull
+    @Nullable
     <T extends IBuildingModule> T getModuleMatching(Class<T> clazz, Predicate<? super T> modulePredicate);
 
     /**
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
@@ -210,7 +210,7 @@ public abstract class AbstractBuilding extends AbstractBuildingContainer
         return Optional.empty();
     }
 
-    @Nullable
+    @NotNull
     @Override
     public <T extends IBuildingModule> T getModuleMatching(final Class<T> clazz, final Predicate<? super T> modulePredicate)
     {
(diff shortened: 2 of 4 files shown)


src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -77,10 +77,10 @@ public interface IBuilding extends IBuildingContainer, IRequestResolverProvider,
      * Get a module matching a certain predicate.
      * @param clazz the class of the module.
      * @param modulePredicate the predicate to match.
-     * @param <T> the optional type.
-     * @return optional of the matching predicate (could be empty).
+     * @param <T> the module type.
+     * @return the first matching module (could be null).
      */
-    @NotNull
+    @Nullable
```
marchermans: Then this needs to be one an optional.
It is either not null or returning an optional.....

From my side you can create an overload that returns the optional and throw an exception on the first if the module is not found.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
```diff
@@ -210,7 +210,7 @@ public <T extends IBuildingModule> Optional<T> getFirstOptionalModuleOccurance(f
         return Optional.empty();
     }
 
-    @NotNull
+    @Nullable
```
marchermans: See interface comment

marchermans (dismissed):

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
uecasm: I don't disagree with that, but that's actually an interface change and not just a doc/attribute change.  And one which @Raycoms previously objected to when I previously suggested it, IIRC.

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
uecasm: As noted above, this does not change the implementation at all, it just corrects the declaration to agree with the actual implementation.

src/api/java/com/minecolonies/api/colony/buildings/IBuilding.java
Raycoms: This is NotNull on purpose, we call this if we expect this to exist and throw an exception otherwise.
(AI dependent calls)

Raycoms: For anything else, create an getOptionalModuleMatching method

uecasm: Again, go look at the actual implementation, which I did not change.  The interface was lying.

Raycoms: I know this, I implemented this. It's because it has to return something if it can't find it. But this should NEVER happen, because the place that calls it relies on it.

uecasm: That's the wrong way to do things.

Raycoms: It's intentional, I don't want to do unpacking everywhere for something that cannot be null because it has to be there. For anything else there is the getOptional method.

Raycoms: And I discussed this enough on the original PR already =P

uecasm: That's fine, but if you're going to do that then the annotation should still be correct.  Either it should be `@Nullable` because it can return null (and it's just the caller that's either expecting it or not expecting it, and the latter might give it an NPE), or it should be `@NotNullable` and it should throw an exception instead of returning null if it can't find what it's looking for.  There are no other correct options.

uecasm: And, FWIW, `IBuildingView.getModuleViewMatching()` is also `@Nullable`, and I didn't change that.  So this change makes it more consistent with that.

Raycoms: Then make it throw an exception if not there

Raycoms: But gotta make sure its handled correctly everywhere then

marchermans (approved):

PR #8366, opened 2022-05-12
Fix claim issues with overlapping colonies

someaddons:
Fix claim issues with multiple colonies, overlapping claims and building claims not getting removed at all.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/IColonyTagCapability.java
+++ b/src/api/java/com/minecolonies/api/colony/IColonyTagCapability.java
@@ -2,10 +2,11 @@ package com.minecolonies.api.colony;
 
 import com.minecolonies.api.util.BlockPosUtil;
 import com.minecolonies.api.util.NBTUtils;
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
 import net.minecraft.nbt.CompoundTag;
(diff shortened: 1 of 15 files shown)


src/api/java/com/minecolonies/api/colony/IColonyTagCapability.java
```diff
@@ -178,33 +207,53 @@ public void addBuildingClaim(final int colonyId, final BlockPos pos, final Level
...
+                                owningColony = colonies.iterator().next();
+                            }
+                        }
                     }
                     else
                     {
-                        setOwningColony(claimingBuildings.keySet().toArray(new Integer[0])[new Random().nextInt(claimingBuildings.size())], chunk);
+                        for (final Map.Entry<Integer, Set<BlockPos>> colonyEntry : claimingBuildings.entrySet())
+                        {
+                            final IColony colony = IColonyManager.getInstance().getColonyByDimension(colonyEntry.getKey(), chunk.getLevel().dimension());
```
Raycoms, line 241: maybe we do a) The colony null check before the loop, and b) We remove the buildings when they're null?

someaddons: ye makes sense

Raycoms (approved):

PR #8370, opened 2022-05-14
add new crusher research

Raycoms:
Closes #
Closes #
Closes #

- Adds a new research that allows the crusher to convert: 2:1 deepslate to tuff and 2:1 tuff to cobble
- Adjusts collision timeout slightly. 10s gives me a feeling of "stuck".
-

Review please


uecasm:
Perhaps we should also add fancy stone (granite etc) to cobble too?  Maybe even stone to cobble, although that might risk a recipe cycle.

someaddons (dismissed):

PR #8376, opened 2022-05-15
Update BlockPosUtil.java

### LINKED ISSUE #8360: Game crashes whenever I open the "Fields" page of the Farm.

I am also running the latest (possible) versions of other mods that are part of my problem.

- [X] I checked the MineColonies/Structurize [Wiki](https://wiki.minecolonies.ldtteam.com/) or I asked on [discord](https://discord.minecolonies.com/). Or I was sent from discord to open an issue here.

### Minecraft Version

1.16

### MineColonies Version

minecolonies-1.16.5-1.0.712-ALPHA

### Structurize Version

structurize-1.16.5-1.0.392-RELEASE

_No response_

### Current Behavior

Whenever I go to my farm, and I open the "Fields" page, the game saves and quits, and crashes.

### Expected Behavior

I expected it to open the "Fields" page.

### Reproduction Steps

1. Go to the farm
2. Open the farm
3. Go to the fields page

### Logs

https://github.com/YoTroops/Crash/commit/c02643c84dfe9e1ba84e798ac46c9d803c453599

### Anything else?

_No response_

### Footer

---
## Viewers

Thodor12:
Closes #8360

- Remove faulty X coord check in the first if statement
- The `calcDirection` method is currently not returning a result for every possible combination, this is causing a crash in (for example) the farmer fields window because `setText` is receiving a null argument, the method has been reworked to:
  - If X and Z are identical, check for a change vertically (either directly above or below the building, this is here because this is a valid position, but still doesn't have a horizontal direction)
  - If Z is greater or smaller than the building pos, make it either north or south (+/- 1 was not needed because it only leaves a gap around the building itself)
  - If X is greater or smaller than the building pos, make it either east or west (combining it with the result of north/south)
  - If all conditions fail return a text saying it's the identical location and there's no direction (this is a fallback so it still won't return null, even if the positions are identical)

Review please


Nightenom (changes requested):
The method can still return null if pos.xz = building.xz

Thodor12:
What scenario should xz return, since this is only possible to reproduce when placed directly above the hut block.
I don't think an empty string is a great idea

MotionlessTrain:
If pos is within one block of the building, it is also returning null
(e.g. `pos.getX() == building.getX() && pos.getZ() == building.getZ()`)
Maybe a default value of a text component of an empty string is better, or an ">=" check could help as well for this specific case

Thodor12:
> If pos is within one block of the building, it is also returning null (e.g. `pos.getX() == building.getX() && pos.getZ() == building.getZ()`) Maybe a default value of a text component of an empty string is better, or an ">=" check could help as well for this specific case

Talking to Ray, am already gonna remove the +1 on either side as these aren't really needed

Thodor12:
Ray suggested a vertical check if `xz` matches the building `xz` exactly, thoughts?

MotionlessTrain:
That may work, but there is still the edge case that both x, y and z are equal

Thodor12:
> That may work, but there is still the edge case that both x, y and z are equal

If that ever occurs I'm just going to return a sarcastic message, "how the hell did you manage that?"

Raycoms:
> That may work, but there is still the edge case that both x, y and z are equal

Not possible, the hut block is there =D

src/api/java/com/minecolonies/api/util/BlockPosUtil.java
```diff
@@ -711,45 +716,57 @@ public static Direction getXZFacing(final BlockPos pos, final BlockPos neighbor)
...
             }
             else
             {
-                compoment = new TranslationTextComponent(DIRECTION_WEST);
+                component = Optional.of(new TranslationTextComponent(DIRECTION_WEST));
             }
         }
 
-        return compoment;
+        return component.orElse(new TranslationTextComponent(DIRECTION_NONE));
```
Raycoms: Oh no, don't package this through an optional. null checks are fine, no need to create wrappers everywhere

Raycoms: can always component == null ? NONE : component;

MotionlessTrain:
> > That may work, but there is still the edge case that both x, y and z are equal
> 
> Not possible, the hut block is there =D

I agree, in that case the function was called wrongly, so that is more a problem at the call side :D

Thodor12:
> > > That may work, but there is still the edge case that both x, y and z are equal
> > 
> > 
> > Not possible, the hut block is there =D
> 
> I agree, in that case the function was called wrongly, so that is more a problem at the call side :D

I'm returning a text now just in case it happens, at least you don't get an outright crash, which is always better.

Nightenom (dismissed):

Raycoms (dismissed):

marchermans (changes requested):
This looks to me like an untested not required change.

Please list the issue number this relates to, describe your changes better and add comments in the code why the check needs to be this way.

Thodor12:
I have tested this and I've commented in the other issue towards this with the related problem

marchermans:
If this is referenced somewhere else, then please state so in the pr description, on the phone i can not see those directly.

Additional code comments are for sure required though

Thodor12:
> If this is referenced somewhere else, then please state so in the pr description, on the phone i can not see those directly.
> 
> Additional code comments are for sure required though

I updated the PR description since the changes updated a bit since PR creation, code comments shouldn't be necessary though as the code is pretty straightforward

marchermans:
Code comments are needed since this pr changes a rather complicated if switch. Please document properly as to why this change is needed in the code.

This is not much work, it takes 10 seconds to do, but Dave's others who are reading that code during debugging sessions from having to look it up here, if they can even easily find it.

marchermans:
Additionally your pr description is still not very good. The four bullet points don't describe what you are actually doing to what method and why.

Please elaborate your reasoning as to why you think those four points are sufficient.

Thodor12:
@OrionDevelopment I updated the description again, is it good enough now?

marchermans (approved):

PR #8396, opened 2022-05-22
Courier delivers only necesary items

ImEr1c:
-Made the delivery man / courier deliver only the requested items with the right amounts

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
@@ -350,17 +350,17 @@ public class EntityAIWorkDeliveryman extends AbstractEntityAIInteract<JobDeliver
         boolean extracted = false;
         final IItemHandler workerInventory = worker.getInventoryCitizen();
         final List<ItemStack> itemsToDeliver = job.getTaskListWithSameDestination((IRequest<? extends Delivery>) currentTask).stream().map(r -> r.getRequest().getStack()).collect(Collectors.toList());
-        final Map<Item, Integer> itemsCount = new HashMap<>();
+        final Map<ItemStorage, Integer> itemsCount = new HashMap<>();
 
         for (ItemStack itemStack : itemsToDeliver)
         {
-            if (itemsCount.containsKey(itemStack.getItem()))
+            if (itemsCount.containsKey(new ItemStorage(itemStack)))
             {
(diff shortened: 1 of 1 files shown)


Raycoms:
Thanks for the PR. Could you give us a small desc of "what was the problem - always happened when" ?

ImEr1c:
As i saw in the code, when the delivery man moves the items to the building that it needs to, it moves all the items, not only the ones that it needs, like if it would pickup an item or someone would give them an item before it arrives to the building, it would also tranfer that item
Also saw a todo for something like this

Raycoms:
> As i saw in the code, when the delivery man moves the items to the building that it needs to, it moves all the items, not only the ones that it needs, like if it would pickup an item or someone would give them an item before it arrives to the building, it would also tranfer that item Also saw a todo for something like this

Ah okay, not a bug fix, just making sure only the "requested items" are also delivered.

ImEr1c:
Yeah, so its ok?

src/main/java/com/minecolonies/coremod/entity/ai/citizen/deliveryman/EntityAIWorkDeliveryman.java
```diff
@@ -345,14 +349,44 @@ private IAIState deliver()
         boolean success = true;
         boolean extracted = false;
         final IItemHandler workerInventory = worker.getInventoryCitizen();
+        final List<ItemStack> itemsToDeliver = job.getTaskListWithSameDestination((IRequest<? extends Delivery>) currentTask).stream().map(r -> r.getRequest().getStack()).collect(Collectors.toList());
+        final Map<Item, Integer> itemsCount = new HashMap<>();
```
Raycoms: I understand the reason behind the complexity, but I would simplify this grossly.
We mostly want to avoid them delivering things they picked up along the way. So we just map the List to a map of ItemStorage here (new ItemStorage(stack)).

And in the bottom we just check for a contains new ItemStorage(stack).

(p.s.: Item map key doesn't work, because of items with nbts etc)

batmanatorul: Ok, I'm gonna do that tomorrow, its 3 am for me

ImEr1c:
Forgot to put my name in git when committed

Raycoms:
> Forgot to put my name in git when committed

double check my comment again, I think we can further simplify this. Aside of this, please re-make the PR then with the updated name so that you can sign the CLA

Raycoms:
No prob

------- Original Message -------

PR #8405, opened 2022-05-27
small PR to ignore those blocks

Raycoms:
This is part one of the specific compat for vanilla structure handling.
This allows setting up the jigsaw blocks etc within a schematic area but then have it ignored by the builder.

I want to invite a two-fold discussion here.

a) Do we want to render them?
b) Do we want to make this maybe a block tag?
c) If we make this a block tag, we can also use this tag for conditional rendering then (similar as the void blocks).

So maybe two tags?

One for conditional rendering.
One for "ignored blocks" ?

Furtheremore, for part two of the compat I had the following planned:

a) A "special scan tool" 
- This scan tool will create a vanilla structure file, however it will do two things:
a) It will fill in the necessary data of the jigsaw blocks (tool will have building/road/field toggles) and a textfield for "stylename" and one for "buildingname".
b) It will also apply the correct tags to the buildings.

This would make it trivial to setup the worldgen stuff for new styles.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/placementhandlers/BuilderIgnorePlacementHandler.java
+++ b/src/main/java/com/minecolonies/coremod/placementhandlers/BuilderIgnorePlacementHandler.java
@@ -1,0 +1,76 @@
+package com.minecolonies.coremod.placementhandlers;
+
+import com.ldtteam.structurize.placement.handlers.placement.IPlacementHandler;
+import com.ldtteam.structurize.util.BlockUtils;
+import com.ldtteam.structurize.util.PlacementSettings;
+import com.minecolonies.api.util.Log;
+import com.minecolonies.api.util.WorldUtil;
+import net.minecraft.core.BlockPos;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.JigsawBlock;
+import net.minecraft.world.level.block.StructureBlock;
+import net.minecraft.world.level.block.StructureVoidBlock;
+import net.minecraft.world.level.block.state.BlockState;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collections;
+import java.util.List;
+
+import static com.ldtteam.structurize.placement.handlers.placement.PlacementHandlers.handleTileEntityPlacement;
+
+/**
+ * Handler for some specific blocks that we want only to paste/cost in the complete mode.
+ */
+public class BuilderIgnorePlacementHandler implements IPlacementHandler
+{
+    @Override
+    public boolean canHandle(@NotNull final Level world, @NotNull final BlockPos pos, @NotNull final BlockState blockState)
+    {
+        return blockState.getBlock() instanceof JigsawBlock || blockState.getBlock() instanceof StructureBlock || blockState.getBlock() instanceof StructureVoidBlock;
+    }
+
+    @Override
+    public ActionProcessingResult handle(
+      @NotNull final Level world,
+      @NotNull final BlockPos pos,
+      @NotNull final BlockState blockState,
+      @Nullable final CompoundTag tileEntityData,
+      final boolean complete,
+      final BlockPos centerPos,
+      final PlacementSettings settings)
+    {
+        if (complete)
+        {
+            WorldUtil.setBlockState(world, pos, blockState, com.ldtteam.structurize.api.util.constant.Constants.UPDATE_FLAG);
+            if (tileEntityData != null)
+            {
+                try
+                {
+                    handleTileEntityPlacement(tileEntityData, world, pos, settings);
+                    blockState.getBlock().setPlacedBy(world, pos, blockState, null, BlockUtils.getItemStackFromBlockState(blockState));
+                }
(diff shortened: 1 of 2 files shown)


uecasm:
> Do we want to render them?

Not by default, but having a button for it might be nice.  On a related note, the tan placeholders should probably not be shown by default either.  (Especially in the L5 previews... they're occasionally interesting, but mostly they just get in the way.)

> Do we want to make this maybe a block tag?

I don't think so.  Jigsaw blocks can't be ignored -- they must explicitly convert to the block specified in their internal replace data (which will usually, but not always, be air), whenever placed by builder or right-paste, but not left-paste or worldgen.  This is still different treatment from any other block, so there's not really any benefit in tagging it.

> It will fill in the necessary data of the jigsaw blocks

I'm also a little dubious about this.  Something to help fill them in might be useful, especially if they're left blank, but it should be allowed for the style designer to set them up how they like (including different replace blocks) without getting overwritten automatically.

src/main/java/com/minecolonies/coremod/placementhandlers/BuilderIgnorePlacementHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Handler for some specific blocks that we want only to paste/cost in the complete mode.
+ */
+public class BuilderIgnorePlacementHandler implements IPlacementHandler
+{
+    @Override
+    public boolean canHandle(@NotNull final Level world, @NotNull final BlockPos pos, @NotNull final BlockState blockState)
+    {
+        return blockState.getBlock() instanceof JigsawBlock || blockState.getBlock() instanceof StructureBlock || blockState.getBlock() instanceof StructureVoidBlock;
```
uecasm, line 32: ... which makes this incorrect, because it shouldn't be ignoring them...

Structure Void should really be treated exactly identical to the tan placeholders.  Regular structure blocks shouldn't appear inside schematics at all (but if they do, it's probably reasonable to treat them like placeholders too).  Jigsaws are an entirely different animal, however.

marchermans:
> > Do we want to render them?
> 
> Not by default, but having a button for it might be nice.  On a related note, the tan placeholders should probably not be shown by default either.  (Especially in the L5 previews... they're occasionally interesting, but mostly they just get in the way.)
> 
> > Do we want to make this maybe a block tag?
> 
> I don't think so.  Jigsaw blocks can't be ignored -- they must explicitly convert to the block specified in their internal replace data (which will usually, but not always, be air), whenever placed by builder or right-paste, but not left-paste or worldgen.  This is still different treatment from any other block, so there's not really any benefit in tagging it.
> 
> > It will fill in the necessary data of the jigsaw blocks
> 
> I'm also a little dubious about this.  Something to help fill them in might be useful, especially if they're left blank, but it should be allowed for the style designer to set them up how they like (including different replace blocks) without getting overwritten automatically.

I agree with this analysis

Raycoms:
> > Do we want to render them?
> 
> Not by default, but having a button for it might be nice. On a related note, the tan placeholders should probably not be shown by default either. (Especially in the L5 previews... they're occasionally interesting, but mostly they just get in the way.)
> 

I want to handle those the same way I want to handle the light placeholder blocks (enable/disable on the same toggle).
So, technically there could be a tag for all the blocks we want to render conditionally in this category.

> > Do we want to make this maybe a block tag?
> 
> I don't think so. Jigsaw blocks can't be ignored -- they must explicitly convert to the block specified in their internal replace data (which will usually, but not always, be air), whenever placed by builder or right-paste, but not left-paste or worldgen. This is still different treatment from any other block, so there's not really any benefit in tagging it.
> 

The idea is to make this as simple as possible, I know that in theory, we can set replacement blocks in the jigsaw blocks, but it's behavior we don't necessarily need for our worldgen colonies and even less need for our builder.

> > It will fill in the necessary data of the jigsaw blocks
> 
> I'm also a little dubious about this. Something to help fill them in might be useful, especially if they're left blank, but it should be allowed for the style designer to set them up how they like (including different replace blocks) without getting overwritten automatically.

This is easy, if the jigsaw block has its "standards" set and nothing explicit, then it will override. The idea of this system is to allow the default setup for our worldgen colonies and make it as easy as possible, if people want to do things custom they can always fall back to the vanilla system.

uecasm:
> The idea is to make this as simple as possible, I know that in theory, we can set replacement blocks in the jigsaw blocks, but it's behavior we don't necessarily need for our worldgen colonies and even less need for our builder.

Just make an adaptor that can run the standard `JigsawReplacementProcessor` (or any `StructureProcessor`) on the blueprints :)  (I had one of those kicking around in my old worldgen branch, although I'm not sure it would still be applicable here.)

uecasm:
FWIW, you will absolutely want to use the jigsaw replacement-block feature if you want to include a style like Fortress that has a solid wall on the side where you might want to put a jigsaw.

Raycoms:
Yeah, but we don't want worldgen fortress colonies, that'd need a massive flat area. I know in general in the future we want support for that. But for an 200%improvement for the status quo, we can do without and then add it later.

uecasm:
Fortress was an example.  But it applies any time that there's supposed to be a non-air block where the jigsaw needs to go.  Could just as easily be a fence or a placeholder or any other block, since the jigsaws would typically be at the edges of the building.

PR #8435, opened 2022-06-11
Expedition log for nether mine

uecasm:
- Bump structurize to get support for entity icon control and tooltip fix
- Change `MarkBuildingDirtyMessage` to actually mark building as dirty (causing colony data sync), not just marking the chunk as dirty (which doesn't)
    - This is used by a few other resource list displays, but as far as I can tell this was the intended purpose
- Tweaks nether worker behaviour slightly so that they will immediately attempt to equip a replacement tool/armour if their current one breaks in the middle of a fight, instead of waiting until the fight is over; this improves survivability a bit (although only if they had a replacement on hand, which is not common).
- Simplifies and unifies the method used to locate a sword with other equipment; this should make it more compatible with e.g. tinkers swords, although I didn't explicitly test that.
- Add an "expedition log" for the nether miner, which shows what they're up to while they're away in their rabbit hole
    - Locked behind research, though deliberately modest requirements:
        image
    - Shows various bits of info (not pictured: also shows current food item):
        image
        image
    - Persists after death, so you can get a hint why they might have died (although it doesn't show who they last fought, because they didn't survive to report on it):
        image

Review please

RIP Gracelynn.

The expedition log will update "live" while you have the window open, but only then; I tried very hard to minimize how often updates are sent.  As a caveat, the tab will not appear immediately after unlocking the research; you need to do something else to trigger a building update afterward.  Updating the hut inventory (e.g. by satisfying an equipment request) is usually sufficient.

The log itself is designed as generically as I can currently think of, so that it can in future be reused for other workers that do something vaguely similar -- it's up to the AI exactly what gets recorded and when.


Raycoms (dismissed):

someaddons:
I like the overview, looks great. On the last image where she died, do you still get the resources? if so we may want to add an explanation text how one got those

uecasm:
No, you don't get any of the resources when they die -- that hasn't changed from previous behaviour.  It's basically just showing you what you missed out on.

Basically the lore is that they were using the ender eyes and ancient enchantments to have like a walkie talkie or crystal ball or something to constantly report back what they were doing, so you see the monsters they've defeated and the resources they gathered.  But if they get killed, they don't return with the resources and they never reported on the monster that actually defeated them, because they didn't have time.

someaddons:
Its a bit weird then to display the result items you gained, when you do not gain them, might be better to display those only after/if they return

uecasm:
Don't think of them as the items you gained.  They're the items the explorer _found_.  You don't actually get any of them until they return safely.

The explorer is reporting back "hey, I found 6 quartz", "I just defeated a hoglin and got a raw pork chop", "uh oh, my sword broke" etc as things are happening.  So you know about them, but you don't have them in hand until they actually come back.

Talyda:
I know I have nothing to do with this whatsoever, but just wanted to say - I LOVE this!  Really love it!

Raycoms:
> Don't think of them as the items you gained. They're the items the explorer _found_. You don't actually get any of them until they return safely.
> 
> The explorer is reporting back "hey, I found 6 quartz", "I just defeated a hoglin and got a raw pork chop", "uh oh, my sword broke" etc as things are happening. So you know about them, but you don't have them in hand until they actually come back.

Maybe show the loot "crossed out" on death then?

someaddons:
ye I got that its the items the entity "found" in the nether while beeing there, its just the player pov is likely, hey this is the report of the expedition, he encountered x monsters and gave me these resources, so probably want to clarify it a bit

uecasm:
Ok, I've added an indicator with a tooltip that appears when the expedition has ended in death, and also shaded the resource background red as well:
image

Sadly, during the course of this I found a tooltip bug that we should probably wait for (since it makes this less effective): ldtteam/Structurize#519

Raycoms (dismissed):

someaddons (approved):

PR #8500, opened 2022-07-10
Fix/blaze spam

uecasm:
Closes [Discord report](https://discord.com/channels/139070364159311872/915411448342118441/995604056158519316)

- Attempts to stop alchemist from spamming requests for blaze powder

Review please

I have not reproduced or tested this, but it seems like the most likely cause for the reported issue.
(It probably affects other branches too.)

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/citizen/alchemist/EntityAIWorkAlchemist.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/citizen/alchemist/EntityAIWorkAlchemist.java
@@ -5,7 +5,6 @@ import com.minecolonies.api.colony.interactionhandling.ChatPriority;
 import com.minecolonies.api.colony.requestsystem.request.IRequest;
 import com.minecolonies.api.colony.requestsystem.request.RequestState;
 import com.minecolonies.api.colony.requestsystem.requestable.Stack;
-import com.minecolonies.api.colony.requestsystem.requestable.StackList;
 import com.minecolonies.api.colony.requestsystem.requestable.crafting.PublicCrafting;
 import com.minecolonies.api.crafting.IRecipeStorage;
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/entity/ai/citizen/alchemist/EntityAIWorkAlchemist.java
```diff
@@ -510,7 +512,7 @@ private IAIState checkBrewingStandFuel()
 
         if(!InventoryUtils.hasItemInItemHandler(worker.getInventoryCitizen(), Items.BLAZE_POWDER)
              && !InventoryUtils.hasItemInProvider(building, Items.BLAZE_POWDER)
-             && !building.hasWorkerOpenRequestsOfType(worker.getCitizenData().getId(), TypeToken.of(StackList.class)))
+             && !building.hasWorkerOpenRequestsOfType(worker.getCitizenData().getId(), TypeToken.of(Stack.class)))
```
Raycoms, line 515: That's def the reason, copy pasta error. But then we need to use the code other workers use too where we check if the request contains the item

uecasm: I looked at other places `hasWorkerOpenRequestOfType` was called and they don't check.

uecasm: And it doesn't look like the AI creates any other requests at all, let alone other `Stack` requests, so it may be safe.  Provided it doesn't collide with whatever the crafting system does.  (But at worst that should presumably only cause the requests to be processed sequentially rather than in parallel.)

Raycoms (approved):

someaddons (approved):

PR #8533, opened 2022-08-02
Feature/puzzle handling

Raycoms:
- Properly handle jigsaw handling
- Ignore other structure block
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/placementhandlers/JigsawPlacementHandler.java
+++ b/src/main/java/com/minecolonies/coremod/placementhandlers/JigsawPlacementHandler.java
@@ -1,14 +1,20 @@
 package com.minecolonies.coremod.placementhandlers;
 
+import com.google.common.collect.ImmutableList;
 import com.ldtteam.structurize.placement.handlers.placement.IPlacementHandler;
 import com.ldtteam.structurize.util.BlockUtils;
 import com.ldtteam.structurize.util.PlacementSettings;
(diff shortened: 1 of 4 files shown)


marchermans (approved):

src/main/java/com/minecolonies/coremod/placementhandlers/JigsawPlacementHandler.java
```diff
@@ -0,0 +1,125 @@
...
+                    Log.getLogger().warn("Unable to place TileEntity");
+                }
+            }
+            return ActionProcessingResult.SUCCESS;
+        }
+
+        if (tileEntityData != null && tileEntityData.contains("final_state"))
+        {
+            final String stateString = tileEntityData.getString("final_state");
+            BlockStateParser blockstateparser = new BlockStateParser(new StringReader(stateString), false);
```
uecasm, line 71: It might be nice if you could invoke `JigsawReplacementProcessor.processBlock` instead of hard-coding this data/behaviour.  Just in case Mojang decide to change the storage or something in the future.  It's basically the same code as this.

Raycoms: Didn't want to do it because of potential side-effects. If they change the storage, it will just stop working for us and we'll directly notice it, but if they change sth to have side-effects it will be hard to find.

PR #8543, opened 2022-08-05
Fix DO teaching crash

### LINKED ISSUE #8530: Attempting to submit a custom recipe with the top ingredient missing causes an immediate crash/shut down

### Minecraft Version

1.18

### MineColonies Version

minecolonies-1.18.2-1.0.893-ALPHA

### Structurize Version

structurize-1.18.2-1.0.411-ALPHA

Mods are listed in the crash report

### Current Behavior

Submitting a recipe in the Custom Tab (I was in the carpenter hut) with the top ingredient missing caused an immediate hard crash of the game.  I thought I saved the first ingredient and clicked the save/teach button and it just crashed.

### Expected Behavior

Attempting to save/teach a recipe without all the ingredients should just tell you that it's wrong or just not save.

### Reproduction Steps

Attempt to save a recipe for Cobblestone Shingles (or any) but put your ingredients in the bottom two slots vs the top two

### Logs

https://pastebin.com/jPjRQmdF

### Anything else?
I did discuss this in the discord, and they were aware of a previous bug/error where it would crash with no ingredients...however that was supposed to have been fixed according to them.  They said I should put this in as it sounded like a new bug and I'm on the newest version at the time of submission.

uecasm:
Closes #8530

- Fixes crash teaching DO recipes when input omitted

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/modules/DOCraftingWindow.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/modules/DOCraftingWindow.java
@@ -103,18 +103,8 @@ public class DOCraftingWindow extends AbstractModuleWindow
             }
         }
 
-        int inputCount = 0;
-        for (int i = 0; i < inputInventory.getContainerSize(); i++)
-        {
-            if (!inputInventory.getItem(i).isEmpty())
-            {
-                inputCount++;
-            }
-        }
-
-        final List<Integer> inputIndizes = map.get(inputCount);
         final List<ItemStorage> input = new ArrayList<>();
-        for (int i = 0; i < inputCount; i++)
+        for (int i = 0; i < inputInventory.getContainerSize(); ++i)
         {
             final ItemStack atPos = inputInventory.getItem(i).copy();
             if (!ItemStackUtils.isEmpty(atPos))
@@ -122,6 +112,12 @@ public class DOCraftingWindow extends AbstractModuleWindow
                 atPos.setCount(1);
                 input.add(new ItemStorage(atPos));
             }
+        }
+
+        final List<Integer> inputIndizes = map.get(input.size());
+        if (inputIndizes == null)
+        {
+            return;
         }
 
         final List<ItemStack> additionalOutput = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/client/gui/modules/DOCraftingWindow.java
```diff
@@ -103,18 +103,8 @@ private void addRecipe()
...
-            if (!inputInventory.getItem(i).isEmpty())
-            {
-                inputCount++;
-            }
-        }
-
-        final List<Integer> inputIndizes = map.get(inputCount);
         final List<ItemStorage> input = new ArrayList<>();
-        for (int i = 0; i < inputCount; i++)
+        for (int i = 0; i < inputInventory.getContainerSize(); ++i)
```
Raycoms, line 107: I'm not sure this works reliably. I remember that I added this on purpose because else, in some situations some things wouldn't work like:
- Having multiple items in the input, but some of the recipe subsets only require a subset of those items, etc.

uecasm: I changed it because the previous version was not correct if there were any empty items (although there probably shouldn't have been, which is why it mostly worked).

It first looped through the entire inventory, counting how many items were non-empty (N).  Then it looped through only the first N slots of the inventory, again ignoring empty items, but also ignoring any non-empty >N slots.

(So it would be wrong if for example the inventory had only slot 1 and slot 3 filled with slot 2 empty.  Though having said that, this would never actually happen since `list` is empty in that scenario, which is what caused the crash.)

This code only ever added one recipe, it didn't support subsets.  The UI is the same -- when you put in one slot it will show a bunch of panels, but these go away once you fill in two slots.  It also doesn't show outputs with flipped inputs, only the different combos with the exact same inputs in the same order.

It looked ok when I tested it, but if there's a specific combo you think might be problematic, I can try it out and see if it works.

uecasm: ~Having said that, I'm seeing a different bug while testing this further (not related to these changes, it happens in a clean build too): when I add a recipe to a building it plays the happy noise but does not appear in the recipe list until I relog.  (Also not DO-specific, it happens for regular recipes too.)~
Weirdly, that was happening reliably across multiple game restarts, but then I went to a different hut and it was working, and now it's working again at the original hut.  So no idea what was up with that.

Raycoms: Isn't container size always 3? And what happens if you add/remove inputs?

uecasm: There's still an empty slot check, so it still calculates `inputCount` (though here now `input.size()`) correctly.  I just merged the two loops you had into one, which avoided a potential bug.

Raycoms: ahk

Raycoms (approved):

PR #8546, opened 2022-08-06
Migrate CompatibilityManager to fully server-sided

uecasm:
- CompatibilityManager discovery is now performed only on server-side, with the results synched to clients (if dedicated server) via network packets.
    - This makes things a lot cleaner; previously it was calculating on both sides and some parts were running more than once.
    - Hopefully this should make `!nofuel` a thing of the past.
- Simplifies global research synching similarly.
- Removes some JEI startup code that previously worked around some weird message timing on dedicated servers; with the above changes and/or other upgrades in Forge, these appear to no longer be needed, and it's much tidier without it.

Review please

I've done basic testing on this with both single-player and a local server, and both initial startup and `/reload`.  I have not tested LAN or a remote server, or with a larger modpack.


src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -217,12 +348,14 @@ public void discover(@NotNull final RecipeManager recipeManager)
     @Override
     public List<ItemStack> getListOfAllItems()
     {
+        if (allItems.isEmpty()) Log.getLogger().error("getListOfAllItems when empty");
```
uecasm, line 351: These things are some debug code that I've sort of half-deliberately left in.  I never saw any of them in testing (suggesting that the timing is good) but there's a remote possibility that they could be useful to indicate problems if they appear in logs with larger installations or other scenarios that I didn't exercise.

Having said that, they probably wouldn't be super useful other than as a basic indicator, since they lack a stack trace; so I'm happy to just remove them if they don't seem sufficiently worthwhile.

Raycoms:
I'm kinda concerned about this, the reason to not sync it was that it could cause long login times in big packs because an enormous ton of data has to be synced over.

uecasm:
I did avoid including what I could (such as the list of all items, since that's "client-safe").  I haven't done specific time or data measurements but I don't think this ought to add all that much to either.  Most of the data is sent as integer arrays, and is a lot smaller than the research data we were already sending.  It could potentially be reduced even further if we could guarantee that there weren't any items with NBT, but I don't think that's a safe assumption.

It could be worthwhile trying it out in a large pack and seeing what happens.

uecasm:
On further reflection, this doesn't actually need to sync `FurnaceRecipes` because we can populate that later from the vanilla recipe sync.  (I was stuck thinking that the `CompatibilityManager` needed those _first_ since they get used in discovery, but of course that doesn't matter when that happens server-side anyway, so it's safe for `FurnaceRecipes` to be populated later than `CompatibilityManager` on client side.)

FWIW, with the standard dev environment the complete `CompatibilityManager` sync packet is 3966 bytes.  I'm reasonably certain this is a lot smaller than the vanilla recipe sync.  (It's certainly a _lot_ smaller than the `FurnaceRecipes` sync was.)  But if there's something else you'd like me to measure, let me know.

someaddons:
could we delay sending this to after player login? think the data isnt that critical, defaults with empty lists or so should work till it arrives

uecasm:
> could we delay sending this to after player login?

No, that won't work, it has to arrive before JEI inits.  Besides, this is happening on the event specifically designated for mods to send this sort of thing.  It's already basically happening as late as possible.

marchermans (approved):
This looks fine, lgtm

someaddons (approved):

Raycoms:
I'd like to merge this after sunday beta only

PR #8556, opened 2022-08-11
Code only

Raycoms:
- Adjust to the new structurize changes
- Special survival placement
- Special creative placement
- Special handling of huts (leveled, named, etc)

Review please


gradle.properties
```diff
@@ -27,9 +27,9 @@ mappingsChannel=official
 mappingsVersion=1.18.2
 
 dataGeneratorsVersion=0.1.48-ALPHA
-blockUI_version=1.18.2-0.0.55-ALPHA
-structurize_version=1.18.2-1.0.424-ALPHA
-domumOrnamentumVersion=1.18.2-1.0.41-ALPHA
+blockUI_version=1.18.2-0.0.58-ALPHA
+structurize_version=1.18.2-1.0.440-PERSONAL
```
uecasm, line 31: I assume this should be a draft then 😛

Raycoms: yes, this won't be merged

src/api/java/com/minecolonies/api/compatibility/newstruct/BlueprintMapping.java
```diff
@@ -0,0 +1,173 @@
...
+        styleMapping.put("incan", "Incan");
+        styleMapping.put("nordic", "Nordic Spruce");
+        styleMapping.put("sandstone", "Desert Oasis");
+        styleMapping.put("spacewars", "Spacewars");
+        styleMapping.put("warped", "Warped Netherlands");
+        styleMapping.put("truedwarven", "Stalactite Caves");
+        styleMapping.put("stone", "Minecolonies Original");
+        styleMapping.put("acacia", "Urban Savanna");
+        styleMapping.put("wooden", "Minecolonies Original");
+        styleMapping.put("asian", "Pagoda");
```
uecasm, line 41: Still not really a fan of all this stuff.  Any way it could be put into the style pack metadata (json file or something) instead?  That way each new style can indicate what its legacy upgrade was, which would also support upgrades for out-of-jar styles.  You have to assume that these will not upgrade simultaneous with this landing, but a little later on, and that might rename/move some things that hopefully shouldn't break as a result.

Raycoms:
At the very least the default mapping should probsbly stay in code, if you want to add a databased approach to this pr you can do so. Just need a datagen too though

Sent from Proton Mail mobile

uecasm:
I mean it should be in the pack.json or whatever, or perhaps an extra mcol-specific file.  I don't see how a datagen would help with that since they're not datapacks.

someaddons (approved):

uecasm:
Speaking of datapacks, it still seems unfortunate that you're not either using the vanilla `PackResources` or at least mirroring its code structure (you could do that even if you didn't want to use its loader).  With the way the structurize code is currently structured it would be a lot harder (though not impossible, of course) to add support for zip-file packs in the future, and that would definitely be a very useful feature to have for external pack distribution.  Heck, it would be a useful feature for the existing server->client custom packs too, since you're already compressing things internally.
(Technically this is a review comment for Structurize, but that's already merged, so...)

PR #8568, opened 2022-08-17
Serveral performance improvements

someaddons:
- Avoiding triggering chunk loading/future gen when checking existance
- Fix structure AI's lagging too much in material gathering
- Fix pathnavigator taking nearly as much time as AI
- Replaced some very frequently used streams on short lists, showing up on profilings where they should not

Review please


marchermans (approved):
Very nice finds :D

Raycoms (approved):

PR #8580, opened 2022-08-24
Several schematic fixes in 1.19

MotionlessTrain:
Closes #
Closes #
Closes #

- Port of #8579 :
- Fix an errant jigsaw block in an Urban Birch guard tower
- Fix an escape path in the Lost Mesa City shephard, due to an unlucky placed dirt slab
- Fix several Fortress mine bugs (e.g. expensive blocks at level 1, wrong scan height)
- Fix schematics with corrupted hut block block entity (missing blueprintDataProvider)

Review please


uecasm:
FWIW, it should no longer be necessary to "fix" errant jigsaw blocks, they should disappear on their own when not-left-pasted.

MotionlessTrain:
I know that, but I still wanted to fix it for consistency reasons, and the jigsaw block was set to replace to air instead of light placeholder block, what it should have been. So it needed to be changed anyway

PR #8586, opened 2022-08-26
Fix raids more

uecasm:
- Clean up some unused rotation/mirror params, and avoid double-loading ship blueprint
- Fix incorrect offset to spawn location due to anchor placement
- Use proper RNG to pick rotation angle
- Allow picking 270° rotation
- Display entire "raid arrived" message in red (rather than just the part before the first argument)

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/managers/EventStructureManager.java
+++ b/src/main/java/com/minecolonies/coremod/colony/managers/EventStructureManager.java
@@ -93,7 +93,7 @@ public class EventStructureManager implements IEventStructureManager
         final Level world = colony.getWorld();
 
         final int y = BlueprintTagUtils.getNumberOfGroundLevels(structure, 4) - 1;
-        final BlockPos spawnPos = targetSpawnPoint.offset(0, -y, 0).offset(structure.getPrimaryBlockOffset());
+        final BlockPos spawnPos = targetSpawnPoint.below(y).above(structure.getPrimaryBlockOffset().getY());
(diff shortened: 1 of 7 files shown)


src/main/java/com/minecolonies/coremod/colony/managers/EventStructureManager.java
```diff
@@ -97,7 +93,7 @@ public boolean spawnTemporaryStructure(
         final Level world = colony.getWorld();
 
         final int y = BlueprintTagUtils.getNumberOfGroundLevels(structure, 4) - 1;
-        final BlockPos spawnPos = targetSpawnPoint.offset(0, -y, 0).offset(structure.getPrimaryBlockOffset());
+        final BlockPos spawnPos = targetSpawnPoint.below(y).above(structure.getPrimaryBlockOffset().getY());
```
Raycoms, line 96: wasn't this on purpose to make the main spawn-point the primary offset of the structure?

uecasm: That was probably the intent, but that's what my version _actually_ does.  The schematic loads around its anchor point anyway, so we just need to line up the `y` properly and it's already correct.

Raycoms: ahhh, ok

Raycoms (approved):

someaddons (approved):

PR #8601, opened 2022-08-30
Override `canBeLeashed` flag for civilian entities

### LINKED ISSUE #713: Civilian entities should override `canBeLeashed`
### Prerequisites 

- [X] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure that my issue is not covered there.
- [X] I made sure that **this change request is not a duplicate of any existing change request.**

### Context 

- Minecraft Version: 1.18 (1.19 is probably also affected)
- MineColonies Version: [ad5fce1dd2d0fc64fbee521d29f9a4def0ae00ff](https://github.com/ldtteam/minecolonies/commit/de35782c4ca6aaa932a9f9e80fd4143138649578) built from source
- Related Mods and their Versions: -

Civilian entities inherit from `Mob`. Mobs are generally assumed to be "leashable" unless they are enemies. This behavior of the `boolean canBeLeashed(Player)` method should be overridden for mobs for which this is not the case. Vanialla villages (`AbstractVillager`) is an example of a type of mob that overrides this behavior. Currently, this check is not overridden for civilian entities. Consequentially, code using the check to determine the behavior of an entity assumes that the civilians can be leashed.

In practice, this is not a problem when running with MineColonies only because the interaction is consumed when a GUI is opened. However, this flag may be used by other mods to determine how an entity should behave. For example, I published a tiny QoL mod a while back that blocks entities that can be leashed from passing through fence gates if are not on a leash. Unless the behavior for an entity type is explicitly configured, the mod tries to interfere the correct behavior using this check. See https://github.com/TheMrMilchmann/TheFenceUnleashed/issues/2 for further context.

I propose that civilian entities override this check for better interop with other mods. An implementation is provided in https://github.com/ldtteam/minecolonies/pull/8601.

---
## Viewers

TheMrMilchmann:
This PR adds a simple override for the `boolean canBeLeashed(Player)` method to `AbstractCivilianEntity` that always returns false.

In practice, this is not a problem when running with MineColonies only because the interaction is consumed when a GUI is opened. However, this flag may be used by other mods to determine how an entity should behave. For a specific use case, see https://github.com/ldtteam/minecolonies-features/issues/713.

Closes ldtteam/minecolonies-features#713

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/citizen/AbstractCivilianEntity.java
+++ b/src/api/java/com/minecolonies/api/entity/citizen/AbstractCivilianEntity.java
@@ -87,6 +87,12 @@ public abstract class AbstractCivilianEntity extends AgeableMob implements Npc, 
     public abstract void setCitizenId(int id);
 
     @Override
+    public boolean canBeLeashed(Player player)
+    {
+        return false;
+    }
+
+    @Override
     public boolean canBeStuck()
     {
         return canBeStuck;


Thodor12:
Can you add the #closes to your PR description so it'll indicate that it will auto close your linked issue?

TheMrMilchmann:
Sure, done!

marchermans (approved):

someaddons (approved):

TheMrMilchmann:
Quick follow-up question: Will this be merged into 1.19 during your development flow at some point or should I open a separate PR for that?

PR #8607, opened 2022-08-31
Shape and supply tool updates

uecasm:
- Compatibility with ldtteam/Structurize#530
- Allow build tool "assign to builder" to actually assign to builder as a decoration again, instead of always treating it as a special item paste.
- Moves supply camp/ship style selection to the top of the screen to avoid overlapping the settings menu.

Review please

~~(Draft pending actual Structurize build number)~~


Raycoms (dismissed):
Could you just check onto the settings location, it's off here too now

uecasm:
> Could you just check onto the settings location, it's off here too now

Done.

PR #8657, opened 2022-09-16
Fix waypoints being invisible in the build tool

MotionlessTrain:
Closes #
Closes #
Closes #

- Fix the waypoints not being selectable in the build tool.
  They were in Minecolonies Original -> infrastructure -> roads, together with a few sub folders (with types of roads).
  I moved them to Minecolonies Original -> infrastructure -> misc, as they aren't really roads.

-
-

Review please


Raycoms:
This is being queried for waypoint rendering as well, gotta rename that too

MotionlessTrain:
Good catch! I didn't know how that part worked.
I changed those as well

PR #8695, opened 2022-10-04
Cavern changes

MotionlessTrain:
Closes #
Closes #
Closes #

- Remove a creative-only block from the Cavern megahall level 5 university
- Add little gardens to the Cavern supply chests
- Add the correct Cavern megahall apiary. Somehow the older one got added to the mod
  The older one will still be available as alternative apairy.
  NOTE that for users who already built this, they may need to deconstruct their apiary
    to switch to the alternative if they want to keep in on the floor,
    or move it to the ceiling with the newer megahall apiary
- Port of the fixes in #8699

Review please


MotionlessTrain:
One more change coming

PR #8707, opened 2022-10-13
stop cheating Shindri

Raycoms:
Closes #
Closes #
Closes #

- No free custom camps
-
-

Review please


uecasm (approved):

Moobien:
what is to stop her doing it with the supplyship/camp from an addon style like steampunk etc

MotionlessTrain:
> what is to stop her doing it with the supplyship/camp from an addon style like steampunk etc

I mean, if you really want to, you can just make your own style, and in single player, it is easy enough to cheat anyway.

This is mostly for servers, where all style packs (with the exception of the personal ones) are on the server. Players can't edit the ones on the server (unless they are the server owner, but in that case the earlier statement applies)

PR #8712, opened 2022-10-15
upgrade children > parent

Raycoms:
Closes #
Closes #
Closes #

- Allow upgrading child when parent building is maxed (e.g. tavern level 3)
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractBuilding.java
@@ -803,7 +803,7 @@ public abstract class AbstractBuilding extends AbstractBuildingContainer
         {
             requestWorkOrder(WorkOrderType.BUILD, builder);
         }
-        else if (getBuildingLevel() < getMaxBuildingLevel() && (parentBuilding == null || getBuildingLevel() < parentBuilding.getBuildingLevel()))
+        else if (getBuildingLevel() < getMaxBuildingLevel() && (parentBuilding == null || getBuildingLevel() < parentBuilding.getBuildingLevel() || parentBuilding.getBuildingLevel() >= parentBuilding.getMaxBuildingLevel()))
         {
             requestWorkOrder(WorkOrderType.UPGRADE, builder);
         }


uecasm:
I know what motivated this, but I'm not sure it's a useful feature.  Tavern level 3 is usually not actually constructed with "level 3" materials -- despite it only actually counting as level 3 housing, it does still produce high-stat visitors and most styles use level 5 materials.  (i.e. build-wise, the tavern is treated as 1-3-5 and not 1-2-3.)  As such people will typically not build the level 3 tavern until they're at level 4 or 5 everything else.

This in turn means that either this style is "doing it wrong" or all the other styles are.

There is one other context where this *might* be reasonable -- where the parent is a decoration that wants to be upgradable but only have two levels.  But for that case I would argue that it still doesn't make sense to limit the child buildings to level 1 until the parent is actually upgraded.  So perhaps a better solution for this might be to have a tag in the parent which disables the level restriction entirely, for cases when the style designer decides it is more harmful than helpful.  Rather than this.

MotionlessTrain:
> This in turn means that either this style is "doing it wrong" or all the other styles are.

Note that this is not the only restaurant in Shire, fortunately. So it doesn't matter *that* much that it may be limited and to that point, it is not that much of an issue.
If you want to have a level 5 restaurant, you can build a second "normal" version

Raycoms:
> I know what motivated this, but I'm not sure it's a useful feature. Tavern level 3 is usually not actually constructed with "level 3" materials -- despite it only actually counting as level 3 housing, it does still produce high-stat visitors and most styles use level 5 materials. (i.e. build-wise, the tavern is treated as 1-3-5 and not 1-2-3.) As such people will typically not build the level 3 tavern until they're at level 4 or 5 everything else.
> 
> This in turn means that either this style is "doing it wrong" or all the other styles are.
> 
> There is one other context where this _might_ be reasonable -- where the parent is a decoration that wants to be upgradable but only have two levels. But for that case I would argue that it still doesn't make sense to limit the child buildings to level 1 until the parent is actually upgraded. So perhaps a better solution for this might be to have a tag in the parent which disables the level restriction entirely, for cases when the style designer decides it is more harmful than helpful. Rather than this.

I mean what matters here is not if the parent uses level 5 materials, but that the child in the tavern has sufficient complexity that it is not cheaty, but in the end that's sth that we have to review anyway

PR #8736, opened 2022-10-30
Feature/1.19 stats

Raycoms:
Closes #
Closes #
Closes #

- Add a way to track statistics in a colony
- Add a few statistics to some jobs
- Add display to the townhall UI

Future plans: Add more fine grained statistic display opportunities to the UI

Review please


src/main/java/com/minecolonies/coremod/colony/managers/StatisticsManager.java
```diff
@@ -0,0 +1,154 @@
...
+    public void increment(final @NotNull String id)
+    {
+        incrementBy(id, 1);
+    }
+
+    @Override
+    public void incrementBy(final @NotNull String id, int qty)
+    {
+        final Map<Short, Integer> innerMap = stats.computeIfAbsent(id, k -> new HashMap<>());
+        innerMap.put(colony.getDay(), innerMap.getOrDefault(colony.getDay(), 0) + qty);
```
someaddons: I'm not sure if I'm happy with the data structure, using a map of SHort,Integer means you always have to do unboxing for both values, which isnt really great performance wise, and increment should be a very fast operation.
I also feel we shouldnt keep track of all data, but just last 30? days. Wonder if we could use a limited size array as data structure then, or at least use Long/Short specialized collections to avoid so much unboxing

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -1090,7 +1107,7 @@ public void removeFreeBlock(@NotNull final Block block)
      * Any per-world-tick logic should be performed here. NOTE: If the Colony's world isn't loaded, it won't have a world tick. Use onServerTick for logic that should _always_
      * run.
      *
-     * @param event {@link TickEvent.LevelTickEvent}
+     * @param event {@link TickEvent.WorldTickEvent}
      */
     @Override
     public void onWorldTick(@NotNull final TickEvent.LevelTickEvent event)
```
uecasm: wat?

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractEntityAICrafting.java
```diff
@@ -471,6 +472,7 @@ public IAIState afterDump()
             // Fallback security blanket. Normally, the craft() method should have dealt with the request.
             if (currentRequest.getState() == RequestState.IN_PROGRESS)
             {
+                worker.getCitizenColonyHandler().getColony().getStatisticsManager().incrementBy(ITEMS_CRAFTED, job.getMaxCraftingCount());
```
uecasm: This can't be right, it's guaranteed to be adding 0 at this point.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/cook/EntityAIWorkCook.java
```diff
@@ -214,6 +215,7 @@ private IAIState serveFoodToCitizen()
                     }
                 }
             }
+            worker.getCitizenColonyHandler().getColony().getStatisticsManager().increment(FOOD_SERVED);
```
uecasm: Shouldn't this be only when `stack.isEdible()` and hence food was actually served?
(Side note: if the stack is _not_ edible due to some mismatch or race condition, doesn't that code destroy an item?)

src/main/java/com/minecolonies/coremod/entity/ai/citizen/cook/EntityAIWorkCook.java
```diff
@@ -229,6 +231,7 @@ else if (InventoryUtils.hasItemInItemHandler(handler, stack -> CAN_EAT.test(stac
             removeFromQueue();
             return getState();
         }
+        worker.getCitizenColonyHandler().getColony().getStatisticsManager().increment(FOOD_SERVED);
```
uecasm: Meanwhile this only counts by one "serving event", not the number of food actually served.  Which might be unexpected.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/RangerCombatAI.java
```diff
@@ -356,5 +356,6 @@ protected void onTargetDied(final LivingEntity entity)
     {
         parentAI.incrementActionsDoneAndDecSaturation();
         user.getCitizenExperienceHandler().addExperience(EXP_PER_MOB_DEATH);
+        user.getCitizenColonyHandler().getColony().getStatisticsManager().increment(MOBS_KILLED);
```
uecasm, line 359: The nether miner should probably have one of these too.  Unless you want to count that separately.  Lots of possibly interesting expedition counters.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/miner/EntityAIQuarrier.java
```diff
@@ -702,4 +704,15 @@ public BlockState getSolidSubstitution(final BlockPos ignored)
...
+
+    @Override
+    protected void triggerMinedBlock(@NotNull final BlockState blockToMine)
+    {
+        super.triggerMinedBlock(blockToMine);
+        if (IColonyManager.getInstance().getCompatibilityManager().isOre(blockToMine))
+        {
+            building.getColony().getStatisticsManager().increment(ORES_MINED);
+        }
+        building.getColony().getStatisticsManager().increment(BLOCKS_MINED);
```
uecasm, line 716: I wonder if these should be mutually exclusive.  Technically a mined ore is also a mined block, but for the purposes of showing these one after the other in a stats list I think most people would assume they're not overlapping.

src/main/java/com/minecolonies/coremod/colony/Colony.java
Raycoms: oopsy forward port

src/main/java/com/minecolonies/coremod/entity/ai/citizen/cook/EntityAIWorkCook.java
Raycoms: Yeah, I thought a bunch about this, but there is no "nondisruptive" way to count this else (without significantly reworking the existing code). So it's "Food served"

src/main/java/com/minecolonies/coremod/entity/ai/citizen/guard/RangerCombatAI.java
Raycoms: Separately probably

src/main/java/com/minecolonies/coremod/entity/ai/citizen/miner/EntityAIQuarrier.java
Raycoms: Doesn't matter a lot tbh, the difference is gonna be so large, it's barely noticeable.

src/main/java/com/minecolonies/coremod/entity/ai/citizen/cook/EntityAIWorkCook.java
uecasm: Could change `transferFoodUpToSaturation` to return number of items served, with 0 being equivalent to the current false return.  Or return the `foundSaturation` directly and have the stat be "food shanks gained from eating" or shorter words to that effect.

src/main/java/com/minecolonies/coremod/colony/managers/StatisticsManager.java
Raycoms: Yeah, I wondered about this too, but I can't come up with anything more efficient. Could obviously insert a tuple, but then we'd have to do a lot more work on value read.

MotionlessTrain: Not sure how large those values get, but Integers <= 127 (and likely the same for Shorts) get cached, which saves the creation of objects at least.

uecasm: Not sure if they're loaded in-game or not, but there's a bunch of classes like `Short2LongSortedMap` in the JDK that try very hard to avoid boxing.

Or as someaddons suggested, you could use a `long[]` instead of a map, either as a circular buffer with "current day index" or shifting all the cells on day rollover so the current day is always `[0]`.  When a day falls off the end, you either discard it or you add it to a total rollup.

Another option is to use an enum key for your parent map instead of a string, which might be a little faster (and still supports stringification for serialization).

Raycoms: Wanted to avoid enum keys for easy extension, I'll look into Short2IntOpenHashMap

marchermans (dismissed):

src/main/java/com/minecolonies/coremod/colony/managers/StatisticsManager.java
uecasm: A sorted map might be more appropriate for this case, as I suggested.  Your day indexes are likely contiguous, it has methods to easily fetch a subrange of data, and if you ever want to do something like graph it you're going to want it to be ordered anyway.  (And if you don't want to do something like that at some point, then why separate it in the first place?)

src/main/java/com/minecolonies/coremod/colony/managers/StatisticsManager.java
```diff
@@ -51,14 +52,14 @@ public void increment(final @NotNull String id)
     @Override
     public void incrementBy(final @NotNull String id, int qty)
     {
-        final Map<Short, Integer> innerMap = stats.computeIfAbsent(id, k -> new HashMap<>());
+        final Map<Short, Integer> innerMap = stats.computeIfAbsent(id, k -> new Short2IntOpenHashMap());
```
uecasm: You don't get most of the performance benefits unless you access it as its own class and not the base `Map<>` class.  Otherwise you're still (un)boxing.

src/main/java/com/minecolonies/coremod/colony/Colony.java
```diff
@@ -1902,10 +1925,16 @@ public CitizenNameFile getCitizenNameFile()
...
     public boolean isTicketedChunksDirty()
     {
         return ticketedChunksDirty;
     }
+
+    @Override
+    public short getDay()
+    {
+        return day;
+    }
```
uecasm: This might be best changed to an integer anyway.  On a server, chunk-loaded, a colony will eat through 65k days in a little under 2 years real time.  Which may sound like a lot, but given that it is intended to take several months to play a colony through, it wouldn't surprise me if some people are still playing colonies that long.  Certainly I still have some colonies that are that old, although they haven't been running continuously and so wouldn't have counted that far yet.

uecasm: Actually, 11 months, when using a signed short...

src/main/java/com/minecolonies/coremod/colony/managers/StatisticsManager.java
Raycoms: done

Raycoms: sth

someaddons (dismissed):

PR #8750, opened 2022-11-04
Make it easier to teach crafters how to Ornamentum

uecasm:
- Adds a "requests" button to the custom DO crafting "teach" page.
    - Clicking this displays a flat list of all unhandled (same rules as clipboard) requests for DO blocks:
        image
    - It only shows blocks that can be learned by this crafter.
    - Clicking the Select button will populate the previous page with the matching blocks:
        image
        (you can see the new button in the top right of this screenshot too)
    - At this point the player just needs to click Teach to actually learn the recipe

Review please (could be backported)


src/main/java/com/minecolonies/coremod/client/gui/modules/DOCraftingWindow.java
```diff
@@ -69,12 +86,132 @@ public DOCraftingWindow(final IBuildingView building, final CraftingModuleView v
...
         registerButton(BLOCK3_ADD, this::addBlock3);
         registerButton(ADD, this::addRecipe);
+        registerButton(BUTTON_REQUEST, this::showRequests);
+    }
+
+    /**
+     * Creates a new crafting module of the correct type but not connected to the building (since we can't get the
+     * real module from the view, and we're on client side anyway.  But this is safe because JEI does it too).
+     */
+    private ICraftingBuildingModule createFakeCraftingModule()
```
Raycoms: I don't like this, but there is probably no other way that wouldn't require a big rework. 
Could we abstract the "is this recipe a match" stuff into a different part of the code that we could use on client and server alike? So that we can get it from the CraftingBuildingModuleView too?

uecasm: Well, a predicate can't be serialized across, nor can the results of the predicate without generating a set of candidate items.  Which is hard for DO items.

*Most* of the logic for matching is already in a shared function.  Could make it statically callable, but then you'd still need to know the class type, so instance polymorphic is best.  And currently a common class is used on the view side and not polymorphic by building type, nor does the view itself have access to call the underlying module class (nor the module having the opportunity to pass data to the view constructor, due to how they're constructed).  So I'm not really seeing any nice alternatives here.  If you can think of something, I'm all ears.

I _think_ the only reasonably viable option would be to check all the requests on server side and sync across their ids whenever the set of possible matches changes... but it would probably have to do that even if nobody is looking at the GUI, which is a bit ugly.  Still, I can give that a try.

uecasm: Done.  Though it does slow things down a bit, but *most* of the time you're going to take a bit of time to notice the request at the other hut and then come over to your crafter to teach it, so it should be fine.

Side note: simply logging in does not add a player to close subscribers so they do not receive any building view updates.  They need to cross chunk boundaries before they become a subscriber.  Again probably not a problem in practice, but it may occasionally cause confusion.

uecasm: Now switched back to client side (no latency, other than existing request syncs) and using static validator parameter.

Raycoms (approved):

someaddons (approved):

PR #8759, opened 2022-11-09
But I *just* cleaned the floor!

uecasm:
Closes discord request

- Adds recipes for mud
    - Farmer inherently knows how to make mud from dirt and water bottles
    - Farmer can be taught how to make packed mud and mud bricks
    - Baker can no longer be taught how to make packed mud (mud pies are not very tasty, despite what the kids say)
- Sawmill can be taught how to add chests to boats, because why not
- JEI now shows that the plantation is a source of bamboo, sugar cane, and cactus

Review please (do not backport; at least not in current form)

[Before/after spreadsheet](https://docs.google.com/spreadsheets/d/1DaGFlFcnce4bHXHylzztr5_Rwnm3P42c6o_xXqgyyr8/edit?usp=sharing)
Another change from that spreadsheet we might want to consider is which of the new `enchanted_book`s we should add to the enchanter... or perhaps come up with a system to do this automatically so that it supports modded enchants too (without needing a datapack)...

### DIFF
--- a/src/datagen/generated/minecolonies/data/minecolonies/tags/items/baker_product_excluded.json
+++ b/src/datagen/generated/minecolonies/data/minecolonies/tags/items/baker_product_excluded.json
@@ -3,6 +3,7 @@
     "minecraft:bread",
     "minecraft:cake",
     "minecraft:cookie",
-    "minecraft:pumpkin_pie"
+    "minecraft:pumpkin_pie",
+    "minecraft:packed_mud"
   ]
 }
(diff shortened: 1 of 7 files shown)


MotionlessTrain:
A lot of the newer enchantments are not available via the enchantment table, probably on purpose. For those, either bartering is required (soul speed), or exploring ancient cities (1.19 ones). Not sure if we should include those in the enchanter's spell book

PR #8762, opened 2022-11-10
Crafting tools

uecasm:
Closes ldtteam/minecolonies-features#719

- Allows `crafterrecipes` to specify a `tool` which is required to perform the craft.
    - The worker will request the tool if they don't have one; it follows the usual rules of min/max tool levels.
        - They won't attempt to craft a different recipe while waiting for the tool (unless some other recipe becomes "first fulfillable" in the meantime) -- i.e. having the tool is not considered a fulfillable precondition, unlike tools in the recipe grid itself.
    - While crafting, the worker will hold the tool.
    - It will take durability damage (or not, due to research/enchants) as items are crafted.
        - I haven't tested this, but I suspect if the tool breaks during the craft they might end up over-crafting, since they stop the craft there, the already-crafted items are not necessarily picked up for a while, and the original request (and count) will be restarted once they get another tool.  Not sure if there's a good way to fix this, but it should otherwise be mostly harmless.
- The required tool (if any) is shown in JEI and listed in the hut recipe list.
- Changes the forester's log stripping recipes to require an axe.
- Changes the farmer's pumpkin carving recipe to require shears.
- Adds recipes to deoxidize copper blocks (and cut copper, at level 2+) to the mechanic
    - Also gives them a few bonus recipe slots to offset the additional recipes
    - _Only_ the plain copper blocks can be deoxidized -- this doesn't support the full range of items, such as doing stairs and slabs too.
    - Waxing blocks is not inherent but is teachable (stonemason) as before.
    - Currently nobody has or can learn recipes to scrape off wax (but adding those might cause some duplicate output ambiguity or crafting loops?).

Review please (probably could backport)


src/main/java/com/minecolonies/coremod/client/gui/modules/WindowListRecipes.java
```diff
@@ -196,11 +196,17 @@ public void updateElement(final int index, @NotNull final Pane rowPane)
...
+                intermediate.setVisible(false);
+                if(recipe.getRequiredTool() != ToolType.NONE)
+                {
+                    intermediate.setText(recipe.getRequiredTool().getDisplayName());
+                    intermediate.setVisible(true);
+                }
+                else if(recipe.getIntermediate() != Blocks.AIR)
                 {
                     intermediate.setText(recipe.getIntermediate().getName());
+                    //intermediate.setVisible(true);
```
Raycoms, line 208: uncommented?

src/api/java/com/minecolonies/api/crafting/IRecipeStorageFactory.java
```diff
@@ -74,6 +76,11 @@ default RecipeStorage getNewInstance(@NotNull final IFactoryController factoryCo
             throw new IllegalArgumentException("Ninth parameter is supposed to be a ResourceLocation or Null!");
         }
 
+        if (context.length > MIN_PARAMS_IRECIPESTORAGE + 6 && context[9] != null && !(context[9] instanceof IToolType))
```
Raycoms, line 79: Gotta make sure this loads old worlds well too

Raycoms (dismissed):
Just two minor comments

src/main/java/com/minecolonies/coremod/client/gui/modules/WindowListRecipes.java
uecasm: I left it in but commented because the previous code here would never have set it to true.  I assumed this was intentional.

src/api/java/com/minecolonies/api/crafting/IRecipeStorageFactory.java
uecasm: Existing worlds don't use this constructor, they use the one that restores from NBT.  Then it will also build the new recipe (with tool) from the datapack and the auto-recipe code will replace the old recipe with the new one.

In testing, this just works.  Unless you can think of something I might have overlooked?

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeFactory.java
```diff
@@ -33,15 +31,24 @@ public class CustomRecipeFactory implements IFactory<FactoryVoidInput, CustomRec
 {
     private final static String CUSTOM_RECIPE_ID_PROP = "id";
 
+    private static boolean checkContext(final Object[] context, final Class<?>... types)
```
marchermans: Is this not a duplicate from the other one you added?

src/main/java/com/minecolonies/coremod/client/gui/modules/WindowListRecipes.java
Raycoms: was probably more of a derp tbh. should be true here and then false in the else =D

uecasm: I also figured that it was kinda unnecessary now that _most_ of the smelting recipes are on a separate tab (though there are a very few exceptions).  Since all this would have done is shown "Furnace" or blank.

src/main/java/com/minecolonies/coremod/colony/crafting/CustomRecipeFactory.java
uecasm: Good catch.  In fact neither of these ended up being used -- I had some early code to support distinguishing between multiple sets of different parameter types, but in the end it was simpler to just change all the callers to use the new args.

In an ideal world, they should probably just use a regular constructor instead of the type-erased factory anyway.

Shadizar:
Sorry to be a bother...in vanilla one uses shears to carve a pumpkin, just checking if your variance is intentional?

uecasm:
> Sorry to be a bother...in vanilla one uses shears to carve a pumpkin, just checking if your variance is intentional?

It wasn't; thanks for the catch.

Mekle001 (approved):

PR #8782, opened 2022-11-24
The goggles, they do nothing

uecasm:
- Refactors the neighbor-building rendering to be a little more generic
    - Also fixes a bunch of things that seemed like bugs, though I didn't specifically verify in prior testing.  For example it looked like if the blueprint future was particularly slow to complete, it would keep queueing up infinite requests.  And if any blueprint failed to load it would not render any subsequent.
- Adds the Builder's Goggles, which use this new implementation
     - image
     - image
     - They're a head-slot item which provides no armor at all.
     - They're also very fragile, and can't be repaired -- the glass and mechanisms are quite delicate, so only put them on when you really need to, and stay out of trouble while wearing them.
     - They are fairly well enchantable though, so you might be able to improve their durability a bit that way.
     - While worn, the goggles will display the bounding box and blueprint preview for any nearby active work orders in the colony.
         - There's a client config setting to specify how close is "nearby".
         - This includes both buildings and decorations.  It will also render (box only) removal orders, and where the miner is currently working.
         - If you're holding shift, it will also show a box around the anchor (if it has one), helping you to find it again if you lose track of it.
             - This also highlights (only) the anchor of placed but as yet unbuilt buildings, which can help find hut blocks lost underground.

I considered giving this a research requirement or making it only colonist-craftable, but I figure it's going to be quite useful in early game too.  I tried to make the cost a little expensive but not especially so, with the balancing caveat that it will break easily and can't be repaired.

I also considered implementing these as a Curio (as optional dependency), but I figured that would be more likely for people to wear them all the time, and would likely avoid the damage penalty.

Currently the goggles will work in any colony, without any permission requirements.  I couldn't decide if perhaps they should be restricted.


Raycoms (dismissed):

Raycoms:
Was wanting something like this for a while. It's a nice immersive way to see things. Cost/destroy should be enough from a balance perspective ye.

someaddons (dismissed):

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
```diff
@@ -0,0 +1,381 @@
...
+
+        if (activeRules.isEmpty())
+        {
+            blueprintCache.clear();
+            alreadyRequestedStructures.clear();
+            lastCacheRebuild = null;
+            return;
+        }
+
+        if (lastCacheRebuild == null || Minecraft.getInstance().level.getGameTime() % 20 == 0)
```
Nightenom: lastCacheRebuild is not used at all btw

imho this should be something like: if moved horizonzally (vertical is irrelevant?) for at least cca 8 to 10 blocks or else if 30 secs have passed
since you want to prefer the move rule, the time rule is really minor probability - work orders don't change that often (afaik)

Nightenom (dismissed):
otherwise kk

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
uecasm: This part was carried over from the previous implementation (which is hard to see on GH, although I did tag it as a rename on my end).

It's not _quite_ unused, it's acting like a glorified boolean at present.  I suspect it did previously have some kind of min movement distance check like you suggested.

The pure time basis is sort of necessary, because this is also how often it processes its queue of pending blueprints -- it only updates at most one per timeslot, since that potentially includes a rotate/mirror operation (might be nice if that happened on the I/O thread instead....)  That part is also the same as before, the only thing I changed is that it can now process more than one queue slot if it didn't actually load a blueprint.

Nightenom: loading queue should be disconnect from update rules, usually updating rules (= generating list of visible/loadable blueprints) requires (a lot?) searching and thus you dont want to run them so often - also they are position based, so it makes really small sense (for me) to make them time dependent (work order state change is only thing i can see happening - when you stand at the same pos)

Nightenom: > which is hard to see on GH, although I did tag it as a rename on my end

afaik you cant tag something as rename, it should be just diff percent threshold

> That part is also the same as before, the only thing I changed is that it can now process more than one queue slot if it didn't actually load a blueprint.

well, Ray obviously (just like always :D ) doing things that are weird
the old impl was `if (lastCacheRebuild == null || !lastCacheRebuild.closerThan(activePosition, CACHE_RESET_RANGE))`
but Ray changed it to current one during blueprint loading update

uecasm: > afaik you cant tag something as rename, it should be just diff percent threshold

Yeah, that's what I meant, it was showing as a (delete+add =>) rename in the commit preview.  Looking at the history it seems it decided it wasn't after all, which is mildly annoying.  If I'd noticed that sooner I might have forced it via a helper commit, though it's not super important I guess.

Pushing a change in a minute that splits the cache regen from the queue processing, and cleans up a few other stray bits.  Slight regression in this version is that on login while wearing the goggles you won't see anything until you move past the threshold distance, though that's not too terrible.

marchermans (approved):

PR #8783, opened 2022-11-24
Dynmap compatability

Thodor12:
- Ability to render borders of colonies on Dynmap using a new marker set "Colonies".
- Renders the colony name, owner and population count on the description of the colony.
- Automatically updates Dynmap when the following occur:
  - The colony name gets changed
  - The colony it's team color is changed
  - The colony borders change when a building is constructured/repaired/removed
  - The colony it's population count changes, either when colonists are born or die.

What is not added in the pull request:
- Ability for Minecolonies specific blocks (primarly hut blocks) to render on Dynmap
- Ability for DO blocks to render on Dynmap (this would have to be implemented from within DO, should not appear here)

Review please

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -11,6 +11,10 @@ repositories {
         name "JourneyMap"
         url "https://jm.gserv.me/repository/maven-public/"
     }
+    maven {
+        name "Dynmap"
+        url = "https://repo.mikeprimm.com/"
+    }
 }
 
 dependencies {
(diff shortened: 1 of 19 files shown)


uecasm:
You actually did want `version/main`.  `version/1.18.2` is ooooooold.

Thodor12:
Darn, I thought version/main was updated to 1.19.2. I'll see if I can merge it into main

gradle/dependencies.gradle
```diff
@@ -7,6 +7,10 @@ repositories {
         name "JourneyMap"
         url "https://jm.gserv.me/repository/maven-public/"
     }
+    maven {
+        name "Dynmap"
+        url = "https://repo.mikeprimm.com/"
+    }
```
uecasm, line 17: Quoth the properties file,
> Dynmap doesn't have a maven

Thodor12: I'll edit that comment a bit to state explicitly they do not have release artifacts stored on Maven, only API artifacts

Thodor12:
I'll reopen since that swapping base caused some issues

PR #8796, opened 2022-12-02
Minimum Order Quantity research

uecasm:
Closes ldtteam/minecolonies-features#733

- Adds Minimum Order Quantity research
    image
    image
    - When unlocked, buildings will not immediately request replacement items when they fall below their set Min Stock level -- instead they will wait until they fall below 75% of the Min Stock level (and then order enough to return to 100%).  So for example if the Min Stock is 64 items then it will wait until it has 47 items or fewer before ordering enough to get back to 64.
        - The order will not cancel if the level rises above 75% but still less than 100%, so if you take some items out at just the wrong time before putting some but not all back again, then you might get an overstock.
    - And yes, this _will_ steal your colony-configured clipboard if you have it on you.  Just go make another one.

Review please (could be backported)

Open to alternative research costs/location/descriptions.

~~Also yes, technically "minimum order quantity" is the wrong term for what this is doing (since the actual order could still be only a small number of items), but it sounds cool.~~


AntoniBilicki:
I suggest not locking this behind research - maybe instead add a research that allows configuring the % at which the restock occurs?
#Edit: My reasoning is that this should be the default restock functionality, as it matches the expected behavior much more closely

uecasm:
Having research to make an existing mechanic work "better" is kinda what research is for -- and it's a level 3 research so it's not super hard to get to.  A lot of similar things (builder modes, warehouse stock-based priority) are also at level 3.

Mekle001 (approved):

PR #8799, opened 2022-12-02
Periodically rebalance courier-claimed requests

uecasm:
Closes ldtteam/minecolonies-features#708

- Periodically checks task queue of couriers and moves tasks to a different courier if that seems like a good idea (only within a single warehouse, though)
    - Notably, if you pause a courier then the others will steal all their tasks; when you unpause then they'll steal some of them back again.
    - The normal scoring/ordering rules are followed; and a task will never be moved to a later queue position than it already is in (unless the worker isn't working).

I'm not entirely sure I've covered all the bases here, but I have tested what I can think of and it appears to behave itself reasonably.


src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/DeliverymenRequestResolver.java
```diff
@@ -203,4 +201,64 @@ public MutableComponent getRequesterDisplayName(
...
+        final Set<IRequest<?>> alreadyMoved = new HashSet<>();
+        final List<ICitizenData> workers = getResolveAbleDeliverymen(manager);
+        for (final ICitizenData worker : workers)
+        {
+            final JobDeliveryman workerJob = (JobDeliveryman) worker.getJob();
+            final List<IToken<?>> taskQueue = workerJob.getTaskQueue();     // assumed immutable copy, not live list
+            final boolean isWorking = worker.isWorking() && !worker.isPaused();
+
+            // don't attempt to reassign the current task unless the worker isn't actually working
+            for (int requestIndex = isWorking ? 1 : 0; requestIndex < taskQueue.size(); ++requestIndex)
```
Raycoms, line 220: Seems complicated, why not just if notWorking, return?

Raycoms: I think rebalancing 1 task every 25seconds is also okay, tasks usually need around that to be done anyway

uecasm: > Seems complicated, why not just if notWorking, return?

Because we **do** want to go through all the requests of the non-working dmen.  We just skip the first request of the working one because they're probably out doing it right that moment and we don't want to rugpull them.

uecasm: > I think rebalancing 1 task every 25seconds is also okay, tasks usually need around that to be done anyway

I was more sort of talking about how it seemed a little odd that the whole `colonyStateMachine` thing appears to be set up to do each task on a different tick to spread out the processing, but then `worldTickSlow` calls about fifteen different things all on the same tick...

someaddons: its because inherently worldTickSlow only had light updates... and the biggest point is to spread all colonies from doing it on the same tick... Also do you see how using ... gives a neat impression of passive agressiveness?... I'd suggest we do not use them...

someaddons (approved):

src/main/java/com/minecolonies/coremod/colony/requestsystem/resolvers/DeliverymenRequestResolver.java
Raycoms, line 220: This is a problem with the courier prefetching (courier might be handling multiple request deliveries at the same time).

uecasm: That's why it has the condition to only move tasks that don't have matching source+destination pairs.  This also avoids splitting up shareable tasks even if they're not currently active.

Raycoms: Ah okay, Ive been thinking a bit more about this and thought:
- Not rebalance ALL courier tasks but rather only rebalance if there are gross queue size differences
and Only rebalance a queue when more than x items (or x complexity)
- When we rebalance, not rebalance all at the same time but rather something like the bottom 10% 

- When triggering a pickup, actually do an estimate on how much has to be picked up and give a pickup task a "complexity" number which we then, when comparing queue sizes on assignment can use to balance tasks better.

uecasm: idk, I think that might be overcomplicating it.  It's very simple atm, and self-adjusting without any need to calculate complexities manually.

It does have rules explicitly to prevent tasks bouncing between couriers repeatedly.

Raycoms: The complexities would then be task in general, and we would use them for the actual normal balancing already to reduce the need to rebelance so much

uecasm:
Superseded by #9969.

PR #8802, opened 2022-12-04
Search for buildings

uecasm:
- ~~Bumps structurize (the same as the other PR so hopefully it won't get too upset about conflicts)~~
- Adds a building browser / search.
    - With a hut block in hand, right-click in air:
        image
    - After about 5-10s (might be faster/slower depending on CPU, disk, and number of installed style packs) it loads:
        image
    - This shows the style pack, path, size, and level of all blueprints containing that hut.
        - Main building blueprints are in black
        - Parent building/deco blueprints (i.e. those that have that hut as a child) are in green.
        - Invisible buildings are hidden (in survival) or shown in blue (in creative).
    - The search is carried out purely client-side.
    - The search is cached (per login session) so if you reopen the window (even for a different hut) it will load immediately, but after logout it will take time the first time you search again.
        - This does mean that (unlike before caching) it won't update with recently-scanned buildings, but that's probably no great loss.

Note that the search has to go through and load all blueprints, though ~~it ends up discarding most of the info.  (So searching again with the same or different building still takes the same time.)  Performance could potentially be improved in future with background caching, or by having some build process store additional metadata in the pack json (at least for the in-mod styles).  However 10s on background threads doesn't seem too terrible to start with.  And~~ currently you don't pay any time if you don't use the feature.


src/main/java/com/minecolonies/coremod/client/gui/WindowBuildingBrowser.java
```diff
@@ -0,0 +1,345 @@
...
+
+        // to reduce total search time, kick work off to several worker threads (currently 4 threads ~= 5s on a decent CPU and slow disk)
+        final ExecutorService packPool = Executors.newFixedThreadPool(WORKER_THREADS, runnable ->
+        {
+            final Thread thread = new Thread(runnable, "Minecolonies Building Browser Worker");
+            thread.setDaemon(true);
+            thread.setUncaughtExceptionHandler((thread1, throwable) -> Log.getLogger().error("Minecolonies Building Browser errored! ", throwable));
+            return thread;
+        });
+        final Map<StructurePackMeta, Future<List<BuildingInfo>>> packFutures = StructurePacks.getPackMetas().stream()
```
Raycoms: Couldnt we use the IO threadpool minecraft already has here?

uecasm: Possibly.  I didn't find one when I went looking, but perhaps I just overlooked it.  Do you recall where it is?

uecasm: Still not seeing anything on client side.  There is one on server side, but I'm deliberately not going there.

Raycoms: public final class IOPool in Structurize should also do the trick tbh

uecasm: The top level does run on structurize's IOPool.  But that's only a single thread, so it couldn't fork for the subtasks, it just got wedged.

Nightenom: Minecraft class itself used to be an executor (like at least two versions ago it was), also there was something in util(s) class

uecasm: That's not a threadpool, that makes it do work on the client/render thread, which is the opposite of the goal.

uecasm:
Implemented some caching.

Raycoms (approved):

someaddons (approved):

PR #8829, opened 2022-12-21
Don't target invisible citizens

uecasm:
Closes [discord report](https://discord.com/channels/139070364159311872/244066364573089793/1054920276938342410)

- Prevents raiders and regular mobs from targeting invisible citizens.
    - This is mostly intended to avoid targeting a netherworker who is "away", but it seems reasonable to avoid targeting anyone affected by an invisibility potion too.  (Perhaps a research could make druids do that deliberately in the future.)

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/RaiderRangedAI.java
+++ b/src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/RaiderRangedAI.java
@@ -163,7 +163,7 @@ public class RaiderRangedAI<T extends AbstractEntityMinecoloniesMob & IThreatTab
     @Override
     protected boolean isAttackableTarget(final LivingEntity target)
     {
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/coremod/entity/mobs/aitasks/RaiderRangedAI.java
```diff
@@ -163,7 +163,7 @@ protected PathResult moveInAttackPosition(final LivingEntity target)
     @Override
     protected boolean isAttackableTarget(final LivingEntity target)
     {
-        return target instanceof EntityCitizen || (target instanceof Player && !((Player) target).isCreative() && !target.isSpectator());
+        return (target instanceof EntityCitizen && !target.isInvisible()) || (target instanceof Player && !((Player) target).isCreative() && !target.isSpectator());
```
uecasm, line 166: Another option might be to use `TargetingConditions` here, which can do partial invisibility based on range if we want to distinguish between invisibility potions and away-in-nether.

someaddons (approved):

PR #8839, opened 2022-12-26
Plantation field rework

Thodor12:
- Converts the plantation to a field-like setup similar to the farm
  - New block: the plantation field, is used in a free form schematic to provide a field to the colony on placement, these fields can be assigned through the fields window in the plantation, similar to how farms work
  - Plantation fields included to handle:
    - Sugar cane
    - Cactus
    - Bamboo
    - Cocoa
    - Vines
    - Kelp
    - Seagrass
    - TODO: Sea pickles
    - Glowberries
    - Weeping vines
    - Twisting vines
    - Crimson roots/fungi
    - Warped roots/fungi
  - Plantation uses a modular approach to handling new plantation fields based on generic functionality modules (ex sugar cane is a form of "vertically growing plant")
  - Several new researches are included to extend the functionality of plantations with more fields (not all fields listed above are unlocked immediately)
  - TODO: Basic schematics for every field type will be provided
  - Plantation field block has a renderer which shows a basic model
  - TODO: Plantation field has a basic GUI that shows the currently attached worker, what plant is being cultivated there aswell as a repair button to fix the schematic if it's broken due to raiders or similar
- Farmer is reworked to utilize new abstract field logic
  - Everything related to fields is now turned abstract (ex FarmerFieldModule > FieldModule)
  - Fields are now their own class, stored in the RegisteredStructureManager that keep track of which citizen is working on it, where it's located and which plant is being cultivated on the field
  - Fields are also stored in the IColonyView and are transferred to the client similar to buildings by the use of a FieldRegistry class which stores all of the field and field view classes.
  - Fields will no longer keep a reference to the person that owns the field, rather to the building that owns it (to prevent assignment/unassignment funky behaviour)
- Farmers and plantations now have fine (and enforced) control of how many fields are assigned to the building, these conditions are validated both on the server and client to ensure a building cannot inappropriately obtain more fields then it should.
- Farmers and plantations can also control a maximum amount of concurrent worked on plants
  - The farmer keeps backwards compatability by retaining both of these values identical to the building level
- NBT data of fields (both farmer and plantation) is now stored on the colony, not on the tile entities anymore, allows AI to keep working on fields regardless of whether the tile entity is loaded or not.
- TODO: NBT conversion of old farmer/plantation to new setup


Thodor12:
Appreciated if already a first look could be taken at this as it's already growing into a substantial PR, whilst I work on the remaining topics

MotionlessTrain:
I think the base branch needs to be changed to version/1.18.2

Thodor12:
Closing because this is not going back into 1.18

PR #8842, opened 2022-12-27
Fix building advancements

uecasm:
- Fix inconsistency between datagen and output due to #8748 manually editing output without editing code
- Advancements for buildings will now more reliably trigger, especially when building alternatives, since it now triggers on the default blueprint name and not the actual blueprint name.  (Decorations continue to trigger on the actual name.)

### DIFF
--- a/src/main/java/com/minecolonies/coremod/placementhandlers/main/SurvivalHandler.java
+++ b/src/main/java/com/minecolonies/coremod/placementhandlers/main/SurvivalHandler.java
@@ -134,11 +134,11 @@ public class SurvivalHandler implements ISurvivalBlueprintHandler
 
                 if (tempColony != null)
                 {
-                    AdvancementUtils.TriggerAdvancementPlayersForColony(tempColony, playerMP -> AdvancementTriggers.PLACE_STRUCTURE.trigger(playerMP, ((AbstractBlockHut<?>) anchor.getBlock()).getHutName()));
+                    AdvancementUtils.TriggerAdvancementPlayersForColony(tempColony, playerMP -> AdvancementTriggers.PLACE_STRUCTURE.trigger(playerMP, ((AbstractBlockHut<?>) anchor.getBlock()).getBlueprintName()));
                 }
                 else
                 {
-                    AdvancementTriggers.PLACE_STRUCTURE.trigger((ServerPlayer) player, ((AbstractBlockHut<?>) anchor.getBlock()).getHutName());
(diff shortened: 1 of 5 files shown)


src/main/java/com/minecolonies/coremod/placementhandlers/main/SurvivalHandler.java
```diff
@@ -134,11 +134,11 @@ public void handle(
 
                 if (tempColony != null)
                 {
-                    AdvancementUtils.TriggerAdvancementPlayersForColony(tempColony, playerMP -> AdvancementTriggers.PLACE_STRUCTURE.trigger(playerMP, ((AbstractBlockHut<?>) anchor.getBlock()).getHutName()));
+                    AdvancementUtils.TriggerAdvancementPlayersForColony(tempColony, playerMP -> AdvancementTriggers.PLACE_STRUCTURE.trigger(playerMP, ((AbstractBlockHut<?>) anchor.getBlock()).getBlueprintName()));
                 }
                 else
                 {
-                    AdvancementTriggers.PLACE_STRUCTURE.trigger((ServerPlayer) player, ((AbstractBlockHut<?>) anchor.getBlock()).getHutName());
+                    AdvancementTriggers.PLACE_STRUCTURE.trigger((ServerPlayer) player, ((AbstractBlockHut<?>) anchor.getBlock()).getBlueprintName());
```
Raycoms, line 141: Mightve merged too soon, sure about blueprintname? The filename is not heavily tied to the building type anymore.

uecasm: Yes, but the blueprint name is nicer than the hut name, because the hut name has "blockhut" prefix and the blueprint name doesn't.  They're otherwise identical and not necessarily really related to the actual blueprint filenames (which was one of the bugs that this fixes -- the advancements would previously only trigger if the player did build something that happened to have the "right" filename, whereas now it will work regardless of filename).

Further evidence that this produces nice names is that none of the generated files actually changed content except for the town hall.

If one of these methods might be getting removed in a future version then we could revisit the choice then, but it was important to ensure that the datagen and the trigger use the same value.  (Currently, `getBlueprintName()` actually just returns the `BuildingEntry` registry key, which seems like the most reasonable choice anyway.)

Raycoms: :+1:

Raycoms: We might want to rename those methods in the future to better reflect what they actually return. Like getBlockHutName() and getBuildingName()

PR #8847, opened 2022-12-29
Allow building search to also search for fields

uecasm:
- Right-clicking a field (scarecrow) item in air will also search for field blueprints.


src/api/java/com/minecolonies/api/items/ItemBlockBuildingSearchable.java
```diff
@@ -0,0 +1,45 @@
...
+import net.minecraft.world.item.BlockItem;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.Block;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A custom item class for searchable blocks.
+ */
+public class ItemBlockBuildingSearchable extends BlockItem
```
Raycoms: would be neat if we could make this an interface so we can easily slap it on stuff

uecasm: I thought about that, or implementing it via a client event handler so it didn't need to modify the item implementation.  But couldn't think of any other blocks that weren't already huts that it would be useful for.

It's not really intended as a general-purpose search for any block usage, even if the internals could technically do that.

Raycoms: Was mostly thinking that because we got the plantation changes coming up with a similar usecase

uecasm: Yeah.  For that it's mostly a two-liner to add an extra searchable block -- one method override in the block to register it as `ItemBlockBuildingSearchable` instead of a pure `BlockItem`, and one line in the building browser window to cache that block type.

The latter could possibly be made automated if it searched for things that used that base class, although that's slower than just listing them in the code directly.  Might be worthwhile if we anticipate more such searchable-but-not-hut blocks in addon mods rather than direct.

Thodor12: I did take a look at this but I don't think this is going to conflict with my changes massively, as I don't touch the Block class of the scarecrow that much. I've only really introduced changes on the tile entity.

uecasm: I've switched it to an interface now.  So now all you need is for the block to `implements IBuildingBrowsableBlock` in any mod and the magic happens.

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildingBrowser.java
```diff
@@ -212,6 +220,8 @@ private void rebuildCache()
             return;
         }
 
+        final List<Block> browsableBlocks = findBrowsableBlocks();
```
Raycoms, line 223: is this cached?

uecasm: Not inherently, but it's only called once per login since the method as a whole is part of the cache rebuild.  This is more often than strictly necessary since the result can't change without swapping mods, but it seems sufficiently harmless.

Raycoms: kk

Raycoms (dismissed):

PR #8858, opened 2023-01-02
Fix use of existing items in buildings when crafting

uecasm:
Closes [discord report](https://discord.com/channels/139070364159311872/835241758861557811/1058844352555798528)

- Fixes a build error introduced by 4118e0b (bad import merge)
- Crafters will now recognise that they already have part of a crafting request in their building inventory instead of always demanding extra items from the warehouse.
    - This fixes some commonly reported problems with Croptopia food crafting, among other things.

The test case was a taught recipe for bread + beef + pan => hamburger + pan.  (With all disabled as regular food so that the cook doesn't interfere.)  The pan is not consumed by the craft and is not considered a "tool" since it doesn't have durability.  The existing code does correctly recognise that it only requires one of these for any single request (regardless of count).  However these secondary outputs are left in the crafting building (neither returned to the warehouse [as might be reasonable] nor delivered to the crafting requester [wouldn't be correct]) -- itself this is probably not a bug (and is actually good in this case, as with most tool/catalyst style crafts, although might be more debatable for things like buckets and empty bottles) but it causes problems for the next crafting operation requiring it, as it was not recognising that the pan was already in the hut inventory.

For example, a postbox request for 2 hamburgers will create 3 child requests: 2x bread, 2x beef, 1x pan.  Previously, it would ignore whether any of these were in the building already and demand them from the warehouse only.  If they were all in the warehouse, they would be delivered, crafted, then the hamburgers would be delivered to the postbox and the pan would remain in the restaurant.  Eventually you would run out of pans in the warehouse (or run out of room in the restaurant, since that doesn't keep a lot of spare space anyway) and the craft will no longer work.  Or alternatively perhaps the bread or beef were already delivered to the restaurant due to the generic food request and those would be the blockers since they were no longer in the warehouse.

There are two key implementation changes here:
- Removed a condition on the building resolver that pretty much always blocked it recognising "child requests" as valid, since the request was not previously associated with any particular citizen at that building.
    - This could perhaps be added back if needed by walking up the parent requests until it does find one with an associated citizen, but I couldn't see any obvious reason for it to be checking if the citizen were loaded here -- as far as I could tell this was only called for the single building that was attempting a craft anyway, which only happens when there's a citizen AI running.
    - The last thing that touched this particular condition was #7503, if that helps.
- Added a filter to the "reserved stacks" check to exclude recipes related to the request we're currently processing -- otherwise since one crafting request was already pending it would treat all the inputs as reserved and refuse to craft until there were double the number of ingredients actually delivered (i.e. two pans and 2x the requested output count of the other things).

**Not** included here is any changes related to returning the "extra" items to the warehouse for future crafts by other huts.  As previously noted, for these types of tool/catalyst crafts the current behaviour is good -- it will speed up future crafting and it's likely that only one crafter would use that particular item anyway or won't be much of a hardship to give each crafter its own "dedicated" item.  However there may be merit in having something to return empty buckets/bowls/bottles -- but that'd be a separate issue anyway.

I have not tested these changes with other request scenarios, but they do fix this case.


src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
```diff
@@ -6,6 +6,7 @@
 import com.ldtteam.structurize.util.PlacementSettings;
 import com.minecolonies.api.IMinecoloniesAPI;
 import com.minecolonies.api.MinecoloniesAPIProxy;
+import com.minecolonies.api.blocks.interfaces.IBuildingBrowsableBlock;
```
Raycoms, line 9: imports in this file?

Raycoms (commented):
I'm not 100% sure this will work alright without breaking some stuff. Can you try some requests where child request consume a similar resource to the parent? and partially take away resources and see if it recovers etc?

src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java
uecasm: That's the hotfix for the 1.19 merge that you broke 😛  Feel free to apply that commit separately to get that working faster if you want to delay merging this PR.

uecasm:
> Can you try some requests where child request consume a similar resource to the parent? and partially take away resources and see if it recovers etc?

Any specific ideas?  I'm having a failure of imagination.

someaddons (approved):

PR #8885, opened 2023-01-10
Datapack more recipes for 1.19.3

uecasm:
- Ports #8884 
- Refactors custom recipe datagen so that the actual recipe code is the same as in 1.19.2
- Changes a few other datagens so that the parameters are consistent


marchermans:
I feel a bit weird about this pr.
Especially on 1.19.3.

I am not sure how much you refactored here tk support the changes. But as far as I can tell it was a relatively large amount, with the recipeJson and all.

uecasm:
All the code changes are in separate commits, so they should be relatively straightforward to examine.  The only one that's unique to 1.19.3 is [715739d](https://github.com/ldtteam/minecolonies/pull/8885/commits/715739d2d5a2a81bf8b3408ae781527dddc9920a), which just hides the Future stuff back in the base class where it belongs, so the actual crafter-specific code is basically the same between both 1.19.x versions.

Raycoms (approved):

PR #8905, opened 2023-01-18
Fix NBT assignment corruption

Thodor12:
In some cases it's possible that NBT data gets corrupted and causes citizens their jobs to no longer match the expected job in the building they are assigned to.
This fix will prevent any loading issues by unassigning the worker when an improper link between the citizen it's job and building is found, to prevent any crashes.


src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -599,6 +599,7 @@
   "com.minecolonies.coremod.gui.debugscreen.noclosecolony": "No close colony!",
   "com.minecolonies.coremod.gui.debugscreen.nextcolony": "Next colony is %d blocks away (too close to found a new one)!",
   "com.minecolonies.coremod.gui.debugscreen.blocksfromcenter": "%d blocks from the center.",
+  "com.minecolonies.coremod.debug.citizenloadfailure": "Worker %s has been unassigned from his job, a problem was found during load.",
```
MotionlessTrain: Small nitpick: I think that "their job" instead of "his job" is more correct. Unless it has been proven that this can only happen for male citizens, which sounds very unlikely

Thodor12: That's a nitpick indeed, I'll change that tomorrow morning, can't be bothered to turn my pc back on for that lol

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -449,6 +450,29 @@ public void initEntityValues()
         citizen.getEntityData().set(DATA_JOB, getJob() == null ? "" : getJob().getJobRegistryEntry().getKey().toString());
         citizen.getEntityData().set(DATA_STYLE, colony.getTextureStyleId());
 
+        // Safety check that ensure the citizen its job matches the work building job
+        if (getJob() != null && workBuilding != null)
+        {
+            try
+            {
+                workBuilding.getModuleMatching(WorkerBuildingModule.class, m -> m.getJobEntry().equals(getJob().getJobRegistryEntry()));
```
Raycoms: Btw, is there no "getOptionalModuleMatching" method atm?

Thodor12: No, there's only overloads of these methods that do not have a predicate match in them.
I can theoretically use `.getModules()` and iterate all of them to find the first problematic module, but that would seem less than ideal.

Thodor12: I changed up the logic a bit, is it better now?

Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -453,12 +453,19 @@ public void initEntityValues()
...
+                {
+                    hasCorrectJob = true;
+                    break;
+                }
             }
-            catch (Exception ex)
+
+            if (!hasCorrectJob)
             {
+                workBuilding.getModuleMatching(WorkerBuildingModule.class, m -> m.getJobEntry().equals(getJob().getJobRegistryEntry()));
```
MotionlessTrain: Wouldn't this line throw an exception when `hasCorrectJob == false`?

Thodor12: Good catch, that's not supposed to be there anymore

Thodor12:
Ran the broken world again just now to ensure that after the changes everything still works, and it does.
I got a message in chat aswell as the message in the log:
```
[19jan.2023 16:25:07.182] [Server thread/ERROR] [minecolonies/]: Worker Jerome P. Clopton has been unassigned from his job, a problem was found during load. Worker job: minecolonies:lumberjack; Building: minecolonies:cook
```

marchermans (approved):

PR #8929, opened 2023-02-07
Extend research unlocked loot condition

uecasm:
Closes ldtteam/minecolonies-features#747

- Check additional loot parameters such that `research_unlocked` works for crafting recipes too
- Allow checking for specific research effect strength ranges too

### DIFF
--- a/src/api/java/com/minecolonies/api/loot/ResearchUnlocked.java
+++ b/src/api/java/com/minecolonies/api/loot/ResearchUnlocked.java
@@ -4,7 +4,9 @@ import com.google.gson.JsonDeserializationContext;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonSerializationContext;
 import com.minecolonies.api.colony.IColony;
+import com.minecolonies.api.colony.IColonyManager;
 import com.minecolonies.api.entity.citizen.AbstractEntityCitizen;
+import net.minecraft.core.BlockPos;
 import net.minecraft.resources.ResourceLocation;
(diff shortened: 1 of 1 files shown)


marchermans (commented):
In general this is a good addition, but personally I would prefer if it is a seperate kind of loot condition

uecasm:
> In general this is a good addition, but personally I would prefer if it is a seperate kind of loot condition

What do you mean by separate?  I don't think it makes much sense in having two conditions that both check for research unlocks... or do you mean one that checks purely for any unlock and another that checks for strength?  I don't really see why that would be useful to separate, that would just make the code more complicated.

sweg4552:
Is there a reason that the research_unlocked checks the effect and not just the research itself?
Unlockable crafting recipes are able to take the research-id without needing an effect at all. 

As a datapacker it would be most intuitive to have a research_unlocked that takes research id & checks as a bool, and a effect_strength that takes effect id & checks it's integer value
If that's possible

uecasm:
Because the way the research system is structured, the effects are intended to be the thing that actually has the gameplay effect (hence the name), while the research just unlocks the effects (and a bit of lore).

One idea is that you could make a "total conversion" research datapack that changes all the research but keeps the same effects.  (While you can change the effects that only affect recipes and loot tables, most of the others are used in the code and so have to remain essentially as-is anyway.)

The recipes were actually recently changed so that (while they still can refer to either research or effect) all the default recipes exclusively use effects now.

PR #8968, opened 2023-02-25
Use Forge tool types for shears and fishing rods

uecasm:
- Use new Forge tool types to detect compatible shears and fishing rods

There is a tool type for shields as well, but I haven't used that since there are other more complex dependencies on `ShieldItem` in particular.

I have ~not~ tested this ~even~ slightly, but it seems plausible.  If someone wants to suggest specific tests before accepting, please do.


Shadizar:
I think one of the more common "mess with fishing rods" mods is Aquaculture 2, if that helps

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -303,51 +303,34 @@ else if (Compatibility.isTinkersTool(stack, toolType))
...
+            return ((TieredItem) stack.getItem()).getTier().getLevel();
+        }
+        else if (toolType.equals(ToolType.FISHINGROD))
         {
             return getFishingRodLevel(stack);
         }
         else if (toolType.equals(ToolType.SHEARS))
         {
-            return stack.getItem() instanceof ShearsItem ? 0 : -1;
+            return 0;
```
Raycoms, line 326: I remember that without this piece, results were sometimes iffy

uecasm: It's already done the `isTool` check by this point, so it can't reach here unless that decided it was valid shears.  (That in turn requires an item that does both `SHEARS_DIG` and `SHEARS_HARVEST`, but does not require that it be a `ShearsItem`.  Perhaps notably we also don't check for `SHEARS_CARVE` or `SHEARS_DISARM` either, but the latter is unlikely to be appropriate and while the former kind of is, it seems a little unnecessary to check everything.)

The rearrangement of the code does mean that it might sometimes return a value higher than 0, if for example a mod introduces new `TieredItem` shears with different material levels.  (But that's good, because they probably also introduced blocks that require it.)  But if `isTool` passed then there's no reason for this to return -1.

uecasm:
Seems to work fine, at least with vanilla tools.  (Haven't tested with extra mods.)

image

Made a level 2 forester, they asked for an axe and shears, put them in the warehouse, courier delivered them, they started working.  (Perhaps slightly misleading is that a level 1 forester with "maximal level:stone" will accept vanilla shears made with iron, since they're considered "wood or gold" tier.  But it does list them as acceptable in the icon.)

(Side note: this does perhaps argue that level 1 workers should not be asking for shears at all, though, since it does still require the player to have iron technology...)

image

Ditto fishing rod with level 1 fisher.

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -303,51 +303,34 @@ else if (Compatibility.isTinkersTool(stack, toolType))
...
+                || ToolType.LEGGINGS.equals(toolType))
         {
             if (stack.getItem() instanceof ArmorItem)
             {
                 final ArmorItem ArmorItem = (ArmorItem) stack.getItem();
                 return getArmorLevel(ArmorItem.getMaterial());
             }
         }
-        else if (stack.getItem() instanceof FishingRodItem)
+        else if (stack.getItem() instanceof TieredItem)  // most tools
```
Nightenom: pattern matching
(btw instanceof variables can be finalized like normal ones)

src/main/java/com/minecolonies/coremod/entity/NewBobberEntity.java
```diff
@@ -310,8 +308,8 @@ public boolean shouldStopFishing()
     {
         final ItemStack itemstack = this.angler.getMainHandItem();
         final ItemStack itemstack1 = this.angler.getOffhandItem();
-        final boolean flag = itemstack.getItem() instanceof net.minecraft.world.item.FishingRodItem;
-        final boolean flag1 = itemstack1.getItem() instanceof net.minecraft.world.item.FishingRodItem;
+        final boolean flag = itemstack.canPerformAction(net.minecraftforge.common.ToolActions.FISHING_ROD_CAST);
```
Nightenom: import (no longer name clutch?)

src/api/java/com/minecolonies/api/util/ItemStackUtils.java
uecasm: That was mostly on lines I wasn't changing, but sure.

src/main/java/com/minecolonies/coremod/entity/NewBobberEntity.java
uecasm: Done.

PR #8985, opened 2023-03-02
Fix sleeping AI

someaddons:
- Fix sleeping AI to no longer update state during movement

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
+++ b/src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
@@ -734,16 +734,6 @@ public abstract class AbstractEntityCitizen extends AbstractCivilianEntity imple
      */
     public abstract void setFleeingState(final boolean fleeing);
 
-    /**
-     * Setter for the citizen pose.
-     *
-     * @param pose the pose to set.
-     */
-    public void updatePose(final Pose pose)
-    {
-        setPose(pose);
-    }
-
     @Override
     public void detectEquipmentUpdates()
     {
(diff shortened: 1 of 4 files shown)


src/api/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
```diff
@@ -734,16 +734,6 @@ public boolean checkCanDropLoot()
      */
     public abstract void setFleeingState(final boolean fleeing);
 
-    /**
-     * Setter for the citizen pose.
-     *
-     * @param pose the pose to set.
-     */
-    public void updatePose(final Pose pose)
```
Raycoms, line 742: Pretty sure now if you relog or reload the chunk they are sleeping in they will stand in their beds

someaddons: ? This is just a duplicate setter we were using lol :D

Raycoms: Meant more in general the pr. Just commented on the first file. The pose methods come from a time the setPose method was protected. Did you test with reloading and similar stuff to avoid back crawling citizens, etc?

someaddons: logic did not change, it simply got moved to the proper place

Raycoms: Yeah, but at the other place it was constantly checked without other AIs potentially blocking the check.

PR #9002, opened 2023-03-12
Add easy scan command for nbt file

Raycoms:
Closes #
Closes #
Closes #

- Add command to not only scan the blueprint file, but also to generate the nbt file, set the tags, and setup jigsaws
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
+++ b/src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
@@ -232,7 +232,7 @@ public class ScanCommand extends AbstractCommand
                     }
                     else
                     {
-                        jigsawBlockEntity.setPool(ResourceKey.create(Registry.TEMPLATE_POOL_REGISTRY, new ResourceLocation("minecolonies:asian/roads")));
+                        jigsawBlockEntity.setPool(ResourceKey.create(Registry.TEMPLATE_POOL_REGISTRY, new ResourceLocation("minecolonies:" + style+ "/roads")));
                         jigsawBlockEntity.setName(new ResourceLocation("minecolonies:street"));
                         jigsawBlockEntity.setTarget(new ResourceLocation("minecolonies:street"));
                     }
--- a/src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
+++ b/src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
@@ -217,7 +217,10 @@ public class ScanCommand extends AbstractCommand
                 JigsawBlockEntity jigsawBlockEntity = (JigsawBlockEntity) world.getBlockEntity(mutablePos);
                 if (jigsawBlockEntity.getName().getPath().isEmpty() || jigsawBlockEntity.getName().getPath().equals("empty"))
                 {
-                    jigsawBlockEntity.setFinalState("minecraft:structure_void");
+                    if (jigsawBlockEntity.getFinalState().equals("minecraft:air"))
+                    {
+                        jigsawBlockEntity.setFinalState("minecraft:structure_void");
+                    }
                     if (isHut)
                     {
                         jigsawBlockEntity.setPool(ResourceKey.create(Registry.TEMPLATE_POOL_REGISTRY, new ResourceLocation("minecolonies:" + style+ "/roads")));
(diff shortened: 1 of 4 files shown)


src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
```diff
@@ -0,0 +1,385 @@
...
+                    }
+                    else if (jigsawBlockEntity.getPool().location().getPath().contains("building"))
+                    {
+                        jigsawBlockEntity.setPool(ResourceKey.create(Registry.TEMPLATE_POOL_REGISTRY, new ResourceLocation("minecolonies:" + style+ "/buildings")));
+                        jigsawBlockEntity.setName(new ResourceLocation("minecolonies:building_entrance"));
+                        jigsawBlockEntity.setTarget(new ResourceLocation("minecolonies:building_entrance"));
+                    }
+                    else
+                    {
+                        jigsawBlockEntity.setPool(ResourceKey.create(Registry.TEMPLATE_POOL_REGISTRY, new ResourceLocation("minecolonies:asian/roads")));
```
MotionlessTrain: Testing code, I assume?
(I would have expected something like `"minecolonies:" + style + "/roads"`)

Raycoms: oopsy

someaddons (approved):

src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
```diff
@@ -0,0 +1,388 @@
...
+            BlockState state = world.getBlockState(mutablePos);
+            if (state.getBlock() == Blocks.JIGSAW)
+            {
+                JigsawBlockEntity jigsawBlockEntity = (JigsawBlockEntity) world.getBlockEntity(mutablePos);
+                if (jigsawBlockEntity.getName().getPath().isEmpty() || jigsawBlockEntity.getName().getPath().equals("empty"))
+                {
+                    if (jigsawBlockEntity.getFinalState().equals("minecraft:air"))
+                    {
+                        jigsawBlockEntity.setFinalState("minecraft:structure_void");
+                    }
```
uecasm, line 223: `structure_void` is equivalent to the light placeholder, not to air.  This means that this is possibly going to leave a solid block in that location when that was probably not intended.  Why is this not just leaving it as air?

People can set it to `structure_void` themselves if they want it to be a placeholder.  (Or, presumably, to a light placeholder, although that might not work depending on what order it runs the processors.)

src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        final BlockState primary = world.getBlockState(zero.offset(bp.getPrimaryBlockOffset()));
+        final boolean isHut;
+        if (primary.getBlock() instanceof AbstractBlockHut<?>)
+        {
+            isHut = true;
+            final AbstractTileEntityColonyBuilding building = (AbstractTileEntityColonyBuilding) world.getBlockEntity(zero.offset(bp.getPrimaryBlockOffset()));
+            building.addTag(new BlockPos(0, 0, 0), "deactivated");
+            building.addTag(new BlockPos(0, 0, 0), style);
```
uecasm, line 205: `TileEntityColonyBuilding.reactivate()` assumes that the style tag is the first non-`deactivated` tag.  However this code will add it to the end, which means it won't work if the hut block has some other tags already (e.g. `leisure` or `invisible`)...

It also appears to expect a more complicated tag value than just `style` -- it wants `Long Pack Name:filename` (unless it's a built-in pack with a default building, then you can cheat a bit).

This whole thing seems a bit fragile.  Perhaps there's a better way to store it?  Is there some reason it doesn't use the existing hut NBT properties?

src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
Raycoms: all the current ones in the mod use structure void at the place, thats why I went with this

src/main/java/com/minecolonies/coremod/commands/generalcommands/ScanCommand.java
Raycoms: I'll look again into it when I come home. I'll make sure to add it to the front.

PR #9018, opened 2023-03-19
Lumberjack zone is actually 3D

uecasm:
- The lumberjack zone tool claims to be specifying a 2D area, but it was actually storing a 3D volume restriction.  This meant that unless the trunk base was at the same level (or possibly 1 above too?) as the clicked block, they would reject it and stay "searching for trees" forever.
    - ~~The zone area will have to be set again after this fix is applied to correct the Y matching.  If the lumberjack zone was already working before, it will continue to work.~~
    - This updates the messages to show that it is actually a 3D area.
      image

### DIFF
--- a/src/main/resources/assets/minecolonies/lang/manual_en_us.json
+++ b/src/main/resources/assets/minecolonies/lang/manual_en_us.json
@@ -73,8 +73,8 @@
   "item.minecolonies.scepterlumberjack": "Forest Tool",
   "item.minecolonies.scepterlumberjack.usedend": "Position A set. Select Position B with a right-click or left-click the same position to finish.",
   "item.minecolonies.scepterlumberjack.usedstart": "Position B set. Select Position A with a left-click or right-click the same position to finish.",
-  "item.minecolonies.scepterlumberjack.restrictiontoobig": "An area of %d blocks is too large. The maximum restricted area for the Forester is %d blocks.",
-  "item.minecolonies.scepterlumberjack.restrictionset": "Forester restricted area set (x: %d to %d, z: %d to %d, area: %d blocks). The max size is %d blocks.",
+  "item.minecolonies.scepterlumberjack.restrictiontoobig": "An volume of %d blocks is too large. The maximum restricted volume for the Forester is %d blocks.",
+  "item.minecolonies.scepterlumberjack.restrictionset": "Forester restricted volume set (x: %d to %d, y: %d to %d, z: %d to %d, volume: %d blocks). The max size is %d blocks.",
 
   "minecolonies.config.maxtreesize": "Max Tree Size",
   "minecolonies.config.maxtreesize.comment": "Max log count in one tree for the Forester to check during their tree search.",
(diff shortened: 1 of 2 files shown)


Raycoms:
Then we should fix the messages, its on purpose 3d to allow cave areas etc as well.

uecasm:
> Then we should fix the messages, its on purpose 3d to allow cave areas etc as well.

Alrighty.

src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -73,8 +73,8 @@
   "item.minecolonies.scepterlumberjack": "Forest Tool",
   "item.minecolonies.scepterlumberjack.usedend": "Position A set. Select Position B with a right-click or left-click the same position to finish.",
   "item.minecolonies.scepterlumberjack.usedstart": "Position B set. Select Position A with a left-click or right-click the same position to finish.",
-  "item.minecolonies.scepterlumberjack.restrictiontoobig": "An area of %d blocks is too large. The maximum restricted area for the Forester is %d blocks.",
```
Raycoms, line 76: Maybe we call it box or so? I think volume might confuse some players

uecasm: A "box" could still be 2D though.

MotionlessTrain: Or "zone", since the setting to turn it on is called "zoning mode"

uecasm: That's also usually a 2D concept.

someaddons: area is probably still the best for 3d, btw your changed version now has "An volume"

MotionlessTrain: I apparently wasn't awake enough yet 🤦
Area sounds more like 2D than 3D in my opinion

uecasm: I've tweaked the wording a bit to use a mixture of "zone", "3D box", and "volume" where each seems most appropriate.

Talyda: Sorry, cos I am confused now reading through this - do we now need to click one corner higher than the other? Cos it didn't matter if we clicked the same Y level before, it did the area from bedrock to sky limit, (actually, it may have done lower Y level to sky limit, can't recall! But I know it did to sky limit!) but now the wording makes it sound like we need to click (for example) y64 one corner, and y74 other corner to get the height in?

*edits* I commented wrong place =P
This wording is what's confusing me:
> "A zone of %d blocks is too large. The maximum restricted volume for the Forester is %d blocks.",
>   "item.minecolonies.scepterlumberjack.restrictionset": "Forester restricted zone set (x: %d to %d, y: %d to %d, z: %d to %d, total: %d blocks). The max size is %d blocks.",

uecasm: The actual behaviour has not been changed by this PR.  I originally thought as you did that it was a 2D region only (bedrock to sky) but it turns out that this is not the actual implementation.  The first version of this changed it back to a 2D region, but Ray said that being 3D was intended, so now it just updates the messages to make that clearer.

So yes, you need to click at different Y levels to get them to register a 3D box properly.  Having said that, there is a bit of a fudge built in, and it's only the base of the tree that has to be within the box, so the previous behaviour of clicking twice at the same ground Y still works to select an area if you're clicking on the same ground level that all of your trees are planted directly on.  But this will not find any trees planted 1y above the ground that you clicked on, which is how I noticed this issue in the first place.

i.e. if you click on a ground block at Y=63 twice, this will only accept the tree base (bottom log) being at Y=62-64 and will not accept Y=65 (i.e. ground, another block, then log).  This might be the source of some of the "my forester doesn't find trees if I put them up in planters or on hilly terrain" issues.

Raycoms:
We had this 2d in the very beginning but this caused big issues in underground colonies

src/main/resources/assets/minecolonies/lang/manual_en_us.json
Talyda: ~~I'm gonna have to test it. Cos unless I am misremembering, I am sure I set a zone for my forester in an underground colony. Forester kept dying above ground - I had to remove all the trees above ground between the x and z coordinates so that they would only chop trees where I planted them. Was doing an underground Hive style colony, so it was either in 1.18 or 1.16.  Unless things changed since then?  Will try and test it anyway.~~
Never mind, just saw ray's comment =D

Talyda:
> We had this 2d in the very beginning but this caused big issues in underground colonies

Oh! ok, sorry, this comment didn't show up for me. That makes sense then, it did change =)

Raycoms:
Sent from Proton Mail mobile

marchermans (approved):

PR #9025, opened 2023-03-22
Plantation field rework

Thodor12:
- Converts the plantation to a field-like setup similar to the farm
  - New block: the plantation field, is used in a free form schematic to provide a field to the colony on placement, these fields can be assigned through the fields window in the plantation, similar to how farms work
  - Plantation fields included to handle:
    - Sugar cane
    - Cactus
    - Bamboo
    - Cocoa
    - Vines
    - Kelp
    - Seagrass
    - Sea pickles
    - Glowberries
    - Weeping vines
    - Twisting vines
    - Crimson roots/fungi
    - Warped roots/fungi
  - Plantation uses a modular approach to handling new plantation fields based on generic functionality modules (ex sugar cane is a form of "vertically growing plant")
  - Several new researches are included to extend the functionality of plantations with more fields (not all fields listed above are unlocked immediately)
  - Basic schematics for every field type will be provided
  - Plantation field block has a renderer which shows a basic model
  - Plantation field has a basic GUI that shows the currently attached worker, what plant is being cultivated there aswell as a repair button to fix the schematic if it's broken due to raiders or similar
  - Plantation fields are considered as leisure sites aswell, so schematic designers can make nice spaces as well
- Farmer is reworked to utilize new abstract field logic
  - Everything related to fields is now turned abstract (ex FarmerFieldModule > FieldModule)
  - Fields are now their own class, stored in the RegisteredStructureManager that keep track of which citizen is working on it, where it's located and which plant is being cultivated on the field
  - Fields are also stored in the IColonyView and are transferred to the client similar to buildings by the use of a FieldRegistry class which stores all of the field and field view classes.
  - Fields will no longer keep a reference to the person that owns the field, rather to the building that owns it (to prevent assignment/unassignment funky behaviour)
- Farmers and plantations now have fine (and enforced) control of how many fields are assigned to the building, these conditions are validated both on the server and client to ensure a building cannot inappropriately obtain more fields then it should.
- Farmers and plantations can also control a maximum amount of concurrent worked on plants
  - The farmer keeps backwards compatability by retaining both of these values identical to the building level
- NBT data of fields (both farmer and plantation) is now stored on the colony, not on the tile entities anymore, allows AI to keep working on fields regardless of whether the tile entity is loaded or not.
- NBT conversion of old farmer/plantation to new setup


src/api/java/com/minecolonies/api/colony/IColonyManager.java
```diff
@@ -371,7 +372,41 @@ static IColonyManager getInstance()
...
+     * Returns result of {@link IColonyView#handleColonyFieldViewMessage} (BlockPos, FriendlyByteBuf)} if {@link #getColonyView(int, ResourceKey)} gives a not-null result. If {@link
+     * #getColonyView(int, ResourceKey)} is null, returns null.
+     *
+     * @param colonyId ID of the colony.
+     * @param dim      the dimension.
+     * @param type     the field type.
+     * @param position the field position.
+     * @param buf      {@link FriendlyByteBuf} with field data.
+     */
+    void handleColonyFieldViewMessage(
```
Raycoms: We try to keep those things out of the colony manager these days. Rather put that in the building manager or so and enough logic in the message itself instead.

src/api/java/com/minecolonies/api/colony/IColonyView.java
```diff
@@ -266,6 +270,40 @@ public interface IColonyView extends IColony
...
     IMessage handleColonyBuildingViewMessage(BlockPos buildingId, @NotNull FriendlyByteBuf buf);
 
+    /**
+     * Update a ColonyView's fields given a network data ColonyView update packet. This uses a full-replacement - fields do not get updated and are instead overwritten.
+     *
+     * @param type     the field type.
+     * @param position the field position.
+     * @param buf     buffer containing field information.
+     */
+    void handleColonyFieldViewMessage(@NotNull FieldRegistries.FieldEntry type, @NotNull BlockPos position, @NotNull FriendlyByteBuf buf);
```
Raycoms, line 290: Same thing, this should be in the manager that holds the fields and not fill up the colony view and colony if possible

src/api/java/com/minecolonies/api/colony/buildings/workerbuildings/plantation/PlantationFieldType.java
```diff
@@ -0,0 +1,21 @@
+package com.minecolonies.api.colony.buildings.workerbuildings.plantation;
+
+/**
+ * All the field types that the plantation supports.
+ */
+public enum PlantationFieldType
```
Raycoms: Registry

src/api/java/com/minecolonies/api/colony/fields/AbstractFieldMatcher.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public final @NotNull BlockPos getPosition()
+    {
+        return position;
+    }
+
+    @Override
+    public boolean matches(@Nullable final IField other)
```
Raycoms: Why is this necessary? Cant we just have that logic in the field itself?

src/api/java/com/minecolonies/api/colony/fields/AbstractFieldView.java
```diff
@@ -0,0 +1,132 @@
...
+     * Building id of the building owning the field.
+     */
+    private BlockPos buildingId;
+
+    /**
+     * Constructor used in deserialization.
+     *
+     * @param colony the colony this field belongs to.
+     */
+    protected AbstractFieldView(@NotNull IColonyView colony, @NotNull BlockPos position)
```
Raycoms: Nothing of these values cannot be sent to the client side, we can reuse the same field class for both sides and make it IColony

src/api/java/com/minecolonies/api/colony/fields/registry/FieldRegistries.java
```diff
@@ -0,0 +1,139 @@
...
+    public static IForgeRegistry<FieldEntry> getFieldRegistry()
+    {
+        return IMinecoloniesAPI.getInstance().getFieldRegistry();
+    }
+
+    /**
+     * Entry for the {@link IField} registry. Makes it possible to create a single registry for a {@link IField}. Used to lookup how to create {@link IField} and {@link
+     * IFieldView}.
+     */
+    public static class FieldEntry
```
Raycoms, line 71: I would get rid of the field type enum and just make the field type enum a Resourcelocation, as we already got this here. Its one FieldEntry per enum type, right? So a resource location per enum is enough.

src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
```diff
@@ -102,33 +100,11 @@ public static void init()
           });
 
         InteractionValidatorRegistry.registerStandardPredicate(Component.translatable(NO_FREE_FIELDS),
-          citizen -> citizen.getWorkBuilding() instanceof BuildingFarmer && ((BuildingFarmer) citizen.getWorkBuilding()).getFirstModuleOccurance(FarmerFieldModule.class).hasNoFields());
+          citizen -> citizen.getWorkBuilding() instanceof BuildingFarmer && citizen.getWorkBuilding().getFirstModuleOccurance(FieldsModule.class).hasNoFields());
 
         InteractionValidatorRegistry.registerStandardPredicate(Component.translatable(INVALID_MINESHAFT),
           citizen -> citizen.getWorkBuilding() instanceof BuildingMiner && citizen.getJob() instanceof JobMiner && (((BuildingMiner) citizen.getWorkBuilding()).getCobbleLocation() == null || ((BuildingMiner) citizen.getWorkBuilding()).getLadderLocation() == null));
 
-        InteractionValidatorRegistry.registerPosBasedPredicate(Component.translatable(NO_SEED_SET),
```
Raycoms, line 110: Is it a default seed now?

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildDecoration.java
```diff
@@ -107,10 +108,15 @@ public class WindowBuildDecoration extends AbstractWindowSkeleton
...
+    public WindowBuildDecoration(
+      final BlockPos pos,
+      final String packMeta,
+      final String path,
+      final Rotation rotation,
+      final boolean mirror,
+      Function<BlockPos, IMessage> buildRequestMessage)
     {
         super(Constants.MOD_ID + BUILDING_NAME_RESOURCE_SUFFIX);
-        this.packMeta = packMeta;
```
Raycoms, line 113: Why?

src/main/java/com/minecolonies/coremod/client/gui/WindowBuildDecoration.java
```diff
@@ -130,20 +136,24 @@ public WindowBuildDecoration(final BlockPos pos, final String packMeta, final St
         blueprintFuture = StructurePacks.getBlueprintFuture(cleanedPackName, path);
         this.rotation = rotation;
         this.mirror = mirror;
+        this.buildRequestMessage = buildRequestMessage;
     }
 
-    @Override
-    public void onOpened()
+    private void confirmedBuild()
```
Raycoms: Please dont reorganize existing methods in a PR, that makes reading it very hard and creates big conflict chunks

src/main/java/com/minecolonies/coremod/client/gui/WindowDecorationController.java
```diff
@@ -126,34 +127,56 @@ public WindowDecorationController(final BlockPos b)
...
      */
-    private void cancelClicked()
+    private void buildClicked()
     {
+        final int level = Utils.getBlueprintLevel(this.controller.getBlueprintPath());
+
+        final String path = controller.getBlueprintPath().replace(level + ".blueprint", (level + 1) + ".blueprint");
+
         close();
+        new WindowBuildDecoration(controller.getBlockPos(),
```
Raycoms, line 147: Same thing with reorg

src/main/java/com/minecolonies/coremod/client/gui/modules/FarmFieldsModuleWindow.java
```diff
@@ -0,0 +1,238 @@
...
+
+        updateUI();
+    }
+
+    /**
+     * Fetches the list of fields, specifically filtered for the farm field.
+     *
+     * @return the filtered list of fields.
+     */
+    private List<FarmField.View> getFields()
```
Raycoms: Please include some screenshots as well of the UIs

src/main/java/com/minecolonies/coremod/client/gui/modules/PlantationFieldsModuleWindow.java
```diff
@@ -0,0 +1,232 @@
...
+        });
+
+        updateUI();
+    }
+
+    @Override
+    public void onUpdate()
+    {
+        super.onUpdate();
+        updateUI();
```
Raycoms, line 227: Whats the dif of the farmfieldsmodule and plantationfieldsmodule windows?

src/main/java/com/minecolonies/coremod/colony/buildings/moduleviews/FieldsModuleView.java
```diff
@@ -0,0 +1,230 @@
...
+     * @return a text component that should be shown if there is a problem for the specific field, else null.
+     */
+    @Nullable
+    public MutableComponent getFieldWarningTooltip(IFieldView field)
+    {
+        if (!FieldsModule.checkFieldCount(getOwnedFields().size(), maxFieldCount))
+        {
+            return Component.translatable(FIELD_LIST_WARN_EXCEEDS_FIELD_COUNT);
+        }
+        //else if (!FieldsModule.checkPlantCount(getWorkedPlants().size(), maxConcurrentPlants))
```
Raycoms: remove uncommented code

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/plantation/PlantationModuleRegistry.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Registry class for all different plantation module types.
+ */
+public class PlantationModuleRegistry
+{
+    /**
+     * Map containing the field types as well as the modules belonging to it.
+     */
+    private static final EnumMap<PlantationFieldType, PlantationModule> plantationModules = new EnumMap<>(PlantationFieldType.class);
```
Raycoms: So, not a real registry. Cant this be part of our initial registry. Im a bit confused for the need for so many types and registries. Isnt it enough to have one Entry

BerryEntry, ThisEntry, ThatEntry, they also represent the types, and they also can get you the modules?

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingPlantation.java
```diff
@@ -88,315 +77,308 @@ public BuildingPlantation(final IColony c, final BlockPos l)
...
-            this.position = position;
-            this.combination = combination;
+            super.deserialize(buf);
+            maxConcurrentPlants = buf.readInt();
         }
 
-        /**
-         * Get the position.
-         *
-         * @return the position.
```
Raycoms, line 306: reorg of methods makes this unreadable

src/main/java/com/minecolonies/coremod/colony/managers/RegisteredStructureManager.java
```diff
@@ -621,91 +561,30 @@ public IBuilding addNewBuilding(@NotNull final AbstractTileEntityColonyBuilding
     }
 
     @Override
```
Raycoms: Too much method reorg, reorg should be separate PRs

Raycoms (changes requested):
I didnt get fully through it yet, but the reorganization of existing classes makes this almost impossible to review in parts. Please undo the reorganizations of methods and fields in existing classes so that we can actually see what changed and review that accordingly.

There also seem to be too many registries that can all be united into the same thing imo.

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingPlantation.java
Thodor12: This specific class is actually not even reorganized, there's just *that* many changes. Almost nothing of the original class survived.

src/api/java/com/minecolonies/api/colony/fields/AbstractFieldMatcher.java
Thodor12: The idea of the field matcher was that you could create an instance of them, without needing the data of the entire field, mainly meant for messages back to the server.
For example when I want to change the plant on a farm field, I'd make a farm field matcher and give it the position, instead of having to give it the entire field instance.

Raycoms: Fields are so tiny, not worth the added complexity.

src/api/java/com/minecolonies/api/colony/IColonyManager.java
Thodor12: The client doesn't have a building manager, this is done just like the buildings, where the colony view keeps a map of them.
```java

    @Override
    public IRegisteredStructureManager getBuildingManager()
    {
        return null;
    }
```

Raycoms: okay, then just remove it from the colony manager and colony and leave it here. Although we could start thinking of managers for the client side here too to make the class not too big.

Thodor12: I will have to keep them for now, as the window actually pulls the field instances from here, so that I have a single point of truth where I know the correct values are.
Alternatively I'd have to rework it back to where the field modules are keeping the instances saved, which I found to not always work as reliably.

src/main/java/com/minecolonies/coremod/client/gui/modules/PlantationFieldsModuleWindow.java
Thodor12: Slightly different UI conditions, for example the farmer no longer has a plant limit, that has moved solely to the plantation.
The changes are marginal, but instead of making it so there's a few methods which change functionality, I figured completely extracting the classes separately would give more freedom (also later for pastures)

src/api/java/com/minecolonies/api/colony/IColonyManager.java
Thodor12: Discussed in DM, for now this is going to stay, but the client in the future will likely implement the building manager class, causing a big refactor in the colony view.

Raycoms: The colony manager doesnt need it, the colonyview can keep it.

src/api/java/com/minecolonies/api/colony/modules/ModuleContainerHandlers.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Class that contains generic logic for handling with modules.
+ */
+public class ModuleContainerHandlers
+{
+    private ModuleContainerHandlers()
+    {
+    }
+
+    public static <T> boolean hasModule(final Collection<T> modules, Class<? extends T> clazz)
```
Raycoms: javadoc

Raycoms: Would also call the class more sth like ModuleContainerUtils

src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
Raycoms: ?? Dont we need an interaction still for this?

src/main/java/com/minecolonies/coremod/colony/ColonyView.java
```diff
@@ -1415,160 +1358,235 @@ public void setColonyColor(final ChatFormatting color)
...
     @Override
-    public ICitizenDataView getVisitor(final int citizenId)
+    public void setTextureStyle(final String style)
     {
-        return visitors.get(citizenId);
+        this.textureStyle = style;
     }
 
     @Override
-    public void setNameStyle(final String style)
```
Raycoms: undo reorg

src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
Thodor12: You cannot assign fields who don't have a seed set to begin with, making this redundant

Raycoms: Is there sth telling the player about that?

Thodor12:
Added some images of the interfaces as they are currently made

2023-06-16 20_03_35-Minecraft_ 1 19 2 - Singleplayer
2023-06-16 20_03_43-Minecraft_ 1 19 2 - Singleplayer
2023-06-16 20_03_56-Minecraft_ 1 19 2 - Singleplayer
2023-06-16 20_04_10-Minecraft_ 1 19 2 - Singleplayer
2023-06-16 20_04_20-Minecraft_ 1 19 2 - Singleplayer
2023-06-16 20_04_32-Minecraft_ 1 19 2 - Singleplayer
2023-06-16 20_04_33-Minecraft_ 1 19 2 - Singleplayer

src/api/java/com/minecolonies/api/colony/modules/IModuleContainer.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
+
+/**
+ * Default interface for objects that contain module instances.
+ */
+public interface IModuleContainer<T>
```
someaddons, line 12: kinda an odd class, you're forcing a bunch of implementations of lookups, but at the same time all those implementations redirect to your static util, maybe simply use this as module getter? Or just scrap it?

src/api/java/com/minecolonies/api/colony/modules/ModuleContainerUtils.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
+
+/**
+ * Class that contains generic logic for handling with modules.
+ */
+public class ModuleContainerUtils
```
someaddons, line 15: Too much boilerplate code in this implementation, added streams optionals, please use the previous lookups those are much faster

src/api/java/com/minecolonies/api/colony/IColonyView.java
```diff
@@ -266,6 +273,40 @@ public interface IColonyView extends IColony
...
+
+    /**
+     * Update a ColonyView's fields given a network data ColonyView update packet. Removing the passed field instance.
+     *
+     * @param type     the field type.
+     * @param position the field position.
+     * @param buf     buffer containing field information.
+     */
+    void handleColonyRemoveFieldViewMessage(@NotNull FieldRegistries.FieldEntry type, @NotNull BlockPos position, @NotNull FriendlyByteBuf buf);
```
someaddons, line 299: documentation and usage/name of this function are odd, is it an update function (add AND remove) or just remove?
and if just remove, how do we add fields then? If you resend the full list on add you can do the same on remove

src/api/java/com/minecolonies/api/colony/fields/AbstractField.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.Optional;
+import java.util.function.Predicate;
+
+import static com.minecolonies.api.util.constant.NbtTagConstants.TAG_OWNER;
+
+/**
+ * Abstract implementation for field instances.
+ * Contains some basic mandatory logic for fields.
+ */
+public abstract class AbstractField implements IField
```
someaddons: is there a reason for this imlementation class to be within API? Same for the abstract FieldModule

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/plantation/modules/generic/DownwardsGrowingPlantModule.java
```diff
@@ -0,0 +1,259 @@
...
+ * Plantation module for plants that grow vertically downwards, similar to sugar glowberries.
+ * <br/>
+ * Requirements:
+ * <ol>
+ *     <li>Grow vertically downwards without any outcroppings.</li>
+ *     <li>Grow uninterrupted (no gaps in between the plant).</li>
+ *     <li>Must break all blocks below when an upper block is destroyed.</li>
+ * </ol>
+ */
+public abstract class DownwardsGrowingPlantModule extends AbstractPlantationModule
```
someaddons, line 27: Modules are data holders/containers, the logic in here belongs into AI classes instead, same goes for all your other modules

src/api/java/com/minecolonies/api/colony/fields/AbstractField.java
```diff
@@ -0,0 +1,230 @@
...
+{
+    /**
+     * Set of field modules this field has.
+     */
+    private final List<IFieldModule> modules = new ArrayList<>();
+
+    /**
+     * Colony owning the field.
+     */
+    private final IColony colony;
```
someaddons: Do fields ever need a colony reference? Since they're not active by themselves I'd think they're always queried form a context with colony

src/api/java/com/minecolonies/api/colony/fields/AbstractField.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    @Override
+    public final boolean isTaken()
+    {
+        return buildingId != null;
+    }
+
+    @Override
+    public final int getDistance(final IBuildingView building)
```
someaddons: getSqDistance

src/api/java/com/minecolonies/api/colony/fields/IField.java
```diff
@@ -0,0 +1,123 @@
...
+     * Gets the position of the field.
+     *
+     * @return central location of the field.
+     */
+    @NotNull BlockPos getPosition();
+
+    /**
+     * Getter for the colony of the field.
+     *
+     * @return the int id.
```
someaddons: not an int id

src/api/java/com/minecolonies/api/colony/fields/modules/IFieldModule.java
```diff
@@ -0,0 +1,24 @@
...
+     */
+    IField getField();
+
+    /**
+     * Set the field of the module.
+     *
+     * @param field the field to set.
+     * @return the module itself.
+     */
+    IFieldModule setField(final IField field);
```
someaddons: not sure if we should be allowed to change the field a module is associated with, should probably not exist/field is final and loaded on creation

src/api/java/com/minecolonies/api/colony/fields/plantation/BasicPlanterAI.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public interface BasicPlanterAI
+{
+    /**
+     * Makes the planter walk to a given block.
+     *
+     * @param blockPos the block to walk to.
+     * @return true whilst the planter is walking, false when he reached the position.
+     */
+    boolean planterWalkToBlock(final BlockPos blockPos);
```
someaddons: walkto block already exists in the default AI, which this call gets redirected to, so not sure what the purpose of this is? Same probably with most other methods in here, think they should be just utilities?

src/api/java/com/minecolonies/api/colony/fields/plantation/IPlantationModule.java
```diff
@@ -0,0 +1,348 @@
...
+        /**
+         * Reset the current field.
+         */
+        FIELD
+    }
+
+    /**
+     * Class containing possible states that the planter AI can be in.
+     */
+    class PlanterAIModuleResult
```
someaddons: should probably be dropped alongside the AI in modules

src/api/java/com/minecolonies/api/colony/fields/registry/IFieldDataManager.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    IField fromBuffer(@NotNull final IColony colony, @NotNull final FriendlyByteBuf buf);
+
+    /**
+     * Generate a byte buffer from a field instance.
+     *
+     * @param field the field instance.
+     * @return the byte buffer.
+     */
+    FriendlyByteBuf toBuffer(@NotNull final IField field);
```
someaddons: Looks like a static util? the field Data manager instance isnt even used in serialization of a field, either make it a static util or put it within the field, same with createCompound below

src/api/java/com/minecolonies/api/tileentities/AbstractTileEntityPlantationField.java
```diff
@@ -0,0 +1,82 @@
...
+import net.minecraft.world.level.block.state.BlockState;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * The abstract implementation for plantation field tile entities.
+ */
+public abstract class AbstractTileEntityPlantationField extends BlockEntity implements IBlueprintDataProviderBE, IRotatableBlockEntity
```
someaddons, line 23: dont think we need this impl in the api either

src/api/java/com/minecolonies/api/tileentities/AbstractTileEntityScarecrow.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.minecolonies.api.colony.IColony;
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+
+/**
+ * The abstract implementation for farmer field tile entities.
+ */
+public abstract class AbstractTileEntityScarecrow extends BlockEntity
```
someaddons, line 11: same here probably not needed within api

src/main/java/com/minecolonies/coremod/colony/buildings/modules/FieldsModule.java
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * Method called to assign a field to the building.
+     *
+     * @param field the field to add.
+     */
+    public void assignField(final IField field)
+    {
+        if (canAssignField(field))
+        {
+            building.getColony().getBuildingManager().getField(otherField -> otherField.equals(field)).ifPresent(f -> {
```
someaddons: shouldnt this just be field.setBuilding(building.getID()) ?

src/main/java/com/minecolonies/coremod/colony/buildings/modules/FieldsModule.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Method called to free a field.
+     *
+     * @param field the field to be freed.
+     */
+    public void freeField(final IField field)
+    {
+        Optional<IField> existingFieldOpt = building.getColony().getBuildingManager().getField(otherField -> otherField.equals(field));
```
someaddons: same as above? do you have multiple instances of fields which are equal with each other? seems quite odd

src/api/java/com/minecolonies/api/colony/fields/registry/FieldRegistries.java
```diff
@@ -0,0 +1,210 @@
...
+     */
+    public static IForgeRegistry<FieldEntry> getFieldRegistry()
+    {
+        return IMinecoloniesAPI.getInstance().getFieldRegistry();
+    }
+
+    /**
+     * Entry for the {@link IField} registry. Makes it possible to create a single registry for a {@link IField}. Used to lookup how to create {@link IField}.
+     */
+    public static class FieldEntry
```
Nightenom, line 71: this could definitely be record, since they have built-in equals/hashes

src/main/resources/assets/minecolonies/gui/windowplantationfield.xml
```diff
@@ -0,0 +1,35 @@
...
+    <label size="130 11" pos="30 14" color="red" textalign="MIDDLE" label="$(block.minecolonies.blockhutplantationfield)"/>
+
+    <label id="not-in-colony" size="164 11" pos="13 55" color="black" textalign="MIDDLE"
+           label="$(com.minecolonies.coremod.gui.workerhuts.fields.notincolony)"/>
+
+    <label id="plants-header" size="130 11" pos="30 30" color="black" textalign="MIDDLE"
+           label="$(com.minecolonies.coremod.gui.workerhuts.plantationfields.plants)"/>
+
+    <list id="plants" size="138 32" pos="30 45">
+        <box size="130 32" linewidth="0">
```
Nightenom: use pure view if linewidth = 0

Nightenom (commented):
just two random fast quickies, btw, disable auto-format on huge projects (eg. use format selection)

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/plantation/AbstractPlantationModule.java
```diff
@@ -0,0 +1,181 @@
...
+import net.minecraft.world.level.Level;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Base class for planter modules that determines how the AI should work specific fields.
+ */
+public abstract class AbstractPlantationModule extends AbstractFieldModule implements IPlantationModule
```
someaddons, line 24: as mentioned on other places, this is probably not needed as AI/behaviour should be in the AI classes, modules should be providing the data necessary to make the decisions

src/main/java/com/minecolonies/coremod/entity/ai/citizen/herders/AbstractEntityAIHerder.java
```diff
@@ -696,8 +693,7 @@ protected void butcherAnimal(@Nullable final Animal animal)
         if (animal != null && !walkingToAnimal(animal) && !ItemStackUtils.isEmpty(worker.getMainHandItem()))
         {
             worker.swing(InteractionHand.MAIN_HAND);
-            final FakePlayer fp = FakePlayerFactory.getMinecraft((ServerLevel) worker.getCommandSenderWorld());
-            final DamageSource ds = DamageSource.playerAttack(fp);
+            final DamageSource ds = DamageSource.playerAttack(getFakePlayer());
```
someaddons, line 686: as mentioned already, I don't think we should be creating player specific drops

src/main/java/com/minecolonies/coremod/tileentities/TileEntityPlantationField.java
```diff
@@ -0,0 +1,401 @@
...
+     */
+    public TileEntityPlantationField(final BlockPos pos, final BlockState state)
+    {
+        super(MinecoloniesTileEntities.PLANTATION_FIELD.get(), pos, state);
+    }
+
+    @Override
+    public Set<FieldRegistries.FieldEntry> getPlantationFieldTypes()
+    {
+        return tagPosMap.values().stream()
```
someaddons: cache the result

src/main/java/com/minecolonies/coremod/tileentities/TileEntityPlantationField.java
```diff
@@ -0,0 +1,401 @@
...
+                 .flatMap(Collection::stream)
+                 .map(this::getPlantationFieldEntryFromFieldTag)
+                 .filter(Objects::nonNull)
+                 .collect(Collectors.toSet());
+    }
+
+    @Override
+    public List<BlockPos> getWorkingPositions(final String tag)
+    {
+        return tagPosMap.entrySet().stream()
```
someaddons: cache the result

someaddons (commented):
Overall pretty decent, but theres a few things I'm not a fan off, left notes :)

src/main/java/com/minecolonies/coremod/colony/buildings/modules/FieldsModule.java
Thodor12: The reason it's done like this is because the field instance provided here usually comes from serialized sources.
Meaning that this input field value is not actually tied to a field as present in the building manager, if I were to update that instance directly, nothing would happen.

someaddons (dismissed):

src/main/java/com/minecolonies/coremod/colony/buildings/workerbuildings/BuildingPlantation.java
```diff
@@ -86,317 +71,415 @@ public BuildingPlantation(final IColony c, final BlockPos l)
...
         @Override
-        public OptionalPredicate<ItemStack> getIngredientValidator()
+        protected int getMaxFieldCount()
         {
-            return CraftingUtils.getIngredientValidatorBasedOnTags(CRAFTING_PLANTATION).combine(super.getIngredientValidator());
+            int allowedPlants = (int) Math.ceil(building.getBuildingLevel() / 2D);
+            return building.getColony().getResearchManager().getResearchEffects().getEffectStrength(PLANTATION_LARGE) > 0
+                     ? allowedPlants + 1
+                     : allowedPlants;
         }
```
uecasm, line 250: Was this intended to be multiplication rather than division?  Because currently it's limiting to 1 plant and 2 fields regardless of level.

Also, if doing the research only adds a single extra field, it doesn't seem very worthwhile (especially if each level does add 20 fields, if it had been multiplication -- which also seems wrong).

Thodor12: Yes, this works as intended. The intended field limits are: `1 1 2 2 3`. This is what the division is for, for example level 3: `3 / 2 = 1.5 = 2`.

The research adds an extra field across the board, giving you access to 4 fields at max, you're confused with planting positions per field, each field can have a max amount of work tags, which is what you got the 20 from.

uecasm: Ohhh, that's a `D`.  I read it as `20`.  Might have been better to write it as `2.0` instead.

Thodor12: 2.0 is the notation for a float no? Reason I used D is to have it cast to a double.

uecasm: No.  `2.0` is a double unless you add an `F` at the end.  It's pretty rare to use the `D` suffix.

Thodor12: For some reason I had in my mind that writing decimal points in Java caused it to be a float, maybe am confused with C# here.

uecasm: All the C-based languages work the same as Java in that respect.

MotionlessTrain: I don't think float would matter in this case, as floats are cast to double automatically by java, when calling `Math.ceil`

PR #9044, opened 2023-04-04
Don't allow kids to be born into explicitly manual housing.

uecasm:
Closes [discord report](https://discord.com/channels/139070364159311872/1003402520846139533/1092714238012817478)

- Adds a new Locked mode to residence/tavern, that otherwise mostly behaves like Manual.
- If a residence/tavern is set explicitly to ~~Manual~~ Locked, then kids won't be born into it, even if it has spare beds.
    - Kids will still be born if set to ~~Default~~ any other mode, regardless of whether the colony as a whole is set to Manual or Auto.

Review please (could port)

In particular, this allows you to deliberately leave some housing vacant, such as leaving a tavern empty once you have level 4/5 housing, or having a single-person house at the edge of your colony for the colonist with the funny smell... (without making them unhappy that they only have a level 1 house)

### DIFF
--- a/src/api/java/com/minecolonies/api/colony/buildings/HiringMode.java
+++ b/src/api/java/com/minecolonies/api/colony/buildings/HiringMode.java
@@ -9,7 +9,8 @@ public enum HiringMode
 {
     DEFAULT(HIRING_MODE_DEFAULT),
     AUTO(HIRING_MODE_AUTOMATIC),
-    MANUAL(HIRING_MODE_MANUAL);
+    MANUAL(HIRING_MODE_MANUAL),
+    LOCKED(HIRING_MODE_LOCKED);
 
     private final String translationKey;
 
--- a/src/api/java/com/minecolonies/api/util/constant/TranslationConstants.java
+++ b/src/api/java/com/minecolonies/api/util/constant/TranslationConstants.java
@@ -442,6 +442,8 @@ public final class TranslationConstants
     @NonNls
     public static final String HIRING_MODE_MANUAL                                                   = "com.minecolonies.coremod.gui.hiringmode.manual";
     @NonNls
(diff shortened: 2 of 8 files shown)


Raycoms (dismissed):
Had this on my own todo list too, nice =D

someaddons:
the functionality is nice, but we need a way to communicate this to the player, else the next report will be just "I have open beds in the townhall, why do I not get kids". Maybe display it in the Townhall population? Or add a note next to putting hosues on manual

Raycoms:
> the functionality is nice, but we need a way to communicate this to the player, else the next report will be just "I have open beds in the townhall, why do I not get kids". Maybe display it in the Townhall population? Or add a note next to putting hosues on manual

I think he pushed something to the townhall display in this regard to.

Raycoms:
> > the functionality is nice, but we need a way to communicate this to the player, else the next report will be just "I have open beds in the townhall, why do I not get kids". Maybe display it in the Townhall population? Or add a note next to putting hosues on manual
> 
> I think he pushed something to the townhall display in this regard to.

But yes, something at the option in housing should maybe clarify this too

someaddons:
maybe just a simple "Manual(no kids)" for just residential houses

Raycoms:
We could even have for residential houses an additional setting? Automatic/Default/Manual/Manual-NoKids

Talyda:
> We could even have for residential houses an additional setting? Automatic/Default/Manual/Manual-NoKids

Need to add it to the tavern as well if that's how you want to deal with it

uecasm:
image

MotionlessTrain:
Locked means it works the same as manual mode (just no kids)?

uecasm:
Yes, I already updated the description to clarify that.

Raycoms (approved):

someaddons (approved):

PR #9046, opened 2023-04-04
Feature/quests

Raycoms:
Closes #
Closes #
Closes #

- Adds quests to Minecolonies.
- Adds forge registries to handle the different triggers etc loading from json.
- Adds data driven research system made up currently of:

Triggers:
- Random
- Colony State
- Citizen State

Objectives:
- Arbitrary Dialogue Trees
- Delivery
- Block Breaking
- Entity Killing

Rewards:
- Items
- Happiness
- Research
- Raid
- Relationship status change
- Skill change

Different answer results for dialogues:
- Return (do quest later)
- Cancel (put back into pool)
- GoTo (go to a specific objective from here)
- More Dialogue

I also added some example quests and a template with all options.

What will be done in the future:

- Objective based side effects (make citizen walk to pos x ob objective start, or make citizen nor work for x days etc)
- Moar Quests
- Moar Triggers
- Moar Objectives
- Moar Rewards

And at some point, we need a system to translate quests without making the jsons unreadable.

Review please


src/api/java/com/minecolonies/api/colony/IColony.java
```diff
@@ -475,4 +478,17 @@ default List<BlockPos> getWayPoints(@NotNull BlockPos position, @NotNull BlockPo
...
+     * @return the quest manager.
+     */
+    IQuestManager getQuestManager();
+
+    /**
+     * Get citizen from colony.
+     * @param id the id of the cit.
+     * @return the cit.
+     */
+    ICitizen getCitizen(int id);
```
marchermans, line 493: Is this not handled by the citizen manager?
If not I think you should move it there.

src/api/java/com/minecolonies/api/entity/citizen/happiness/StaticHappinessModifier.java
```diff
@@ -10,7 +10,7 @@ public class StaticHappinessModifier extends AbstractHappinessModifier
     /**
      * The supplier to get the happiness factor.
      */
-    private final DoubleSupplier supplier;
+    protected DoubleSupplier supplier;
```
marchermans: This feels wrong.
I see no inheritor actually using this supplier.

And even if they did, you might want to write a tiny abstraction over this instead of just exposing the supplier directly.

src/api/java/com/minecolonies/api/quests/IQuestObjective.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    default void onAbort(final IColonyQuest colonyQuest) { }
+
+    /**
+     * On world load trigger.
+     * @param colonyQuest the quest.
+     */
+    default void onWorldLoad(IColonyQuest colonyQuest) { };
+
+    @Nullable
```
marchermans: Missing documentation.

src/api/java/com/minecolonies/api/quests/ITerminalAnswerResult.java
```diff
@@ -0,0 +1,14 @@
+package com.minecolonies.api.quests;
+
+import net.minecraft.world.entity.player.Player;
+
+public interface ITerminalAnswerResult extends IAnswerResult
```
marchermans: Missing documentation.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -272,6 +276,22 @@ public class CitizenData implements ICitizenData
...
+    /**
+     * The list of participating quests the citizen can give out.
+     */
+    private final List<ResourceLocation> participatingQuests = new ArrayList<>();
+
+    /**
+     * Tracking quests the citizen participated in and finished their participation.
+     */
+    private final List<ResourceLocation> finishedQuests = new ArrayList<>();
+    private final List<ResourceLocation> finishedQuestParticipation = new ArrayList<>();
```
marchermans, line 296: Missing documentation.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractAISkeleton.java
```diff
@@ -124,6 +124,12 @@ public final void stop()
         worker.getCitizenData().setVisibleStatus(null);
     }
 
+    @Override
+    public boolean requiresUpdateEveryTick()
+    {
+        return true;
```
marchermans: Why is this the new default?
Some comment as to why the default is to update everytick would be nice to prevent confusion in the future.

src/main/java/com/minecolonies/coremod/quests/rewards/HappinessReward.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Target to affect.
+     */
+    private final int target;
+
+    /**
+     * Setup the research reward.
+     */
+    public HappinessReward(final int target, final int qty, final int days)
```
marchermans: Would it possibly be a good idea to expose creating these somehow in the API?

marchermans (changes requested):
Minor nitpickings and questions. All in all looks good.

src/main/java/com/minecolonies/coremod/entity/ai/basic/AbstractAISkeleton.java
Raycoms: Was merged in another branch into main already this week too. We used to tick every tick in the past but mojang added this to reduce ticks of some of their entities and also nerfed ours

src/main/java/com/minecolonies/coremod/quests/rewards/HappinessReward.java
Raycoms: Happiness changes?

src/api/java/com/minecolonies/api/entity/citizen/happiness/StaticHappinessModifier.java
Raycoms: I added one to write to this supplier for the happiness reward.

src/api/java/com/minecolonies/api/colony/IColony.java
Raycoms: Its because of the client/server separation. citizen manager doesnt exist on the client. Would require quite the large rework to make sure we dont expose methods to the client that are unusable.

src/api/java/com/minecolonies/api/colony/ICitizenData.java
```diff
@@ -370,4 +373,24 @@ default boolean hasCustomTexture()
     {
         return false;
     }
+
+    /**
+     * Sets the citizen idle for a certain amount of days
+     *
+     * @param days amount of days
+     */
+    void setIdleDays(int days);
```
uecasm, line 382: A getter would be nice as well.

src/api/java/com/minecolonies/api/colony/interactionhandling/IInteractionResponseHandler.java
```diff
@@ -27,6 +27,16 @@ public interface IInteractionResponseHandler extends INBTSerializable<CompoundTa
      */
     Component getInquiry();
 
+    /**
+     * The inquiry of the GUI to the player. This is the key for the interaction, functions as id.
```
uecasm, line 31: "Functions as id" doesn't seem right given that the other methods are using integer ids now.  And there's a separate `getId`.

src/api/java/com/minecolonies/api/colony/interactionhandling/IInteractionResponseHandler.java
```diff
@@ -129,4 +139,20 @@ default ResourceLocation getInteractionIcon()
...
+    /**
+     * Trigger on opening the interaction.
+     */
+    default void onOpened(final Player player) {}
+
+    /**
+     * Get the id.
+     * @return the id.
+     */
+    Component getId();
```
uecasm, line 157: Why is the id a `Component`?  Shouldn't it be an `int` or `ResourceLocation`?

src/api/java/com/minecolonies/api/entity/citizen/happiness/StaticHappinessModifier.java
uecasm: It does seem like it would be nicer if you added a new modifier class that was mutable or replaced the old instance with a new immutable instance whenever the reward changed, rather than adding a setter and mutable supplier to all of them.

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
```diff
@@ -0,0 +1,117 @@
...
+     * Get the colony matching the quest.
+     * @return the colony.
+     */
+    IColony getColony();
+
+    /**
+     * Get the player UUID that accepted the quest.
+     * @return the player uuid.
+     */
+    UUID getAssignedPlayer();
```
uecasm: What about co-op quests?  Alice accepts the quest, but Bob helps them collect whatever and wants to turn it in, but can't?

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * A dialogue element in the dialogue objective.
+     */
+    class DialogueElement implements IAnswerResult
+    {
+        /**
+         * The text the participant says.
+         */
+        private final String text;
```
uecasm: Should this perhaps be a Component, to allow complex formatting (or e.g. showing items with tooltips when asking for certain things).  Could use the default JSON serialization/deserialization.

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * An answer element part of a dialogue path.
+     */
+    class AnswerElement
+    {
+        /**
+         * The text the player displays.
+         */
+        private final String text;
```
uecasm: Likewise.  Though I assume these are probably shorter.

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
```diff
@@ -0,0 +1,136 @@
...
+
+        /**
+         * Parse the answer element from json.
+         * @param jsonObject the json obj.
+         * @return the answer element.
+         */
+        public static AnswerElement parse(final JsonObject jsonObject)
+        {
+            final JsonObject resultObj = jsonObject.getAsJsonObject(RESULT_ID);
+            return new AnswerElement(jsonObject.get(ANSWER_ID).getAsString(), IMinecoloniesAPI.getInstance().getQuestDialogueAnswerRegistry().getValue(new ResourceLocation(resultObj.get(TYPE_ID).getAsString())).produce(resultObj));
```
uecasm: If this is called to parse objects that go into a registry (as it looks like), you shouldn't be directly consuming another registry, because they load in indeterminate order AFAIK (even more so in 1.19.3+).  You should probably just be storing the `ResourceLocation` and looking up the real value later.

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestDialogueInteraction.java
```diff
@@ -0,0 +1,268 @@
...
+        if (colonyQuest == null)
+        {
+            colonyQuest = data.getColony().getQuestManager().getAvailableOrInProgressQuest(questId);
+        }
+        if (currentElement != null && colonyQuest != null)
+        {
+            final IAnswerResult result = this.currentElement.getOptionResult(responseId);
+            if (result instanceof ITerminalAnswerResult)
+            {
+                ((ITerminalAnswerResult) result).applyToQuest(player, data.getColony().getQuestManager().getAvailableOrInProgressQuest(questId));
```
uecasm: Pattern matching.

(I'll stop pointing these out now, but the IDE should be pointing out more cases -- at least mine does.)

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return answers.stream().map(answerElement -> Component.literal(answerElement.text)).collect(Collectors.toList());
+        }
+
+        /**
+         * Get the matching answer result.
+         * @param responseId the triggered response to match.
+         * @return the next answer.
+         */
+        public IAnswerResult getOptionResult(final int responseId)
```
uecasm: Mark as `@Nullable` so you get more warnings for unchecked usage.

src/api/java/com/minecolonies/api/quests/IQuestData.java
```diff
@@ -0,0 +1,67 @@
...
+     * Get the objective count of this quest.
+     * @return the count.
+     */
+    int getObjectiveCount();
+
+    /**
+     * The name of the quest.
+     * @return the name of the quest.
+     */
+    String getName();
```
uecasm: Should this be `ResourceLocation` or `Component`?

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -988,6 +1007,18 @@ public void serializeViewNetworkData(@NotNull final FriendlyByteBuf buf)
         }
         buf.writeUtf(parents.getA());
         buf.writeUtf(parents.getB());
+
+        buf.writeInt(availableQuests.size());
+        for (final ResourceLocation av : availableQuests)
+        {
+            buf.writeResourceLocation(av);
```
uecasm, line 1018: When the resource location represents an entry in a synched registry, you should use `writeRegistryIdUnsafe` instead of `writeResourceLocation`.  (I disagree with the "unsafety" of that, it's only unsafe if you somehow use entirely the wrong registry at the other end, which is always a coding bug anyway, and using the "safe" version negates all benefits.)

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -1644,4 +1736,77 @@ public void setParents(final String firstParent, final String secondParent)
...
+    @Override
+    public void assignQuest(final IColonyQuest quest)
+    {
+        this.availableQuests.add(quest.getId());
+    }
+
+    @Override
+    public void openDialogue(final IColonyQuest quest, final int index)
+    {
+        final Component comp = Component.literal(quest.getId().toString());
```
uecasm, line 1759: This is the other half of "why is the interaction system using a Component as a key?"  It should only be used for display text, not as a lookup key and not to wrap a resource location.

src/api/java/com/minecolonies/api/quests/IQuestManager.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param questID the id of the quest.
+     */
+    void deactivateQuest(ResourceLocation questID);
+
+    /**
+     * Get the currently available or in progress quest with a given id.
+     * @param questId the id.
+     * @return the quest.
+     */
+    IColonyQuest getAvailableOrInProgressQuest(final ResourceLocation questId);
```
uecasm: `@Nullable`?

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestActionInteraction.java
```diff
@@ -0,0 +1,156 @@
...
+        if (currentElement != null && colonyQuest != null)
+        {
+            final IAnswerResult result = this.currentElement.getOptionResult(responseId);
+            if (result instanceof ITerminalAnswerResult)
+            {
+                if (result instanceof IResolveResult)
+                {
+                    if (((IQuestActionObjective) objective).isReady(player, colonyQuest) && ((IQuestActionObjective) objective).tryResolve(player, colonyQuest))
+                    {
+                        ((ITerminalAnswerResult) result).applyToQuest(player, data.getColony().getQuestManager().getAvailableOrInProgressQuest(questId));
```
uecasm: `result` can be pattern matched.

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestActionInteraction.java
```diff
@@ -0,0 +1,156 @@
...
+                    }
+                }
+                currentElement = startElement;
+            }
+            else if (result instanceof DialogueObjective.DialogueElement)
+            {
+                this.currentElement = (DialogueObjective.DialogueElement) result;
+                if (data != null && data.getJob() != null)
+                {
+                    ((AbstractEntityAIBasic) data.getJob().getWorkerAI()).setDelay(TICKS_SECOND * 3);
```
uecasm: Consider adding to API rather than downcasting?  Unless you don't want addons to implement their own AI from the interface...

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestActionInteraction.java
```diff
@@ -0,0 +1,156 @@
...
+     * Trigger response state.
+     *
+     * @param player    the player for.
+     * @param objective the objective to check.
+     */
+    private void triggerResponseState(final Player player, final IQuestObjective objective)
+    {
+        if (objective instanceof IQuestActionObjective)
+        {
+            if (((IQuestActionObjective) objective).isReady(player, colonyQuest))
```
uecasm: Pattern matching

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestDialogueInteraction.java
```diff
@@ -0,0 +1,268 @@
...
+        if (colonyQuest == null && citizen != null)
+        {
+            colonyQuest = citizen.getColony().getQuestManager().getAvailableOrInProgressQuest(questId);
+        }
+    }
+
+    @Override
+    public Component getId()
+    {
+        return Component.literal(this.questId.toString());
```
uecasm, line 171: 😢

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestDialogueInteraction.java
```diff
@@ -0,0 +1,268 @@
...
+        return tag;
+    }
+
+    @Override
+    public void deserializeNBT(final @NotNull CompoundTag compoundNBT)
+    {
+        super.deserializeNBT(compoundNBT);
+        this.questId = new ResourceLocation(compoundNBT.getString(TAG_QUEST_ID));
+        this.index = compoundNBT.getInt(TAG_QUEST_INDEX);
+        this.currentElement = ((DialogueObjective) IQuestManager.GLOBAL_SERVER_QUESTS.get(questId).getObjective(index)).getDialogueTree();
```
uecasm: This won't cope with the quest no longer existing.

src/main/java/com/minecolonies/coremod/colony/interactionhandling/RequestBasedInteraction.java
```diff
@@ -164,9 +164,9 @@ public void onWindowOpened(final BOWindow window, final ICitizenDataView dataVie
 
     @Override
     @OnlyIn(Dist.CLIENT)
-    public boolean onClientResponseTriggered(final Component response, final Player player, final ICitizenDataView data, final BOWindow window)
+    public boolean onClientResponseTriggered(final int responseId, final Player player, final ICitizenDataView data, final BOWindow window)
     {
-        if (response.equals(Component.translatable("com.minecolonies.coremod.gui.chat.fulfill")))
+        if (((TranslatableContents) getPossibleResponses().get(responseId).getContents()).getKey().equals("com.minecolonies.coremod.gui.chat.fulfill"))
```
uecasm, line 169: 😢
Also, this cast will fail if the response might be a quest literal resource location.

src/main/java/com/minecolonies/coremod/datalistener/QuestJsonListener.java
```diff
@@ -0,0 +1,366 @@
...
+    /**
+     * Our universal apply.
+     * @param jsonElementMap the map.
+     */
+    private static void apply(final Map<ResourceLocation, JsonElement> jsonElementMap)
+    {
+        Log.getLogger().info("Loading quests from data");
+
+        // We start by clearing all old requests.
+        IQuestManager.GLOBAL_SERVER_QUESTS.clear();
```
uecasm, line 100: Any particular reason this whole thing is not also a forge registry?

src/main/java/com/minecolonies/coremod/datalistener/QuestJsonListener.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    /**
+     * Recursively parses a string condition to a predicate
+     *
+     * @param colony the colony.
+     * @param data split string data
+     * @return predicate from data
+     */
+    private static List<ITriggerReturnData> evaluate(final IColony colony, final Map<String, IQuestTrigger> triggerMap, final List<String> data, final List<ITriggerReturnData> lastReturnData)
```
uecasm: Wouldn't it be nicer to datapack authors to use existing conventions (e.g. advancement triggers or list-of-lists)?

src/main/java/com/minecolonies/coremod/permissions/ColonyPermissionEventHandler.java
```diff
@@ -244,7 +261,23 @@ else if (event.getState().getBlock() instanceof BlockDecorationController)
...
         else
         {
-            checkEventCancelation(Action.BREAK_BLOCKS, event.getPlayer(), event.getPlayer().getCommandSenderWorld(), event, event.getPos());
+            if (!checkEventCancelation(Action.BREAK_BLOCKS, event.getPlayer(), event.getPlayer().getCommandSenderWorld(), event, event.getPos()) && event.getPlayer() != null)
+            {
+                final Block block = event.getLevel().getBlockState(event.getPos()).getBlock();
+                if (breakBlockObjectives.containsKey(block) && breakBlockObjectives.get(block).containsKey(event.getPlayer().getUUID()))
+                {
+                    final IColonyQuest colonyQuest =  breakBlockObjectives.get(block).get(event.getPlayer().getUUID());
+                    final IQuestObjective objective = IQuestManager.GLOBAL_SERVER_QUESTS.get(colonyQuest.getId()).getObjective(colonyQuest.getIndex());
```
uecasm: unchecked

src/main/java/com/minecolonies/coremod/permissions/ColonyPermissionEventHandler.java
```diff
@@ -599,4 +632,70 @@ public void on(final AttackEntityEvent event)
...
         }
     }
+
+    @SubscribeEvent
+    public void on(final LivingDeathEvent event)
+    {
+        if (event.getSource().getEntity() instanceof Player && entityKillObjectives.containsKey(event.getEntity().getType()) && entityKillObjectives.get(event.getEntity().getType()).containsKey(event.getSource().getEntity().getUUID()))
+        {
+            final IColonyQuest colonyQuest =  entityKillObjectives.get(event.getEntity().getType()).get(event.getSource().getEntity().getUUID());
+            final IQuestObjective objective = IQuestManager.GLOBAL_SERVER_QUESTS.get(colonyQuest.getId()).getObjective(colonyQuest.getIndex());
```
uecasm: unchecked

src/main/java/com/minecolonies/coremod/quests/QuestData.java
```diff
@@ -0,0 +1,121 @@
...
+    private final int maxOccurrence;
+
+    private final String name;
+
+    /**
+     * How long the quest stays active/available.
+     */
+    private final int questTimeout;
+
+    // We want two different things. a) A global "quest holder" kind of thing that has all the fields that we read from the json (e.g. the necessary triggers, side effects, etc)
```
uecasm: There are two types of people: a) those who finish lists

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
```diff
@@ -0,0 +1,220 @@
...
+        if (inProgressQuests.containsKey(questId))
+        {
+            inProgressQuests.remove(questId);
+            finishedQuests.put(questId, finishedQuests.getOrDefault(questId, 0) + 1);
+        }
+        else if (availableQuests.containsKey(questId))
+        {
+            // When a player short-cut quits a job without accepting it. (E.g. been there, done that options).
+            availableQuests.remove(questId);
+            finishedQuests.put(questId, finishedQuests.getOrDefault(questId, 0) + 1);
```
uecasm, line 72: Should this be counted as "finished" if they skipped it?

src/main/java/com/minecolonies/coremod/quests/objectives/BreakBlockObjective.java
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * @param target       the target citizen.
+     * @param blocksToMine the number of blocks to mine.
+     * @param blockToMine  the block to mine.
+     * @param rewards the rewards this unlocks.
+     */
+    public BreakBlockObjective(final int target, final int blocksToMine, final Block blockToMine, final int nextObjective, final List<Integer> rewards)
+    {
+        super(target, new DialogueElement("I am still waiting for you to mine %d " + blockToMine.getName().getString() + " !",
+          List.of(new AnswerElement("Sorry, be right back!", new IAnswerResult.ReturnResult()), new AnswerElement("I don't have time for this!", new IAnswerResult.CancelResult()))), rewards);
```
uecasm: Translation?

src/main/java/com/minecolonies/coremod/quests/objectives/DeliveryObjective.java
```diff
@@ -0,0 +1,131 @@
...
+        this.buildDialogueTrees();
+    }
+
+    private void buildDialogueTrees()
+    {
+        this.readyDialogueElement = new DialogueElement("Oh hey, you brought " + item.getDisplayName().getString() + " can I have it?",
+          List.of(new AnswerElement("Yes, here you are!", new IAnswerResult.GoToResult(this.nextObjective)), new AnswerElement("No, wait!", new IAnswerResult.ReturnResult())));
+
+        this.waitingDialogueElement = new DialogueElement("I am still waiting for " + item.getDisplayName().getString() + " !",
+          List.of(new AnswerElement("Sorry, be right back!", new IAnswerResult.ReturnResult()), new AnswerElement("I don't have any of it!", new IAnswerResult.CancelResult())));
```
uecasm: Translation?  Also, consider using item links, if blockui supports rendering them with tooltips properly.

src/main/java/com/minecolonies/coremod/quests/objectives/KillEntityObjective.java
```diff
@@ -0,0 +1,165 @@
...
+     * Parse the mine block objective from json.
+     * @param jsonObject the json to parse it from.
+     * @return a new objective object.
+     */
+    public static IQuestObjective createObjective(final JsonObject jsonObject)
+    {
+        JsonObject details = jsonObject.getAsJsonObject(DETAILS_KEY);
+        final int target = details.get(TARGET_KEY).getAsInt();
+        final int quantity = details.get(QUANTITY_KEY).getAsInt();
+        final EntityType<?> entityType = ForgeRegistries.ENTITY_TYPES.getHolder(new ResourceLocation(details.get(ENTITY_TYPE_KEY).getAsString())).get().get();
```
uecasm: This seems more convoluted than necessary.

src/main/resources/data/minecolonies/quests/general/hungrycourier.json
```diff
@@ -0,0 +1,128 @@
...
+  "rewards": [
+    {
+      "type": "minecolonies:item",
+      "details": {
+        "item": "minecraft:gold_ingot",
+        "qty": 1
+      }
+    },
+    {
+      "type": "minecolonies:happiness",
```
uecasm, line 119: Shouldn't this also reward hunger points?

src/main/resources/data/minecolonies/quests/general/thezombiemenace.json
```diff
@@ -0,0 +1,103 @@
...
+          }
+        }
+      }
+    }
+  ],
+  "objectives": [
+    {
+      "type": "minecolonies:dialogue",
+      "target": 0,
+      "text": "We have a problem. The Zombies in the nearby forest are causing trouble. Can you help us take care of them?",
```
uecasm, line 35: inb4 desert island colony

src/api/java/com/minecolonies/api/colony/ICitizenData.java
Raycoms: Part of the next part with side-effects.

src/api/java/com/minecolonies/api/colony/interactionhandling/IInteractionResponseHandler.java
Raycoms: Returns have int ids, the interaction still has those IDs. And making it other ids would be very very difficult and prob not backwards compat.

src/api/java/com/minecolonies/api/colony/interactionhandling/IInteractionResponseHandler.java
Raycoms: Returns have int ids, the interaction still has those IDs. And making it other ids would be very very difficult and prob not backwards compat.

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
Raycoms: Delivery handin can be done by others. Just action tracking like killing entities and breaking blocks has to be the assigned player.

src/api/java/com/minecolonies/api/entity/citizen/happiness/StaticHappinessModifier.java
Raycoms: H onestlyhappiness needs a whole rework to work properly. So, its better like this for now.

Raycoms: Adjusted based on convo

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
Raycoms: General question that is related to all of the quest system. Atm its all 100% json generated and 100% full text in the quests. So, at least for this version I dont intend to put translation constants all over the json files.
If we do this, we need a better way to ship translations for quests that doesnt involve spreading them all over the place as this makes the quest files unreadable.

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
Raycoms: Im not sure I got what you meant here, can we chat on discord about this?

src/api/java/com/minecolonies/api/quests/IQuestData.java
Raycoms: All related to translation. String for now.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
Raycoms: Yep, but that would be a non backwards compatible change to the system. So, for now we have to do this.

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestActionInteraction.java
Raycoms: I guess this is a todo for someaddons ai rework.

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestDialogueInteraction.java
Raycoms: Interaction will be killed when that happens yes. Handled

src/main/java/com/minecolonies/coremod/colony/interactionhandling/RequestBasedInteraction.java
Raycoms: Old stuff, not gonna rework another entire system in this already too big pr =P

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
Raycoms: Yes. So that it doesnt show up anymore, etc.

Raycoms: but still unlocks the fullowups.

src/main/java/com/minecolonies/coremod/quests/objectives/BreakBlockObjective.java
Raycoms: Same as everything, future problem.

src/main/java/com/minecolonies/coremod/quests/objectives/DeliveryObjective.java
Raycoms: Same as everything, future problem. Nothing is translated, wont make sense to have it half translated

src/main/java/com/minecolonies/coremod/quests/objectives/KillEntityObjective.java
Raycoms: If you find a better way

src/main/resources/data/minecolonies/quests/general/thezombiemenace.json
Raycoms: Always some trees around xD

src/main/resources/data/minecolonies/quests/general/hungrycourier.json
Raycoms: Why?

src/main/java/com/minecolonies/coremod/datalistener/QuestJsonListener.java
Raycoms: Didnt see a reason for it to be.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
Raycoms: We dont look it up from here, so it should be fine.

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
uecasm: I don't mean for translation constants specifically, but rather for styled text and item links.  (You know, how you can put an `[Item]` in chat and it has a tooltip showing the item details.  Would be especially important if at some point someone wants to make a "get me this specific DO block" quest...)

Having said that, at least for quests included in mods rather than standalone it would be trivial to support translation as well -- just specify a translation key and use placeholders for the entity names.  This will work if the text is just hard-coded too, since it automatically uses it as a literal if the key isn't found as is.

There even is a syntax for doing entity name replacements already, although it's probably a bit too awkward and verbose to want to write into the JSON manually.  But with a datagen though.... 😁 (Pity you can't add custom contents to components though.)

A datagen would help with "spreading translation keys over files" too, at least at the source level, similar to how the research system does it.  Alternatively you could have a map of translations directly within the quest files, but then you'd have to implement the lookup yourself since it's non-standard (and it will be trickier to get it to work properly since you have to do the translation client-side).

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
uecasm: It means you're sending 200 bytes instead of 20.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
uecasm: Why does that matter?  Change the legacy ones first (separate PR if needed).  Nothing outside of mcol itself depends on it (since it can't) so you don't risk breaking anything.  The only possible risk is for serialization, but I wouldn't have thought these were currently serialized.

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
uecasm: But why unlock the followup?  Someone has a quest "please deliver this apple for me", you respond "no, i don't have time, go away".  Then the target gives a quest "thanks for giving me that apple from X, can you give them this diamond in response?"  It makes no sense.

src/main/java/com/minecolonies/coremod/quests/objectives/KillEntityObjective.java
uecasm: `ForgeRegistries.ENTITY_TYPES.getValue(id)`

src/main/resources/data/minecolonies/quests/general/hungrycourier.json
uecasm: Because they asked for a potato because they were hungry.  Presumably, they would then eat it and be less hungry.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
Raycoms: They are serialized, thats what I meant

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
Raycoms: Well, this is up to the quest designer to design this. Else they just make it a cancel not a skip.

src/main/resources/data/minecolonies/quests/general/hungrycourier.json
Raycoms: Can be a future additional reward type.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
Raycoms: Ah, its not a registry entry, its just a quest id.

src/api/java/com/minecolonies/api/entity/citizen/happiness/IHappinessModifier.java
```diff
@@ -58,4 +58,11 @@
      * @return the days.
      */
     int getDays();
+
+    /**
+     * Method to set the values at a later day.
+     * @param qty the supplier value.
+     * @param days the period.
+     */
+    void set(int qty, int days);
```
someaddons: too generic name "set" and qty probably does not belong here? at least the supplier you're talking about is a specific isntance of a happinessmodifier not all of them

src/api/java/com/minecolonies/api/quests/IAnswerResult.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.google.gson.JsonObject;
+import net.minecraft.world.entity.player.Player;
+
+import static com.minecolonies.api.util.constant.QuestParseConstant.GO_TO_ID;
+
+/**
+ * Possible answer results in a dialogue tree.
+ */
+public interface IAnswerResult
```
someaddons: different name here, maybe IQuestDialogueAnswer?

src/api/java/com/minecolonies/api/quests/IAnswerResult.java
```diff
@@ -0,0 +1,71 @@
...
+{
+    /**
+     * Type that will close the quest without alteration.
+     */
+    class ReturnResult implements ITerminalAnswerResult
```
someaddons: FinalAnswer?

src/api/java/com/minecolonies/api/quests/IAnswerResult.java
```diff
@@ -0,0 +1,71 @@
...
+        public void applyToQuest(final Player player, final IColonyQuest quest)
+        {
+            // Do nothing, just close UI.
+        }
+    }
+
+    /**
+     * Will advance the quest to another objective.
+     */
+    class GoToResult implements IResolveResult
```
someaddons: NextObjective?

src/api/java/com/minecolonies/api/quests/IAnswerResult.java
```diff
@@ -0,0 +1,71 @@
...
+        public void applyToQuest(final Player player, final IColonyQuest quest)
+        {
+            quest.advanceObjective(player, nextObjective);
+        }
+    }
+
+    /**
+     * Cancel request and remove from in-progress and available pool for now.
+     */
+    class CancelResult implements ITerminalAnswerResult
```
someaddons: Cancel/Deny-Answer

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @return the id of the quest giver.
+     */
+    IQuestGiver getQuestGiver();
+
+    /**
+     * Id of questgiver.
+     * @return the id.
+     */
+    int getQuestGiverId();
```
someaddons: doesnt this belong into the questgiver itself? and then you use getQuestGiver.getId

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Check if the expiration date relative to the colony day count was reached.
+     * @param colony the colony to check the validity for.
+     * @return true if so.
+     */
+    boolean isValid(IColony colony);
```
someaddons: isExpired? would need to invert return value. Or checkTime()

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Get the quest id of the quest.
+     * @return the id.
+     */
+    ResourceLocation getId();
+
+    /**
+     * On deletion of the quest.
+     */
+    void onDeletion();
```
someaddons: is that called on user cancellation? if so should probably be called onCancelled

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * On question completion call.
+     */
+    void onCompletion();
+
+    /**
+     * Get the current objective index.
+     * @return the index number.
+     */
+    int getIndex();
```
someaddons: Are we limited to one objective active at a time? So no concurrent ones like: Gather 10 apples and 5 carrots

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param target the target participant id.
+     * @return the quest participant.
+     */
+    IQuestParticipant getParticipant(int target);
+
+    /**
+     * Get the full list of quest participants.
+     * @return the list of participants.
+     */
+    List<Integer> getParticipants();
```
someaddons: Are participants other assigned players?

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
```diff
@@ -0,0 +1,138 @@
...
+import com.minecolonies.api.IMinecoloniesAPI;
+import net.minecraft.network.chat.Component;
+import net.minecraft.resources.ResourceLocation;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static com.minecolonies.api.util.constant.QuestParseConstant.*;
```
someaddons: wonder if we maybe want to merge this and interactions, kinda the same system

src/api/java/com/minecolonies/api/quests/IObjectiveData.java
```diff
@@ -0,0 +1,22 @@
+package com.minecolonies.api.quests;
+
+import net.minecraft.nbt.CompoundTag;
+import net.minecraftforge.common.util.INBTSerializable;
+
+/**
+ * Objective data type to take track of activities.
+ */
+public interface IObjectiveData extends INBTSerializable<CompoundTag>
```
someaddons: IObjectiveType

src/api/java/com/minecolonies/api/quests/IQuestActionObjective.java
```diff
@@ -0,0 +1,29 @@
+package com.minecolonies.api.quests;
+
+import net.minecraft.world.entity.player.Player;
+
+/**
+ * Quest objective interface only for action types (e.g. deliveries).
+ */
+public interface IQuestActionObjective
```
someaddons: action sounds arbitrary, what kind of actions would this be?

src/api/java/com/minecolonies/api/quests/IQuestActionObjective.java
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * Check if the objective is ready to move on.
+     * @param colonyQuest the objective belongs to.
+     * @return true if so.
+     */
+    boolean isReady(final Player player, final IColonyQuest colonyQuest);
```
someaddons: isFulfilled?

src/api/java/com/minecolonies/api/quests/IQuestActionObjective.java
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Attempt to resolve an objective.
+     * @param colonyQuest the objective belongs to.
+     * @return true if so.
+     */
+    boolean tryResolve(final Player player, final IColonyQuest colonyQuest);
```
someaddons: Resolve and objective to what and why? Not sure why this exists

src/api/java/com/minecolonies/api/quests/IQuestActionObjectiveData.java
```diff
@@ -0,0 +1,13 @@
+package com.minecolonies.api.quests;
+
+/**
+ * Quest objective data. For in progress quests (e.g. kill x monsters, mine x blocks, etc).
+ */
+public interface IQuestActionObjectiveData
```
someaddons: Action/Data too generic here too, ObjectiveProgress?

src/api/java/com/minecolonies/api/quests/IQuestData.java
```diff
@@ -0,0 +1,67 @@
...
+import com.minecolonies.api.colony.IColony;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.player.Player;
+
+import java.util.List;
+
+/**
+ * Quest Data Instance.
+ */
+public interface IQuestData
```
someaddons: isnt this simply IQuest?

src/api/java/com/minecolonies/api/quests/IQuestObjective.java
```diff
@@ -0,0 +1,45 @@
+package com.minecolonies.api.quests;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+/**
+ * Quest objective interface for all objective types.
+ */
+public interface IQuestObjective
```
someaddons: IQuestObjectiveType?

src/api/java/com/minecolonies/api/quests/IQuestObjective.java
```diff
@@ -0,0 +1,45 @@
...
+{
+    /**
+     * Initialization of an objective.
+     *
+     * @param colonyQuest the colony quest it belongs to.
+     * @return potentially related objective data.
+     */
+    @Nullable
+    IObjectiveData init(final IColonyQuest colonyQuest);
```
someaddons: createObjectiveFor?

src/api/java/com/minecolonies/api/quests/IQuestParticipant.java
```diff
@@ -0,0 +1,50 @@
...
+
+import net.minecraft.resources.ResourceLocation;
+
+/**
+ * Type of entity that participates somehow in quests.
+ */
+public interface IQuestParticipant
+{
+    /**
+     * Check if the quest giver is still available.
```
someaddons: *participant not quest giver I assume

src/api/java/com/minecolonies/api/quests/IQuestParticipant.java
```diff
@@ -0,0 +1,50 @@
...
+     * Notify quest participant about their participation.
+     * @param quest the quest to assign to the entity.
+     */
+    void addQuestParticipation(final IColonyQuest quest);
+
+    /**
+     * Method for cleanup purposes on quest deletion.
+     * @param questId the id of the quest.
+     */
+    void onQuestDeletion(final ResourceLocation questId);
```
someaddons, line 26: onQuestCancelled

src/api/java/com/minecolonies/api/quests/IQuestParticipant.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Check if the citizen has a quest open.
+     * @param questId the id of the quest.
+     * @return true if so.
+     */
+    boolean hasQuestOpen(final ResourceLocation questId);
```
someaddons: isParticipantFor(

src/api/java/com/minecolonies/api/quests/IQuestParticipant.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Initiates a dialogue at the citizen.
+     * @param quest the quest to open it for.
+     * @param index the current objective index
+     */
+    void openDialogue(final IColonyQuest quest, final int index);
+
+    /**
+     * Get a display name for the quest giver.
```
someaddons: *participant not q giver

src/api/java/com/minecolonies/api/quests/IQuestReward.java
```diff
@@ -0,0 +1,18 @@
+package com.minecolonies.api.quests;
+
+import com.minecolonies.api.colony.IColony;
+import net.minecraft.world.entity.player.Player;
+
+/**
+ * Quest objective interface for all objective types.
+ */
+public interface IQuestReward
```
someaddons: javadoc doesnt fit the name, c&p?

src/api/java/com/minecolonies/api/quests/IQuestTrigger.java
```diff
@@ -0,0 +1,127 @@
...
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+import com.minecolonies.api.colony.IColony;
+import net.minecraft.nbt.*;
+
+/**
+ * Quest trigger interface.
```
someaddons: What does a quest trigger do? Use that as doc instead

src/api/java/com/minecolonies/api/quests/IQuestTrigger.java
```diff
@@ -0,0 +1,127 @@
...
+ */
+public interface IQuestTrigger
+{
+    /**
+     * Check if the quest trigger condition is fulfilled.
+     * @return true if so.
+     */
+    ITriggerReturnData isFulfilledForColony(final IColony colony);
```
someaddons: canTrigger(colony)?

src/api/java/com/minecolonies/api/quests/IQuestTrigger.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Match a nbt tag and a json element tag.
+     * @param nbtTag the nbt tag to check.
+     * @param matchTag the element tag to check.
+     * @return true if the matchTag fits into the nbtTag or if they match.
+     */
+    default boolean matchNbt(final Tag nbtTag, final JsonElement matchTag)
```
someaddons: too generic, for what purpose are we matching NBT here? This looks more like a generic static Utility function

src/api/java/com/minecolonies/api/quests/IResolveResult.java
```diff
@@ -0,0 +1,9 @@
+package com.minecolonies.api.quests;
+
+/**
+ * Positive terminal answer result.
+ */
+public interface IResolveResult extends ITerminalAnswerResult
```
someaddons: What does this do? Unclear naming/doc

src/api/java/com/minecolonies/api/quests/ITerminalAnswerResult.java
```diff
@@ -0,0 +1,17 @@
+package com.minecolonies.api.quests;
+
+import net.minecraft.world.entity.player.Player;
+
+/**
+ * Terminal type answer. Will close the interaction.
+ */
+public interface ITerminalAnswerResult extends IAnswerResult
```
someaddons: do we need this? couldnt the interaction auto-close if the answer has no further chain?

src/api/java/com/minecolonies/api/quests/ITriggerReturnData.java
```diff
@@ -0,0 +1,19 @@
+package com.minecolonies.api.quests;
+
+/**
+ * Custom return data for triggers.
+ */
+public interface ITriggerReturnData<T>
```
someaddons, line 6: What is this used for? Naming is rather generic, same for the calls

src/api/java/com/minecolonies/api/util/constant/QuestParseConstant.java
```diff
@@ -0,0 +1,40 @@
+package com.minecolonies.api.util.constant;
```
someaddons: can we move this into a quest related package instead? Better than gathering all constants in a folder

src/main/java/com/minecolonies/coremod/client/gui/WindowInteraction.java
```diff
@@ -97,6 +102,7 @@ private void setupInteraction()
             button.setTextAlignment(Alignment.MIDDLE);
             button.setText(component);
             group.addChild(button);
+            button.setTextScale(Math.min(1, 20.0 / component.getString().length()));
```
someaddons, line 105: just a note: we should add the scaling automatically on BlockUI

src/main/java/com/minecolonies/coremod/client/gui/WindowInteraction.java
```diff
@@ -121,17 +138,19 @@ public void onButtonClicked(@NotNull final Button button)
...
-                    }
+                    currentInteraction++;
+                    setupInteraction();
+                    return;
                 }
+                setupInteraction();
+            }
+            catch (final Exception ex)
+            {
+                Log.getLogger().warn("Wrong button id of interaction.");
```
someaddons: add the exception to the log too with ", ex"

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -272,6 +275,26 @@ public class CitizenData implements ICitizenData
...
+
+    /**
+     * Tracking quests the citizen was the quest giver in.
+     */
+    private final List<ResourceLocation> finishedQuests = new ArrayList<>();
+
+    /**
+     * Tracking quests the citizen participated in.
+     */
+    private final List<ResourceLocation> finishedQuestParticipation = new ArrayList<>();
```
someaddons, line 296: How about we move all of these to the quests manager instead of splitting data over different instances?

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -1271,11 +1336,18 @@ public void deserializeNBT(final CompoundTag nbtTagCompound)
...
+                {
+                    final ServerCitizenInteraction handler =
+                      (ServerCitizenInteraction) MinecoloniesAPIProxy.getInstance()
+                                                   .getInteractionResponseHandlerDataManager()
+                                                   .createFrom(this, handlerTagList.getCompound(i).getCompound(TAG_CHAT_OPTION));
+                    citizenChatOptions.put(handler.getId(), handler);
+                }
+                catch (final Exception ex)
+                {
+                    Log.getLogger().warn("Failed to load Interaction for a quest. Did the quest vanish?");
```
someaddons: log the exception here too:   ,ex

src/main/java/com/minecolonies/coremod/colony/interactionhandling/QuestActionInteraction.java
```diff
@@ -0,0 +1,156 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.player.Player;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+
+import static com.minecolonies.api.colony.interactionhandling.ModInteractionResponseHandlers.QUEST_ACTION;
+import static com.minecolonies.api.util.constant.Constants.TICKS_SECOND;
+
+/**
+ * A simple interaction which displays until an acceptable response is clicked
```
someaddons: same javadoc as parent class?

src/main/java/com/minecolonies/coremod/datalistener/QuestJsonListener.java
```diff
@@ -0,0 +1,366 @@
...
+
+import java.util.*;
+import java.util.function.Function;
+
+import static com.minecolonies.coremod.quests.QuestParsingConstants.*;
+import static com.minecolonies.coremod.quests.QuestParsingConstants.BRACE_CLOSE;
+import static org.apache.commons.lang3.StringUtils.EMPTY;
+
+/**
+ * Loader for Json based crafter specific recipes
```
someaddons: crafter recipes, I see

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -422,6 +422,8 @@ public InteractionResult checkAndHandleImportantInteractions(final Player player
         if (citizenData != null && citizenData.getJob() != null)
         {
             ((AbstractEntityAIBasic) citizenData.getJob().getWorkerAI()).setDelay(TICKS_SECOND * 3);
+            getNavigation().stop();
+            getLookControl().setLookAt(player);
```
someaddons, line 426: pretty sure we're on client side code here, and those calls are server side :)

src/main/java/com/minecolonies/coremod/permissions/ColonyPermissionEventHandler.java
```diff
@@ -75,6 +82,16 @@
...
 
+    /**
+     * Mine block objective tracker.
+     */
+    private final Map<Block, Map<UUID, IColonyQuest>> breakBlockObjectives = new HashMap<>();
+
+    /**
+     * Entity kill objective tracker.
+     */
+    private final Map<EntityType<?>, Map<UUID, IColonyQuest>> entityKillObjectives = new HashMap<>();
```
someaddons: Those definitly do no belong here, you're just re-using an existing hook for block breaking. Not related to permissions. Move all of this into its own subscribed handler

src/main/java/com/minecolonies/coremod/quests/ColonyQuest.java
```diff
@@ -0,0 +1,334 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import static com.minecolonies.api.util.constant.NbtTagConstants.*;
+
+/**
+ * Instance of a specific quest type
+ */
+public class ColonyQuest implements IColonyQuest
```
someaddons: are there non-colony quests? Could just call it Quest

src/main/java/com/minecolonies/coremod/quests/ColonyQuest.java
```diff
@@ -0,0 +1,334 @@
...
+{
+    /**
+     * Colony reference
+     */
+    private final IColony colony;
+
+    /**
+     * The id of this quest
+     */
+    private ResourceLocation questID;
```
someaddons: isnt this the id of the Quest's type? Maybe just reference the quest type instead. If not should be renamed to reflect its not actually this quest instances ID

src/main/java/com/minecolonies/coremod/quests/QuestData.java
```diff
@@ -0,0 +1,130 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.player.Player;
+
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Instance of a specific quest type
+ */
+public class QuestData implements IQuestData
```
someaddons: QuestType

src/main/java/com/minecolonies/coremod/quests/QuestData.java
```diff
@@ -0,0 +1,130 @@
...
+        this.maxOccurrence = maxOccurrence;
+        this.objectives = questObjectives;
+        this.questTimeout = questTimeout;
+        this.questRewards = questRewards;
+    }
+
+    @Override
+    public IColonyQuest attemptStart(final IColony colony)
+    {
+        final List<ITriggerReturnData> triggerReturnData = questTriggerList.apply(colony);
```
someaddons: apply -> check since it checks a colony for conditions and doesnt really apply a colony to something

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
```diff
@@ -0,0 +1,220 @@
...
+        {
+            return false;
+        }
+        this.inProgressQuests.put(questID, quest);
+        this.availableQuests.remove(questID);
+        return true;
+    }
+
+    @Override
+    public void concludeQuest(final ResourceLocation questId)
```
someaddons: I like completeQuest more :D

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
```diff
@@ -0,0 +1,220 @@
...
+            final IColonyQuest colonyQuest = quest.getValue().attemptStart(colony);
+            if (colonyQuest != null)
+            {
+                this.availableQuests.put(quest.getKey(), colonyQuest);
+            }
+        }
+
+        for (final Map.Entry<ResourceLocation, IColonyQuest> availableQuest : new ArrayList<>(availableQuests.entrySet()))
+        {
+            if (!GLOBAL_SERVER_QUESTS.containsKey(availableQuest.getKey()) || !availableQuest.getValue().isValid(colony))
```
someaddons, line 112: do we actually need to do live-rechecking of the global quest type still exists? I think this should be guranteed and done at loading time, deserialize quest -> check type

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
```diff
@@ -0,0 +1,220 @@
...
+                this.inProgressQuests.remove(inProgressQuest.getKey());
+            }
+        }
+    }
+
+    @Override
+    public void deactivateQuest(final ResourceLocation questID)
+    {
+        this.availableQuests.remove(questID);
+        this.inProgressQuests.remove(questID);
```
someaddons, line 133: if you keep the quest data within citizens, which I suggested to remove above then this probably would need to cleanup that data aswell

src/main/java/com/minecolonies/coremod/quests/objectives/BreakBlockObjective.java
```diff
@@ -0,0 +1,167 @@
...
+        final int target = details.get(TARGET_KEY).getAsInt();
+        final int quantity = details.get(QUANTITY_KEY).getAsInt();
+        final Block block = ForgeRegistries.BLOCKS.getHolder(new ResourceLocation(details.get(BLOCK_KEY).getAsString())).get().get();
+        final int nextObj = details.has(NEXT_OBJ_KEY) ? details.get(NEXT_OBJ_KEY).getAsInt() : -1;
+
+        return new BreakBlockObjective(target, quantity, block, nextObj, parseRewards(jsonObject));
+    }
+
+    @Override
+    public IObjectiveData init(final IColonyQuest colonyQuest)
```
someaddons: IObjectiveData-> IObjective and Objective -> ObjectiveType

src/main/java/com/minecolonies/coremod/quests/objectives/BreakBlockObjective.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Nullable
+    @Override
+    public IObjectiveData getObjectiveData()
+    {
+        return new BlockMiningProgressData();
+    }
+
+    @Override
+    public void onAbort(final IColonyQuest colonyQuest)
```
someaddons: onCancelled

src/main/java/com/minecolonies/coremod/quests/objectives/BreakBlockObjective.java
```diff
@@ -0,0 +1,167 @@
...
+    {
+        cleanupEvent(colonyQuest);
+    }
+
+    private void cleanupEvent(final IColonyQuest colonyQuest)
```
someaddons: javadoc, and different name since its not an event: cleanupListeners()?

src/main/java/com/minecolonies/coremod/quests/objectives/BreakBlockObjective.java
```diff
@@ -0,0 +1,167 @@
...
+        {
+            // Only serverside cleanup.
+            ((Colony) colonyQuest.getColony()).getEventHandler().addQuestObjectiveListener(this.blockToMine, colonyQuest.getAssignedPlayer(), colonyQuest);
+        }
+    }
+
+    /**
+     * Progress data of this objective.
+     */
+    public class BlockMiningProgressData implements IObjectiveData
```
someaddons: Maybe move those out of the type classes, it is kinda confusing having Type and Instance classes in the same file

src/api/java/com/minecolonies/api/quests/IQuestReward.java
```diff
@@ -0,0 +1,18 @@
...
+{
+    /**
+     * Apply the reward to colony and player.
+     * @param colony the involved colony.
+     * @param player the involved player.
+     * @param colonyQuest the related quest.
+     */
+    void applyReward(final IColony colony, final Player player, final IColonyQuest colonyQuest);
+}
```
someaddons: should all rewards maybe get a displaye component they return for displaying them easier?

someaddons (commented):
Looks ok, mostly got some comments on structure or naming

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
Raycoms: The questgiver is the citizen. But we cannot hold a citizen here because that causes trouble.

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
Raycoms: There are other things that can make it invalid, like the citizens involved being gone

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
Raycoms: Several possible things. Expiration, Cancellation, etc

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
someaddons: oh thought questgiver would be citizendata

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
Raycoms: At this moment, yes. But we can have objectives be an array of objectives too. Or a multiObjective type even.

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
Raycoms: No, other assigned citizens

src/api/java/com/minecolonies/api/quests/IDialogueObjective.java
Raycoms: Interactions are kinda a generic event handler for potential dialogues. So I would still keep them separate

src/api/java/com/minecolonies/api/quests/IQuestData.java
Raycoms: Wanted to avoid IQuest, because else it could seem that a IQuest is assigned to a colony. This is just a global data holder without contextual quest data yet.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
Raycoms: I want to be able to easily include them in citizen matching

src/main/java/com/minecolonies/coremod/quests/ColonyQuest.java
Raycoms: Wanted to keep names more verbose. Like QuestData, is just a generic questData without context, and ColonyQuest is an instance of a quest for a colony. Quest seems kinda inbetween for me

src/main/java/com/minecolonies/coremod/quests/ColonyQuest.java
Raycoms: Could call it quest type too

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
Raycoms: you can have data reload, do you want this to trigger all colonies to check if all colonies and everything they hold is still fine?

Raycoms: and potentially on restart when colonies are not fully constructed yet

src/api/java/com/minecolonies/api/quests/IColonyQuest.java
Raycoms: Yes, it is. But this exists on both client and serverside. We might not have that

src/main/java/com/minecolonies/coremod/datalistener/QuestJsonListener.java
Raycoms: Pft, your leftover javadoc copies!

src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
Raycoms: Gets triggerd on both sides afaik. checking null on citizen data will guarantee that too

src/main/java/com/minecolonies/coremod/quests/QuestData.java
Raycoms: Quest Trigger List is a function Function<IColony, List<ITriggerReturnData>> questTriggerList;

I cannot rename the way to call that =D

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
Raycoms: What calls this already cleans it up

src/api/java/com/minecolonies/api/quests/IQuestReward.java
Raycoms: Can be added in the future when we add things to display them

src/api/java/com/minecolonies/api/quests/IObjectiveData.java
Raycoms: This is just a data holder to keep track of the progress. Not a type.

src/api/java/com/minecolonies/api/quests/IQuestTrigger.java
Raycoms: It is a utility function yes

src/main/java/com/minecolonies/coremod/quests/objectives/BreakBlockObjective.java
Raycoms: Didnt see this as an instance, just as a small tracking container. Not all Objectives have and need it.

src/main/java/com/minecolonies/coremod/colony/CitizenData.java
someaddons: oh I wouldnt change that, I'd just do a Map<Citizen, thiese lists here and redirect lookups to the manager when a citizen is queried. So functionally its the very same, the citizens just lookup things in the managers map instead of their own list, just so you don't accidently create mismatched data between the quest manager and the citizen store

Raycoms: Yeah, but that makes matching them not possible. At the very least not for finished participation.

Active one we could move, but would go through additional corners of querying from interactions etc.

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
someaddons: I'd do it during colony construction, when its loaded from nbt, the quests get loaded from nbt and during that they check hey am I still an existing quest?

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
someaddons: so it would break if someone else tried to stop a question from elsewhere in code and just calls this?^^

Raycoms: Can also just call this some sort of cleanUpFromManager to make that more clean. Actual cleanup is through the quest itself.

src/main/java/com/minecolonies/coremod/quests/QuestManager.java
Raycoms: Its a O(1)= operation and allows us easier cleanup.Anything that is done on NBT load requires load order to be done correctly. Like, if the global server store is not loaded at the right point everything falls apart else.

someaddons: we can do both I dont mind that, but I absolutely love a proper load order isntead of needing to cleanup later :)

src/api/java/com/minecolonies/api/quests/IDialogueObjectiveTemplateTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static com.minecolonies.api.quests.QuestParseConstant.*;
+
+/**
+ * Dialogue type of objective interface.
+ */
+public interface IDialogueObjectiveTemplateTemplate extends IQuestObjectiveTemplate
```
uecasm: Why are these `TemplateTemplate`s?

src/main/java/com/minecolonies/coremod/quests/ColonyQuest.java
someaddons: quest template id now ^^

src/api/java/com/minecolonies/api/quests/IDialogueObjectiveTemplateTemplate.java
Raycoms: intellij refactor fail

someaddons (approved):

PR #9075, opened 2023-04-16
Jade integration

uecasm:
- Adds integration with Jade (aka the new WAILA/HWYLA).
    - Hut blocks show the building level, style pack, and list of citizens (the inventory was pre-existing)
        image
    - If you give a building a custom name, that appears as the main name with the building name below:
        image
    - Citizens display their gender and job (health/armor display was pre-existing):
        image
    - Visitors display their recruitment cost as well:
        image

Review please (could port, but need to adjust versions)

Suggestions for additional interesting values (or alternative icons) welcome.


Raycoms (dismissed):

gradle/dependencies.gradle
```diff
@@ -57,6 +63,8 @@ dependencies {
     compileOnly fg.deobf("mezz.jei:jei-${project.jei_mcversion}-forge-api:${project.jei_version}")
     runtimeOnly fg.deobf("mezz.jei:jei-${project.jei_mcversion}-forge:${project.jei_version}")
 
+    implementation fg.deobf("curse.maven:jade-324717:${project.jade_id}")
```
marchermans: No, our CI system has problems with curse dependencies since it can not cache them
Find an official maven.

marchermans (changes requested):
Outside of the dependency coming from curse, this looks good.
Can be merged when you provide me with an official maven repo.

gradle/dependencies.gradle
uecasm: Sadly, cursemaven **is** the official maven repo.  [Reference](https://jademc.readthedocs.io/en/latest/plugins19/getting-started/)

uecasm: Could we perhaps just cache it manually in the LDT maven?  What files would you need for that?

marchermans: No we can't we tried that with dynamic trees and it failed horribly.

uecasm: Why?  It's not like maven dir structures are all that complicated.  It's trivial to get a copy of the release jar and corresponding pom.  It's a bit harder if it needs the supplemental jars as well, but even that isn't all that hard, since only source-compatibility is needed and not binary-compatibility.

marchermans: As already said we tried this with dynmap that failed horribly. Then we tried to setup our own ci for their stuff that failed too because version numbers did not line up.


The onyk way this works properly is if they get a normal proper maven with proper coordinates and proper Metadata. This has been discussed to death and this is our official stance.

Raycoms: This is not the official curse maven btw. The official one seems to use the proper name.

marchermans: The official one is probably fine. I would need to see it. However we just had problems with this one.

uecasm: iwfm (new commit)

gradle/dependencies.gradle
```diff
@@ -13,6 +13,13 @@ repositories {
         name "Rei"
         url "https://maven.shedaniel.me/"
     }
+    maven {
+        name "Mirality"
+        url "https://maven.mirality.co.nz/"
```
marchermans, line 18: Dis you now manually upload this to your maven?

marchermans (changes requested):

gradle/dependencies.gradle
uecasm: I did.  And I verified that it works perfectly fine with both the version from cursemaven and the "real" curseforge mod.  It's no different than any other maven.

But if you want to host the files on the LDT maven or elsewhere, that's fine too, I can send them to you or you can pull them yourself.

marchermans: It is not about it working in your case.
It needs to work with our artifact cache

Yes your maven would probably work if it supplied hashing files and etags which I doubt you do. However, you need to manually manage this now undocumented external system which is a no go to start with.

And as @Raycoms discussed there is an official solution from curse with better maven Metadata. So that is what we need you to try.

To be clear: your manual uploaded maven artifact is just as hard a no as the cursemaven solution you had before!

Raycoms: I think the official cf maven doesnt have metadata either, but it uses names and stuff

Raycoms: Atm the PR doesnt build, but I guess if it builds we should be fine, right?

marchermans: Names is also Metadata it needs to be properly identifyable by our cache and I would love it if they support hash files so that we can properly detect cache invalidation. But proper naming is at least a must. And it had to be automated. We are not doing manual stuff cause then you can just put it in a libs folder and load it from there.

uecasm: It has all the files that ForgeGradle builds for maven:
image

I agree that an external maven is not ideal, but you already veto'd the curse maven, so idk what you're even asking any more.  The only difference between this and any other maven (like the one for JEI, for example) is that I uploaded it instead of the mod author.  But if you insist on using only the mod author's own upload, that's what cursemaven was.

Although I just noticed that the build failed in `javadoc`, which I don't habitually run locally.  It looks like the problem there is that the API jar had external links, which is readily solved.

Raycoms: Now it did build properly, so it should be fine

uecasm: The build succeeded from cursemaven too 😛

uecasm: Cursemaven just works (I've never had any problems including all kinds of mods with it), and literally everybody that I can find on the entire internet recommends using it over the curseforge.com maven "API", because nobody can figure out how to even access anything using that.  If you can tell me what dependency name it should use, I can change it to that, but what little docs there are on it gives me negative data and the endpoint itself is not explorable, so it's completely worthless.

For my maven, I did actually rebuild the mod (from the matching commit), mostly to get it to generate source packages so that Find Usages works better.  (These *could* work through cursemaven too, except that people don't usually upload them because the Forge MDK doesn't generate them by default.)  But it remains compatible with the "real" mod, so it still works fine.

uecasm:
FYI it looks like CurseForge have decided that CurseMaven is their official maven API:

<https://www.curseforge.com/minecraft/mc-mods/jade/files/4433884>
image

So should I change it back to that?  Slight downside is that you lose Find Usages support since the source jar isn't included, but otherwise it works fine.

marchermans:
Sadly that does not solve the issue. The maven artifact id is. The numbers are not really something our system can properly handle when it tries to cache them so that it does not need to download them over and over again.

Yes they are valid maven coordinates but jfrogs caching filters were not designed for it and I have to clear it out regularly.

So no as it states right now this pr is on hold indefinitely

Raycoms:
> Sadly that does not solve the issue. The maven artifact id is. The numbers are not really something our system can properly handle when it tries to cache them so that it does not need to download them over and over again.
> 
> Yes they are valid maven coordinates but jfrogs caching filters were not designed for it and I have to clear it out regularly.
> 
> So no as it states right now this pr is on hold indefinitely

Isnt there something we can do about this? This kinda kills of 98% of any potential compatibility things on our side. And you were wanting more compat things on our side.

marchermans:
> > Sadly that does not solve the issue. The maven artifact id is. The numbers are not really something our system can properly handle when it tries to cache them so that it does not need to download them over and over again.
> > Yes they are valid maven coordinates but jfrogs caching filters were not designed for it and I have to clear it out regularly.
> > So no as it states right now this pr is on hold indefinitely
> 
> Isnt there something we can do about this? This kinda kills of 98% of any potential compatibility things on our side. And you were wanting more compat things on our side.

98% is wildly blown out of proportions. Most proper mods with which we should integrate, have their own maven with proper ids.

But right now, unless curse provides proper NAMES as their maven coordinates, no.

Actively I would need to rewrite the caching layer. Which is planned in the future. Likely around the end of the year, but not before that.

uecasm:
So what was wrong with the maven I made then?  It had names and hashes.

uecasm:
https://www.curseforge.com/minecraft/mc-mods/jadecolonies

PR #9082, opened 2023-04-20
Quick fix for instant xp gain in library if worker has no paper/book …

mrlamb:
…in own inventory and paper/book exists in building

Observed some behavior that made me look at the Study AI. Essentially workers without studyItems where study items are present in the hut inventory go into a pacing behavior where they rapidly gain xp. 
The go to the bookshelf, get the xp, see that there are items to grab, go to the hut block then return out of the function before grabbing anything, and then go back to the bookshelf to start over getting xp on each trip.

I've added comments to the code to explain it a bit better.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
+++ b/src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
@@ -136,12 +136,15 @@ public class EntityAIStudy extends AbstractEntityAISkill<JobStudent, BuildingLib
                 final int bSlot = InventoryUtils.findFirstSlotInProviderWith(building, studyItem.getItem());
                 if (bSlot > -1)
                 {
+                    //Need to move this here. Since worker's previous position was a bookshelf, he'll always walk to the building and then return getState, never actually grabbing the item.
+                    //This causes him to loop back and forth getting the free xp on line 131 and never hitting the delay on 173
+                    takeItemStackFromProvider(building, bSlot); 
                     if (walkToBuilding())
                     {
                         setDelay(WALK_DELAY);
                         return getState();
                     }
-                    takeItemStackFromProvider(building, bSlot);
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/entity/ai/citizen/student/EntityAIStudy.java
```diff
@@ -136,12 +136,15 @@ private IAIState study()
                 final int bSlot = InventoryUtils.findFirstSlotInProviderWith(building, studyItem.getItem());
                 if (bSlot > -1)
                 {
+                    //Need to move this here. Since worker's previous position was a bookshelf, he'll always walk to the building and then return getState, never actually grabbing the item.
+                    //This causes him to loop back and forth getting the free xp on line 131 and never hitting the delay on 173
+                    takeItemStackFromProvider(building, bSlot); 
                     if (walkToBuilding())
```
Raycoms, line 142: WalkToBuilding should return true until it arrived at the building and then take the stack on arrival only. The current code will get it from the building even before they started to walk

Raycoms: We should probably redirect this to its own state. We already have code for pickup stuff. So instead just set the pickup predicate and put them into pickup state.

mrlamb: That makes sense, I guess walking is it's own state? When it finishes the walking what state does it default into? Observation says it goes back to STUDY because it never actually grabs the item and he just walks back to the bookshelf for more sweet sweet xp :D

What I suggested was the simplest "fix" I could think of. If he's going to get the item anyway, it didn't seem to matter if he teleported it to himself before walking and then walked there. It would still resolve the loop. (at least I thought)

Raycoms:
fix incoming from myself

PR #9086, opened 2023-04-21
Quest translation deux

### LINKED ISSUE #9080: [BUG] Missing translation interface for mod's quest text

### Minecraft Version

1.19.2

### MineColonies Version

1.0.1371-ALPHA

### Structurize Version

1.0.491-ALPHA

_No response_

### Current Behavior

1.0.1371-ALPHA version adds some quests, but I can't translate them on Crowdin, because Crowdin doesn't have them.
I can only override them if I want to translate. There's no second way. Once the mod updated the newer version, I have to override them again, that's too inconvenience.

### Expected Behavior

I hope quest texts can be translated to other language on Crowdin. 

### Reproduction Steps

1. Open `minecolonies.jar` by 7z;
2. Open `data\minecolonies\quests` folders, I found some json files;
3. These texts are all English.

### Logs

Not crash

### Anything else?

uecasm:
Closes #9080

- Fixes #9085 (vanilla regex bug breaks `%d` in translation resources)
- Hotfix nullptr
- Fix resource file exclusions:
    - The build script was trying to exclude `default.json` from the jar, but didn't.  Now it does.
    - It also now excludes the new `quests.json` and the old `manual_en_us.json`.
    - This leaves `entries.json` and `en_us.json`, identical files.  I'm not sure why the first one exists.  Something crowdin related maybe?  But does it need to go in the jar?
- Magic translation datagen for quest json
    - The quest jsons themselves are still exactly as before.
    - During datagen it processes these files and generates translation keys for the dialogue automatically.  (This does unfortunately mean they have more generic names than if done manually, but this does still allow following the conversation path in the translation file.)
    - It doesn't make any attempt to de-duplicate the strings.

Review please (do not port, datagen will need some code changes for 1.19.3+)


marchermans:
I am not sure about this PRs approach.
Does it not make sense to implement datagen once and for all properly? And handle it that way? Instead
of this backwards way where you make a "magic" translation datagen only?

Raycoms:
I think quests are too unique and code writing dialogues is kinda a big nono

Sent from Proton Mail mobile

uecasm:
My main thought here is that we seem to be wanting the community to write a lot of quests, and that seems like it would be easier with self-contained json files rather than with codegen.

Although a counterargument might be that the codegen could be written in a sufficiently fluent style that a non-coder would be able to follow it.  And that might let them avoid json's quirks.  Though modpack authors would have to use the json format regardless.

marchermans:
I understand. It is a balancing act I agree.
Maybe ask some people in the community what they want.

Raycoms:
> I understand. It is a balancing act I agree. Maybe ask some people in the community what they want.

They would optimally want to send us text files, but that would then put a lot of load on us, json files are okay too. The format is pretty easy to grasp for everything put the triggers

marchermans:
> > I understand. It is a balancing act I agree. Maybe ask some people in the community what they want.
> 
> They would optimally want to send us text files, but that would then put a lot of load on us, json files are okay too. The format is pretty easy to grasp for everything put the triggers

Oke, then I am not sure what to do.....
A webservice?

Raycoms:
> > > I understand. It is a balancing act I agree. Maybe ask some people in the community what they want.
> > 
> > 
> > They would optimally want to send us text files, but that would then put a lot of load on us, json files are okay too. The format is pretty easy to grasp for everything put the triggers
> 
> Oke, then I am not sure what to do..... A webservice?

Nah, they will give us the jsons and we include the jsons and datagen the translations from there then.

uecasm:
We could come up with some third vaguely-plain-text file format, parse _that_, and datagen it into json and translations.

But I don't think that really gains us much over just the jsons, especially as (like the pure codegen) it's still _different_ from what modpacks would have to do anyway.  And adventure/gating modpacks are likely to be the most interested in creating quests, aside from whatever's in the mod itself.

src/main/java/com/minecolonies/coremod/generation/defaults/QuestTranslationProvider.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.file.Path;
+
+import static com.minecolonies.api.quests.QuestParseConstant.*;
+import static com.minecolonies.api.quests.registries.QuestRegistries.DIALOGUE_OBJECTIVE_ID;
+import static com.minecolonies.api.util.constant.Constants.MOD_ID;
+import static com.minecolonies.coremod.quests.QuestParsingConstants.*;
+
+public class QuestTranslationProvider implements DataProvider
```
Raycoms, line 31: needs javadoc

Raycoms (commented):
Just needs a bit of javadoc

PR #9126, opened 2023-05-07
Tools & Armour category in JEI

uecasm:
- Adds a page to JEI that shows the tools & armour colonists will accept at each hut level.
image
image
image
- The first row is base tools, the second row are tools with some example enchants of each level (not comprehensive, just illustrative).
- As best I can tell, this correctly replicates the logic that the mod itself uses to decide if tools are the right level or not.  Interestingly, this differs from what the wiki and discord claim about how enchantments are treated.
    - In particular: any number of level 1 enchants can be applied without increasing the tool level; any number of level 2 enchants increase the level by 1 (in total, not each), etc.

Review please (could port)

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/crafting/ToolUsage.java
+++ b/src/main/java/com/minecolonies/coremod/colony/crafting/ToolUsage.java
@@ -1,0 +1,36 @@
+package com.minecolonies.coremod.colony.crafting;
+
+import com.minecolonies.api.util.constant.ToolType;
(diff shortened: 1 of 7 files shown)


marchermans (dismissed):
Very nice work!  Looks amazing!

uecasm:
Added audit file and updated [the spreadsheet](https://docs.google.com/spreadsheets/d/1DaGFlFcnce4bHXHylzztr5_Rwnm3P42c6o_xXqgyyr8/edit#gid=1551679170) to show all base tools.

Raycoms (approved):

PR #9150, opened 2023-05-22
Ask if bug reports are for SP or MP

uecasm:
- Explicitly ask whether bug reporters are playing SP or MP

Review please (could port)


Raycoms (approved):

uecasm:
Kinda tempted to also move the description field to the top, just so the discord channel display is more useful. 😁 
But this is kinda a more logical order as is.

someaddons (approved):

marchermans:
Would am option "Both" make sense

Raycoms:
Its checkboxes, so they can just hit both

marchermans (approved):

PR #9183, opened 2023-06-13
Parent-child buildings infinite recursion

Thodor12:
Prevent parent-child schematics which both contain the child buildings inside the schematic, from causing infinite recursion problems.

Fixes https://discord.com/channels/139070364159311872/1003402520846139533/1118131704524185730

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
@@ -297,7 +297,7 @@ public abstract class AbstractSchematicProvider implements ISchematicProvider, I
     private void validateParent()
     {
         final IBuilding building = colony.getBuildingManager().getBuilding(parentSchematic);
-        if (building != null && (building.getID().equals(getID()) || building.hasParent()))
+        if (building == null || building.getID().equals(getID()) || building.hasParent())
         {
             parentSchematic = BlockPos.ZERO;
         }
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
@@ -1,6 +1,5 @@
 package com.minecolonies.coremod.colony.buildings;
 
-import com.google.common.collect.ImmutableSet;
 import com.ldtteam.structurize.blockentities.interfaces.IBlueprintDataProviderBE;
 import com.ldtteam.structurize.blueprints.v1.Blueprint;
 import com.ldtteam.structurize.storage.StructurePacks;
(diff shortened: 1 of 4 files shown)


src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
```diff
@@ -306,53 +290,26 @@ public boolean hasParent()
     @Override
     public void setParent(final BlockPos pos)
     {
-        if (!pos.equals(getID()))
```
Raycoms, line 309: Not sure about this part, wont this maybe override a valid parent in the worst case?

Thodor12: That check was moved to `validateParent()`

Raycoms: yeah I know, but after it derped the parent up, a potentially valid parent was replaced

Thodor12: Ah, for context this function is only called once on a block at most. So that shouldn't be an issue in that case.
Also if I set a parent I'd expect it to change it, not keep the old one.

someaddons (approved):

src/main/java/com/minecolonies/coremod/colony/buildings/AbstractSchematicProvider.java
```diff
@@ -306,53 +289,25 @@ public boolean hasParent()
...
-        if (toRemove != null)
-        {
-            final Set<BlockPos> oldPositions = new HashSet<>(this.childSchematics);
-            oldPositions.removeAll(toRemove);
-            this.childSchematics = ImmutableSet.copyOf(oldPositions);
-        }
-
-        return childSchematics;
+        final IBuilding building = colony.getBuildingManager().getBuilding(position);
+        return building != null && !building.getID().equals(getID()) && !building.hasParent();
```
uecasm, line 301: Won't this block the case where a parent has a child that has another child?  That ought to still be valid.

Thodor12: I asked Ray this and he basically said that wouldn't ever happen

uecasm: I don't think that's true.  It sounds like a reasonable thing to want to do.  It wouldn't even surprise me if a style like Byzantine already does it, since they have quite complex nesting (though I have not checked if they actually do).

Raycoms:
Sent from Proton Mail mobile

uecasm:
Ray, your comment didn't comment 😛

PR #9240, opened 2023-07-10
Events for Dynmap implementation

Thodor12:
- Introduces a load of events that the Dynmap integration will be able to hook into

Review please


src/api/java/com/minecolonies/api/compatibility/Compatibility.java
```diff
@@ -2,16 +2,17 @@
...
-import net.minecraft.world.level.block.Block;
-import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.NonNullList;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
-import net.minecraft.core.NonNullList;
-import net.minecraft.core.BlockPos;
-import net.minecraft.world.level.LevelAccessor;
 import net.minecraft.world.level.Level;
```
Raycoms: undo changes to file you didnt change

Raycoms (dismissed):

marchermans (dismissed):
Principally this looks like very fine work.
Not sure if you covered all the basis. But perfect is the enemy of good and this is certainly good enough to get started.

src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowInfoPage.java
```diff
@@ -86,14 +87,14 @@ private void createAndSetStatistics()
         final int citizensSize = building.getColony().getCitizens().size();
         final int citizensCap;
 
-        if(MinecoloniesAPIProxy.getInstance().getGlobalResearchTree().hasResearchEffect(CITIZEN_CAP))
+        if (MinecoloniesAPIProxy.getInstance().getGlobalResearchTree().hasResearchEffect(CITIZEN_CAP))
         {
-            citizensCap = (int) (Math.min(MineColonies.getConfig().getServer().maxCitizenPerColony.get(),
-              25 + this.building.getColony().getResearchManager().getResearchEffects().getEffectStrength(CITIZEN_CAP)));
```
someaddons, line 92: is it intended to drop the static +25 here?

src/main/java/com/minecolonies/coremod/generation/defaults/DefaultResearchProvider.java
```diff
@@ -75,7 +76,9 @@ public Collection<ResearchEffect> getResearchEffectCollection()
         effects.add(new ResearchEffect(BLOCK_ATTACKS).setTranslatedName("Knight Shield Blocking Chance +%3$s%%").setLevels(new double[] {0.05, 0.1, 0.25, 0.5}));
         effects.add(new ResearchEffect(BLOCK_BREAK_SPEED).setTranslatedName("Citizen Block Break Speed +%3$s%%").setLevels(new double[] {0.1, 0.25, 0.5, 1, 2}));
         effects.add(new ResearchEffect(BLOCK_PLACE_SPEED).setTranslatedName("Citizen Block Place Speed +%3$s%%").setLevels(new double[] {0.1, 0.25, 0.5, 1, 2}));
-        effects.add(new ResearchEffect(CITIZEN_CAP).setTranslatedName("Increase Max Citizens +%s").setLevels(new double[] {25, 75, 125, 1000}));
+        effects.add(new ResearchEffect(CITIZEN_CAP).setTranslatedName("Increase Max Citizens +%s")
+                      .setLevels(new double[] {CitizenConstants.CITIZEN_LIMIT_OUTPOST, CitizenConstants.CITIZEN_LIMIT_HAMLET, CitizenConstants.CITIZEN_LIMIT_VILLAGE,
```
uecasm, line 80: You need to also `runData` if you want this change to actually do anything.  (Otherwise your other changes will reduce the citizen cap.)

Also the `+%s` part is no longer correct with this change, you should change it to `Increase Max Citizens to %s`.

src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowInfoPage.java
uecasm: I believe that is intended, given the way that the `CITIZEN_LIMIT_*` constants are defined.  However as is it's not quite correct, as per my comment below.

Thodor12: It'll always be limited by the `CITIZEN_LIMIT_DEFAULT`.

```
final int max = Math.max(CitizenConstants.CITIZEN_LIMIT_DEFAULT, (int) this.building.getColony().getResearchManager().getResearchEffects().getEffectStrength(CITIZEN_CAP));
```
Ensures that the highest between the research, or the default is taken, it the research is not yet done, it'll return 0, meaning the max allowed citizen count will be 25, afterwards the config is taken to determine the minimum.

uecasm (changes requested):
Bump since I think the issue above might have been overlooked.

Thodor12:
@uecasm I'm aware of the datagen problem you mentioned. I'll get to it tomorrow because I was working on the plantation.

PR #9244, opened 2023-07-14
Custom quarry shafts

uecasm:
- Shrinks the building path font size in the building browser (right click hut block) to make longer path names easier to read.
- Adds support for multiple custom quarry shafts (rather than using a single fixed path for each quarry type):
    - By default, it will look for a name matching the name of the quarry building:
        - `infrastructure/mineshafts/simplequarry1.blueprint` => `infrastructure/mineshafts/simplequarryshaft1.blueprint`
        - `infrastructure/mineshafts/mediumquarry1.blueprint` => `infrastructure/mineshafts/mediumquarryshaft1.blueprint`
        - `infrastructure/mineshafts/fancyquarry1.blueprint` => `infrastructure/mineshafts/fancyquarryshaft1.blueprint`
        - `infrastructure/pitofdoom.blueprint` => `infrastructure/pitofdoomshaft.blueprint`
        - `some/other/folder/bighole1.blueprint` => `some/other/folder/bigholeshaft1.blueprint`
    - It's legal to have the quarry in any folder.  It's legal to have it with or without the `1` at the end (provided it's consistent).  It's not legal to have a different number at the end.
        - It's probably still recommended to keep the 1 at the end -- I think the neighbour rendering possibly among other things might be annoyed if it's omitted.  Though it does build ok otherwise.
    - As an extra bonus, it's possible to put a special tag onto the quarry hut block itself:
        - `infrastructure/mineshafts/simplequarry1.blueprint` with `shaft=foo` => `infrastructure/mineshafts/foo.blueprint`
        - `infrastructure/simplequarry1.blueprint` with `shaft=hole1` => `infrastructure/hole1.blueprint`
        - `infrastructure/simplequarry1.blueprint` with `shaft=decorations/pit_c` => `decorations/pit_c.blueprint`
    - If you omit the `/`, it will look in the same folder.  If you include it, it will look in that absolute folder (not relative).  You can't use `\` in the tag.  It's legal for the shaft to not have a `1` at the end when explicitly named like this.

This is technically a behaviour change that could potentially break existing style packs -- however at least all the in-mod packs use the default quarry building name, which still works as before.  It would only break things if the quarry had a non-standard name, and this could be fixed by either renaming or tagging.

Review please (do not backport; can forward port)

An alternative rule that I was considering, but haven't implemented, is to still always use `infrastructure/mineshafts/` as a base (but still using the custom name) rather than matching the quarry hut folder (unless overridden by tag).  But the above behaviour seemed more consistent with the way that parent/child buildings currently work -- but not consistent with how the mine works.


Raycoms (dismissed):

src/main/java/com/minecolonies/coremod/entity/ai/citizen/miner/EntityAIQuarrier.java
```diff
@@ -139,14 +141,55 @@ public IAIState loadRequirements()
...
+            final Tuple<String, String> shaft = getShaftPath(quarry);
+            final WorkOrderMiner wo = new WorkOrderMiner(quarry.getStructurePack(), shaft.getA(), shaft.getB(), quarry.getRotation(), quarry.getPosition().below(2), false, building.getPosition());
             building.getColony().getWorkManager().addWorkOrder(wo, false);
             job.setWorkOrder(wo);
         }
 
         return super.loadRequirements();
     }
 
+    private Tuple<String, String> getShaftPath(@NotNull final IBuilding quarry)
```
someaddons, line 162: missing javadoc

src/main/java/com/minecolonies/coremod/entity/ai/citizen/miner/EntityAIQuarrier.java
```diff
@@ -139,14 +141,55 @@ public IAIState loadRequirements()
...
+    {
+        String path = "infrastructure/mineshafts/" + quarry.getSchematicName() + "shaft1.blueprint";
+
+        final AbstractTileEntityColonyBuilding tileEntity = quarry.getTileEntity();
+        if (tileEntity != null)
+        {
+            path = quarry.getBlueprintPath().replace('\\', '/')
```
someaddons, line 169: are those seperators the same on all OS's? or should we be using the Files.seperator thing instead

uecasm: Windows will produce backslashes but will accept both, which is one reason this converts strictly to forward slashes.

Linux will produce forward slashes and only accepts forward slashes.

MacOS classic used a different separator but MacOS BSD also uses forward slashes.

So forward slashes are safe everywhere, and it makes life easier when the tag can assume a single format.

(It's technically possible for this to misbehave on a Linux system with a blueprint that actually contains a genuine backslash in its path, but I think it's reasonable to ignore that case, as there's lots of other code that will also break if people use that kind of silly filename.)

someaddons (approved):

PR #9246, opened 2023-07-14
Froglights

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/473575384445878273/1129239041531969566)

- Makes JEI realise that froglights aren't actually a real drop from magma cubes, at least during nether expeditions.
- Adds a custom crafting recipe for froglights to the netherworker (unlocked at level 3).
    - This is somewhat deliberately a bit more complicated/expensive than it might be to actually go find a frog and magma cube yourself, but is perhaps more easily mass-producible.
    - All recipes require a slime block and a magma cream.
    - Sadly there aren't really any "frog" items that could be used, so it instead uses a representative block from typical frog biomes:
        - Lily pad for ochre
            image
        - Magma block for pearlescent
            image
        - Snow block for verdant
            image

Review please (could port either way)

Alternative recipe suggestions welcome.

### DIFF
--- a/src/datagen/generated/minecolonies/data/minecolonies/crafterrecipes/netherworker/ochre_froglight.json
+++ b/src/datagen/generated/minecolonies/data/minecolonies/crafterrecipes/netherworker/ochre_froglight.json
@@ -10,7 +10,7 @@
     },
     {
       "count": 4,
-      "item": "minecraft:magma_block"
+      "item": "minecraft:lily_pad"
     }
   ],
   "intermediate": "minecraft:air",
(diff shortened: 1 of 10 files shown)


someaddons:
I don't think we should make froglights craftable, its one of the few resources requiring more player investment to obtain, which we also factor in as cost when using it in schematics. This would break both the cost and the player experience

MotionlessTrain:
Not sure whether all style designers factor that in. They may be used too often for that (and getting a frog into the nether, or magma cubes in the overworld is not very easy or obvious for new players)

someaddons:
at least the three of us on stream do, why do we need this to begin with?

Raycoms:
I dont have an issue with a rare drop from the netherminer, but recipes would be too easy

uecasm:
Closed in favour of #9249.

PR #9249, opened 2023-07-15
Froglights v2

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/473575384445878273/1129239041531969566)

- Makes JEI realise that froglights aren't actually a real drop from magma cubes, at least during nether expeditions.
- Factors the number of loot table rolls into the JEI probability calculations.
    - In particular, for the nether miner this means that JEI now shows the probability for a whole expedition, not just a part of it.  Although since it uses max values for both item and roll counts, it's a "best case", not an "average case", so typical results will be lower.
    - Bonus rolls are still ignored; these are rarely used, but do produce extra rolls for high-skill netherminers, so there's still a hefty amount of approximation in the display anyway and some specific miners might produce even better results.
    - It's perhaps a little debatable whether the rolls act as a multiplier on the probability or on the resulting item count, but it's easier to show probability rather than item count.  But this does mean that some "100% probability" items are not actually guaranteed to drop at all, and some items will show at "432% probability" (and still theoretically could not drop any, though that's very unlikely).
        - I kinda would have preferred increasing the count instead of the probability, but that has ItemStack serialization and rendering complications that make the implementation more painful.
- Adds froglight blocks as L3+ drops.  (Not dependent on encountering magma cubes.)
    - The ochre ones are more common, since the corresponding frogs spawn in more common biomes.
    - With the new roll calculations, JEI estimates a 7% or less chance to get one non-ochre froglight block per expedition, so they're still pretty rare.  (It's the same chance as getting ancient debris, although that can drop x2.)

Review please (could port either way)


src/main/java/com/minecolonies/coremod/colony/crafting/LootTableAnalyzer.java
```diff
@@ -166,6 +170,20 @@ private static List<LootDrop> entryToDrops(@Nullable final LootTables lootTableM
         return drops;
     }
 
+    private static boolean conditionsSeemImpossible(@NotNull final JsonArray conditions)
```
someaddons, line 210: missing javadoc

uecasm: A lot of private methods lack javadoc.  But k.

someaddons: good call, if you want you could add those aswell

uecasm: I meant in general in the entire mod 😛

someaddons: hey I wouldnt complain if you added those :)

PR #9258, opened 2023-07-20
Add scan analyzer feature

someaddons:
- New feature for style builders to allow schematics to be inspected easily and compared with eachother:
image
- Block tier lists are introduced which sort blocks according to building level, displayed on the block via tooltip for builders(scan tool in inv). Also replaces the old, outdated, barely used spreadsheets builders got.
- Comparison score is calculated by a bunch of factors, including weighted block tiers, schematic size, amount of buildings included and block variety. It is meant to be able to roughly compare the difficulty of schematics, to be improved in the future.
- Adds a new creative only tool the schematic analyzer which grants access to the analyzation UI and allows manual selection of a region to compare against

Review please


Nightenom:
Didnt read the code much, but there was (is?) issue with uncraftable blocks, if so: since you made tiers for virtually anything, can you add infested etc. blocks to uncraftable tiers and put them on top of list?

someaddons:
I've been thinking about a tag for uncraftables ones too yea, we'd probably also want to use it to complain on schematic creation so ppl notice right away

Nightenom:
schematic creation is problematic unless we've mcol specific scan tool (which im not aware of)

Raycoms:
We have specific buildtool handlers. We could have specific scan tool handlers, and do the test on the minecol sides.

someaddons:
planning to add some automated scans of all schematics anyways, can easily include a scan for uncraftables

uecasm:
Uncraftables is a bit of a tricky definition.  You could check if there's a recipe definition, but there are lots of things that don't have recipes but are still readily available, so that will have lots of false positives.  Or you could rely on explicitly tagging dodgy blocks, but that won't pick up when a modpack does add a recipe, or new blocks that don't, so that will have false negatives.

(Still, between the two I prefer a block tag.  At least pack authors can add new blocks to the list easily that way, though removing is a little harder.)

Raycoms:
Yeah, non craftable it probably the wrong tag. Maybe creative-only or unobtainable

uecasm:
Still a slippery definition, though.  e.g. dragon egg.

src/main/java/com/minecolonies/coremod/client/gui/WindowBlueprintSelector.java
```diff
@@ -0,0 +1,1116 @@
+package com.minecolonies.coremod.client.gui;
```
Raycoms: is this just a copy? =O

someaddons: modified copy ye, since the original is final and you didnt want to have modifications I recall :D

Raycoms: Would have moved some of the bigger functions into statics or so. Or created a new abstract version on the struct side, where one is strictly "Loading" and the other part is "placement"

Raycoms (approved):

src/main/resources/data/minecolonies/tags/blocks/tier1blocks.json
```diff
@@ -0,0 +1,71 @@
...
+    "minecraft:red_stained_glass_pane",
+    "minecraft:pink_stained_glass_pane",
+    "minecraft:hay_block",
+    "minecraft:coal_block",
+    "minecraft:fern",
+    "minecraft:grass",
+    "minecraft:dead_bush",
+    "minecraft:dandelion",
+    "minecraft:poppy",
+    ^^
```
Nightenom, line 42: ?

Raycoms: bugsy?

MotionlessTrain: That doesn't look very JSON-y to me :D

someaddons: well :D good catch

PR #9275, opened 2023-07-28
Fix default style pack selection in townhall

### LINKED ISSUE #9130: Cannot change the default style of buildings via the Town Hall

### Minecraft Version

1.19.2

### MineColonies Version

1.0.1392-ALPHA

### Structurize Version

1.0.491-ALPHA

_No response_

### Current Behavior

So I started MineColonies with a desecrated village. Decided to go do my own thing instead. The village had the Medieval Oak style. I checked for what style I wanted to use, decided it would be Nordic Spruce. I built a lot of stuff as of now, and my Town Hall is level 4. I go in there, click settings, chose to Nordic Spruce theme. As I click on anything else, if I go back to the settings, it's still Medieval Oak.

### Expected Behavior

If everything worked well, it would just save the state at which it is (with the correct choice) so I wouldn't have to do it again and failing.

### Reproduction Steps

Build a town hall.
Go in there, click the book.
Go into settings.
Under "Switch Colony Pack", pick a new one.
Click on any other tab or exit the book screen.
Go back there, the style should be the one that wasn't saved before.

### Logs

https://gist.github.com/erroroid/a4a05a679a439de9b8fa98e9b18a1b70

### Anything else?

Thodor12:
Closes #9130 

- Removes the check which forces buildings to be level 0 or deconstructed for the town hall.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowSettingsPage.java
+++ b/src/main/java/com/minecolonies/coremod/client/gui/townhall/WindowSettingsPage.java
@@ -12,10 +12,9 @@ import com.minecolonies.coremod.Network;
 import com.minecolonies.coremod.client.gui.WindowBannerPicker;
 import com.minecolonies.coremod.colony.buildings.workerbuildings.BuildingTownHall;
 import com.minecolonies.coremod.network.messages.server.colony.*;
-import com.minecolonies.coremod.network.messages.server.colony.building.BuildingSetStyleMessage;
+import net.minecraft.ChatFormatting;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.screens.Screen;
-import net.minecraft.ChatFormatting;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.contents.TranslatableContents;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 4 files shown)


Thodor12:
Does the logic in the else statement also have to be ran if it's a town hall.
I'd presume the pack would still have to change if you were to deconstruct the townhall and change the style of it?

Also, wouldn't this be better to just be 2 separate message instances, because this is also used in WindowBuildBuilding, acting as the initial setter for the pack.

This way you could also take down the town hall without automagically modifying the default style.

uecasm:
Given that there's an associated bug reported recently with changing the TH style making it load the wrong blueprints when upgrading the TH itself, perhaps the Real Bug™ is that the TH style selector affects the TH's own style in some way?  It really ought to be completely independent.

Or rather, whatever style the TH was built in is the default value, then they're entirely separate after that.

Thodor12:
> Given that there's an associated bug reported recently with changing the TH style making it load the wrong blueprints when upgrading the TH itself, perhaps the Real Bug™ is that the TH style selector affects the TH's own style in some way? It really ought to be completely independent.
> 
> Or rather, whatever style the TH was built in is the default value, then they're entirely separate after that.

I'm unsure whether this will fix that. But perhaps just pulling those 2 message classes apart will already go a long way, technically they don't have anything to do with one another.

PR #9285, opened 2023-08-01
Improve statistics manager serilization

someaddons:
- Improve statistics manager serilization
- Add better safeguard against pirates spawning without water
- Stop citizens from regenerating health from food during fighting
- Avoid marking citizen data dirty when health changes
Review please


Raycoms:
The stats manager stuff looks fine. The rest Im not so sure. We at least should sometimes update the citizen health data on the client side, they serialize quite rarely otherwise afaik. And players should be able to have fairly up to date data.

And on the healing, afaik that turns off all healing mid fight, is that intended?

Raycoms (approved):

PR #9305, opened 2023-08-13
Forge+neoforge compat

someaddons:
Closes #
Closes #
Closes #

- Allows loading minecolonies on both neoforge and newer forge versions
- Temporarily removed the entity size event

Review please

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@ modBaseName=minecolonies
 javaVersion=17
 useJavaToolChains=true
 fml_range=[45,)
-forge_range=[46.0.1,47.1.3]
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -9,7 +9,7 @@ modBaseName=minecolonies
 javaVersion=17
 useJavaToolChains=true
 fml_range=[45,)
-forge_range=[46.0.1,47.1.3]
+forge_range=[46.0.1]
```
MotionlessTrain, line 12: Is it intended that minecolonies will only accept Minecraft 1.20 going forward? Or does NeoForge use 46.x.x for 1.20.1?

PR #9310, opened 2023-08-16
Work overlays

uecasm:
- Closes [discord request](https://discord.com/channels/472875599422291968/1131962240627654706/1131962240627654706)
- Closes [discord report](https://discord.com/channels/472875599422291968/486940969737125922/1145498378797600899)

- Prevents the quarrier being auto-hired before the quarry is built (since you can't access the Manage Workers button before then anyway).
- Holding shift while wearing Builder's Goggles also shows the current block that the builder/miner/quarrier is working on.
    - This can help watch their progress, or find where they're stuck.  It won't show every block due to sync delays, but it should still give you a reasonable idea.
    - The idea is to in future extend this to other kinds of worker progress, such as tracking the forester tree while holding a forester tool.
- Adds a hotkey (disabled by default) that toggles the goggles -- while still wearing them, they won't show the blueprint preview, but will still show the boxes.
    image
    - The current state is also shown in the goggles icon and tooltip -- the tooltip also now hints that it shows previews, for the people who take the lore text too seriously.
    - This toggle also affects the neighbour rendering (even if you're not wearing goggles), but "toggle goggles" sounds funner.
    - The state is deliberately non-persistent; it will reset to showing the blueprints each restart (but not reconnect).
- Fixes some internal state tracking for work orders (quarry orders especially) and makes the previews update faster when starting/cancelling a work order.
    - Pre-existing quarry builds won't track active position as a result of this; you'd need to cancel and restart the work order, e.g. by firing the quarrier from the quarry and re-assigning them.
- Fixes an exception in the nearby-neighbours preview in a particularly weird edge-case.

Review please (could port)


src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
```diff
@@ -69,16 +96,42 @@ public class ColonyBlueprintRenderer
...
      *
      * @param ctx rendering context
      */
     static void renderBlueprints(final WorldEventContext ctx)
     {
+        if (ModKeyMappings.TOGGLE_GOGGLES.get().consumeClick())
+        {
+            shouldRenderBlueprints = !shouldRenderBlueprints;
+
+            ctx.clientPlayer.playNotifySound(SoundEvents.NOTE_BLOCK_PLING, SoundSource.NEUTRAL, 1.0F, shouldRenderBlueprints ? 0.75F : 0.25F);
```
Nightenom, line 127: I would not play sound

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
```diff
@@ -47,16 +56,34 @@ public class ColonyBlueprintRenderer
...
+
+    /**
+     * The cached list of boxes to be rendered.
+     */
+    private static Map<BlockPos, BoxRenderData> boxRenderCache = new HashMap<>();
+
+    /**
+     * The cache of blueprint data.
+     */
+    private static final LoadingCache<BlueprintCacheKey, BlueprintPreviewData> blueprintDataCache = CacheBuilder.newBuilder()
```
Nightenom, line 70: Probably join with `blueprintRenderCache` using tuple or sth like that, it's hard to understand when you have three caches
I see you want to cache loaded blueprints, but at least don't go over rendering cache expire duration (it makes no sense)

Nightenom (commented):
Just looked through blueprint rendering, had hard time to understand what are you trying to achieve here, so dont mind my comments if they are off

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
uecasm: This only occurs when the player presses a hotkey, and the effect is otherwise invisible (if there isn't a blueprint in the current field of view), so auditory feedback seemed useful.

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
uecasm: There's three caches because they're caching different things.  Technically only this one is actually a cache, the others just store different kinds of "current data" that's recalculated infrequently.

The `blueprintDataCache` is the "real" cache, which keeps around a loaded `Blueprint` based on its path so that it doesn't need to keep reloading it from disk.  It's a `LoadingCache` so it gets populated implicitly by usage in the other cache.

The `blueprintRenderCache` is storing a list of blueprints that ought to be rendered right now, and where.  This is populated explicitly, so can't be a `LoadingCache`.

The data cache is expected to have a longer expiration than the render cache, because buildings go in and out of the render cache as the player moves around the world, but it will usually be the same ones coming back in that left a little while ago.

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
```diff
@@ -159,96 +223,110 @@ static void renderBoxes(final WorldEventContext ctx)
...
+        final Future<Blueprint> blueprintFuture = StructurePacks.getBlueprintFuture(key.packName(), key.path());
+
+        final BlueprintPreviewData blueprintPreviewData = new BlueprintPreviewData();
+        blueprintPreviewData.setBlueprintFuture(blueprintFuture);
+        blueprintPreviewData.setPos(BlockPos.ZERO);
+        if (key.orientation().mirror() != Mirror.NONE)
+        {
+            blueprintPreviewData.mirror();
+        }
+        blueprintPreviewData.rotate(key.orientation().rotation());
```
uecasm: BTW @Nightenom, while you're here: this looks like a gap in the Structurize API, that I can't just `setRotationMirror` but have to break it into its components.

Also, it would be nice if there was a setter that does **not** trigger `syncChangesToServer()`, because that's completely pointless in this case.  Perhaps the caller should be doing that where needed instead of putting it in that class?

Nightenom: gap is intentional, will break at major, will cause quite huge pr so I delayed it

for the second we can have boolean in ctor which may disable syncing permanently for that instance (I wouldnt do enable/disable methods bcs the decision needs to know context of origin)

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
Nightenom: struct render cache keeps it as long as you want to render it (regardless visibility, if you do visibility checks you should not cuz structurize do that itself)

and the box cache? doesnt each blueprint provide the info anyway? so why keep it standalone

uecasm: The structurize render cache still requires you to have a blueprint instance, or to give it a unique "slot" name (the latter being what the tools do).  This cache is what stores those instances to avoid having to come up with unique names (and caching those instead).  It also allows it to use the (perhaps more efficient) `renderStructureAtPosList` call instead of repeated `renderStructureAtPos`.

Also, the structurize cache is designed to be synced from server to client, whereas this is a pure client-side thing that should never sync to server.

This doesn't do visibility checks for occlusion, but it does have an effective range beyond which it will not try to load and render blueprints, to avoid loading the entire colony.  (For context: this code is what does the "level 5 neighbour" rendering, as well as the builder's goggles.)

The box cache exists because it's pointless to recalculate that every render tick, so it does that up front at a slower pace instead.  As for why it's a separate cache, it's because each blueprint can generate multiple boxes, and some boxes may not have a blueprint to render, so they have different arity and different lookup keys.

uecasm:
The [fields exception](https://discord.com/channels/472875599422291968/486940969737125922/1145498378797600899) is a bit weird.  At first I thought it meant farmer fields have an `IBuildingView` now, but it doesn't look like they actually do.

So it looks like somehow their colony has some other building anchored at the same location as a scarecrow?  Probably corrupt data.  Still, the code improvement seemed worthwhile anyway.

Raycoms (dismissed):

Raycoms:
got a little conflict

uecasm:
_coughs_

marchermans (approved):

PR #9327, opened 2023-08-22
Fix field sync logic

Thodor12:
- Mainly on request of Sam to improve the field sync logic, manages to fix a sync issue on automatic field assignment

Review please


src/main/java/com/minecolonies/coremod/colony/ColonyView.java
```diff
@@ -1106,20 +1104,10 @@ public IMessage handleColonyBuildingViewMessage(final BlockPos buildingId, @NotN
...
-    }
-
-    @Override
-    public void handleColonyRemoveFieldViewMessage(final @NotNull FieldRegistries.FieldEntry type, final @NonNull BlockPos position, @NotNull final FriendlyByteBuf buf)
-    {
-        final IField field = type.produceField(position);
-        field.deserialize(buf);
-        fields.remove(field);
+        this.fields.clear();
+        this.fields.addAll(fields);
```
Raycoms, line 1110: The problem I have with this approach is that it invalidates existing UIs that cached things client side.
So, I would do sth rather: Copy fields over to temp field. Clear fields. For each field check if we got a temp, if so, serialize into it, and add to list.

Thodor12: Nothing is cached, this set of fields is the only source of truth

Raycoms: While you got a UI open for a field, that should be at least temporarily cached there

someaddons: why would you cache it? caching only makes sense if a lookup is expensive

Raycoms: In all buildings at least, in the window constructor we query the building and then operate on it. Else, we have to constantly query the list, go through all items in the list and get the correct building. Every frame.

Thodor12: I have to do so anyway because the field list updates on each frame anyway, and there's no way to even stop that anyway.

If I then cache it I have to constantly request the new state back from the colony view anyway to get the new active state.
So instead of caching it just immediately asks it directly from the colony view.

Raycoms: So each field, goes every frame to the colonyview and asks it for the field list and sorts out their own instance as a basis for truth?

someaddons: we're doing way more expensive stuff on other UIs, this is fine xD

Raycoms: Seems odd, especially as we do it differently everywhere else. If we rely in some areas on the instance to remain "fresh" and do it like this in others this invites future bugs to happen.

Thodor12: I actually got more issues back then with having several sources of truth. Previously assigned fields were stored within the fields module for example. But I got so many sync related issues I got completely tired of it, hence why I moved every instance of fields to the structure manager and colony view, and everything must be obtained through there.

someaddons (dismissed):

src/main/java/com/minecolonies/coremod/colony/ColonyView.java
Raycoms: No, I agree, the previous tileEntity system was a mess. I am talking about handling how it is also handled with buildings and colonyviews.

Raycoms (dismissed):

PR #9330, opened 2023-08-22
new pr template

Raycoms:
Closes #
Closes #
Closes #

- New PR template
-
-

Review please

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -7,4 +7,7 @@ Closes #
 -
 -
 
+
+[ ] Yes I tested this before submitting it.
+
 Review please
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -9,5 +9,6 @@ Closes #
 
 
 [ ] Yes I tested this before submitting it.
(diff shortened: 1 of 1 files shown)


MotionlessTrain:
Just in case: does it matter I don't test all buildings before submitting a PR to change them?
I'm afraid I don't get to check that box often

Raycoms:
> Just in case: does it matter I don't test all buildings before submitting a PR to change them? I'm afraid I don't get to check that box often

No, its fine, its just so we got an idea. Building PRs dont have to be tested

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -7,4 +7,7 @@ Closes #
 -
 -
 
+
+[ ] Yes I tested this before submitting it.
```
Nightenom, line 11: well, better add sth like this:
a) tested in single new
b) tested in single existing
c) tested in multi

Raycoms: I adjusted it.

MotionlessTrain:
Aren't github templates just used from the default branch? It is now merged to version/1.29, which I don't think has an effect on pull requests at all

Raycoms:
Ill merge it to the main too

PR #9337, opened 2023-08-26
Fix not catching all errors

someaddons:
- Catch all throwables during creative tab errors, to include class def not found errors

Review please


src/api/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -641,7 +640,7 @@ private void discoverAllItems(final Level level)
                             stacks.add(stack);
                         });
                     }
-                    catch (final Exception ex)
+                    catch (final Throwable ex)
```
Nightenom, line 643: Honestly i would just add the specific error classes into clause, this will also catch memory/... errors

someaddons: that is fine, no matter what error some other mod has when gathering tabs we want to catch it. We're still logging the stacktrace anyways

MotionlessTrain: I have the feeling that if they are severe errors like OutOfMemorry, that Java will crash right after that with the same error anyway

PR #9363, opened 2023-09-07
Fix research citizen cap addition [1.20]

Thodor12:
Closes [discord issue](https://discord.com/channels/472875599422291968/486940969737125922/1149189690310799480)

- Added a new effect strenght level to citizen cap of 25, this is so the "relative" changes in the UI aren't messed up
- Text in the research is set back to "by", indicating relative change.
- Effect strenghts for researches have been shifted by 1, so that researches will not pick the correct level in the future.

Review please


src/main/java/com/minecolonies/coremod/client/gui/WindowResearchTree.java
```diff
@@ -632,8 +632,8 @@ private void generateResearchTooltips(final Button tipItem, final IGlobalResearc
             // CITIZEN_CAP's meaningful effect range is controlled by configuration file settings. Very low values will necessarily make their researches a little weird, but we should at least handle 'sane' ranges.
             // Only change the effect description, rather than removing the effect, as someone may plausibly use the research as a parent research.
             // I'd rather make these modifications during ResearchListener.apply, but that's called before config files can be loaded, and the other workarounds are even uglier.
-            if(research.getEffects().get(txt).getId().equals(CITIZEN_CAP)
-                 && (((GlobalResearchEffect)research.getEffects().get(txt)).getEffect() + 25) > IMinecoloniesAPI.getInstance().getConfig().getServer().maxCitizenPerColony.get())
+            if (research.getEffects().get(txt).getId().equals(CITIZEN_CAP)
```
Raycoms, line 635: Okay, but if nothing is researched, the level is 0 and the cap is 0?

Thodor12: Doesn't actually mean anything for this case, this only determines the text that should render on the information lines. Whether it should contain the actual number or 'MAX"

Raycoms: What if no research, does it work fine?

Thodor12: This only iterates the effects linked to a research, it doesn't get any values for a colony.
0 cannot occur here because there's no effect with a value of 0, only 25 (which is not used), 50, 100, etc

Raycoms: I mean, not here necessarily, but at the other places we query the limits

Thodor12: I did not change anything as to how the limits are calculated in the colony itself, that logic works fine and has remained the same.
The addition of this extra effect level is only visual for the research tree, this new effect level of 25 can never be researched.

Thodor12: All other researches for citizen cap it's effects were actually shifted by one, so the first citizen cap addition research actually gives you effect level 2, instead of 1, so it's immediately raised to 50.
The code still handles the initial +25

Raycoms (dismissed):

PR #9391, opened 2023-09-20
Dedicated buildings

uecasm:
- Consolidates "should try to auto-hire" logic to a single place, for convenience.
- Allows creation of "dedicated buildings" -- i.e. where only specific jobs are allowed.
    - For example, a style might want to make separate blueprints for Miner and Quarrier -- while both are technically a mine hut, they will only allow that specific job.
    - Another example might be a dedicated building for a particular guard type, with appropriate decorations.
    - To use, apply one or more tags directly to the hut block of the format `job=<job>`, where `<job>` is the internal id of the job (with or without namespace) that should be allowed.  For example, the currently reasonable tags to use are:
        - Knight: `job=knight`
        - Archer: `job=ranger`
        - Druid: `job=druid`
        - Miner: `job=miner`
        - Quarrier: `job=quarrier`
    - It's possible to make it so that nobody can be hired there, if you mess up the tags.  Don't do that.

Review please (do not port)


src/main/java/com/minecolonies/coremod/util/BuildingUtils.java
```diff
@@ -39,4 +53,48 @@ public static ItemStack getItemStackForHutFromInventory(final Inventory inventor
...
+     * @param building   the building to check.
+     * @param hiringMode the current hiring mode of the job.
+     * @param job        the job to hire, or null for a non-specific check.
+     * @return           true if automatic hiring is allowed.
+     */
+    public static boolean canAutoHire(@NotNull final IBuilding building,
+                                      @NotNull final HiringMode hiringMode,
+                                      @Nullable final JobEntry job)
+    {
+        return ((building.getBuildingLevel() > 0 && building.isBuilt()) || building instanceof BuildingBuilder)
```
Raycoms: Since you touch this anyway, I kinda feel like, we should actually make this a method in the building, and have an override for the builder.

uecasm: Interestingly, that already exists, but only in the view.  But it probably does make sense to move it to the building instead.

someaddons:
I like the idea, my only concern is how we make it clear for the players that this building e.g. only allows knights to be hired.
Some would be confused by it "My guard tower is broken, I can only hire knights!". Maybe you got some idea on how to transfer that information?

Raycoms:
> I like the idea, my only concern is how we make it clear for the players that this building e.g. only allows knights to be hired. Some would be confused by it "My guard tower is broken, I can only hire knights!". Maybe you got some idea on how to transfer that information?

I was thinking about this, aside of the blueprint name, I think we could probably display some information in the hiring window. So, there is a bit of info pre placement and then that info is post placement in the hiring window.

Thodor12:
Would this be limited to only work on certain buildings, not that you can accidently do this for a restaurant and disallow assistant cooks from working there or something?

uecasm:
> I like the idea, my only concern is how we make it clear for the players that this building e.g. only allows knights to be hired.
Some would be confused by it "My guard tower is broken, I can only hire knights!". Maybe you got some idea on how to transfer that information?

The job doesn't show up when it's not enabled, which I thought was a reasonably good indication by itself. 😀 But yeah, I'm mostly expecting that it will be obvious from the building name, like `archeryrange1` is unsurprisingly restricted to only rangers and `quarrier1` is only for quarriers.  I didn't really want to clutter up the UI with additional warnings or indicators; that might still just lead to "why is this disabled?" questions, and the only answer is "because the designer said so", which isn't especially satisfactory.

For upgrade safety if the style designer derps the tags (and because the player can add/remove the tags themselves, although they'll be replaced on upgrade), if someone is actually hired in a disabled job then it will still appear regardless, though it won't auto-hire a replacement, and if they die or are fired the job will disappear the next time you open the window.

> Would this be limited to only work on certain buildings, not that you can accidently do this for a restaurant and disallow assistant cooks from working there or something?

Currently there is deliberately no such restriction, so you can do that if you really want to.  It should even work basically as you'd expect, although it won't disable the assistant-only UI (perhaps that could be an extension, though, whenever the job is disabled or perhaps even just when there's no worker hired).

Operationally there isn't really any functional difference between the job being disabled and having it enabled but set to manual and with no worker hired; this is purely a UI thing, other than affecting auto-hiring.

someaddons:
Talked a bit with ray about the building names, I think it can work though it'll need some changes to achieve it. 
image
That is how our current building names look like, they're translated names not so much file names. 
Since we already have Job specific translation keys we could do with Job key + building key for specialized buildings for one specific job, else we'd have schematics add new translation entries which would be a pain.

So I think you should add that those schematic display nicely like that in the build tool UI + the built buildings default name

uecasm:
Yeah, but when you click the building name, you get the list of alternates, and the tooltip for those is the actual blueprint name.  I think it's likely that there will always be alternates when there are job-specific buildings like this.

I *do* think that we should probably just do away with the "alternate 1", "alternate 2" etc button labels and just put the blueprint name directly on that instead of only as a tooltip.  But that's a Structurize change, so unrelated to this PR.

Raycoms:
> Yeah, but when you click the building name, you get the list of alternates, and the tooltip for those is the actual blueprint name. I think it's likely that there will always be alternates when there are job-specific buildings like this.
> 
> I _do_ think that we should probably just do away with the "alternate 1", "alternate 2" etc button labels and just put the blueprint name directly on that instead of only as a tooltip. But that's a Structurize change, so unrelated to this PR.

I already did alter that

uecasm:
Then it should be fine, as long as there's alternates.

Raycoms:
We talked about this a bit more and the avg player will be a bit too confused if there is no in main GUI notification that this is a "single type only thing" we were thinking about having proper job icons to solve that

uecasm:
I don't think it would be all that confusing.  Provided that there are alternates, the schematic name will be visible, and should presumably be sufficient indication as already specified.

For example, it wouldn't really make sense for a style to *only* have a ranger guard tower -- it's more likely they'd have one of each, so the player would first click "guard tower" and then click one of "melee", "ranged", or "support" (or better names).

I need to update ldtteam/Structurize#614 but once that's merged it opens another possibility to have an explicit custom name, which could help in the rare case someone wants a dedicated building without alternates.

Raycoms:
> I don't think it would be all that confusing. Provided that there are alternates, the schematic name will be visible, and should presumably be sufficient indication as already specified.
> 
> For example, it wouldn't really make sense for a style to _only_ have a ranger guard tower -- it's more likely they'd have one of each, so the player would first click "guard tower" and then click one of "melee", "ranged", or "support" (or better names).
> 
> I need to update [ldtteam/Structurize#614](https://github.com/ldtteam/Structurize/pull/614) but once that's merged it opens another possibility to have an explicit custom name, which could help in the rare case someone wants a dedicated building without alternates.

The schematic name is not visible in buildings afaik, which is the problem.

uecasm:
> The schematic name is not visible in buildings afaik, which is the problem.

It is, though, as long as there are alternates.  Or if you use the custom names PR 😁

Raycoms:
> > The schematic name is not visible in buildings afaik, which is the problem.
> 
> It is, though, as long as there are alternates. Or if you use the custom names PR 😁

No, I mean, after placed.

uecasm:
> No, I mean, after placed.

Then you can see it in the Manage Workers screen.  And presumably by the themed decoration (or obvious presence or lack of a mineshaft) that the style designer used to differentiate them as well.

uecasm:
Closed in favour of #9653 since 1.19 is feature-locked now.

PR #9412, opened 2023-09-26
Few raid improvements

someaddons:
- Add officer command now also adds as important packet receiver
- Raiders now spread into bigger groups and spread out less
- Raiders now spawn less far away
- Guards get alerted when a nearby building is visited by raiders
- Pathing cost for doors and other toggleables increased
- Fix gates not counted as door for vanilla break AI

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/mobs/AbstractEntityMinecoloniesMob.java
+++ b/src/api/java/com/minecolonies/api/entity/mobs/AbstractEntityMinecoloniesMob.java
@@ -249,7 +249,6 @@ public abstract class AbstractEntityMinecoloniesMob extends Mob implements IStuc
             newNavigator.setSwimSpeedFactor(getSwimSpeedFactor());
             this.newNavigator.getNodeEvaluator().setCanPassDoors(true);
             newNavigator.getPathingOptions().withDropCost(1.3D);
-            newNavigator.getPathingOptions().withDropCost(1.3D);
             PathingStuckHandler stuckHandler = PathingStuckHandler.createStuckHandler()
                                                  .withTakeDamageOnStuck(0.4f)
                                                  .withBuildLeafBridges()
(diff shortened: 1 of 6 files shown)


src/api/java/com/minecolonies/api/entity/mobs/AbstractEntityMinecoloniesMob.java
```diff
@@ -259,6 +249,7 @@ public AbstractAdvancedPathNavigate getNavigation()
             newNavigator.setSwimSpeedFactor(getSwimSpeedFactor());
             this.newNavigator.getNodeEvaluator().setCanPassDoors(true);
             newNavigator.getPathingOptions().withDropCost(1.3D);
+            newNavigator.getPathingOptions().withDropCost(1.3D);
```
Raycoms: twice the same?

src/api/java/com/minecolonies/api/entity/pathfinding/PathingOptions.java
```diff
@@ -43,7 +43,7 @@ public class PathingOptions
     /**
      * Cost to traverse trap doors
      */
-    public double traverseToggleAbleCost = 2D;
+    public double traverseToggleAbleCost = 10D;
```
Raycoms, line 46: I would only set that for raiders I think?

someaddons: nah I think both is fine, going through doors for citizens is also rather difficult

src/api/java/com/minecolonies/api/entity/mobs/AbstractEntityMinecoloniesMob.java
someaddons: ops

src/api/java/com/minecolonies/api/entity/pathfinding/PathingOptions.java
Raycoms: Yeah, but I think for citizens its nice if "they take the main door" instead of taking the back entrance for example =D

someaddons: well normally both of them have doors^^ also 10 isnt that high, its kinda check 10 blocks to the side before going here

PR #9428, opened 2023-10-03
added editorconfig for formatting in "all" IDEs

efinder2:
- adding an [editorconfig ](https://editorconfig.org/#pre-installed) file. This is a file format which could be read by almost all IDEs. For some (e.g. Eclipse) a plugin is needed. The big advantage is [IntelliJ ](https://www.jetbrains.com/help/idea/editorconfig.html) overrules the project and user settings, if the file is present.  So even if the configuration of the local configuration in the IDE is changed, the standard in the file is used. 

I imported the IntelliJ coding spec xml and exported them as editorconfig. It not only wrote the Java spec from the xml but also for all other languages with the default values in IntelliJ. It might be debatable whether only currently used formats should be specified or also all other kinds.

In the future it would be an option to verify the code style against the config during build. Discussions in pull requests for formatting would not be necessary.

^^ Don't apply. Is a coding change.

Review please


Thodor12:
We already have style configuration files for IntelliJ and other most used IDE's for Java, so this likely will not happen.

(Do note that Editorconfig does not support every IntelliJ or Eclipse specific rules, and exported Editorconfig `ij` rules do not work in Eclipse etc)

PR #9430, opened 2023-10-03
Fix Alex's Mobs Lasso breaking when lasso-ing citizens

Thodor12:
Closes #9369 
Closes several discord reports

# Changes proposed in this pull request:
- Changes the logic in the ChunkCache to not be side dependent anymore
  - This did create a change in the chunk loading logic, where the ServerChunkCache attempts to do a blocking load, but considering that there is a chunk load protection around the method, that seems a non-issue to Mirality

[X] Yes I tested this before submitting it.
[ ] I also did a multiplayer test.

Review please

PR #9431, opened 2023-10-03
Fix Alex's Mobs Lasso breaking when lasso-ing citizens [1.20]

Thodor12:
See #9430

### DIFF
--- a/src/main/java/com/minecolonies/coremod/entity/pathfinding/ChunkCache.java
+++ b/src/main/java/com/minecolonies/coremod/entity/pathfinding/ChunkCache.java
@@ -1,59 +1,55 @@
 package com.minecolonies.coremod.entity.pathfinding;
 
 import com.minecolonies.api.util.WorldUtil;
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
 import net.minecraft.core.Holder;
 import net.minecraft.core.RegistryAccess;
-import net.minecraft.server.level.ChunkHolder;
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/coremod/entity/pathfinding/ChunkCache.java
```diff
@@ -69,9 +67,9 @@ public ChunkCache(Level worldIn, BlockPos posFromIn, BlockPos posToIn, int subIn
         {
             for (int l = this.chunkZ; l <= j; ++l)
```
Raycoms, line 68: Im not sure this wouldnt just break somewhere else. We should probably prevent this much earlier to reach this place in the first place then

Thodor12: Right now this is just meant to prevent the classcast exception, I tested it and it still works.
It's up to Alex's Mobs to just not call it on the client.

Raycoms: Did you test that it doesnt just now crash further down the road because of the missing chunkcache?

Thodor12: Yep, I threw it on them, and dragged them out for multiple chunks far, nothing special happened, not even anything present in the log files that would indicate an issue.

After releasing them they just continued to path normally,

PR #9439, opened 2023-10-09
Add Block Hut name to the description.

alexandersokol:
Hello!

This is kinda micro QaL update. For some translations full hut names were not fully fit and sometimes it wasn't easy to understand what kind of building is it (especially for newcomers).

Ukrainian translation for Composter's hut just for sample:
image
image

- Added hut name to the hint pop-up

Review please


Thodor12:
Hello

Could you modify the translation key to just use `getBuildingEntry().getTranslationKey()`, like the line directly below.
It's better if you stick to existing methods instead of introducing your own dynamic keys, which would only add more technical debt in case things have to change.

Additionally, are you able to provide screenshots in English?

alexandersokol:
Hello!


To be honest I took this line from the method in the current file from here:
https://github.com/ldtteam/minecolonies/blob/6b43da64056662236224e0cea9615cef4f17f0ed/src/api/java/com/minecolonies/api/blocks/AbstractBlockHut.java#L427

Mostly English texts fit perfectly because it has shorter names.

I will update it a bit later.

Thank you for the review.

alexandersokol:
Hi @Thodor12 !

I checked the keys that are used for the title and description, looks like they are generated in different ways. To use `getBuildingEntry().getTranslationKey()` might require changing translation keys in JSON for all huts and all code that relies on these keys.

```
Component.translatable("block." + Constants.MOD_ID + "." + getHutName()).toString()
// translation{key='block.minecolonies.blockhutbuilder', args=[]}
```
vs
```
Component.translatable(getBuildingEntry().getTranslationKey() + ".desc")
// translation{key='com.minecolonies.building.builder.desc', args=[]}
```

Also attaching the screenshot with English translation:
image

Thodor12:
These translation keys point to exactly the same values, you also shouldn't put `.desc` behind it.

Raycoms:
We did an alteration of translation keys for a few of these things quite a while ago, so, we have some duplicate values with different keys.

Thodor12:
Yeah block and building keys should be exactly the same, I checked in the translation files, if they do not match for some buildings then that's a bug.
(I'd prefer to actually see the `block.` translation keys disappear entirely)

MotionlessTrain:
> These translation keys point to exactly the same values, you also shouldn't put `.desc` behind it.

Fwiw, the key with ".desc" is the part in the tooltip that didn't change. Removing that .desc will break all descriptions of buildings in the tooltips

alexandersokol:
Thanks!
I pushed new changes.

I updated block hut name with translation key as you previously said, and I left `+ ".desc"` as it was.

Raycoms (approved):

Thodor12 (approved):
Looks good, although the second one wasn't strictly needed I won't object to it either

PR #9486, opened 2023-10-18
Trigger build event for creative tabs too

uecasm:
Closes uecasm/colony4cc#25

- Also fire the `BuildCreativeModeTabContentsEvent` when building creative tabs, same as Forge.
    - ~~This also uses public methods instead of private internals.~~
    - Caveat: as before, this fires on server (and datagen), not on client as some mods will be expecting, so this might add some more mod incompatibilities (if they do the "wrong" things) or make no difference (if they register the event only client-side).
- Hides the "gamemaster only" items (light, barrier, jigsaw, command block, etc) from the item list (and postbox), since they're never going to be craftable anyway.

Review please (do not port)


someaddons:
would be interesting to see if it works better or not in an actual modpack, due to the compat with other mods (esp on server side)

Raycoms:
> would be interesting to see if it works better or not in an actual modpack, due to the compat with other mods (esp on server side)

Did you test this in a bigger pack? Would be interesting to load atm on 1.20 with this change to see if anything breaks

uecasm:
Welp, turns out this just crashes servers, due to an unanticipated classloading issue.  Other than that, though, ATM9 works just fine. 🤣 

With the rewrite, ATM9 loads without errors (from this, anyway, they have plenty of other errors) on server now too.

Raycoms:
> Welp, turns out this just crashes servers, due to an unanticipated classloading issue. Other than that, though, ATM9 works just fine. 🤣
> 
> With the rewrite, ATM9 loads without errors (from this, anyway, they have plenty of other errors) on server now too.

Can we try catch that?

uecasm:
> Can we try catch that?

Try catch what?  There's no errors any more.

Raycoms:
conflict

Raycoms (dismissed):

uecasm:
> conflict

_shakes fist at runData_

someaddons (approved):

PR #9490, opened 2023-10-19
Research costs based on multiple items/tags

Thodor12:
- Created a new registry for research costs, which has 3 possible entries
  - Simple item costs: These are the old type of costs, which can still be found in datapacks, they no longer appear in our own json files
  - List item costs: These costs allow for a list of items to appear in the cost, these are also used by single item costs too, for simplicity sake
  - Tag item costs: These costs use a tag to allow any kind of item belonging to that given tag, for example `diamond_ores`.
-  Applied tag research for ores, planks and concrete
- Ran datagen to update all research files
- Ensured migration of old research types still works (mainly for datapacks)

Review please


src/main/java/com/minecolonies/coremod/client/gui/blockui/RotatingItemIcon.java
```diff
@@ -0,0 +1,110 @@
...
+import net.minecraft.world.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Variation of the ItemIcon which accepts multiple items and slowly rotates between them.
+ */
+public class RotatingItemIcon extends ItemIcon
```
Raycoms, line 17: Feel like this should be in BlockUI

Nightenom: Intentionally here, on my to-do list

Raycoms: Okay, then just add a "todo to be moved to blockui by nightenom" here

Raycoms (dismissed):

PR #9493, opened 2023-10-19
Settings views rework

Thodor12:
- Settings views have been updated so the controls/inputs are on separate lines from the text itself
- Settings now always have a label above them to indicate what the setting is for, this was previously missing for a few settings (most notably the builder)
- Settings will (by default) no longer be hidden from the settings list, instead they will appear disabled, optionally also giving a reason as to why they are disabled (for example missing research, etc)
  - Settings may still be hidden entirely by turning a different method it's return value to true, this is true for some settings like Dynamic tree compat for the forester, useful in case of dependencies on other mods
- Migration done from an existing colony against the current previous version, where parsing at least one setting of each setting type, all setting values retained properly

Guard tower:

2023-10-19 21_58_00-Minecraft_ 1 19 2 - Singleplayer

Crusher:

2023-10-19 21_58_16-Minecraft_ 1 19 2 - Singleplayer

Builder:

2023-10-19 21_58_32-Minecraft_ 1 19 2 - Singleplayer
2023-10-19 21_58_38-Minecraft_ 1 19 2 - Singleplayer

Review please


uecasm:
> Settings views have been updated so the controls/inputs are on separate lines from the text itself

I'm not sure "takes up more vertical space" is a feature. 😆  Though granted this probably allows for easier translation.

someaddons (approved):

PR #9499, opened 2023-10-20
Fix crash when selecting warehouse in another colony using resource scroll

Thodor12:
Closes #9496 

# Changes proposed in this pull request:
- Fix crash when selecting a warehouse in another colony for a snapshot
- Extract logic for calculating snapshot to the item, not the window

[X] Yes I tested this before submitting it.
[ ] I also did a multiplayer test.

Review please

PR #9500, opened 2023-10-20
Fix crash when selecting warehouse in another colony using resource scroll [1.20]

Thodor12:
See #9499


Raycoms (commented):
That seems like a lot of changes for such a small fix?

Thodor12:
> That seems like a lot of changes for such a small fix?

It was easier this way, previously all of the warehouse handling was done in the window, all on the client, which wasn't ideal, had some troubles getting the fix in that way.

So I opted to pull the logic up to the item itself, prior to even opening the window, that way the window code is not so polluted and all the special conditions were easier to make before opening the window.

Raycoms:
> Reference

that was intentional though, no need to involve the server with something that could be done entirely on the client.

Thodor12:
> > Reference
> 
> that was intentional though, no need to involve the server with something that could be done entirely on the client.

That's not entirely true, the client was modifying the snapshot, which required a message class to persist that on the server, else the snapshot data would not be present.
Doing that on the server to begin with means I can also remove the message class (which I forgot, let me do that)

Raycoms:
> > > Reference
> > 
> > 
> > that was intentional though, no need to involve the server with something that could be done entirely on the client.
> 
> That's not entirely true, the client was modifying the snapshot, which required a message class to persist that on the server, else the snapshot data would not be present. Doing that on the server to begin with means I can also remove the message class (which I forgot, let me do that)

Yeah, but I did not want to do this relatively expensive call to happen on the server. The entire indexing warehouse stuff for all the items in the list, not that amazing to do on the server if we can compute this entirely on the client and just synch the result.

Raycoms (changes requested):
This intentionally was computed on the client

Thodor12:
Then I'll modify it so the info is calculated upon calling openWindow, but not in the window itself.
I just got rid of all that code mess in there

Raycoms:
> Then I'll modify it so the info is calculated upon calling openWindow, but not in the window itself. I just got rid of all that code mess in there

What does it matter if the exact same code is in the window, or in the item? From a maintenance perspective its easier in the window because thats where we expect code related to this feature, the item has a bunch of other features aside of that.

Thodor12:
In the window it's a little bit messy, because I had to load the building as part of the constructor, which means it could optionally be null.

That meant in every single function I had to account for the building being null as well, which quickly becomes annoying.
Putting all that logic before the window constructor is even called means I can guarantee the building in the window is even there to begin with.
That also means I don't have to pass the compound data of the item around into the window itself (seperation of concern), the item itself should manage the compound data, not the window.

Raycoms:
> In the window it's a little bit messy, because I had to load the building as part of the constructor, which means it could optionally be null.
> 
> That meant in every single function I had to account for the building being null as well, which quickly becomes annoying. Putting all that logic before the window constructor is even called means I can guarantee the building in the window is even there to begin with. That also means I don't have to pass the compound data of the item around into the window itself (seperation of concern), the item itself should manage the compound data, not the window.

I mean we can just do a colony id verification on the server before opening. Or, in the constructor once verify the building is not null, and else exit

Raycoms (approved):

someaddons (approved):

PR #9515, opened 2023-10-23
Citizen sync imp

someaddons:
Closes #
Closes #
Closes #

- Add a few more citizen improvements for server performance
- Change citizen markDirty to markDirty(time) to allow a more on demand sync style by giving it ticks to update

Review please


src/main/java/com/minecolonies/coremod/colony/CitizenData.java
```diff
@@ -1451,6 +1457,15 @@ public void tick()
             return;
         }
 
+        if (dirty > 0)
+        {
```
Raycoms: is this per tick or already only per s?

someaddons: oh good catch, its per 3 seconds

Raycoms (dismissed):

PR #9540, opened 2023-10-30
Rework delete command so it's less complex to use

Thodor12:
- No longer requires 2 rudimentary string arguments for "delete Buildings", which serve absolutely no purpose
- Removed some unused duplicate execution code in the command logic

New changes post review:
- The command has still been modified, the command now only takes 2 boolean arguments
- If both arguments have been omitted, you will enter a "guided" command, which basically means you will get a set of 2 chat messages, asking you what you want to do.
  - The first chat message will ask you whether you want to keep or delete the buildings
  - The second chat message is a general verification, "Are you sure you want to do this?" (it will optionally state in bold if you chose to delete buildings, else it does not state anything)
- The full command is intended to be used by OPS in console, or you know what you're doing. In that case you just type `/mc colony delete <colonyId> (true|false) true`

2023-11-02 12_54_19-Window

Review please


MotionlessTrain:
Note that the "delete buildings" and "keep buildings" part was deliberate, to make it less easy to make mistakes and delete those by accident

Thodor12:
Except this is not enforced by any means.
You could type `/mc colony delete 1 true x x` and it'd do the exact same

Raycoms:
Ye, I dont see an issue with that too

Raycoms:
> Except this is not enforced by any means. You could type `/mc colony delete 1 true x x` and it'd do the exact same

Yes, intentionally

Raycoms:
There is the "I know what I am doing command" and then the auto complete which is more clear

Thodor12:
Would it work if the argument is optional, defaulting to false (if that's even possible)

someaddons:
ye this was fully intentional, players are using our commands during gameplay and deleting buildings is not reversible even with backups

Raycoms:
> ye this was fully intentional, players are using our commands during gameplay and deleting buildings is not reversible even with backups

Although the command went derp, int he past we had this boolean to bypass instant delete.
So, normally the command asks you "are you sure" and the boolean bypasses this.

Couldnt find this logic anymore on a quick look. Mustve been lost at some port

someaddons:
s[0] = "true delete buildings";
        s[1] = "false keep buildings";
        
        it still does, its either or

Raycoms:
> ```
>     s[0] = "true delete buildings";
>     s[1] = "false keep buildings";
>     
>     it still does, its either or
> ```

I mean, it asked before doing the action even.

someaddons:
ye t hats like the super old 1.12 one

Thodor12:
Updated description, please re-check @Raycoms @someaddons @MotionlessTrain

MotionlessTrain:
Just to be sure: The first boolean argument is to delete or keep buildings?
And the second is "Are you sure?"?
What happens if you omit the second boolean?

Thodor12:
> Just to be sure: The first boolean argument is to delete or keep buildings? And the second is "Are you sure?"? What happens if you omit the second boolean?

The commands are chained after one another.
If you were to omit the second boolean, you will simply get the confirm question only.

The first is still whether to keep (false) or delete (true)

MotionlessTrain (approved):

Talyda:
Why is it changing? Just curious! Isn't that command same as the old one basically?  I mean, I know one person had trouble with it the other day, but that was because they didn't read the example right next to the command they were using, but I don't recall many other people having any trouble with it

uecasm:
I don't have any strong preference for changing it (it then becomes a version-specific difference you have to explain on the wiki).  But the guided links are probably slightly more user-friendly than the auto-complete.

Raycoms (approved):

someaddons (approved):

PR #9553, opened 2023-11-06
A try & catch to avoid game crashes

HexatomicRing:
Closes #
Closes #
Closes #

-
-
-

IDK why my client sometimes crashes while placing a blue print. I'm using the same mod files that on server. It seems a blueprint nearby cannot find its path. Then the program tries to cutoff the last char of the empth path, and throws a StringIndexOutOfBoundExcepption.

By the way, I'm using Birch Urban pack, and the mod seems to try searching  /blueprints/minecolonies/birch (and throws NoSuchFileException).

Review please


src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
```diff
@@ -395,8 +396,11 @@ public Map<BlockPos, PendingRenderData> getDesiredBlueprints(final WorldEventCon
                     {
                         String schemPath = buildingView.getStructurePath();
                         schemPath = schemPath.replace(".blueprint", "");
-                        schemPath = schemPath.substring(0, schemPath.length() - 1) + buildingView.getBuildingMaxLevel() + ".blueprint";
-
+                        try{
+                            schemPath = schemPath.substring(0, schemPath.length() - 1) + buildingView.getBuildingMaxLevel() + ".blueprint";
```
Raycoms, line 399: If it crashes here its probably because schemPath is empty. So we can do an isEmpty check and add some useful logging.

gradle.properties
```diff
@@ -3,7 +3,7 @@ org.gradle.daemon=false
 
 modId=minecolonies
 modGroup=com.ldtteam
-modVersion=0.0.1-LOCAL
+modVersion=1.1.297-BETA
```
Thodor12: Can you leave this as it is?

HexatomicRing: Do I need to re-pull a request?

Raycoms: No you can push to your local branch just the changes

src/main/java/com/minecolonies/coremod/client/render/worldevent/ColonyBlueprintRenderer.java
Thodor12: Also, just noticed that this entire section has not had the formatter run over it.

Raycoms:
solved

PR #9564, opened 2023-11-12
Change surnames and firstnames to Chinese characters (1.20.1)

mc-kaishixiaxue:
Test screenshot


Raycoms:
@mc-kaishixiaxue can you adjust the PR to add it as a second chinese set instead of replacing it?

Raycoms (approved):

PR #9609, opened 2023-12-08
properly handle child adjustment, and drop gear on death

Raycoms:
Closes #
Closes #
Closes #

- Children when growing up should properly adjust dimensions/model
- When guards die, they drop their gear
-

Review please


src/main/java/com/minecolonies/coremod/entity/citizen/EntityCitizen.java
```diff
@@ -1110,11 +1110,13 @@ public void setIsChild(final boolean isChild)
         {
             if (!isChild && this.child)
             {
+                this.child = isChild;
```
someaddons, line 1113: copy paste error?

someaddons (approved):

PR #9618, opened 2023-12-12
Settings and serialization fixes

someaddons:
Closes #
Closes #
Closes #

Fix missing settings
Fix network errors not displaying
Fix Hire fire message error
Improve chunk claim error logging

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/colony/buildings/modules/BuildingModules.java
+++ b/src/main/java/com/minecolonies/coremod/colony/buildings/modules/BuildingModules.java
@@ -459,6 +459,7 @@ public class BuildingModules
       new BuildingEntry.ModuleProducer<>("miner_settings", () -> new SettingsModule()
         .with(AbstractCraftingBuildingModule.RECIPE_MODE, new CrafterRecipeSetting())
         .with(BuildingMiner.FILL_BLOCK, new BlockSetting((BlockItem) Items.COBBLESTONE))
+        .with(BuildingBuilder.BUILDING_MODE, new BuilderModeSetting())
         .with(BuildingMiner.MAX_DEPTH, new IntSetting(-100))
         .with(AbstractBuilding.USE_SHEARS, new BoolSetting(true)), () -> SettingsModuleView::new);
     public static final BuildingEntry.ModuleProducer<IBuildingModule,MinerGuardAssignModuleView> MINER_GUARD_ASSIGN    =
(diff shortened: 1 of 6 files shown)


Raycoms (approved):

Thodor12 (approved):

src/main/java/com/minecolonies/coremod/colony/buildings/modules/BuildingModules.java
```diff
@@ -459,6 +459,7 @@ public class BuildingModules
       new BuildingEntry.ModuleProducer<>("miner_settings", () -> new SettingsModule()
         .with(AbstractCraftingBuildingModule.RECIPE_MODE, new CrafterRecipeSetting())
         .with(BuildingMiner.FILL_BLOCK, new BlockSetting((BlockItem) Items.COBBLESTONE))
+        .with(BuildingBuilder.BUILDING_MODE, new BuilderModeSetting())
```
uecasm, line 462: It was previously intended that the miner did not have this setting, so that they'd always use "default" for predictable builds.  Not that I'd personally object to allowing it to be selectable.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/BuildingModules.java
```diff
@@ -513,7 +514,7 @@ public class BuildingModules
       new BuildingEntry.ModuleProducer<>("tool_scepterguard_view", null, () -> () -> new ToolModuleView(
         ModItems.scepterGuard));
     public static final BuildingEntry.ModuleProducer<SettingsModule,SettingsModuleView> GUARD_SETTINGS = new BuildingEntry.ModuleProducer<>("guard_settings", () -> new SettingsModule()
-      .with(AbstractBuildingGuards.GUARD_TASK, new GuardTaskSetting(GuardTaskSetting.PATROL, GuardTaskSetting.GUARD, GuardTaskSetting.FOLLOW))
+      .with(AbstractBuildingGuards.GUARD_TASK, new GuardTaskSetting(GuardTaskSetting.PATROL, GuardTaskSetting.GUARD, GuardTaskSetting.FOLLOW, GuardTaskSetting.PATROL_MINE))
```
uecasm, line 517: Firstly, if you want all the values then you can use the parameterless constructor.

Secondly, this will give barracks guards the ability to patrol the mine as well, which was not previously the case and is probably not intended (as you'd still only be able to assign a single guard at the mine itself and then the squad AI would probably get borked).  These settings were deliberately different between guard tower and barracks tower.

src/main/java/com/minecolonies/coremod/colony/buildings/modules/BuildingModules.java
Raycoms: Huh, odd the miner was throwing exception that they "want it"

uecasm: Possibly a side effect of the refactor?  They definitely didn't have that setting before.

MotionlessTrain: Before, it was indeed intended that they used the default one

PR #9625, opened 2023-12-17
Herder improvements

someaddons:
Closes #
Closes #
Closes #

- Herders now start breeding from the center and butcher from the outside, to reduce the likelyhood of reproducing animals outside the pen
- Butchering now also prefers animals which are under some kind of roof
- Added occasionally feeding animals

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/ai/statemachine/states/AIWorkerState.java
+++ b/src/api/java/com/minecolonies/api/entity/ai/statemachine/states/AIWorkerState.java
@@ -368,6 +368,11 @@ public enum AIWorkerState implements IAIState
     /**
      * Pickup items within area.
      */
+    HERDER_FEED_CHILDREN(false),
+
+    /**
+     * Feed animals.
+     */
     HERDER_FEED(false),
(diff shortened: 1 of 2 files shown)


Raycoms (dismissed):
Looks good to me, would only adjust "20" to the "TICKS_SECOND" constant

uecasm:
The change from `maxAnimals` to `chanceToButcher` seems to make it so that instead of keeping 2*LEVEL animals minimum and butchering the extras, the herder will randomly butcher until there are only 3 adults (with a slightly higher chance to butcher if there are more adults and lower chance to butcher at higher level, but non-zero means they will still eventually butcher down to 3 regardless of level).

Is that intended?  It seems especially problematic for shepherds where you want to maintain a specific population size of adults for shearing purposes (and to a lesser extent for the cowboy, for milking).  It also makes it harder to do "peaceful runs" where you disable breeding and keep the population within limits yourself so that the herder never butchers.  (Which again is mostly only useful for sheep and cows, or just for aesthetics rather than drops.)

PR #9633, opened 2023-12-18
Fix composted dirt loot dropping

Thodor12:
- Composted dirt loot table did not exist, not present in DefaultBlockLootProvider
- Improved DefaultBlockLootProvider so blocks can have more complex loot drop conditions, instead of just survives explosion.

Review please


src/main/java/com/minecolonies/coremod/generation/defaults/DefaultBlockLootTableProvider.java
```diff
@@ -55,6 +64,17 @@ protected void registerTables(@NotNull final LootTableRegistrar registrar)
...
+        saveBlock(ModBlocks.blockCompostedDirt, lootPool -> {
+            final AlternativesEntry.Builder alternatives = AlternativesEntry.alternatives();
+
+            final LootPoolSingletonContainer.Builder<?> compostedDirt = LootItem.lootTableItem(ModBlocks.blockCompostedDirt);
+            needsSilkTouch(compostedDirt);
+            alternatives.otherwise(compostedDirt);
+
+            alternatives.otherwise(LootItem.lootTableItem(Blocks.DIRT).when(ExplosionCondition.survivesExplosion()));
+
+            lootPool.add(alternatives);
```
uecasm: Builder methods are fluent and should be called as such without great reason to do otherwise.

Also it's nicer to have lambda arguments at the end of the signature so that you don't get that seemingly orphaned `registrar` argument that's easy to overlook.

Raycoms (approved):

PR #9647, opened 2023-12-25
Introduce placement handler for copper variants

Thodor12:
- Placement handler to make sure that oxidized copper variations of blocks are kept, the handler will either:
  - Replace the block if schematic paste is on, no matter what
  - Replace the block in the world if a non copper block (or wrong copper block variation (full, cut, stair, etc)) is present
  - Replace the block in the world if the oxidized state is lower than the one in the schematic (ex. world contains regular copper but schematic wants oxidized copper)
  - Else keep the block original copper block

Review please

### DIFF
--- a/src/main/java/com/minecolonies/coremod/placementhandlers/PlacementHandlerInitializer.java
+++ b/src/main/java/com/minecolonies/coremod/placementhandlers/PlacementHandlerInitializer.java
@@ -36,5 +36,6 @@ public final class PlacementHandlerInitializer
         PlacementHandlers.add(new LecternPlacementHandler());
         PlacementHandlers.add(new HutPlacementHandler());
         PlacementHandlers.add(new InfestedBlocksPlacementHandler());
+        PlacementHandlers.add(new WeatheredCopperPlacementHandler());
     }
 }
--- a/src/main/java/com/minecolonies/coremod/placementhandlers/WeatheredCopperPlacementHandler.java
+++ b/src/main/java/com/minecolonies/coremod/placementhandlers/WeatheredCopperPlacementHandler.java
@@ -1,0 +1,113 @@
+package com.minecolonies.coremod.placementhandlers;
+
+import com.ldtteam.structurize.placement.handlers.placement.IPlacementHandler;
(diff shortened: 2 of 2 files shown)


Moobien:
But steampunk already accounts for this via placeholders...

MotionlessTrain:
> But steampunk already accounts for this via placeholders...

We did so for preserving copper blocks during upgrades, indeed.
But repairing a building would replace the copper blocks with less oxidised ones, which can be quite annoying for e.g. the mine or the alt barracks

Moobien:
> > But steampunk already accounts for this via placeholders...
> 
> We did so for preserving copper blocks during upgrades, indeed. But repairing a building would replace the copper blocks with less oxidised ones, which can be quite annoying for e.g. the mine or the alt barracks

A repair would only cause that issue for levels where copper gets added, not for any subsequent levelw

Thodor12:
> > > But steampunk already accounts for this via placeholders...
> > 
> > 
> > We did so for preserving copper blocks during upgrades, indeed. But repairing a building would replace the copper blocks with less oxidised ones, which can be quite annoying for e.g. the mine or the alt barracks
> 
> A repair would only cause that issue for levels where copper gets added, not for any subsequent levelw

A repair wouldn't be effective for this as it would keep air blocks in case something was removed, not to mention people that play creative who intend to build a level 5 pasted colony.

Moobien:
> > > > But steampunk already accounts for this via placeholders...
> > > 
> > > 
> > > We did so for preserving copper blocks during upgrades, indeed. But repairing a building would replace the copper blocks with less oxidised ones, which can be quite annoying for e.g. the mine or the alt barracks
> > 
> > 
> > A repair would only cause that issue for levels where copper gets added, not for any subsequent levelw
> 
> A repair wouldn't be effective for this as it would keep air blocks in case something was removed, not to mention people that play creative who intend to build a level 5 pasted colony.

Well they shouldnt be lvl 5 pasting anyway :p

Thodor12:
We literally did so on stream not too long ago to build a sample colony though.

Raycoms (approved):

PR #9663, opened 2023-12-30
Respect minimum animal count

uecasm:
Closes [comment](https://github.com/ldtteam/minecolonies/pull/9625#issuecomment-1872416916) and [discord report](https://discord.com/channels/139070364159311872/1125094832143077456/1190462777597378590)

- When breeding is off, avoids butchering entirely if already at minimum animal count for level.


Raycoms:
I think it was intentional, and I think the chance was supposed to be "low enough" so that it keeps a large amount of animals around (e.g. breeding would be slightly faster than butchering).

I do understand the issue with, previously you could turn off breeding, which would leave a bunch in there, now it will end up with 3

someaddons:
yes it is intentional, with breeding it averages out after a while. only issue probably is that he butchers them when breeding is off too

Raycoms:
> yes it is intentional, with breeding it averages out after a while. only issue probably is that he butchers them when breeding is off too

We can do a "if breeding setting is off, stop butchering earlier"

uecasm:
K, adjusted to check setting.

Raycoms (approved):

someaddons (approved):

PR #9666, opened 2024-01-01
make mangrove roots compostable

IchHabeHunger54:
- Makes Mangrove Roots compostable. They don't really have a use besides being used as fuel, so I figured making them compostable too would be a good idea.

Review please


Talyda:
Yessss! Please! Thanks IHH! =D

Thodor12:
Does this include muddy mangrove roots as well?

IchHabeHunger54:
Not currently, no. I could make Muddy Mangrove Roots compostable as well, though it is worth noting that they are not compostable in the vanilla composter either (while regular Mangrove Roots are).

someaddons (approved):

PR #9680, opened 2024-01-05
Performance improvements and some fixes

someaddons:
- Change base rate of AI updates to 5 ticks, to reduce overhead, AI actions still happen in the same frequency(with a lower limit of every 5 ticks)
- Tickrate state machines now expect to be called at their given tickrate
- Fix crash with some modded bows
- Improve building full check speed
- Improve worldutil performance

[ X] I also did a multiplayer test.

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickingTransition.java
+++ b/src/api/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickingTransition.java
@@ -118,9 +118,9 @@ public class TickingTransition<S extends IState> extends BasicTransition<S> impl
      * @return random
      */
     @Override
-    public int countdownTicksToUpdate()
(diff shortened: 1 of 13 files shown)


src/api/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickingTransition.java
```diff
@@ -118,9 +118,9 @@ public void setTickRate(final int tickRate)
      * @return random
      */
     @Override
-    public int countdownTicksToUpdate()
+    public int countdownTicksToUpdate(final int reduction)
     {
-        return --ticksToUpdate;
+        return ticksToUpdate -= reduction;
```
Raycoms, line 123: A thing I wondered about, what is the tick rate we have a goal to be at, is not multiples of this? Would this still work?

PR #9693, opened 2024-01-08
Update fields list visuals

Thodor12:
- Modified "calcDirection" to return an enum, not a component, which allows for some more flexibility in comparison before showing the component
- Fields list of the plantation and farm have been updated with an improved loo

Farm GUI

Review please


Raycoms (approved):

someaddons:
I think the "taken by other farm already" should be a disabled button thats crossed out instead of checked, checked kinda seems like you can take it once you free up a slot

Raycoms:
> I think the "taken by other farm already" should be a disabled button thats crossed out instead of checked, checked kinda seems like you can take it once you free up a slot

Doubling down on this. Can you make a screenshot of how a field taken up by someone else looks like?

Thodor12:
Fields that are taken by another farm simply do not appear on this list at all

Thodor12:
This was the second farmer I had when I was testing this

someaddons (approved):

PR #9700, opened 2024-01-12
refactor colony chunk cap access

Raycoms:
Closes #
Closes #
Closes #

- Use utils to check the chunk status of a colony instead of doing it directly over the cap
- Replace 0 with the NO_COLONY_ID constant
-

[ x ] I also did a multiplayer test.

Review please


someaddons (dismissed):

someaddons (changes requested):
New PR for the visitor

someaddons (dismissed):

PR #9711, opened 2024-01-15
Visitor types

Thodor12:
This change is needed to support the expedition system, but I pulled this separately to not make the expedition PR too big.

- Visitors can now be defined as separate types (currently only the regular visitor) by using a VisitorType instance, this class controls some special logic to that specific visitor
- Specific data like the custom texture, recruit costs and sitting position has been removed from the visitor data and moved into a settings-esque system where you request "extra data" by giving a property key.
This has been done so the expeditionary can also support their own unique data.
- VisitorData has had a complete reformat since the file was an absolute mess, the main things that actually changed are:
  - NBT loading now uses the visitor type to construct it's visitor data
  - Entity type getter is forwarded to visitor type
  - Extra data
  - (De)Serialization
- VisitorCitizen has been reworked to also receive the visitor type instance, in order to control it's state machine for separate visitor types, which means there's no need for introducing separate entities.

**Migration of the existing visitor and its NBT data is tested and is working, so no faults to existing colonies should happen**

Review please


src/main/java/com/minecolonies/api/colony/ICitizenDataView.java
```diff
@@ -220,4 +220,12 @@ public interface ICitizenDataView extends ICitizen
...
      */
     void setHomeBuilding(BlockPos position);
+
+    /**
+     * Get the colony this citizen belongs to.
+     *
+     * @return the colony view.
+     */
+    @Override
+    IColonyView getColony();
```
Raycoms, line 230: A bit weird. Why is the existing IColony handling one not good enough?

Thodor12: This one ensures that the `getColony()` always returns the IColonyView so you don't have to unnecessarily cast.

PR #9715, opened 2024-01-16
fix raiders walking to zero

someaddons:
Closes #
Closes #
Closes #

-fix raiders walking to zero
- Make chest a button like all others

Review please


src/main/java/com/minecolonies/core/entity/mobs/aitasks/RaiderWalkAI.java
```diff
@@ -195,7 +195,7 @@ private void walkToCampFire()
         }
 
         walkTimer = raider.level.getGameTime() + raider.level.random.nextInt(1000);
-        targetBlock = BlockPosUtil.getRandomPosition(raider.level,
+        final BlockPos targetBlock = BlockPosUtil.getRandomPosition(raider.level,
```
Raycoms: wouldn't it be better to rename it as well then if this overrides something else in the file?

PR #9719, opened 2024-01-17
Fix dangerous blocks

uecasm:
Closes [discord report](https://discord.com/channels/472875599422291968/1193382095851094098/1196714118116294666)

- Unify the checks for blocks that are dangerous to stand on (plus barrels for some reason).
- Add a `minecolonies:dangerousblocks` block tag (empty by default) to allow adding additional blocks that citizens will avoid.
- Since Minecraft physics will damage entities if they walk only just slightly above a dangerous block, don't allow citizens to walk on partial blocks above dangerous blocks either.
    image
    - The panel on the far left would be a walkable path if the magma block was removed, since it's in the top half of the block.
    - The panel and carpet directly on the magma blocks are not walkable since the citizen would still take damage.
    - The brewing stand is also no longer walkable because despite being a full block height, it doesn't have a solid base and so MC decides that will trigger magma block damage too.
    - The bottom half slab on the right is a walkable path because it's tall enough to avoid triggering damage from the magma block below.
    - There's a possibility that some other block combinations that were previously walkable now aren't; I haven't done a comprehensive test.


Raycoms (dismissed):

marchermans (dismissed):

src/api/java/com/minecolonies/api/entity/pathfinding/SurfaceType.java
```diff
@@ -128,6 +136,25 @@ public static SurfaceType getSurfaceType(final BlockGetter world, final BlockSta
...
+     * @return           true if dangerous.
+     */
+    public static boolean isDangerous(final BlockState blockState)
+    {
+        final Block block = blockState.getBlock();
+
+        return  blockState.is(ModTags.dangerousBlocks) ||
+                block instanceof FireBlock ||
+                block instanceof CampfireBlock ||
+                block instanceof AbstractBlockBarrel ||
```
someaddons: dont need the barrel as dangerous block anymore, that was an old legacy check/leftover

PR #9727, opened 2024-01-19
exported despawn time for guard teleportation scroll to config

### LINKED ISSUE #9726: [BUG] Spatial Guard Reinforcement Scroll

- [ ] Single Player
- [X] Multi Player

### Minecraft Version

1.19.2

### MineColonies Version

1.19.2-1.1.360-RELEASE

### Structurize Version

1.19.2-1.0.676-RELEASE

_No response_

### Current Behavior

When you summon a guard with the scroll they get the glow (30s) countdown, and are summoned to guard you outside the colony, but for some reason, the countdown resets back to 30, and the guards don’t go away like the scroll says, it’s supposed to be limited time only, but not unlimited, they follow me through the colony, everywhere I go, even when one dies and I hire a new one, I’ve also tried restarting client/server and the problem still persists. Thanks!!

### Expected Behavior

Guards are summoned, are there for 30 seconds, then you receive a message “Guard names have returned to duty at the Colony!” And are teleported back.

### Reproduction Steps

Have scroll, right click a barracks tower or guard tower, leave the colony and summon guard, the guards won’t leave, they’ll stay with you forever

### Logs

N/A

### Anything else?

lustig7:
Closes #9726

- Exported time value for the guard help scroll which temporarily summons the guards of a registered guard building to the user on follow mode.
- Now the user can setup the time, because 15 minutes are a little bit long.
- Sorry for the whitespace changes :/

Review please

### DIFF
--- a/src/api/java/com/minecolonies/api/configuration/ServerConfiguration.java
+++ b/src/api/java/com/minecolonies/api/configuration/ServerConfiguration.java
@@ -13,272 +13,273 @@ import static com.minecolonies.api.util.constant.Constants.*;
  * Mod server configuration. Loaded serverside, synced on connection.
  */
 public class ServerConfiguration extends AbstractConfiguration {
-  /*
-   * --------------------------------------------------------------------------- *
-   * ------------------- ######## Gameplay settings ######## ------------------- *
-   * ---------------------------------------------------------------------------
-   */
+    /*
+     * --------------------------------------------------------------------------- *
+     * ------------------- ######## Gameplay settings ######## ------------------- *
+     * ---------------------------------------------------------------------------
+     */
 
-  public final ForgeConfigSpec.IntValue initialCitizenAmount;
-  public final ForgeConfigSpec.BooleanValue allowInfiniteSupplyChests;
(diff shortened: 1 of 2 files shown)


Thodor12 (changes requested):
Please undo the formatting changes in the ServerConfiguration, actual changes are unreadable

lustig7:
> Please undo the formatting changes in the ServerConfiguration, actual changes are unreadable

Okay, thanks i removed it.

Raycoms:
I think 15 minutes is fine for time, I don't think having a config for this is super useful. However, I would make the tooltip more clear. The 30s cooldown is the cooldown to call one, not the time the guard stays. And then 15 minutes is the time the guard stays. So I would put that also in the tooltip.

someaddons:
Hi and thanks for the contribution, to fix your formatting issues you can import our formatting stlyes from the "documentation" folder(part of the git).

Now for the functionality I think it doesnt really need a config, as those arent that great for users during play. Instead I think it'd be a good idea to keep the 15minutes and be able to use the scroll a second time to send them back earlier on demand

Raycoms:
Close -> Stale

PR #9731, opened 2024-01-21
Pathing/rally fixes

someaddons:
Closes #
Closes #
Closes #

Add higher pathing cost to caveair
Improve pathfinding block lookup slightly
Fix rally banner causing lag and odd guard fighting behaviour

Review please


src/main/java/com/minecolonies/core/entity/ai/workers/guard/AbstractEntityAIGuard.java
```diff
@@ -724,11 +724,6 @@ public boolean isWithinPersecutionDistance(final BlockPos entityPos, final doubl
      */
     private BlockPos getTaskReferencePoint()
     {
-        final ILocation location = buildingGuards.getRallyLocation();
-        if (location != null)
-        {
-            return buildingGuards.getRallyLocation().getInDimensionLocation();
-        }
```
Raycoms, line 731: Will that not stop them from rallying etc?

someaddons: no that is handled elsewhere, this is just used for combat chaining which we already handle by default so its just not needed/causes odd behaviour

Raycoms (approved):

PR #9736, opened 2024-01-24
Misc fixes

someaddons:
- Fix guards being hit immune when not following or rallied
- Adjust pathing to prefer going down than going up, while also increasing the cost for dropping down multiple blocks
- Add fast client updates for citizens that are being interacted with by a player

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
+++ b/src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
@@ -1,6 +1,5 @@
 package com.minecolonies.core.entity.pathfinding.pathjobs;
 
-import com.ldtteam.domumornamentum.block.AbstractBlockStairs;
 import com.ldtteam.domumornamentum.block.decorative.FloatingCarpetBlock;
 import com.ldtteam.domumornamentum.block.decorative.PanelBlock;
 import com.minecolonies.api.blocks.decorative.AbstractBlockMinecoloniesConstructionTape;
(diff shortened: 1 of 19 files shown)


src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -551,7 +551,7 @@ protected double computeCost(
             }
             else if (pathingOptions.dropCost != 1)
             {
-                cost *= pathingOptions.dropCost * Math.abs(dPos.getY());
+                cost *= pathingOptions.dropCost * Math.abs(dPos.getY() * dPos.getY());
```
Raycoms: could we except stair blocks from this?

uecasm:
> Adjust pathing to prefer going down than going up

I can see the logic for that with ordinary colonies, but I wonder if that might be counterproductive in cavern or other underground colonies.  Perhaps they should favour going toward y=64 when in doubt?  Or even y=townhall.

Raycoms:
> > Adjust pathing to prefer going down than going up
> 
> I can see the logic for that with ordinary colonies, but I wonder if that might be counterproductive in cavern or other underground colonies. Perhaps they should favour going toward y=64 when in doubt? Or even y=townhall.

I understood it in a way that was: Fix pathing that was preferring this

uecasm:
If I'm reading the code correctly, the cost of jumping up one block has changed from 1.1 to 1.5, the cost of dropping down one block has remained at 1.1, and the cost of dropping down two blocks has changed from 2.2 to 4.4.  (Though the last assumes that dPos.y can be -2, which the javadoc suggests should not happen, though perhaps it can.)

So that's _probably_ overall preferring down more than up, though only when not dropping more than one block.  (Also, walking up stairs might still count as jumping.)

src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
someaddons: stair blocks are always max 1 diff, this only affects jumping multiple blocks

someaddons:
closing and reopening since bunch of stuff changed

PR #9784, opened 2024-02-15
Fix double feeding confusion

someaddons:
Closes #
Closes #
Closes #

- Add statistics to herders
- Merge feeding children and normal feeding

Review please


Raycoms (approved):

Thodor12 (approved):

MotionlessTrain:
Does that mean that if you enable breeding, that baby animals are also fed automatically?
(I like to have breeding on, but feeding off myself)

Thodor12:
> Does that mean that if you enable breeding, that baby animals are also fed automatically? (I like to have breeding on, but feeding off myself)

Feeding setting is gone, the herder will now always feed all animals, including children. If a child gets fed, it will trigger the aging process. However unlike the previous feeding children implementation, feeding children is not a continuous action, it will run on the same timeout as the current feeding system does, which also reduces the amount of food that is required by a bit if feeding were on.
If feeding were off it'd slightly increase food amount, but not by much (children could still be fed by the regular feed mechanic)

Breeding is pretty much untouched.

someaddons:
since we currently have non optional interval feeding, I'm dropping the optional child feeding and allowing the already existing feeding to also feed the child

uecasm:
I'd prefer if child feeding were still a setting, because I always want it to be off.  I'd rather just wait the time instead of spending the food.

someaddons:
with this change it does not cost any food, like this literally removes the setting and the food cost overhead

Thodor12:
Children would be fed with the daily feeding cost anyway, on top of what they would get fed with the separate child feeding system. So this would only remove excess wheat and clear out the confusion in the UI regarding what "feeding" does.

MotionlessTrain:
> Children would be fed with the daily feeding cost anyway, on top of what they would get fed with the separate child feeding system. So this would only remove excess wheat and clear out the confusion in the UI regarding what "feeding" does.

Don't they just feed to breed anymore? Is it now using a custom way to breed?
That was what the wheat initially was for, to feed just the adults to invoke the default breeding mechanic. The force feeding of baby animals was added later as extra thing for impatient colony owners who happen to have a backlog of wheat.
Not sure whether it makes sense to always feed the baby animals now (even in the daily routine), unless feeding animals isn't doing anything (just implemented for lore reasons), and the breeding uses a custom behaviour

PR #9786, opened 2024-02-17
Fix armor level checking

Thodor12:
- When checking for armor level, the entire armor set + toughness is taken into consideration as opposed to only the armor level of the chestplate.
- Fixed the MinecoloniesArmorMaterial to updated ArmorMaterials Mojang implementation, so that armor values are applied in the correct order to each armor type.
- Updated plate armor:
  - Durability raised from 20 to 37
  - Armor levels updated from 4,9,7,4 to 3,8,6,3 (helmet, chestplate, leggings, boots)
  - Toughness reduced from 2 to 0
  - Enchantment level raised from 6 to 9
- Updated pirates armors:
  - Pirate armor 1 is now identical to iron with durability of diamond
  - Pirate armor 2 is now identical to diamond with durability of iron
- Fixed other armors in the mod to use the new armor material constructor (santa hat and build goggles)

Review please


src/main/java/com/minecolonies/core/items/MineColoniesArmorMaterial.java
```diff
@@ -1,92 +1,96 @@
...
 
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
 /**
- * Implementation for the IArmorMaterial interface so we can add custom armor types.
+ * Implementation for the ArmorMaterial interface, so we can add custom armor types.
  */
 public class MineColoniesArmorMaterial implements ArmorMaterial
```
Nightenom, line 19: just side question: side you edit this class anyway how about making it true record?

src/main/java/com/minecolonies/core/items/MineColoniesArmorMaterial.java
```diff
@@ -1,92 +1,96 @@
...
-    @Override
     public int getEnchantmentValue()
     {
-        return enchantability;
+        return this.enchantmentValue;
     }
 
-    @Override
-    public SoundEvent getEquipSound()
+    public @NotNull SoundEvent getEquipSound()
```
Nightenom: while this is technically more correct, i think we put annotations on separate lines
edit also where overrides

src/main/java/com/minecolonies/core/items/MineColoniesArmorMaterial.java
someaddons: I prefer to not hide code :D

src/main/java/com/minecolonies/core/items/MineColoniesArmorMaterial.java
Raycoms: Yeah, idea puts it there even though our style is above, idk why

src/main/java/com/minecolonies/core/items/MineColoniesArmorMaterial.java
Thodor12: My intention was to leave it the same implementation as the ArmorMaterials enum so it's easy to spot differences.

Raycoms (approved):

someaddons (approved):

PR #9802, opened 2024-02-27
Improve raider entity spawn position

someaddons:
Closes #
Closes #
Closes #

- Improve raider entity spawn position
- Fix dimensions of a few textures

[x ] I also did a multiplayer test.

Review please


Thodor12 (changes requested):
Not sure it's a smart idea to modify the size of the builder_paper to 256 in height, that's going to break about every window in the mod.

Raycoms:
> Not sure it's a smart idea to modify the size of the builder_paper to 256 in height, that's going to break about every window in the mod.

Please make sure that where it's used we properly give it a size so it can downscale and don't do just 100%

Thodor12:
By changing the dimensions of the background, you not only have to make sure all of the windows get the new correct size, but you have to change *every, single, element* it's position relative to the new size. So that is quite a massive thing to do.

Raycoms:
> By changing the dimensions of the background, you not only have to make sure all of the windows get the new correct size, but you have to change _every, single, element_ it's position relative to the new size. So that is quite a massive thing to do.

No no. We downscale to the size we want.

Thodor12:
But is that not going to stretch the texture?

PR #9832, opened 2024-03-11
bunch of fixes

Raycoms:
Closes #
Closes #
Closes #

- Not so strict on guards not eating/sick handling (as we already got an override on the AI anyway)
- Make sure builders recalc on restart
- Try to improve compat for enchantments, etc
- Fix nametag rename
- Fix crash in research window

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/entity/ai/statemachine/states/AIWorkerState.java
+++ b/src/main/java/com/minecolonies/api/entity/ai/statemachine/states/AIWorkerState.java
@@ -198,7 +198,7 @@ public enum AIWorkerState implements IAIState
     /**
      * Decision state for guards.
      */
-    GUARD_DECIDE(false),
+    GUARD_DECIDE(true),
 
     /**
      * Physically attack the target.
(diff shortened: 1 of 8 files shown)


Thodor12 (dismissed):

src/main/java/com/minecolonies/core/entity/ai/workers/AbstractEntityAIStructureWithWorkOrder.java
```diff
@@ -219,14 +219,24 @@ private void requestMaterialsState()
     @Override
     protected IAIState waitForRequests()
```
someaddons: not sure if the override is still needed at all, since its on load the AI never starts in this state

someaddons (dismissed):

src/main/java/com/minecolonies/api/entity/ai/statemachine/states/AIWorkerState.java
```diff
@@ -223,12 +223,12 @@ public enum AIWorkerState implements IAIState
     /**
      * Sleeping on duty
      */
-    GUARD_SLEEP(false),
+    GUARD_SLEEP(true),
```
someaddons, line 226: actually this not! no eating while sleeping

someaddons (approved):

PR #9833, opened 2024-03-11
Several fixes

Thodor12:
Closes [discord report 1](https://discord.com/channels/139070364159311872/1125094832143077456/1216697323560898633) (unclear function of the hire trainee setting)
Closes [discord report 2](https://discord.com/channels/472875599422291968/486940969737125922/1216598254398078996) (certain bamboo blocks fall to the stonemason instead of the sawmill)

- Modified the hire trainee setting from a bool setting to a dedicated setting, this way the active state can be controlled
  - With this the hire trainee setting now becomes disabled when none of the different guard types are automatic hire
- Modified some tags related to bamboo crafting recipes
- Fixed a few other tags to make them more consistent
- Added the option for the forester to strip bamboo blocks into stripped bamboo blocks

Review please


Raycoms:
What is the problem with the current setting?

Thodor12:
Nothing, this is more an accessibility update, the setting only works when any of the 3 possible guard jobs are on automatic hiring mode.
However this is currently not really communicated and more so falls to the player to just have to discover this.
So I made it that the setting shows a warning message if none of the 3 guard jobs are on automatic, so it's clear this setting won't have any effect.

Example

(For this to work I need to have a complex settings object, can't merely use a regular BoolSetting because I cannot override the active/reason methods for that)

src/main/java/com/minecolonies/core/util/BuildingUtils.java
```diff
@@ -102,4 +103,21 @@ public static boolean canAutoHire(@NotNull final IBuilding building,
...
+
+    /**
+     * Check if the given building should try to automatically hire a new citizen.
+     *
+     * @param building   the building to check.
+     * @param hiringMode the current hiring mode of the job.
+     * @param job        the job to hire, or null for a non-specific check.
+     * @return           true if automatic hiring is allowed.
+     */
+    public static boolean canAutoHire(@NotNull final IBuildingView building,
```
Raycoms: can be removed?

Raycoms (approved):

someaddons (approved):

PR #9837, opened 2024-03-13
Pathfinding restructuring & Fixes

someaddons:
* Added new PathfindingUtils, which hold various common checks used in pathfinding
* Refactored packages of pathfinding classes, they're no longer split
* Removed/Cleaned up a lot of unused/old/legacy code
* Extracted restriction zones to the specific path job types
<p> <br> </p>

* There is now two different constructors for AbstractPathJob for two different types of pathfinding:
* Start + range for exploring types which do now know their destination
* Start + end for go to types which know where they end at
* Cost computation for corner nodes is now based on the parent node, fixing stepup costs for stairs
* Cost computation for ladders now is an entry-only cost, once on a ladder there is no extra to continue using it
<p> <br> </p>

- Added a randomness cost to pathing options, giving each node a small random cost value additionally, which makes paths more natural and breaks off tie issues
- Added the ability to explore additional nodes after finding a possible destination, to allow weak destination conditions to explore a bit further, those are then compared by node result score
- Improved the block lookup cache, allowing it to keep previous values that are in bounds
<p> <br> </p>

- Pathfinding debugging now displays heuristic and cost values
- Pathfinding debug entity tracking now uses UUID isntead of storing players
<p> <br> </p>

- Colony loaded chunks can now be accessed offthread
- Fix heuristics/end node scores in serveral pathjobs
- Fix stuckhandler sometimes resetting paths despite the citizens progressing
- Fix pond search logic
- Fix lumberjack tree search with restrictions failing under some conditions
- Fix nonladder climbables not working
<p> <br> </p>

[X ] Yes I tested this before submitting it.

Review please


src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/PathJobCanSee.java
```diff
@@ -21,37 +25,39 @@ public class PathJobCanSee extends AbstractPathJob
...
     {
-        return BlockPosUtil.distManhattan(start.getX(), start.getY(), start.getZ(), x, y, z);
+        return BlockPosUtil.distManhattan(searchAroundPos.getX(), searchAroundPos.getY(), searchAroundPos.getZ(), x, y, z);
     }
 
     @Override
     protected boolean isAtDestination(final MNode n)
     {
-        if (end.getY() - n.y > 2)
+        if (start.getY() - n.y > 2)
```
Raycoms: this seems weird.

someaddons: ye that is wrong, should be the search around pos (guardpos)

PR #9850, opened 2024-03-18
Move quests to colony/quests

uecasm:
Closes incompatibility with pixelmon

- Moves quests to `colony/quests` dir


MotionlessTrain:
Would this work correctly in existing worlds?

uecasm:
> Would this work correctly in existing worlds?

_points at the not-ticked "tested" checkbox_

But yes, it should, the quest ids are relative so changing the base path doesn't change them.

uecasm:
It _will_ break any existing addon quest datapacks, but I don't think there are currently any of those yet.

Raycoms (approved):

someaddons (approved):

PR #9863, opened 2024-03-23
pathfinding heuristics

someaddons:
- Navigators now keep a long-time average of cost vs distance relation to allow adaptation of heuristics to the terrain (for reachability/nodes explored)
- Pathjobs now re-evalute heuristic values used on reaching the target or when progress is going badly
- Add a node limit hit callback to pathjobs to allow increasing a limit when conditions are met

- Improved pathing stuck handler to not take action while an entity is progressing on a path
- Improved pathing stuck handler to try walking in all 4 directions, instead of only opposite direction
- Add new pathjob to move into a given direction for X range

- Extend debug draw with more types of nodes, path reached nodes are now yellow instead of orange
- Pathfinding costs are now applied in order, boni are multiplied first, mali are additive (reduced base for all by one due to this)
- Navigator now keeps the pathresult of finished paths, to allow comparing new incomming ones
- Fix block lookup cache sometimes shifting previous blockstates to wrong places

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/pathfinding/MNode.java
+++ b/src/main/java/com/minecolonies/core/entity/pathfinding/MNode.java
@@ -222,14 +222,6 @@ public class MNode implements Comparable<MNode>
     }
 
     /**
-     * Sets the node as closed.
-     */
-    public void unSetVisited()
-    {
-        visited = true;
-    }
-
-    /**
      * Getter of the score of the node.
      *
      * @return the score.
--- a/src/main/java/com/minecolonies/core/entity/pathfinding/MNode.java
+++ b/src/main/java/com/minecolonies/core/entity/pathfinding/MNode.java
@@ -222,6 +222,14 @@ public class MNode implements Comparable<MNode>
     }
 
     /**
+     * Sets the node as closed.
+     */
+    public void unSetVisited()
+    {
+        visited = true;
+    }
+
+    /**
      * Getter of the score of the node.
      *
      * @return the score.
@@ -364,6 +372,16 @@ public class MNode implements Comparable<MNode>
     public boolean isCornerNode()
     {
         return isCornerNode;
+    }
+
+    /**
+     * Get the added index
+     *
+     * @return
+     */
+    public int getCounterAdded()
+    {
+        return counterAdded;
     }
 
     /**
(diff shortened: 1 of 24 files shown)


src/main/java/com/minecolonies/core/entity/pathfinding/MNode.java
```diff
@@ -221,6 +221,14 @@ public void setVisited()
         visited = true;
     }
 
+    /**
+     * Sets the node as closed.
+     */
+    public void unSetVisited()
```
Nightenom: hmm, method above has same impl, plus the name doesnt match

someaddons: indeed, bug :)

src/main/java/com/minecolonies/core/network/messages/client/SyncPathMessage.java
```diff
@@ -72,6 +84,24 @@ public void toBytes(final FriendlyByteBuf buf)
         {
             node.serializeToBuf(buf);
         }
+
+        buf.writeInt(debugNodesVisitedLater.size());
```
Nightenom, line 88: in 1.20.4 there is write/readCollection (not sure its in earlier 1.20)

someaddons: only some oddly typed ones that need a buffer writer

src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -276,71 +303,236 @@ protected Path search()
...
+                else
+                {
+                    return a.getCounterAdded() - b.getCounterAdded();
+                }
+            });
+            nodesToVisit.addAll(original);
+
+            while (!nodesToVisit.isEmpty())
+            {
+                if (Thread.currentThread().isInterrupted())
```
Raycoms, line 418: Why this interrupted check here? and why not return finalize?

someaddons: I just copied the check from the main loop :D Afaik this is only triggered when the executor is shutting down, aka game is stopping so returning a path doesnt matter

Raycoms: ah okay, it's just to avoid preventing shutdown with this

someaddons: ye, not sure if its really needed though but w/e

Raycoms (approved):

PR #9879, opened 2024-03-30
Several fixes

Raycoms:
Closes #
Closes #
Closes #

- Make locale consistent, 
- fix workorder display name on resource scroll
- fix eating honey bottles and prevent apiary from eating honey at all
- fix kids display name height

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/util/ItemStackUtils.java
+++ b/src/main/java/com/minecolonies/api/util/ItemStackUtils.java
@@ -1111,11 +1111,11 @@ public final class ItemStackUtils
      *
      * @param foodStack   the itemstack of food.
      * @param citizen     the citizen entity.
-     * @param citizenData the citizen data.
      * @param inventory optional inventory to insert stack into if not citizen.
      */
-    public static void consumeFood(final ItemStack foodStack, final AbstractEntityCitizen citizen, final ICitizenData citizenData, final Inventory inventory)
+    public static void consumeFood(final ItemStack foodStack, final AbstractEntityCitizen citizen, final Inventory inventory)
     {
+        final ICitizenData citizenData = citizen.getCitizenData();
         final FoodProperties itemFood = foodStack.getItem().getFoodProperties(foodStack, citizen);
         ItemStack itemUseReturn = foodStack.finishUsingItem(citizen.level(), citizen);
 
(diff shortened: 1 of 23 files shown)


src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingBeekeeper.java
```diff
@@ -130,6 +130,17 @@ public void serializeToView(@NotNull final FriendlyByteBuf buf, final boolean fu
         }
     }
 
+    @Override
+    public boolean canEat(final ItemStack stack)
+    {
+        if (stack.getItem() == Items.HONEY_BLOCK)
```
MotionlessTrain: Shouldn't this be something like `HONEY_BOTTLE`?
Afaik honey blocks aren't edible

Raycoms: idk why it returned to this. I had made it bottle =D Good point

MotionlessTrain: #blame-autocomplete 😄

src/main/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -1098,5 +1105,55 @@ public static Set<ItemStack> allItemsPlusInventory(@NotNull final Player player)
...
+
+    /**
+     * Consume food helper.
+     *
+     * @param foodStack   the itemstack of food.
+     * @param citizen     the citizen entity.
+     * @param citizenData the citizen data.
+     * @param inventory optional inventory to insert stack into if not citizen.
+     */
+    public static void consumeFood(final ItemStack foodStack, final AbstractEntityCitizen citizen, final ICitizenData citizenData, final Inventory inventory)
```
someaddons: just need either citizen data or citizen entity as input

Raycoms: I know we can technically get it from either, but we need an extra query step which we already done at the places where we call this.

someaddons: ye you make it one param to avoid potential futures issues where someone might use it with a desync of worker/data

someaddons (approved):

PR #9906, opened 2024-04-10
make quests more uncommon, move research

Raycoms:
Closes #
Closes #
Closes #

- Adjust how common quests are by factor 10 at least
- Move stonesmeltery after stonemason level 1, require brick instead of stonebricks (to not be too similar)
-

Review please


Thodor12 (dismissed):

uecasm:
Random thought on the cookies quest: given #9895, wouldn't it make sense to restrict this to level 2 baker?  That way they'll actually know the recipe they're pretending to use in the quest.

someaddons (approved):

PR #9907, opened 2024-04-11
Polymorph-like support for teaching crafting

### LINKED ISSUE #808: Button to Switch Recipe Output when Teaching Colonist
### Prerequisites 

- [ ] I am running the latest **alpha** version of MineColonies for my Minecraft version.

### Context 

- Minecraft Version: 1.19.2
- MineColonies Version: 1.0.1247 (atm8 version 1.0.14)

I want to teach my farmer how to turn grapes into grape seeds, but the result is always purple dye, even when using JEI to try to place the recipe for grape seeds into the farmer's crafting recipe ui.

## Feature description

If a recipe has multiple outputs, add a button or key bind to allow switching of the resulting recipe. Similar to the button that appears when the player is using a crafting table and there are multiple results for the same recipe.

### Notes or related things

---
## Viewers

uecasm:
Closes ldtteam/minecolonies-features#808

- Adds a button to the crafting teaching window to switch output items if there are conflicting recipes.
    - Pictured: added a couple of extra recipes for crafting planks via datapack (as an example; recipes not included):
        craftingswitch
    - This is similar to functionality in the Polymorph mod, but does not require that.
    - Only implemented for the crafting grid teaching; it doesn't work for furnace recipe conflicts, those are more complex.

Review please (could port)


Raycoms (dismissed):

src/main/resources/assets/minecolonies/textures/gui/craftingswitch.png
Sandriell, line 1: Any reason why this is 256x256 when it only needs to be 32x64 ?

Or ideally, 16x32. I threw together a quick button set at that size, if interested.

craftingswitch2

Similar to Polymorph's style (without being a copy) for consistency and what end users would be used to.

Thodor12: When you want to render images on a GUI, you have to use something called a blit, essentially capturing part of an image and drawing it on the screen. By default the blit assumes the texture size is 256x256.

uecasm: Yeah, it's possible to use textures with different sizes, but 256x256 is sort of the standard tile that it defaults to.

Thanks for the images, though; it was somewhat deliberate to use something not identical to Polymorph, but you're right that something closer would probably be more familiar to people.  I didn't use your image as-is but I've incorporated it into mine.

Nightenom (changes requested):
Don't create new atlases by manual stitching, 1.20.4+ would require unstitcher definition...

uecasm:
I don't know what that even means.  It's just using a standard GUI control class.

If you'd like me to make the 1.20.4 port myself (and thus possibly encounter whatever you're talking about), I could.

src/main/resources/assets/minecolonies/textures/gui/craftingswitch.png
Nightenom, line 1: Sorry, was on mobile, but I meant this, go with two images, cuz anything that goes into vanilla ui or our ImageButton in 1.20.4+ must have either unstitcher definition in atlas JSON (see vanilla wiki) or be split into two images

uecasm: In 1.20.1 it won't work unless both images are in a single texture.  That's just how the vanilla `ImageButton` works.  It could be split in 1.20.4 if that's different.

someaddons (approved):

PR #9911, opened 2024-04-13
Improvements for free blocks management

uecasm:
- Adds a button in the TH permissions page to obtain a Permission Scepter (which already existed, but was creative-only and nobody knew about it):
    image
- Changes the free interaction list to display the friendly name of the block (and fix up some size/wrapping issues):
    - You still have to type the block id if you want to add a block via the GUI.
    image
- Entering a non-existent block id is now ignored, instead of adding Air.
- Makes UI buttons disable properly when you don't have Edit Permissions permission.
- The Permission Scepter tooltip shows whether it's in tool or location mode.
    image
- Holding CTRL when clicking a block with the Permission Scepter now removes that block/location from the list.
    - Why not SHIFT?  Because you'll often also have to hold shift in order to avoid triggering interactive blocks -- and interactive blocks are the whole point.
- While holding a Permission Scepter, green boxes will be shown around free interaction blocks/locations.
    - It shows either blocks or locations depending on the current tool mode.
    - Locations will be shown colony-wide.
    - Blocks will only be shown if you're relatively close (for performance reasons).
    image

Review please (could port)


Raycoms:
How do you switch mode on the tool?

Raycoms:
As you're adding a new button, I think we should transform the buttons to our own buttons for consistency sakes

uecasm:
> How do you switch mode on the tool?

Right-click the air, same as before.

> As you're adding a new button, I think we should transform the buttons to our own buttons for consistency sakes

All the TH buttons currently use this style; it's the BlockUI default.

Raycoms:
> > How do you switch mode on the tool?
> 
> Right-click the air, same as before.
> 
> > As you're adding a new button, I think we should transform the buttons to our own buttons for consistency sakes
> 
> All the TH buttons currently use this style; it's the BlockUI default.

I started replacing all these vanilla type blockui buttons with the buttons of our style to make all our UIs consistent. So when we touch the UIs and add new buttons we should already follow the minecol style for that imo

uecasm:
Ok, updated the whole permissions page to use new-style buttons.

Raycoms (approved):

someaddons (approved):

PR #9914, opened 2024-04-13
Avoid error messages/mod compat on duplicate entities

someaddons:
Closes #
Closes #
Closes #

- Avoid error messages/mod compat on duplicate entities

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/event/EventHandler.java
+++ b/src/main/java/com/minecolonies/core/event/EventHandler.java
@@ -10,6 +10,7 @@ import com.minecolonies.api.colony.permissions.Action;
 import com.minecolonies.api.entity.ModEntities;
 import com.minecolonies.api.entity.ai.statemachine.tickratestatemachine.TickRateStateMachine;
 import com.minecolonies.api.entity.citizen.AbstractEntityCitizen;
+import com.minecolonies.api.entity.other.AbstractFastMinecoloniesEntity;
 import com.minecolonies.api.items.ModTags;
 import com.minecolonies.api.util.*;
 import com.minecolonies.api.util.constant.Constants;
@@ -114,7 +115,7 @@ public class EventHandler
      *
      * @param event the event.
      */
-    @SubscribeEvent
+    @SubscribeEvent(priority = HIGHEST)
     public static void onEntityAdded(@NotNull final EntityJoinLevelEvent event)
     {
         if (!event.getLevel().isClientSide())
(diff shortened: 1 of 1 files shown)


Raycoms (approved):

src/main/java/com/minecolonies/core/event/EventHandler.java
```diff
@@ -123,9 +124,15 @@ public static void onEntityAdded(@NotNull final EntityJoinLevelEvent event)
...
             {
-                ((Mob) event.getEntity()).targetSelector.addGoal(6, new NearestAttackableTargetGoal<>((Mob) event.getEntity(), EntityCitizen.class, true, citizen -> !citizen.isInvisible()));
+                ((Mob) event.getEntity()).targetSelector.addGoal(6,
+                  new NearestAttackableTargetGoal<>((Mob) event.getEntity(), EntityCitizen.class, true, citizen -> !citizen.isInvisible()));
                 ((Mob) event.getEntity()).targetSelector.addGoal(7, new NearestAttackableTargetGoal<>((Mob) event.getEntity(), EntityMercenary.class, true));
             }
+
+            if (event.getEntity() instanceof AbstractFastMinecoloniesEntity && ((ServerLevel) event.getLevel()).getEntity(event.getEntity().getUUID()) != null)
+            {
+                event.setCanceled(true);
```
uecasm, line 134: Cancelling the event entirely seems potentially troublesome.  It's still actually adding the entity to the world, and some things might want to be aware of that.  e.g. perhaps mapping mods or entity radars.

uecasm:
Looking at the code of DEUF, the thing it seems to get upset about is when `level.getEntity(uuid)` doesn't match the entity that's being added.  Presumably, the real fix for this issue would be to delete the old entity strictly before spawning the new one?  Then there shouldn't be a conflict/duplication, and no need to cancel this event.

src/main/java/com/minecolonies/core/event/EventHandler.java
someaddons: image

someaddons:
those are entities being added during chunk load, they cannot be the same object references as existing ones the colony respawned on tick

uecasm:
Oh, so this is blocking the old persisted entities from respawning rather than affecting the intended respawn on button click/etc?  That makes more sense.

someaddons:
yep we just manually drop the duplicate entity on load like vanilla does (duplicate uuid message) but by doing it manually we circumvent that "uuid fix" mod and the log message of vanilla

PR #9924, opened 2024-04-18
Support for overworld mushroom trees in forester

### LINKED ISSUE #202: Mushroom Farmer
### Prerequisites 

- [X ] I am running the latest **alpha** version of MineColonies for my Minecraft version.
- [X ] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure that my feature does not already exist.
- [X ] I made sure that **this feature request is not a duplicate of any existing feature request.**

### Context 

- Minecraft Version: 1.16.3
- MineColonies Version: 0.13.452-ALPHA

No, but would allow more uses out of items.

## Feature description

A mushroom farmer to make using rabbit stew as a primary food much easier.

### Notes or related things

The idea came about because I had 3,000 carrots in my warehouse :)
---
## Viewers

uecasm:
Closes ldtteam/minecolonies-features#202
Closes [discord request](https://discord.com/channels/139070364159311872/1228515515945975901)

- Adds the ability to cut and replant overworld mushroom trees to the Forester (since they could already handle nether fungi trees).
    - You can plant the overworld mushrooms on whatever block you like (that they'll grow on), but podzol is the most recommended.  (Mycelium decays back to dirt, so is not sustainable; unlike nylium there's no magic restoration.)
    - As with any of the other tree types, you will need to plant and grow the first one yourself (unless there's a nearby mushroom biome to harvest); after that it should be self-sustaining unless they run out of saplings.
- Adds the overworld and nether mushroom "saplings" to the settings list, so that cutting can be disabled if desired.
    - Known (pre-existing) issue: if any forester in the colony has never cut down a particular overworld tree before the sapling was disabled, then they will still cut at least one down if available; the setting is only respected for the second and subsequent tree of that type.  This is due to threading limitations between the pathfinding system and the loot system.  The nether trees are not affected by this limitation.
- When the "break leaves" setting is enabled (or for overworld/nether mushroom trees, where it's forced regardless of the setting), the Forester will now break the leaves before breaking the logs, so the overall tree cutting occurs top-down.
- Includes leaf blocks only one block above the sapling in the breakable leaves list, since red mushrooms can generate that low.  (Jungle bushes can also generate that low, but only in worldgen and not as proper trees.)
- The internal sapling lookup cache is now based purely on block type rather than block state -- the previous implementation could cause spurious cache misses for entirely unrelated properties (like "waterlogged", and various facing directions of the mushroom blocks).  The state-based implementation predated the Great Flattening; there aren't any vanilla trees where different states of the same block drop different saplings, so this should not be needed any more.
    - For backwards compatibility, persistence of this cache is still state-based, but it will not preserve all historic states on upgrade.
- Attempts to use tags for more things to theoretically support additional mushroom tree types from mods, but some MineColonies-specific compatibility datapacks will still be required.

Review please (could port)

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
@@ -1,13 +1,13 @@
 package com.minecolonies.core.entity.ai.workers.production;
 
+import com.minecolonies.api.colony.IColonyManager;
 import com.minecolonies.api.compatibility.Compatibility;
 import com.minecolonies.api.entity.ai.statemachine.AITarget;
 import com.minecolonies.api.entity.ai.statemachine.states.IAIState;
 import com.minecolonies.api.entity.citizen.VisibleCitizenStatus;
-import com.minecolonies.api.util.constant.ColonyConstants;
-import com.minecolonies.core.entity.pathfinding.pathresults.PathResult;
-import com.minecolonies.core.entity.pathfinding.pathresults.TreePathResult;
+import com.minecolonies.api.items.ModTags;
 import com.minecolonies.api.util.*;
+import com.minecolonies.api.util.constant.ColonyConstants;
 import com.minecolonies.api.util.constant.Constants;
 import com.minecolonies.api.util.constant.ToolType;
 import com.minecolonies.core.MineColonies;
@@ -17,9 +17,11 @@ import com.minecolonies.core.colony.buildings.workerbuildings.BuildingLumberjack
 import com.minecolonies.core.colony.jobs.JobLumberjack;
 import com.minecolonies.core.entity.ai.workers.crafting.AbstractEntityAICrafting;
 import com.minecolonies.core.entity.ai.workers.util.Tree;
+import com.minecolonies.core.entity.pathfinding.PathfindingUtils;
 import com.minecolonies.core.entity.pathfinding.navigation.MinecoloniesAdvancedPathNavigate;
 import com.minecolonies.core.entity.pathfinding.pathjobs.PathJobMoveToWithPassable;
-import com.minecolonies.core.entity.pathfinding.PathfindingUtils;
+import com.minecolonies.core.entity.pathfinding.pathresults.PathResult;
+import com.minecolonies.core.entity.pathfinding.pathresults.TreePathResult;
 import com.minecolonies.core.util.WorkerUtil;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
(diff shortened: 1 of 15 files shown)


src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
```diff
@@ -566,15 +579,6 @@ private IAIState chopTree()
             job.getTree().pollNextLog();
             worker.decreaseSaturationForContinuousAction();
         }
-        else if (job.getTree().hasLeaves() && shouldBreakLeaves)
```
someaddons, line 569: moving this to above will prevent leaf decay/mods like fast leaf decay so he'll spend much more time harvesting leaves

uecasm: That's true, but only when the setting is enabled, and it's usually disabled unless you're not entirely trusting the decay mechanism to drop the saplings quickly enough.  i.e. people who turn it on probably don't have fast decay anyway, and people that do will either leave it off or expect that it does harvest all the leaves first.

I did consider making it switch the orders only for the mushroom trees and not regular trees, but that seemed less consistent.  And requires a more awkward code structure.

someaddons: is there a reason why mushroom trres need to harvest them first?

uecasm: Mostly just that it looks really really silly to chop the trunk first and leave a huge mushroom head entirely unsupported while slowly chipping away at it.

Nether trees also benefit because almost immediately after they break the bottom log and reveal the netherrack they convert it back to nylium and replant the fungus, so you don't have the netherrack visible for a long time.

someaddons: ye makes sense, wonder if normal trees need the leaf break at all then, maybe we just do it for mushroom ones
Players might not recognize the need to turn break leaves on for mushroom trees, since those arent really leaves on them

uecasm: As noted, there's a setting for normal trees.  The setting defaults to off, but people can turn it on if they like (e.g. they don't have fast decay and they don't think they're getting enough saplings, especially for things like dark oak).

For overworld/nether mushroom trees, it always breaks "leaves" regardless of the setting, because they don't otherwise decay.

This is not new behaviour, this is how it already worked prior to this PR.

MotionlessTrain:
> When the "break leaves" setting is enabled (or for overworld/nether mushroom trees, where it's forced regardless of the setting), the Forester will now break the leaves before breaking the logs, so the overall tree cutting occurs top-down.

Would this work correctly when the forester gets interrupted after the leaves have been removed? (E.g. by rain, night or a raid) As I thought they only recognise trees when there is a leaf block at the top (so the tree is not really tree like anymore when they are allowed to work again)

uecasm:
I've tested it across night and they will correctly resume the tree the next morning.  I haven't tested the other cases but it shouldn't be any worse than before -- they would otherwise have had the same problem when there's only leaves and no logs.

someaddons (approved):

PR #9936, opened 2024-04-27
Miner ylevels

someaddons:
Closes #
Closes #
Closes #

- Change miner to select y levels by ore levels
- Fix worker hitting blocks not always playing the swing animation during delay as intended

Review please


src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingMiner.java
```diff
@@ -159,26 +166,25 @@ public CompoundTag serializeNBT()
...
-        }
-        else if (this.getBuildingLevel() == 4)
-        {
-            return normalizeMaxDepth(MAX_DEPTH_MAX, level);
-        }
-        else
+        int buildingY = this.getLadderLocation().getY() - 5;
+
+        int buildingLevels = getBuildingLevel();
+        int yLevel = 0;
```
Raycoms, line 172: shouldnt this be 48?

someaddons: initial doesnt matter since it always takes the first one that is lower

uecasm: Unless `miningLevel < buildingY` is never true.  Which could be the case if the max depth is increased again at some point and the mine is built low enough.

It also returns this initial value directly at building level 0, which seems like a thing that could more easily happen.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingMiner.java
```diff
@@ -159,26 +166,25 @@ public CompoundTag serializeNBT()
...
+        for (final Integer miningLevel : MINING_LEVELS)
         {
-            return normalizeMaxDepth(-100, level);
+            if (miningLevel < buildingY)
+            {
+                yLevel = miningLevel;
+                buildingLevels--;
+            }
+
+            if (buildingLevels == 0)
```
uecasm, line 181: This will loop badly if `getBuildingLevel() == 0` (since `buildingLevels` becomes negative), though the loop will eventually terminate since it's bound by the array.

uecasm (commented):
Overall this seems like the goal is to make it so that a level 1 miner built at y=-20 would be able to dig all the way down to bedrock.  Didn't we decide already on Discord that this was not intended behaviour?

I don't really see any benefit in making this change, because the level 1 miner will still be restricted to stone tools and so won't actually be able to mine out many ores at that level anyway.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingMiner.java
Raycoms: AI should never query this at level 0, but maybe some early return at <= 0 could be nice.

Raycoms:
> Overall this seems like the goal is to make it so that a level 1 miner built at y=-20 would be able to dig all the way down to bedrock. Didn't we decide already on Discord that this was not intended behaviour?
> 
> I don't really see any benefit in making this change, because the level 1 miner will still be restricted to stone tools and so won't actually be able to mine out many ores at that level anyway.

Yeah, we had discussed it and wanted to change this, as height changes, changed quite a lot of things. So now, they are pretty much bound mostly by the tool level (e.g. low level miners will not be able to use iron tools to get the necessary ores)

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingMiner.java
someaddons: well it loops a whopping 4 entries xD so unless you need a specific return on 0 it probably doesnt matter

Talyda:
> Yeah, we had discussed it and wanted to change this, as height changes, changed quite a lot of things. So now, they are pretty much bound mostly by the tool level (e.g. low level miners will not be able to use iron tools to get the necessary ores)

Isn't that just going to cause a whole lot of people to be confused as to why their miner is not getting any ores?  You're going to have people that have a level 1 mine and just getting stuck all the time not being able to mine ores, and asking for a better tool _way_ more often than you do now if you change that.

Talyda:
BTW How will people get the miner to branch out with these changes? Is it still going down to a set Y level per building level before they branch out?  Or like, from a support perspective, how do we explain when the miner will branch out?

Raycoms:
> > Yeah, we had discussed it and wanted to change this, as height changes, changed quite a lot of things. So now, they are pretty much bound mostly by the tool level (e.g. low level miners will not be able to use iron tools to get the necessary ores)
> 
> Isn't that just going to cause a whole lot of people to be confused as to why their miner is not getting any ores? You're going to have people that have a level 1 mine and just getting stuck all the time not being able to mine ores, and asking for a better tool _way_ more often than you do now if you change that.

They do get stuck at the "needing a better tool" then. So, the players have to upgrade the hut either way. The message already says that.

Raycoms:
> BTW How will people get the miner to branch out with these changes? Is it still going down to a set Y level per building level before they branch out? Or like, from a support perspective, how do we explain when the miner will branch out?

They go down to the next ore level like in the list. So if you want branching out, you need to change the setting, which we are also planning on adjusting btw =D

Talyda:
> They go down to the next ore level like in the list. So if you want branching out, you need to change the setting, which we are also planning on adjusting btw =D
The setting in the "max depth" box? 
I was trying to understand how it will work with that code change, but I couldn't work it out. Now it is easy, we have Y levels they will dig down to, once they have done that they branch out (or you can change max depth box). I couldn't work out what the new Y levels are, or how to convey that for support.  Are 48, 16, -16, -100 the new Y levels they go down to? Or it depends on other things?
(should I not ask this until you have finalised it?)

uecasm:
As I mentioned in the other PR, it's nice if the PR description is more descriptive :)

The net effect of these changes (ignoring weirdnesses if it ever checks at hut level 0) is that:
* for a mine at y=64, level 1 will dig to 48, level 2 will dig to 16, level 3 will dig to -16, level 4 or 5 will dig to bedrock
* for a mine ay y=40, level 1 will dig to 16, level 2 will dig to -16, levels 3-5 will dig to bedrock.
* for a mine at y=10, level 1 will dig to -16, levels 2-5 will dig to bedrock.
* for a mine at y=-20, all levels will dig to bedrock.

I still think it's a bit pointless, since the moment they hit a block that requires a better tool, they will stop dead (and complain), at which point you either have to mine it yourself or upgrade the hut anyway.

Talyda:
Thanks Mirality!  That is definitely not going to be as easy to explain!

Would it be any mine above Y 52 will dig to 48
any mine under 52 and over Y 20 will go to 16
any over under Y 20 and over Y -12 will go to -16
anything under Y -16 will go to bedrock?

(I think they need 4 Y levels to make at least one platform?)  Oh, that would only be the Y level of lowest ladder, wouldn't it?

I will wait til it is finalised LOL!

uecasm:
~IIRC it's purely about the hut block itself,  Hut block at y=48 will dig down to 16, hut block at y=49 will dig down to 48 (i.e. not at all, because it doesn't take the ladder pos or node height into account).~

Nope, I recalled wrong.  The magic Y is the ladder location minus 5, for some reason.  Presumably that's related to the first node height under the ladder.

Raycoms:
> IIRC it's purely about the hut block itself, Hut block at y=48 will dig down to 16, hut block at y=49 will dig down to 48 (i.e. not at all, because it doesn't take the ladder pos or node height into account).

It takes ladder pos.

Raycoms (approved):

Talyda:
> It takes ladder pos.

I was just reminded of this PR in my emails =P Can you please tell us how to explain it easily to people if you are going to change it to this?

Do we just tell people "if the lowest ladder is at Y (whatever) it will dig to (wherever)"?  And actually tell us what those numbers are please?

PR #9938, opened 2024-04-28
Reworked fisher pond search

someaddons:
Closes #
Closes #
Closes #

- Reworked fisher pond search

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/util/Pond.java
+++ b/src/main/java/com/minecolonies/api/util/Pond.java
@@ -1,9 +1,11 @@
 package com.minecolonies.api.util;
 
 import net.minecraft.core.BlockPos;
+import net.minecraft.tags.FluidTags;
 import net.minecraft.world.level.BlockGetter;
 import net.minecraft.world.level.block.Blocks;
 import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.material.FluidState;
 import org.jetbrains.annotations.NotNull;
 
 /**
@@ -11,8 +13,6 @@ import org.jetbrains.annotations.NotNull;
  */
 public final class Pond
 {
-    //TODO: Recheck logic/rewrite it simpler, return central pond position instead of the water pos next to the standing for better direction
-
     /**
      * The minimum pond requirements.
      */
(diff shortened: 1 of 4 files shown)


Raycoms (dismissed):

src/main/java/com/minecolonies/api/util/Pond.java
```diff
@@ -16,165 +16,43 @@ public final class Pond
...
-      final int z,
-      final int vector,
-      final int width,
-      final int length)
-    {
-        //Check 6 blocks in direction +/- x
-        for (int dx = x - (width / 2 * vector); dx <= x + (width / 2 * vector); dx++)
-        {
-            if (world.getBlockState(new BlockPos(dx, y, z)).getBlock() != Blocks.WATER)
+            if (world.getBlockState(tempPos.set(water.getX(), water.getY() - i, water.getZ())).getBlock() != Blocks.WATER)
```
uecasm, line 36: This might be problematic, there are some styles that have decorative plants at the bottom of the pond.  Is it intended that this only checks one position?

Also, instead of using `tempPos`, this could just use `water.below(i)`.

src/main/java/com/minecolonies/api/util/Pond.java
```diff
@@ -16,165 +16,43 @@ public final class Pond
...
+            for (int z = -WATER_POOL_LENGTH_REQUIREMENT; z < WATER_POOL_LENGTH_REQUIREMENT / 2; z++)
             {
-                return false;
-            }
-        }
-        //Takes the middle z block and searches 3 water blocks to both sides
-        return checkWaterPoolInDirectionX(world, x, y, z + length * vector, 1) && checkWaterPoolInDirectionX(world, x, y, z + length
-                                                                                                                                * vector, -1);
-    }
+                tempPos.set(water.getX() + x, water.getY(), water.getZ() + z);
```
uecasm, line 46: It seems kinda unnecessary to use `MutableBlockPos` further down as well.

Also, could use `BlockPos.betweenClosed` or `BlockPos.spiralAround` here.

uecasm (commented):
It's nice if the description is more descriptive of the intended changes.  Both for review purposes and future changelogs.

src/main/java/com/minecolonies/api/util/Pond.java
Raycoms: We could use the util for water here?

PR #9949, opened 2024-05-10
sound manager rework

Raycoms:
Closes #
Closes #
Closes #

- Sound Manager Rework:
-> One central sound manager per client
-> Schedule normal citizen sounds there too
-> Max concurrent sounds
-> Builder also now plays normal block place sounds.

Review please


src/main/java/com/minecolonies/core/colony/ColonyManager.java
```diff
@@ -78,6 +78,11 @@ public final class ColonyManager implements IColonyManager
      */
     private boolean capLoaded = false;
 
+    /**
+     * Client side sound manager.
+     */
+    private SoundManager clientSoundManager;
```
someaddons, line 84: dont think a sound manager really belongs under colonymanager, since its a client side singleton why not just make it fully static? Could still do SoundManager.tick() from the hook here

someaddons (approved):

Thodor12 (approved):

PR #9959, opened 2024-05-19
Fix: quarrier's interaction with the zigzag iterator

### LINKED ISSUE #9950: [BUG] Quarrier will not quarry

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.20.1

### MineColonies Version

1.1.566-BETA

### Structurize Version

1.0.731-RELEASE

1. Forge 47.2.20 2. Domum Ornamentum 1.0186-RELEASE
All the mods 9 0.2.58

### Current Behavior

Quarry building is built and quarrier assigned. Quarrier stands at quarry hut tile and does nothing. Does not ask for items or tools. 
He is equipped with tools according to his mine level. Fired and Hired several times over.

### Expected Behavior

Quarrier is expected to begin quarrying out his quarry... 

### Reproduction Steps

1. Build a quarry (using shire). 2. Assign quarrier from mine.  3. Enjoy

### Logs

https://gist.github.com/Putzaroo/8db946617b0ef921a062665087b6d43d

### Anything else?

I have deleted the ATM configs and allowed them to repopulate with the correct settings

MotionlessTrain:
Closes #9950
Closes #
Closes #

- The old row-by-row iterator pattern decremented in opposite order in the X-Z plane than it incremented (starting at the highest Z, and decrementing it over time)
  The zigzag pattern both increments and decrements in the same direction in the X-Z plane, starting at the lowest X and Z coordinate.
  The quarrier was relying on the order of the row-by-row pattern, and only calculated solid blocks and entities for the row with the highest Z-value (as it decrements the iterator in both those cases).
  This led to an infinite loop, where the resources where not calculating correctly. They then started building the first row (using `NULL_POS`, so the first plane was done correctly), and needed blocks that where not known to the building,
    which forced them to recalculate the resources, etc.
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIQuarrier.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIQuarrier.java
@@ -321,7 +321,7 @@ public class EntityAIQuarrier extends AbstractEntityAIStructureWithWorkOrder<Job
                 if (progress.getY() != -1 && result.getIteratorPos().getY() < progress.getY())
                 {
                     structurePlacer.getB().nextStage();
-                    this.storeProgressPos(new BlockPos(0, progress.getY() + 1, 0), structurePlacer.getB().getStage());
+                    this.storeProgressPos(new BlockPos(-1, progress.getY() + 1, 0), structurePlacer.getB().getStage());
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIQuarrier.java
```diff
@@ -474,12 +474,12 @@ public boolean requestMaterials()
 
                 if (requestProgress.getY() != -1 && result.getIteratorPos().getY() < requestProgress.getY())
                 {
-                    requestProgress = new BlockPos(0, requestProgress.getY() + 1, 0);
+                    requestProgress = new BlockPos(-1, requestProgress.getY() + 1, 0);
```
MotionlessTrain, line 477: I noticed during stepping through the code that this caused the block in the north west corner of the layer to get skipped whilst calculating decoration blocks resources (and it looks like it is also skipped during building the quarry). I changed that as well while I was at it

PR #9962, opened 2024-05-24
fix john exploit

Raycoms:
Closes #
Closes #
Closes #

- Fix John bypassing progressing
-
-

Review please


someaddons (dismissed):

src/main/java/com/minecolonies/core/placementhandlers/main/SurvivalHandler.java
```diff
@@ -88,7 +88,7 @@ public void handle(
       final BlockPos blockPos,
       final PlacementSettings placementSettings)
     {
-        if (blueprint == null)
+        if (blueprint == null || blueprintPath.startsWith("scans/"))
```
uecasm: Won't this block building decorations from scans too?  Shouldn't that still be permitted?

(i.e. wouldn't it be better to put this check at line ~141?  That's what the `clientPack` check is intended for.)

PR #9966, opened 2024-05-29
Only hit at line of sight

Raycoms:
Closes #
Closes #
Closes #

- Make Archer Trainees only shoot if there is a line of sight
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/guard/training/EntityAIArcherTraining.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/guard/training/EntityAIArcherTraining.java
@@ -6,6 +6,7 @@ import com.minecolonies.api.entity.ai.statemachine.states.IAIState;
 import com.minecolonies.api.entity.citizen.VisibleCitizenStatus;
 import com.minecolonies.api.util.InventoryUtils;
 import com.minecolonies.api.util.SoundUtils;
+import com.minecolonies.api.util.WorldUtil;
 import com.minecolonies.api.util.constant.Constants;
 import com.minecolonies.api.util.constant.ToolType;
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/core/entity/ai/workers/guard/training/EntityAIArcherTraining.java
```diff
@@ -111,7 +114,8 @@ private IAIState selectTarget()
             return DECIDE;
         }
         final BlockPos targetPos = archeryBuilding.getRandomShootingTarget(worker.getRandom());
-        if (targetPos == null)
+        if (targetPos == null || !WorldUtil.isBlockLoaded(world, targetPos) ||
+              !world.clip(new ClipContext(new Vec3(worker.getX(), worker.getEyeY(), worker.getZ()), new Vec3(targetPos.getX() + HALF_BLOCK, targetPos.getY() + HALF_BLOCK, targetPos.getZ() + HALF_BLOCK), ClipContext.Block.COLLIDER, net.minecraft.world.level.ClipContext.Fluid.NONE, worker)).getBlockPos().equals(targetPos))
```
someaddons, line 118: make sure to test it with different target blocks, it may be depending on the target shape that the check fails

Raycoms: We only use full block target blocks here

someaddons: oh ok thought we had added tags

Raycoms: for the shooting position only

someaddons (approved):

PR #9969, opened 2024-05-30
Central Courier Task Assignment

Raycoms:
Closes #
Closes #
Closes #

- Move Assignment of Courier Tasks to Warehouse
- Couriers pick up to 5 tasks from there to handle (only if tasks go to same destination and its not pickups)
- Generify task list rendering code

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/buildings/moduleviews/RequestTaskModuleView.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/moduleviews/RequestTaskModuleView.java
@@ -4,7 +4,7 @@ import com.ldtteam.blockui.views.BOWindow;
 import com.minecolonies.api.colony.buildings.modules.AbstractBuildingModuleView;
 import com.minecolonies.api.colony.requestsystem.token.IToken;
 import com.minecolonies.api.util.constant.Constants;
-import com.minecolonies.core.client.gui.modules.WindowHutCrafterTaskModule;
+import com.minecolonies.core.client.gui.modules.WindowHutRequestTaskModule;
 import net.minecraft.network.FriendlyByteBuf;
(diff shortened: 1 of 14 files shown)


src/main/java/com/minecolonies/core/colony/buildings/moduleviews/RequestTaskModuleView.java
```diff
@@ -38,4 +41,10 @@ public String getDesc()
     {
         return "com.minecolonies.coremod.gui.workerhuts.crafter.tasks";
     }
+
+    /**
+     * Get the specific task list.
+     * @return the task list.
+     */
+    public abstract List<IToken<?>> getTasks();
```
someaddons, line 49: if this is all you need, I think you should be using an interface and not creating a class hierarchy

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingWareHouse.java
```diff
@@ -8,6 +8,7 @@
 import com.minecolonies.api.colony.IColonyView;
 import com.minecolonies.api.colony.buildings.workerbuildings.IWareHouse;
 import com.minecolonies.api.colony.requestsystem.resolver.IRequestResolver;
+import com.minecolonies.api.colony.requestsystem.token.IToken;
```
Thodor12: Magic import?

src/main/java/com/minecolonies/core/colony/buildings/moduleviews/RequestTaskModuleView.java
Raycoms: The other stuff in the abstract is also needed

someaddons (approved):

PR #9972, opened 2024-06-02
Avoid nearby colony lookup on each frame

someaddons:
Closes #
Closes #
Closes #

- Avoid nearby colony lookup on each frame

Review please


Raycoms (approved):

uecasm:
Wouldn't it make more sense to cache this based on the player crossing chunk boundaries?  There's already some other stuff in here that does that, and the nearby colony is unlikely to change without the player moving around.

someaddons:
The other stuff you're thinking on is serverside, clientside just has one player and data is more volatile due to sync timing. So performance wise there is little benefit to more caching and it'd have a higher risk of problems

Nightenom:
Just wonderin, can't we do once per second vs random out of 20?

uecasm:
No, it's client-side too.  Look inside the `ColonyBorderRenderer` and the `ColonyBlueprintRenderer` for example.

PR #9981, opened 2024-06-09
Improve builder building positions

someaddons:
Closes #
Closes #
Closes #

- Prevent builder from standing where hes mining below
- Prevent builder from dropping down when searching for a building position
- Add strong preference for standing outside of water while building

Review please


src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/PathJobMoveCloseToXNearY.java
```diff
@@ -56,6 +57,11 @@ protected double computeHeuristic(final int x, final int y, final int z)
     @Override
     protected boolean isAtDestination(@NotNull final MNode n)
     {
+        if (desiredPosition.getX() == n.x && desiredPosition.getZ() == n.z)
+        {
+            return false;
```
Raycoms, line 62: why?

Raycoms: ah nvm, its only for the close to task

PR #9986, opened 2024-06-12
Improve fisher

Raycoms:
Closes #
Closes #
Closes #

- Fix fisher running into water to fish from there
- Fix fisher not being able to stand on certain slabs etc to fish.
- Fix fisher derping when going through water when going to a fishing pos
- Fix and cleanup undertaker

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
+++ b/src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
@@ -1059,23 +1059,6 @@ public class MinecoloniesAdvancedPathNavigate extends AbstractAdvancedPathNaviga
         super.stop();
     }
 
-    @Nullable
-    @Override
-    public WaterPathResult searchWater(final int range, final double speed, final List<Tuple<BlockPos, BlockPos>> ponds)
-    {
-        @NotNull final BlockPos start = PathfindingUtils.prepareStart(ourEntity);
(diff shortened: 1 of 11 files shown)


uecasm:
Is it worthwhile to add an optional tag for suggested "stand here" locations when fishing nearby?  (Pick one at random if within X blocks of chosen water spot.)  Some styles have a very well-defined pier or similar that's intended to be used for fishing, but often gets ignored.

Raycoms:
> Is it worthwhile to add an optional tag for suggested "stand here" locations when fishing nearby? (Pick one at random if within X blocks of chosen water spot.) Some styles have a very well-defined pier or similar that's intended to be used for fishing, but often gets ignored.

From my superficial tests this should be much better now as well. They finally use these.

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
```diff
@@ -1061,16 +1061,19 @@ public void stop()
 
     @Nullable
     @Override
-    public WaterPathResult moveToWater(final int range, final double speed, final List<Tuple<BlockPos, BlockPos>> ponds)
+    public WaterPathResult searchWater(final int range, final double speed, final List<Tuple<BlockPos, BlockPos>> ponds)
```
someaddons: I think we shouldnt have this in the navigator, the navigators job is moving the entity/following the path. If you want a result without that just do it in the AI directly

someaddons (approved):

PR #9989, opened 2024-06-16
Some perm changes

Raycoms:
Closes #
Closes #
Closes #

- Add levitation effect after too many times breaking colony perms
- Add entity right click whitelist
- Add new toggleable permission that defaults to true

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/permissions/ColonyPermissionEventHandler.java
+++ b/src/main/java/com/minecolonies/core/colony/permissions/ColonyPermissionEventHandler.java
@@ -189,6 +189,7 @@ public class ColonyPermissionEventHandler
             {
                 MessageUtils.format(PERMISSION_DENIED).sendTo((Player) entity);
                 lastPlayerNotificationTick.put(entity.getUUID(), worldTime);
+                playerAttempts.put(entity.getUUID(), 0);
             }
             else
             {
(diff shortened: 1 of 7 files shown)


src/main/java/com/minecolonies/core/colony/permissions/ColonyPermissionEventHandler.java
```diff
@@ -173,12 +184,23 @@ private void cancelEvent(final Event event, @Nullable final Entity entity, final
...
-                  || lastPlayerNotificationTick.get(entity.getUUID()) + (Constants.TICKS_SECOND * 10)
+                  || lastPlayerNotificationTick.get(entity.getUUID()) + (TICKS_SECOND * 10)
                        < worldTime)
             {
                 MessageUtils.format(PERMISSION_DENIED).sendTo((Player) entity);
                 lastPlayerNotificationTick.put(entity.getUUID(), worldTime);
             }
+            else
+            {
+                if (playerAttempts.compute(entity.getUUID(), (uuid, count) -> count == null ? 1 : count + 1) > 10)
```
MotionlessTrain, line 196: Does this get reset after the notification tick?
Right now, it looks like if you try to do something that is not allowed a couple of times, then wait a minute, then a couple of times again, then wait a minute and continue,
eventually you seem to trigger the levitation, without ever breaking the permission more than ten times within ten seconds. Is it intended to work like that?

Raycoms: good point, changed

someaddons (approved):

PR #9997, opened 2024-06-20
Rack fixes

someaddons:
- Fixes rack Z fighting caused by double rendering a rack

Review please


Raycoms (approved):

MotionlessTrain:
Just in case, because it seems a bit unclear (at least from the announcement message). Should this fix racks which were already placed in the world before, or should this only fix newly placed racks?

someaddons:
newly placed

PR #10004, opened 2024-06-26
courier adjutment + sleep adjustments

Raycoms:
Closes #
Closes #
Closes #

- Citizens don't reset bedpos on restart anymore
- Fix manual player pickup
- A bit of pickup cleanup

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/buildings/AbstractBuilding.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/AbstractBuilding.java
@@ -1553,11 +1553,6 @@ public abstract class AbstractBuilding extends AbstractBuildingContainer
             return false;
         }
 
-        if (pickUpPrio == getPlayerActionPriority(true))
-        {
-            MessageUtils.format(COM_MINECOLONIES_COREMOD_ENTITY_DELIVERYMAN_FORCEPICKUP)
-                                .sendTo(colony.getMessagePlayerEntities());
-        }
         createRequest(new Pickup(pickUpPrio), true);
         return true;
     }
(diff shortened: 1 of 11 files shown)


Thodor12 (dismissed):

Talyda:
Will that fix the stash too Ray?

Thodor12:
Yes, at least according to Ray

Talyda:
Thanks!

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuilding.java
```diff
@@ -1552,7 +1553,12 @@ public boolean createPickupRequest(final int daysToPickup)
             return false;
         }
 
-        createRequest(new Pickup(MAX_BUILDING_PRIORITY), true);
+        if (pickUpPrio == getPlayerActionPriority(true))
+        {
+            MessageUtils.format(COM_MINECOLONIES_COREMOD_ENTITY_DELIVERYMAN_FORCEPICKUP)
```
someaddons: I'd like if we made this a notification only for the player who clicked it, the "A courier will come along shortly" does only confuse you if it happens for other players, which did not do anything. We probably should just do it in the UI instead?

Thodor12: Or we could just pass the initiating player to here, no?

someaddons: not to here, but we could pass it to where the request is being created(though a bit more work, as atm we dont pass the player along on menu creation for stash-inventory based triggers, so maybe just on the place where the request is created search the closest player)

uecasm: Couldn't it just be done in the `ForcePickupMessage` instead?  That's already where it prints a failure message if it doesn't create a request, it can be where it shows a success message when it does too.  At that point, the specific requesting player is known.

I'm assuming that the point of this change is to stop the message being shown by the stash pickup requests, in which case it definitely makes more sense to me to move it closer to the UI that it relates to.

someaddons (approved):

PR #10016, opened 2024-07-05
fixed potential memory leak

cyberNinja6143:
# Fixed a potential memory leak while investigating an issue a user was complaining about.

- Removed a variable that was not being used

[ *] Yes I tested this before submitting it.

Review please


someaddons:
in java that is not a memory leak :D but we can remove the line if it isnt used

cyberNinja6143:
> in java that is not a memory leak :D but we can remove the line if it isnt used

Yeah but it still looks bad in the source code

someaddons:
I'm less concerned about visuals in code

cyberNinja6143:
> I'm less concerned about visuals in code

It is a bit of a silly change but I'm kind of a new developer

Thodor12:
When you make contributions you have to follow our style guide, all code has to be formatted with our formatting rules.
The space you removed is exactly against our formatting rules. So that has to be removed.

cyberNinja6143:
> When you make contributions you have to follow our style guide, all code has to be formatted with our formatting rules. The space you removed is exactly against our formatting rules. So that has to be removed.

My bad, ill fix it

Thodor12 (dismissed):

src/main/java/com/minecolonies/core/entity/ai/minimal/EntityAIInteractToggleAble.java
```diff
@@ -387,7 +387,7 @@ public void tick()
         {
             return;
         }
-        updateTimer = ColonyConstants.rand.nextInt(40 + offSet);
+        updateTimer = ColonyConstants.rand.nextInt(30 - offSet);
```
Raycoms, line 390: if you want to update this please make this a separate PR

PR #10020, opened 2024-07-09
Fix Citizens taking too long to open doors

cyberNinja6143:
# Fixed Citizens taking too long to open doors.

- The offset value was being added instead of subtracted which substantially increased the amount to takes for citizens to open doors

[ *] Yes I tested this before submitting it.

Review please


Raycoms:
That is intentional. Door spam can cause lag. It' a trade-off

PR #10042, opened 2024-07-15
Fix tools with damage value not being used in crafting recipes

### LINKED ISSUE #9764: [BUG] Crafter recipe added by datapck does not ignore damage tag of tool item

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.19.2

### MineColonies Version

1.1.453 - RELEASE

### Structurize Version

1.0.683 - RELEASE

1. Forge Version: 43.3.2
2. BlockUI Version: 1.0.140 - BETA
3. Domum Ornamentum: 1.0.146 - BETA
4. Farmer's Delight: 1.2.4

### Current Behavior

I made a custom craftman recipe for assistant cook via datapack,
This photo is JEI screen for that recipe

And this pastebin link is datapack source
https://pastebin.com/0Y0QTi3K

According to Mirality says to me in discord,
Recipe should ignore nbt tag of "iron knife"

BUT it doesn't, check the nbt tag of "iron knife", if durability is not max they can't use it

### Expected Behavior

Custom recipe should ignore the nbt tag of "iron knife"
Or find another way to add "iron knife" using recipe

### Reproduction Steps

1. make a datapack
2. make a json file and paste datapack code of below pastebin link
https://pastebin.com/0Y0QTi3K
3. put the json file in following directory 
/data/minecolonies/crafterrecipes/assistantcook/

### Logs

https://gist.github.com/supernova0102/24cc6583bae0539368b79c5ca00cedc2

### Anything else?

I'm attatching minecolonies discord message link about discussion of this issue
https://discord.com/channels/139070364159311872/976625343358189618/1203233898570457108

Thodor12:
Closes #9764 

- When tools are detected in crafting recipes, modify the input item to not match on damage

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -252,14 +252,14 @@ public class RecipeStorage implements IRecipeStorage
      */
     private void calculateTools()
     {
-        for (int index = 0; index < this.cleanedInput.size(); index++)
+        for (int index = 0; index < cleanedInput.size(); index++)
         {
-            final ItemStorage item = this.cleanedInput.get(index);
+            final ItemStorage item = cleanedInput.get(index);
             for (ItemStack result : getSecondaryOutputs())
             {
                 if (ItemStackUtils.compareItemStacksIgnoreStackSize(item.getItemStack(), result, false, true) && result.isDamageableItem())
                 {
-                    this.cleanedInput.set(index, new ImmutableItemStorage(new ItemStorage(item.getItemStack(), item.getAmount(), true)));
+                    cleanedInput.set(index, new ImmutableItemStorage(new ItemStorage(item.getItemStack(), item.getAmount(), true)));
                     tools.add(result);
                     secondaryOutputs.remove(result);
                     break;
--- a/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -252,15 +252,14 @@ public class RecipeStorage implements IRecipeStorage
      */
     private void calculateTools()
     {
-        final List<ItemStorage> cleaned = getCleanedInput();
-        for (int index = 0; index < cleaned.size(); index++)
+        for (int index = 0; index < this.cleanedInput.size(); index++)
         {
-            final ItemStorage item = cleaned.get(index);
+            final ItemStorage item = this.cleanedInput.get(index);
             for (ItemStack result : getSecondaryOutputs())
             {
                 if (ItemStackUtils.compareItemStacksIgnoreStackSize(item.getItemStack(), result, false, true) && result.isDamageableItem())
                 {
-                    cleaned.set(index, new ImmutableItemStorage(new ItemStorage(item.getItemStack(), item.getAmount(), true)));
+                    this.cleanedInput.set(index, new ImmutableItemStorage(new ItemStorage(item.getItemStack(), item.getAmount(), true)));
                     tools.add(result);
                     secondaryOutputs.remove(result);
                     break;
--- a/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -252,12 +252,15 @@ public class RecipeStorage implements IRecipeStorage
      */
     private void calculateTools()
     {
-        for(ItemStorage item : getCleanedInput())
+        final List<ItemStorage> cleaned = getCleanedInput();
+        for (int index = 0; index < cleaned.size(); index++)
         {
-            for(ItemStack result: getSecondaryOutputs())
+            final ItemStorage item = cleaned.get(index);
+            for (ItemStack result : getSecondaryOutputs())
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -252,12 +252,15 @@ private List<ItemStack> calculateSecondaryOutputs()
...
+        for (int index = 0; index < cleaned.size(); index++)
         {
-            for(ItemStack result: getSecondaryOutputs())
+            final ItemStorage item = cleaned.get(index);
+            for (ItemStack result : getSecondaryOutputs())
             {
-                if(ItemStackUtils.compareItemStacksIgnoreStackSize(item.getItemStack(), result, false, true) && result.isDamageableItem())
+                if (ItemStackUtils.compareItemStacksIgnoreStackSize(item.getItemStack(), result, false, true) && result.isDamageableItem())
                 {
+                    cleaned.set(index, new ImmutableItemStorage(new ItemStorage(item.getItemStack(), item.getAmount(), true)));
```
MotionlessTrain: Is that the output item (which is slightly more damaged), or the input item (which has not been damaged by the crafting recipe yet)?

uecasm: It's the input item.  At this point it's just telling it that it should ignore existing damage on the item -- actually damaging it happens elsewhere.

uecasm (dismissed):

src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
MotionlessTrain: Ah, I was wondering about the damaging part of that, thank you!

Raycoms: So, on calculate tools, you alter the cleanedInput?

Thodor12: Yes, Mirality found that was the easiest fix to make this work

Raycoms: Why don't we do a toolcheck in calculateCleanedInput()?

Thodor12: You might as well remove the calculate tools function then

Raycoms: We can technically do that indeed. But at the very least it fills up separate collections. But side injecting into a different one is weird.

Thodor12: I tried to check it already and it's quite some dated/complex code, is it fine if we merge this first so the bug is solved, then look at improving the handling in this class?

Raycoms: This side inject is too bug prone. Moving this specific code to the second loop in the calculateCleanedInput is straightforward

Thodor12: I've tested it, it works fine as is, there are no bugs with it.
I made sure to turn it into a numeric for so you cannot have concurrent read/write problems that you get with generic fors.

Raycoms: Yeah, I believe you tested it, but it's still much more bug prone this way because it the main input in an unexpected way for anyone else. Moving this to the calc simplifies this code by a massive amount

uecasm: The "side inject" to the cleaned input is not really any weirder than it altering the tools and secondary outputs -- it's just the thing it needs to do to do the job.  Both methods are only called from the constructor, but in different contexts (so simply moving the code to `calculateCleanedInput()` would not work).

What I would recommend (and might make you happier about it) is to s/`getCleanedInput()`/`cleanedInput`/ in this method -- it's a lot less weird to alter a field directly than to alter the result of a getter method, even if in this case they happen to be equivalent.

src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -252,15 +252,14 @@ private List<ItemStack> calculateSecondaryOutputs()
...
+        for (int index = 0; index < this.cleanedInput.size(); index++)
         {
-            final ItemStorage item = cleaned.get(index);
+            final ItemStorage item = this.cleanedInput.get(index);
             for (ItemStack result : getSecondaryOutputs())
             {
                 if (ItemStackUtils.compareItemStacksIgnoreStackSize(item.getItemStack(), result, false, true) && result.isDamageableItem())
                 {
-                    cleaned.set(index, new ImmutableItemStorage(new ItemStorage(item.getItemStack(), item.getAmount(), true)));
+                    this.cleanedInput.set(index, new ImmutableItemStorage(new ItemStorage(item.getItemStack(), item.getAmount(), true)));
```
uecasm: Using `this` here and not on the other fields is a bit inconsistent.  I'd omit that.

uecasm (dismissed):

uecasm (approved):

Thodor12:
Superseded by tool-handling branch

Issue #10026, opened 2024-07-11
[BUG]Quarrier isn't working saying "I finished the current quarry I'm assigned to, I need more!"


minnie1089:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Are you using the latest MineColonies Version?

- [X] I am running the latest **beta** version of MineColonies for my Minecraft version.

### Did you check on the Wiki? or ask on Discord?

- [X] I checked the MineColonies Wiki and made sure my issue is not covered there. Or I was sent from discord to open an issue here.

### What were you playing at the time? Were you able to reproduce it in both settings?

- [ ] Single Player
- [X] Multi Player

### Minecraft Version

1.20.1

### MineColonies Version

1.20.1-1.1.605-BETA

### Structurize Version

1.20.1-1.0.742-RELEASE

### Related Mods and their Versions

Forge 47.3.0
Domum Ornamentum 1.20.1-1.0.188-BETA

### Current Behavior

![2024-07-11_12 20 53](https://github.com/ldtteam/minecolonies/assets/67824161/4e21cd5d-ebd7-4cd4-a93b-198eede563f4)
The quarrier just stands at their bed and complains

### Expected Behavior

The quarrier should do their job

### Reproduction Steps

First the quarrier was working fine, then they just stopped, so i tried lots of things, i restarted the npc, i tried switching my miner and quarrier around I tried just firing the quarrier, none of this worked. I then broke the quarry block and replaced it with the build tool in the same place and location. The quarrier mined a bit more, then said that the quarry was finished.
![image](https://github.com/ldtteam/minecolonies/assets/67824161/75a9f78e-3d03-4638-b2b1-5e2d40b845db)

### Logs

Attached on the previous field because I can't remember my ip

### Anything else?

I have lots and lots of log files if you need them but I find cli hard, not sure why I manage a Minecraft server on a VPS.

* Add a thumbs-up to the bug report if you are also affected. This helps the bug report become more visible to the team and doesn't clutter the comments.
* Add a comment if you have any insights or background information that isn't already part of the conversation.

Raycoms:
@MotionlessTrain sth must be off with the iterator

MotionlessTrain:
Did the quarrier start with the quarry before you updated?

minnie1089:
yes, I did update the mod and dependencies on the client and server to try and fix the stopped quarrier, I think this was before I replaced the quarry although I am not entirely sure

[...]

PR #10052, opened 2024-07-18
Quarrier tweaks

MotionlessTrain:
Seems to mitigate #10026
Closes #
Closes #

- Don't remove blocks which were just placed 
  As the quarrier places blocks before breaking them, blocks like fences and walls could immediately be invalid because of the surrounding blocks that needed to be removed first
- Don't recalculate level when not needed. Apparently when walking towards where the quarrier was going to mine, they set the level on each citizen tick.
  It is a little wasteful to do the same calculation constantly over again, when the iterator, or its structure handler, has the result already
- Fix z-dimension of the iterator size
  This is the only thing that I could think of that could maybe have something to do with the issue, but I didn't see this being used anywhere, and it seemed to work fine before this change as well

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIQuarrier.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIQuarrier.java
@@ -360,6 +360,20 @@ public class EntityAIQuarrier extends AbstractEntityAIStructureWithWorkOrder<Job
     }
 
     @Override
+    protected boolean skipClearing(final BlueprintPositionInfo info, final BlockPos pos, final IStructureHandler handler)
+    {
+        if (super.skipClearing(info, pos, handler))
+        {
+            return true;
+        }
+        final BlockState state = handler.getWorld().getBlockState(pos);
+        // The blocks have been placed earlier by the quarrier. E.g. fences and walls can have the wrong blockstate
+        // (due to connecting to yet to be mined blocks), which means the block state is incorrect, and they would be mined by the quarrier,
+        // only to be replaced after the next restart. Instead, skip them now if the block itself is correct
+        return state.getBlock() == info.getBlockInfo().getState().getBlock();
+    }
+
+    @Override
     public boolean requestMaterials()
     {
         StructurePhasePlacementResult result;
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/util/LayerBlueprintIterator.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/util/LayerBlueprintIterator.java
@@ -72,9 +72,12 @@ public class LayerBlueprintIterator extends AbstractBlueprintIteratorWrapper
      */
     public void setLayer(final int newLevel)
     {
-        layer = newLevel;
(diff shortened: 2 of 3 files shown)


Thodor12 (approved):

src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIQuarrier.java
```diff
@@ -359,6 +359,20 @@ protected boolean skipBuilding(final BlueprintPositionInfo info, final BlockPos
...
+    {
+        if (super.skipClearing(info, pos, handler))
+        {
+            return true;
+        }
+        final BlockState state = handler.getWorld().getBlockState(pos);
+        // The blocks have been placed earlier by the quarrier. E.g. fences and walls can have the wrong blockstate
+        // (due to connecting to yet to be mined blocks), which means the block state is incorrect, and they would be mined by the quarrier,
+        // only to be replaced after the next restart. Instead, skip them now if the block itself is correct
+        return state.getBlock() == info.getBlockInfo().getState().getBlock();
```
Thodor12, line 373: Only thing I'm unsure about, block classes don't seem to have an equals method, can you guarantee these are identical instances?

MotionlessTrain: That is how blocks work afaik. A different instance means a different block altogether

MotionlessTrain: That is, each dirt block in the world would reference to a block state of the same block instance (in this case Blocks.DIRT). Dirt and cobblestone would both be instances of the base Block class, etc.

Nightenom: Anything with registry can/should be considered for instance equality

Raycoms: ye, this is fine

src/main/java/com/minecolonies/core/entity/ai/workers/util/LayerBlueprintIterator.java
```diff
@@ -122,7 +125,7 @@ public BlueprintPositionInfo getBluePrintPositionInfo(final BlockPos localPos)
     public BlockPos getSize()
     {
         final Blueprint blueprint = originalHandler.getBluePrint();
-        return new BlockPos(blueprint.getSizeX(), blueprint.getSizeY(), blueprint.getSizeY());
+        return new BlockPos(blueprint.getSizeX(), blueprint.getSizeY(), blueprint.getSizeZ());
```
Raycoms, line 128: whoopy

Raycoms (approved):

src/main/java/com/minecolonies/core/entity/ai/workers/util/LayerBlueprintIterator.java
MotionlessTrain: Exactly my reaction, when I saw this 😄

PR #10072, opened 2024-07-24
Add new jugs and seperate loottables

Raycoms:
Closes #
Closes #
Closes #

- Add EmptyJug/WaterJug/MilkJug/SoyMilkJug
- Allow conversion for player from any jug to Water/Milk if right clicking water/milk
- Add recipe replacement for previous water bottle and milk buckets to jug
- Add cowherder setting to allow producing jugs

Review please


src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -2661,5 +2661,14 @@
   "com.minecolonies.coremod.gui.townhall.happiness.desc.greatfood": "Had a tier 3 meal recently.",
   "com.minecolonies.coremod.gui.townhall.happiness.food": "Food Factor",
   "com.minecolonies.coremod.gui.townhall.happiness.greatfood": "Food Factor",
-  "death.attack.nether": "%s succumbed to the depths of the nether."
+  "death.attack.nether": "%s succumbed to the depths of the nether.",
+  "com.minecolonies.creativetab.huts": "Minecolonies Huts",
```
Thodor12, line 2665: Maybe "Hut Blocks"

Thodor12 (dismissed):

someaddons:
petition to rename Jug to Jar :D

Thodor12:
> petition to rename Jug to Jar :D

If we'd add chocolate milk you can name it the jar of dirt, see who will get the reference

Thodor12:
> petition to rename Jug to Jar :D

Though, the term jug is actually correct. Jars are described as "containers with a lid to store items, like storing cookies or what not".
Jugs are containers specifically meant for holding and pouring liquids. Making Jugs the preferred name

someaddons:
well its supposed to be a bottle not a container :D

MotionlessTrain:
Water/milk/soymilk bottle?

Raycoms:
> Water/milk/soymilk bottle?

Overlap with vanilla naming

MotionlessTrain:
Right 🤦

Talyda:
> petition to rename Jug to Jar :D

Pitcher? Flagon? Both work well! (though still not a bottle.)

kezmodius:
Jugs is also a word used for a female body part and it might be better to avoid it. What about Cup? It's a valid cooking measurement.

Shadizar:
Others may include urn, crock, ewer, etc.
<https://www.bing.com/search?q=synonyms%20for%20jug>

Shadizar:
Though I am neutral...just linking to the synonyms

Thodor12:
There is nothing wrong with the word jug unless you want to make it look wrong.
Cup is definitely the wrong measurement since this is not a small little cup, it's a full bottle size.
Urns are open top and not sealed off.
Crocks and ewer's I've never heard off, given the general audience they'd be impossible to find.

src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingModules.java
```diff
@@ -171,7 +171,9 @@ public class BuildingModules
       new BuildingEntry.ModuleProducer<>("cowherder_settings", () -> new SettingsModule().with(AbstractBuilding.BREEDING, new BoolSetting(true))
         .with(BuildingCowboy.MILKING_AMOUNT, new IntSetting(1))
         .with(BuildingCowboy.STEWING_AMOUNT, new IntSetting(1))
-        .with(BuildingCowboy.MILKING_DAYS, new IntSetting(1)), () -> SettingsModuleView::new);
+        .with(BuildingCowboy.MILKING_DAYS, new IntSetting(1))
+       .with(BuildingCowboy.MILK_ITEM, new StringSetting("item.minecolonies.large_milk_bottle", "item.minecraft.milk_bucket")), () -> SettingsModuleView::new);
```
uecasm, line 175: Wouldn't it be nicer to use the registry names here?

src/main/java/com/minecolonies/core/event/EventHandler.java
```diff
@@ -147,15 +151,21 @@ public static void onLootTableLoad(@NotNull final LootTableLoadEvent event)
...
-                                       .add(LootItem.lootTableItem(ModBlocks.blockSoyBean)
-                                              .when(EntityInBiomeTag.of(ModTags.humidBiomes))
-                                              .when(LootItemRandomChanceCondition.randomChance(0.01f))).build());
+            event.getTable().addPool(LootPool.lootPool().when(EntityInBiomeTag.of(ModTags.coldBiomes))
+                                       .add(AlternativesEntry.alternatives()
+                                              .otherwise(LootItem.lootTableItem(ModBlocks.blockCabbage)
+                                                           .when(MatchTool.toolMatches(ItemPredicate.Builder.item().of(ItemTags.HOES)))
+                                                           .when(LootItemRandomChanceCondition.randomChance(0.10f)))
+                                              .otherwise(LootItem.lootTableItem(ModBlocks.blockCabbage)
+                                                           .when(LootItemRandomChanceCondition.randomChance(0.01f)))).build());
```
uecasm, line 160: ~I think you need to list the hoe conditions first; the alternatives are processed in order.~
Actually maybe they are after all, it's harder to read in diff format.

src/main/java/com/minecolonies/core/generation/defaults/workers/DefaultChefCraftingProvider.java
```diff
@@ -248,5 +248,15 @@ protected void registerRecipes(@NotNull final Consumer<FinishedRecipe> consumer)
           .result(new ItemStack(ModItems.pepper_hummus, 2))
           .minBuildingLevel(1)
           .build(consumer);
+
+        CustomRecipeBuilder.create(CHEF, MODULE_CRAFTING, "soy_milk")
+          .inputs(List.of(
+            new ItemStorage(new ItemStack(ModItems.large_water_bottle)),
+            new ItemStorage(new ItemStack(ModBlocks.blockSoyBean)),
+            new ItemStorage(new ItemStack(ModBlocks.blockSoyBean)),
+            new ItemStorage(new ItemStack(ModBlocks.blockSoyBean))))
```
uecasm, line 257: You can add the soy bean once and x3 it rather than adding it three times.  There's no functional difference but it makes the recipe file tidier.

src/main/java/com/minecolonies/core/generation/defaults/DefaultRecipeProvider.java
```diff
@@ -637,6 +637,16 @@ private void buildOtherItems(@NotNull final Consumer<FinishedRecipe> consumer)
                 .unlockedBy("has_boat", has(ItemTags.BOATS))
                 .save(consumer);
 
+        ShapedRecipeBuilder.shaped(RecipeCategory.MISC, ModItems.large_empty_bottle, 4)
+          .pattern("PSP")
+          .pattern("P P")
+          .pattern("PGP")
+          .define('P', Items.GLASS_PANE)
+          .define('G', Items.GLASS)
+          .define('S', ItemTags.WOODEN_SLABS)
```
uecasm, line 646: Maybe a wooden button?  A slab seems a bit big for a bottle, even if you are making 4 at a time.
Though the slab is (oddly) cheaper, so maybe this makes sense.

uecasm: Also buttons are mechanic items, so they're more expensive that way too.

Raycoms: Yeah,  I wanted to make it a button, but in the crafting grid the slab looks nicer

src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingModules.java
Raycoms: I think this was loaded initially at registry loading time, so the item regsitry might not be loaded yet.

src/main/java/com/minecolonies/core/event/EventHandler.java
Raycoms: I did test it and it works fine =D

src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -2661,5 +2661,14 @@
...
+  "com.minecolonies.creativetab.general": "Minecolonies General",
+  "com.minecolonies.creativetab.food": "Minecolonies Food",
+
+  "item.minecolonies.large_empty_bottle": "Large Empty Bottle",
+  "item.minecolonies.large_water_bottle": "Large Water Bottle",
+  "item.minecolonies.large_milk_bottle": "Large Milk Bottle",
+  "item.minecolonies.large_soy_milk_bottle": "Large Soy Milk Bottle",
+  "com.minecolonies.coremod.setting.minecolonies:milk_item": "Milk Item:"
```
someaddons, line 2673: not descriptive enough for the tooltip^^

Raycoms: that's not the tooltip, that's just the setting header

someaddons: guess its missing a tooltip then!

someaddons (approved):

PR #10102, opened 2024-08-09
Allow use bonemeal to crops

gisellevonbingen:
Closes #
Closes #
Closes #

- MinecoloniesCropBlock implements BonemealableBlock
-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
+++ b/src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
@@ -9,6 +9,7 @@ import net.minecraft.core.Direction;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.tags.TagKey;
+import net.minecraft.util.RandomSource;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.InteractionResult;
 import net.minecraft.world.entity.player.Player;
@@ -18,6 +19,7 @@ import net.minecraft.world.level.*;
 import net.minecraft.world.level.biome.Biome;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.Blocks;
+import net.minecraft.world.level.block.BonemealableBlock;
 import net.minecraft.world.level.block.SoundType;
 import net.minecraft.world.level.block.state.BlockBehaviour;
 import net.minecraft.world.level.block.state.BlockState;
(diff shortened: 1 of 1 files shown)


Thodor12:
If you want to make this extend BonemealableBlock you have to remove the `use` method. Compost is already implemented with a useOn that applies bonemeal similar to how vanilla bonemeal works, it checks for BonemealableBlock instances.
You can combine your 2 PR's into one

gisellevonbingen:
I used translator for my poor english.
In Minecraft vanilla, the convention for bonemeal handling is to implement BonemealableBlock.
Also, I have never seen an example of being implemented it directly via the useOn method.

And 2 PR's are logically different features.
So, there are should be separated.

someaddons:
it is intentionally not bonemealable due to other mods using it in very inflationary ways

Thodor12:
> I used translator for my poor english. In Minecraft vanilla, the convention for bonemeal handling is to implement BonemealableBlock. Also, I have never seen an example of being implemented it directly via the useOn method.
> 
> And 2 PR's are logically different features. So, there are should be separated.

I meant the second PR would be automatically fixed if you already did this change alone

Shadizar:
> it is intentionally not bonemealable due to other mods using it in very inflationary ways

It'd seem to be unharmful no? i.e., if folks want to play that way it doesn't really harm anything right?
It does seem this is the sort of thing that pushes folks more into mixins when there's closed-off options?

gisellevonbingen:
> > I used translator for my poor english. In Minecraft vanilla, the convention for bonemeal handling is to implement BonemealableBlock. Also, I have never seen an example of being implemented it directly via the useOn method.
> > And 2 PR's are logically different features. So, there are should be separated.
> 
> I meant the second PR would be automatically fixed if you already did this change alone

2 PR is on each other issue.
and second PR is on another branch where from based on master.

Thodor12:
> it is intentionally not bonemealable due to other mods using it in very inflationary ways

On one hand I get where you're coming from, on the other I feel like this is not really useful.
Remember that this interaction is only for the player growing the crops, colonists do not reuse this logic and can only cope with compost anyway. So at best you want to limit the player in this sense, who already cannot grow these crops.

I think it is fair to allow this to grow since:
A: Make use of vanilla logic
B: People can create bulk compost anyway to mass grow for the colonists

someaddons:
> > it is intentionally not bonemealable due to other mods using it in very inflationary ways
> 
> On one hand I get where you're coming from, on the other I feel like this is not really useful. Remember that this interaction is only for the player growing the crops, colonists do not reuse this logic and can only cope with compost anyway. So at best you want to limit the player in this sense, who already cannot grow these crops.
> 
> I think it is fair to allow this to grow since: A: Make use of vanilla logic B: People can create bulk compost anyway to mass grow for the colonists

please dont use code review to discuss design, do that in discord. Also this is ray choice

Shadizar:
> > > it is intentionally not bonemealable due to other mods using it in very inflationary ways
> > 
> > 
> > On one hand I get where you're coming from, on the other I feel like this is not really useful. Remember that this interaction is only for the player growing the crops, colonists do not reuse this logic and can only cope with compost anyway. So at best you want to limit the player in this sense, who already cannot grow these crops.
> > I think it is fair to allow this to grow since: A: Make use of vanilla logic B: People can create bulk compost anyway to mass grow for the colonists
> 
> please dont use code review to discuss design, do that in discord. Also this is ray choice

Gotcha...I guess I was confused because you commented on design discussion topic which seemed to open that door

PR #10103, opened 2024-08-09
Block to use compost on max age crop

gisellevonbingen:
Closes #
Closes #
Closes #

- Added condition for whether crop is max age on MinecoloniesCropBlock#use method
-
-

Review please


Raycoms (approved):

Thodor12:
I was thinking instead of the weird use method on the crop block we add a special case to the compost item itself since that's already where the bonemeal logic lives

someaddons (approved):

PR #10104, opened 2024-08-09
Fix compost application rules to Minecolonies crops

Thodor12:
Closes #10103 

- Respect the max age of the crop blocks to not unnecessarily consume compost
- Fix the particles and sounds not being shown/heard when using compost
- Remove logic from farmland/crop blocks into compost item

Review please


Raycoms (approved):

src/main/java/com/minecolonies/core/items/ItemCompost.java
```diff
@@ -23,19 +31,52 @@ public ItemCompost(final Item.Properties properties)
...
+    {
+        BlockState state = level.getBlockState(pos);
+        if (state.getBlock() instanceof MinecoloniesCropBlock cropBlock)
+        {
+            if (!cropBlock.isMaxAge(state))
+            {
+                if (level instanceof ServerLevel serverLevel)
+                {
+                    cropBlock.attemptGrow(state, serverLevel, pos);
+                    stack.shrink(1);
```
gisellevonbingen, line 54: please check player is creative mode.

Thodor12: Unnecessary, vanilla bonemeal action also does not take this into account, creative players can just grab a new stack of compost anyway.

Raycoms: afaik vanilla does

Thodor12:
```java
    public static boolean applyBonemeal(ItemStack p_40628_, Level p_40629_, BlockPos p_40630_, Player player) {
        BlockState blockstate = p_40629_.getBlockState(p_40630_);
        int hook = ForgeEventFactory.onApplyBonemeal(player, p_40629_, p_40630_, blockstate, p_40628_);
        if (hook != 0) {
            return hook > 0;
        } else {
            if (blockstate.getBlock() instanceof BonemealableBlock) {
                BonemealableBlock bonemealableblock = (BonemealableBlock)blockstate.getBlock();
                if (bonemealableblock.isValidBonemealTarget(p_40629_, p_40630_, blockstate, p_40629_.isClientSide)) {
                    if (p_40629_ instanceof ServerLevel) {
                        if (bonemealableblock.isBonemealSuccess(p_40629_, p_40629_.random, p_40630_, blockstate)) {
                            bonemealableblock.performBonemeal((ServerLevel)p_40629_, p_40629_.random, p_40630_, blockstate);
                        }

                        p_40628_.shrink(1);
                    }

                    return true;
                }
            }

            return false;
        }
    }
```
That is the vanilla implementation, it always shrinks the input stack regardless of creative or not

Raycoms: Odd, I remember testing stuff with bonemeal and it never ran out

Thodor12: I can boot it up again and check, but the code seems to suggest there's no creative check

Thodor12: Interesting, despite always hitting the stack shrink, if I'm creative the stack does not actually shrink at all.
Must be the inventory class of the player knowing not to consume anything in creative or something?
So anyhow this is a non issue.

someaddons (approved):

PR #10136, opened 2024-08-19
Do not run needless patreon auth checks

Thodor12:
A large majority of players will likely never even have subscribed to Patreon, hence we can try to reduce network traffic on our servers by simply not allowing the mod to make a call to our auth system.
If players do want to play with features enabled, they will have to turn on the config value in the server config in order to start running these checks again.

This is merely a proposal, if we don't want this I'll remove it again.

The secondary check I added is something that prevents running the auth check again if features were previously unlocked already, that should also result in unnecessary calls being removed upon secondary opening of the town hall GUI.
(It seems highly unlikely that a Patreon subscription would expire in the middle of a session)

Review please

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,12 +19,13 @@ on:
 permissions:
   contents: write
   statuses: write
+  packages: write
 
 jobs:
   release:
     uses: ldtteam/operapublicacreator/.github/workflows/gradle.publish.yaml@main
     with:
-      java: 21
+      java: 17
     secrets:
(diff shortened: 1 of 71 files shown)


someaddons:
It has to auto enable for the patreons :)

someaddons (approved):

Nightenom (approved):
Honestly should run once during mod loading but whatever

Raycoms:
> Honestly should run once during mod loading but whatever

We don't do that, so that people that sign up to patreon don't have to relog to get the stuff.

PR #10175, opened 2024-08-27
Make pond actually check for what is on wiki

Nightenom:
Closes issues with fisherman not having first pond

- add waterPathJob (and problematic pond positions) to path tracking debug
- fix pond iterating so it is actually 7x7x2, keep 30% chance to not check the second level, below example of all problems within one checked position (see some blocks on second level not being checked)
- will port once merged

image

Image of successful waterPathJob:
image

Review please


src/main/java/com/minecolonies/api/util/Pond.java
```diff
@@ -17,8 +20,7 @@ public final class Pond
     /**
      * The minimum pond requirements.
      */
-    public static final int WATER_POOL_WIDTH_REQUIREMENT  = 5;
-    public static final int WATER_POOL_LENGTH_REQUIREMENT = 5;
+    public static final int WATER_POOL_WIDTH_REQUIREMENT  = 7;
```
someaddons: it is actually intended that we just check 5 but state 7 on the wiki :)

someaddons (approved):

Nightenom:
Verified on discord <https://discord.com/channels/472875599422291968/1274733488398139497/1279541702209568840>

PR #10188, opened 2024-09-02
NetworkChannel update for custom messages

Wmaxlees:
Updated the NetworkChannel class to allow modders to add new NetworkChannels externally. Also, changed a field in the Crafting GUI open message to be protected so any subclasses can access this.

- Make NetworkChannel.registerMessage public
- Change a field in the Crafting GUI message to protected so it can be subclassed effectively.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/network/messages/server/colony/building/OpenCraftingGUIMessage.java
+++ b/src/main/java/com/minecolonies/core/network/messages/server/colony/building/OpenCraftingGUIMessage.java
@@ -29,7 +29,7 @@ public class OpenCraftingGUIMessage extends AbstractBuildingServerMessage<IBuild
     /**
      * The type of container.
      */
-    private int id;
+    protected int id;
 
     /**
      * Empty public constructor.
--- a/src/main/java/com/minecolonies/core/network/NetworkChannel.java
+++ b/src/main/java/com/minecolonies/core/network/NetworkChannel.java
@@ -86,6 +86,12 @@ public class NetworkChannel
     private final AtomicInteger messageCounter = new AtomicInteger();
 
     /**
+     * Next available free index for network messages.
+     * Index 0 is reserved
+     */
+    private int nextMessageId = 1;
+
+    /**
      * Creates a new instance of network channel.
      *
      * @param channelName unique channel name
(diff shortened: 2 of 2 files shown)


Thodor12:
Before we go ahead and merge this, could you give an example of a place in the code where you require this?

src/main/java/com/minecolonies/core/network/messages/server/colony/building/OpenCraftingGUIMessage.java
```diff
@@ -29,7 +29,7 @@ public class OpenCraftingGUIMessage extends AbstractBuildingServerMessage<IBuild
     /**
      * The type of container.
      */
-    private int id;
+    protected int id;
```
Thodor12, line 32: Is this change strictly needed?
You can just create a custom message class and use that instead of modifying and extending this one.

uecasm:
I already mentioned in Discord that this sort of thing is counterproductive.  If you want to make custom messages, make your own channel for it, using standard Forge code.

You are **absolutely forbidden** by how Forge networking works from adding additional messages into some other mod's channel.

There's no reason you can't override the GUI view and send your own message when the button is clicked.  Or alternatively, the `CraftingType` could be extended with a method to open the appropriate GUI view, though that's trickier because GUIs are client-side-only and you can trigger classloading problems if you're not careful.

Thodor12:
> I already mentioned in Discord that this sort of thing is counterproductive. If you want to make custom messages, make your own channel for it, using standard Forge code.
> 
> You are **absolutely forbidden** by how Forge networking works from adding additional messages into some other mod's channel.
> 
> There's no reason you can't override the GUI view and send your own message when the button is clicked. Or alternatively, the `CraftingType` could be extended with a method to open the appropriate GUI view, though that's trickier because GUIs are client-side-only and you can trigger classloading problems if you're not careful.



> I already mentioned in Discord that this sort of thing is counterproductive. If you want to make custom messages, make your own channel for it, using standard Forge code.
> 
> You are **absolutely forbidden** by how Forge networking works from adding additional messages into some other mod's channel.
> 
> There's no reason you can't override the GUI view and send your own message when the button is clicked. Or alternatively, the `CraftingType` could be extended with a method to open the appropriate GUI view, though that's trickier because GUIs are client-side-only and you can trigger classloading problems if you're not careful.

I mean, realistically if the problem is the OpenCraftingGUIMessage, you just don't use said message.
That message is only implemented to work for smelting, recipes and brewing.

Considering you have to make your own module view for your custom crafting module, you also have to implement a custom message to open a separate view to add a message, and so forth.

marchermans:
I am 100% against this. Our network is our protected namespace. If you need your own payloads, configure your own channel under your own network.

marchermans (changes requested):
This will not be merged if it were up to me as it violates several design objectives, language objectives and forge design and architecture objectives

Wmaxlees:
Understood. I'll close the pull request. Is there a proper channel in the future for me to discuss changes with the dev team before making a pull request? I'd like to save everyone's time if the idea is a no-go. Particularly for bigger changes like converting the IToolType system into a registry (which was also suggested on discord as a PR).

PR #10170, opened 2024-08-26
Simplify item block registry

Nightenom:
# Changes proposed in this pull request:
- api (binary) breaking change, in 1.21 its fine, however when backporting this might potentially hurt existing addons?
- moves item/block to deferred registry, removal of initializer classes
- removal of empty classes for registry things, possible further cleanup: removal of empty block/item classes

[x] Yes I tested this before submitting it.
[ ] I also did a multiplayer test.

Review please

PR #10190, opened 2024-09-03
Tool registry

Wmaxlees:
Change tooltypes from hardcoded to a registry. This allows addon makers to define their own tool types to use in custom recipes + jobs. As part of this, some of the tool level generation + checks for tool type have been unified + updated.

[x] Yes I tested this before submitting it (though this change touches lots of code so I can't personally test it exhaustively).

Review please


src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -189,16 +191,16 @@ public boolean matches(@NotNull final ItemStack stack)
...
                                               .filter(s -> getToolClass().getName().equalsIgnoreCase(s))
-                                              .map(ToolType::getToolType)
-                                              .anyMatch(t -> t != ToolType.NONE && (stack.getDamageValue() > 0 || !stack.isDamaged()) && ItemStackUtils.hasToolLevel(stack,
+                                              .map(ToolTypeEntry::getToolType)
+                                              .anyMatch(t -> t != ModToolTypes.none.get() && (stack.getDamageValue() > 0 || !stack.isDamaged()) && ItemStackUtils.hasToolLevel(stack,
                                                 t,
                                                 getMinLevel(),
                                                 getMaxLevel()));
 
         if (!toolTypeResult)
```
Thodor12: Pretty sure this block is unnecessary given that you now iterate all tool types in the registry, including shields and flint and steel, so this extra if should not be necessary.

src/main/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -1152,5 +1006,63 @@ public static void consumeFood(final ItemStack foodStack, final AbstractEntityCi
...
     }
+
+    /**
+     * Get the tool level for vanilla tools.
+     *
+     * @param toolId The type of vanilla tool
+     * @param itemStack The item stack to check
+     * @return The tool level
+     */
+    public static int vanillaToolLevel(final String toolId, ItemStack itemStack) {
```
Thodor12: Considering these methods are applicable to the builders inside of the initializer class and don't really have a usage outside there, please move.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,201 @@
...
+                .build());
+
+        sword = register(SWORD_ID, () -> new ToolTypeEntry.Builder()
+                .setName(SWORD_ID)
+                .setVariableMaterials(true)
+                .setDisplayName(Component.translatable(ToolTranslationConstants.TOOL_TYPE_SWORD))
+                .setIsTool(itemStack -> ItemStackUtils.canPerformDefaultActions(itemStack, ToolActions.DEFAULT_SWORD_ACTIONS)  || Compatibility.isTinkersWeapon(itemStack))
+                .setToolLevel(itemStack -> {
+                    final String toolId = SWORD_ID;
+                    if (!Compatibility.getMiningLevelCompatibility(itemStack, toolId))
```
Thodor12: This one was unnecessary

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,201 @@
...
+                    return -1;
+                })
+                .build());
+
+        bow = register(BOW_ID, () -> new ToolTypeEntry.Builder()
+                .setName(BOW_ID)
+                .setVariableMaterials(false)
+                .setDisplayName(Component.translatable(ToolTranslationConstants.TOOL_TYPE_BOW))
+                .setIsTool(itemStack -> itemStack.getItem() instanceof BowItem)
+                .setToolLevel(itemStack -> 1)
```
Thodor12: Durability based perhaps?

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,91 @@
...
+import com.minecolonies.api.util.constant.IToolType;
+import net.minecraft.network.chat.Component;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.registries.RegistryObject;
+
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class ToolTypeEntry implements IToolType {
```
Thodor12: Javadoc and formatting (entire class)

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,201 @@
...
+import net.minecraft.world.item.*;
+import net.minecraftforge.common.ToolActions;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.RegistryObject;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+
+public class ModToolTypes {
```
Thodor12: Javadoc and formatting (entire class)

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,91 @@
...
+        this.displayName = displayName;
+        this.isTool = isTool;
+        this.itemLevel = itemLevel;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public boolean hasVariableMaterials() {
```
Thodor12: Variable materials is no longer required.

src/main/java/com/minecolonies/api/util/constant/IToolType.java
```diff
@@ -26,5 +29,27 @@ public interface IToolType
...
+
+    /**
+     * The level the item stack is.
+     *
+     * @param itemStack to test
+     * @return The level the item stack is.
+     */
+    int getMiningLevel(ItemStack itemStack);
+
+    class IToolTypeComparator implements Comparator<IToolType> {
```
Thodor12: Javadoc and formatting

src/main/java/com/minecolonies/api/util/constant/IToolType.java
```diff
@@ -1,6 +1,9 @@
 package com.minecolonies.api.util.constant;
 
 import net.minecraft.network.chat.Component;
+import net.minecraft.world.item.ItemStack;
+
+import java.util.Comparator;
 
 public interface IToolType
```
Thodor12: If you're doing javadoc anyway, give the interface some love as well

src/main/java/com/minecolonies/core/client/gui/modules/WindowListRecipes.java
```diff
@@ -195,7 +195,7 @@ public void updateElement(final int index, @NotNull final Pane rowPane)
 
                 final Text intermediate = rowPane.findPaneOfTypeByID("intermediate", Text.class);
                 intermediate.setVisible(false);
-                if(recipe.getRequiredTool() != ToolType.NONE)
+                if(recipe.getRequiredTool() != ModToolTypes.none.get())
```
Thodor12: Formatting (spacing between if)

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/plantation/modules/specific/CactusPlantModule.java
```diff
@@ -38,8 +39,8 @@ protected boolean isValidHarvestBlock(final BlockState blockState)
     }
 
     @Override
-    public ToolType getRequiredTool()
+    public ToolTypeEntry getRequiredTool()
```
Thodor12: Return IToolType

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/plantation/modules/specific/CocoaPlantModule.java
```diff
@@ -63,9 +63,9 @@ public BlockState getPlantingBlockState(final Level world, final BlockPos workPo
     }
 
     @Override
-    public ToolType getRequiredTool()
+    public ToolTypeEntry getRequiredTool()
```
Thodor12: Return IToolType

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/plantation/modules/specific/GlowBerriesPlantModule.java
```diff
@@ -47,8 +48,8 @@ public ResourceLocation getRequiredResearchEffect()
     }
 
     @Override
-    public ToolType getRequiredTool()
+    public ToolTypeEntry getRequiredTool()
```
Thodor12: Return IToolType

Thodor12 (changes requested):
Overal looks great, thanks for taking the effort of implementing this.
There are a few style changes that need looking at.

Also I noticed that you touched a few things related to data generation, so could you execute `runData` gradle task in IntelliJ?

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/util/ItemStackUtils.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
Wmaxlees: Done. Though the CheckStyle config fails to load for me.

src/main/java/com/minecolonies/api/util/constant/IToolType.java
Wmaxlees: Done.

src/main/java/com/minecolonies/core/client/gui/modules/WindowListRecipes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Done. Though the CheckStyle config fails to load for me.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
Wmaxlees: Removed.

src/main/java/com/minecolonies/api/util/constant/IToolType.java
Wmaxlees: Done. Though the CheckStyle config fails to load for me.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/plantation/modules/specific/CocoaPlantModule.java
Wmaxlees: Done.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/plantation/modules/specific/CactusPlantModule.java
Wmaxlees: Done.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/plantation/modules/specific/GlowBerriesPlantModule.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /**
+     * Determine whether an item stack can perform the default actions of a given tool.
+     *
+     * @param itemStack The item stack to check
+     * @param actions The set of actions to compare
+     * @return Whether the item stack can perform the actions
+     */
+    public static Boolean canPerformDefaultActions(ItemStack itemStack, Set<ToolAction> actions) {
```
Thodor12: Unnecessary boxing of Boolean

Wmaxlees: Done.

Thodor12 (dismissed):

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -184,24 +186,16 @@ public ItemStack getResult()
     @Override
     public boolean matches(@NotNull final ItemStack stack)
     {
-        //API:Map the given strings a proper way.
-        final boolean toolTypeResult = !ItemStackUtils.isEmpty(stack)
+        return !ItemStackUtils.isEmpty(stack)
```
Raycoms: can we untangle this stream to a loop? This is called fairly often and the list is not long, so the performance impact of this is for sure real

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -213,79 +207,12 @@ private Set<String> getToolClasses(final ItemStack stack)
...
-                set.add("boots");
-            }
-            else if (armor.getEquipmentSlot() == EquipmentSlot.HEAD)
-            {
-                set.add("helmet");
-            }
-            else if (armor.getEquipmentSlot() == EquipmentSlot.LEGS)
-            {
-                set.add("leggings");
+        for (RegistryObject<ToolTypeEntry> toolType : ModToolTypes.toolTypes) {
```
Raycoms: formatting

src/main/java/com/minecolonies/api/compatibility/Compatibility.java
```diff
@@ -120,10 +120,10 @@ public static boolean isTinkersWeapon(@NotNull final ItemStack stack)
     /**
      * Check if a certain item stack is a tinkers tool of the given tool type.
      * @param stack the stack to check for.
-     * @param toolType the tool type.
+     * @param toolId the tool type ID.
      * @return true if so.
      */
-    public static boolean isTinkersTool(@Nullable final ItemStack stack, IToolType toolType) { return tinkersCompat.isTinkersTool(stack, toolType); }
+    public static boolean isTinkersTool(@Nullable final ItemStack stack, final String toolId) { return tinkersCompat.isTinkersTool(stack, toolId); }
```
Raycoms: Can't we still hand around the actual type or registry entry? Not a huge fan of handing around the strings

Raycoms (commented):
I really like this, good job.
Some minor comments

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees: Done. I was also able to remove an extra loop!

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees: Done

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -139,37 +168,50 @@ public static Tool deserialize(final IFactoryController controller, final Friend
...
+        }
+
+        if (stack.getCount() < 1)
+        {
+            return false;
+        }
+
+        for (RegistryObject<ToolTypeEntry> toolType : ModToolTypes.toolTypes)
+        {
+            if (!getToolClass().equals(toolType.get()))
```
Raycoms: can't we == those?

Thodor12: It's an interface, should use equals

Thodor12:
Reminder for when formatting, don't blatently run the formatter over the entire file, that introduces a load of unrelated changes and makes it difficult to view actual changes.

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -129,174 +160,46 @@ public static void serialize(final IFactoryController controller, final Friendly
...
-        else if (stack.getItem() instanceof SwordItem || Compatibility.isTinkersWeapon(stack))
-        {
-            set.add("weapon");
-        }
-        else if (stack.getItem() instanceof ShieldItem)
-        {
-            set.add("shield");
-        }
-        else if (stack.getItem() instanceof ArmorItem)
+        for (RegistryObject<ToolTypeEntry> toolType : ModToolTypes.toolTypes)
```
Thodor12: Do not iterate over our `ModToolTypes` class, this should not be used as a container class to get references to all objects.
(Addon mods should not be using our deferred register to create their addon instances, and not inserted them inside our mod container).

Use the Forge Registry instance you create inside `CommonMinecoloniesAPIProxy`:
```java
MinecoloniesAPIProxy.getInstance().getToolTypeRegistry()
```
(Please introduce the correct getters in the interface and implement them inside `CommonMinecoloniesAPIProxy`

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,283 @@
...
+    public static RegistryObject<ToolTypeEntry> helmet;
+    public static RegistryObject<ToolTypeEntry> leggings;
+    public static RegistryObject<ToolTypeEntry> chestplate;
+    public static RegistryObject<ToolTypeEntry> boots;
+    public static RegistryObject<ToolTypeEntry> flint_and_steel;
+
+    /**
+     * The list of all tool types.
+     */
+    public static List<RegistryObject<ToolTypeEntry>> toolTypes = new ArrayList<>();
```
Thodor12: Remove this holder list

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,283 @@
...
+
+    /**
+     * Register a new tool type. This should be used rather than directly adding entries
+     * to the registry.
+     *
+     * @param id       The tool type name
+     * @param supplier A function that provides the ToolTypeEntry
+     * @return The new RegistryObject
+     */
+    public static RegistryObject<ToolTypeEntry> register(final String id, final Supplier<ToolTypeEntry> supplier)
```
Thodor12: Remove this method in favor of just registering to the `DEFERRED_REGISTER` directly.

Thodor12 (changes requested):

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,283 @@
...
+ */
+public class ModToolTypes
+{
+    /**
+     * The Forge registry for tool types. Do not directly add to this registry. Prefer using the
+     * register() function provided here.
+     */
+    private final static DeferredRegister<ToolTypeEntry> DEFERRED_REGISTER = DeferredRegister.create(new ResourceLocation(Constants.MOD_ID, "tooltypes"), Constants.MOD_ID);
+
+    public static String NONE_ID            = "none";
```
Nightenom: Can you please inline these into registry methods? and ideally follow style (for mod[type]s.java) set in #10170
includes: member visibility, final, etc.

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Raycoms: Aahh, I thought we were comparing two types.
But that opens up the question, why isn't getToolClass returning the same type of type?

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,283 @@
...
+    public static String FISHING_ROD_ID     = "rod";
+    public static String SHEARS_ID          = "shears";
+    public static String SHIELD_ID          = "shield";
+    public static String HELMET_ID          = "helmet";
+    public static String LEGGINGS_ID        = "leggings";
+    public static String CHESTPLATE_ID      = "chestplate";
+    public static String BOOTS_ID           = "boots";
+    public static String FLINT_AND_STEEL_ID = "flintandsteel";
+
+    public static RegistryObject<ToolTypeEntry> none;
```
Nightenom: these should be final

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Thodor12: Yeah actually, what is this entire loop good for anyway? It only contains a single IToolType which is checked against all tool types, so what is the point here?

Also tools previously were able to return several tool classes because a tool (like a paxel) could have multiple functionalities, was that logic retained as well?

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,198 @@
...
+import net.minecraftforge.registries.RegistryObject;
+
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * An entry in the ToolType registry that defines the types of
+ * tools within the colony.
+ */
+public class ToolTypeEntry implements IToolType
```
Nightenom: I feel this could be kept in same place as old ToolType which was deleted? To at least keep the source break minimal

Thodor12: I'm actually not a fan where this class is placed because it has nothing to do with items. Maybe move it to `/api/tools/registry/ToolTypeEntry`, and maybe put some other relevant files there?

Nightenom: I would rather keep source break minimal, since this is already binary break, but either dont move or move all

Thodor12: I'd probably move all, we have too many "util" files. IToolType can also easily move to the folder I mentioned.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,198 @@
...
+        this.itemLevel = itemLevel;
+    }
+
+    /**
+     * Get the registered ToolTypeEntry from a tool ID.
+     *
+     * @param toolId The target ID
+     * @return The ToolTypeEntry
+     */
+    public static IToolType getToolType(final String toolId)
```
Thodor12: Remove method and use the API proxy to get the tool types.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,198 @@
...
+        return displayName;
+    }
+
+    /**
+     * Determine whether an item stack works as this tool.
+     *
+     * @param itemStack to test
+     * @return Whether the item stack can act as the tool.
+     */
+    public Boolean checkIsTool(ItemStack itemStack)
```
Thodor12: Unnecessary boxing

src/main/java/com/minecolonies/api/compatibility/Compatibility.java
Wmaxlees: Done. I was worried about capturing the ToolTypeEntry that wasn't assigned yet but I tested and it works fine,

src/main/java/com/minecolonies/api/items/ModToolTypes.java
```diff
@@ -0,0 +1,283 @@
...
+     * The list of all tool types.
+     */
+    public static List<RegistryObject<ToolTypeEntry>> toolTypes = new ArrayList<>();
+    static
+    {
+        none = register(NONE_ID,
+          () -> new ToolTypeEntry.Builder().setName("")
+                  .setDisplayName(Component.translatable(ToolTranslationConstants.TOOL_TYPE_NONE))
+                  .setIsTool(itemStack -> true)
+                  .setToolLevel(itemStack -> itemStack == null ? 0 : 1)
```
Thodor12: Would probably have this return `-1` by default, as tooltype none isn't really intended to be used to begin with.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
Thodor12: Final class

Thodor12 (changes requested):
Considering this is a registry entry now, I don't think we actually need `IToolType` anymore at all. Registry entries are not supposed to be overridable, so if someone wanted to make something custom they'd have to make a `ToolTypeEntry` anyway, so having an interface where the entry extends from feels kinda pointless. (We don't have this for any other registry entry either).

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees: Fixed. This doesn't actually need to loop now with the new logic. And also should work for paxels, etc.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Done

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Done. Actually have a new register function to simplify ID + ResourceLocation.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Based on conversations in Discord, will keep as is with additional finals.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
Wmaxlees: Moved to `/api/tools/registry/`

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/ModToolTypes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/items/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.function.Consumer;
+
+/**
+ * Class used for storing and registering any ToolTypes.
+ */
+public class ModToolTypes
+{
+    public final static DeferredRegister<ToolTypeEntry> DEFERRED_REGISTER = DeferredRegister.create(new ResourceLocation(Constants.MOD_ID, "tooltypes"), Constants.MOD_ID);
+
+    public static final String NONE_ID            = "none";
```
Nightenom: These should be in the register methods
I've seen your comment - in your PR you dont use these elsewhere, plus if someone need the registry key then it should be acquired from registry itself, not via static strings

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
```diff
@@ -0,0 +1,242 @@
...
+    public static final String FISHING_ROD_ID     = "rod";
+    public static final String SHEARS_ID          = "shears";
+    public static final String SHIELD_ID          = "shield";
+    public static final String HELMET_ID          = "helmet";
+    public static final String LEGGINGS_ID        = "leggings";
+    public static final String CHESTPLATE_ID      = "chestplate";
+    public static final String BOOTS_ID           = "boots";
+    public static final String FLINT_AND_STEEL_ID = "flintandsteel";
+
+    public static RegistryObject<ToolTypeEntry> none;
```
Nightenom: Still missing final

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -129,45 +161,39 @@ public static void serialize(final IFactoryController controller, final Friendly
...
+        if (ItemStackUtils.isEmpty(stack))
+        {
+            return false;
+        }
+
+        if (!getToolClass().checkIsTool(stack)) {
+            return false;
+        }
+
+        return stack.getDamageValue() > 0 || !stack.isDamaged() && ItemStackUtils.hasToolLevel(stack, getToolClass(), getMinLevel(), getMaxLevel());
```
Thodor12: I would still remove `stack.getDamageValue() > 0 || !stack.isDamaged()` and just leave this up to the tool level calculation

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -129,45 +161,39 @@ public static void serialize(final IFactoryController controller, final Friendly
...
+    @Override
+    public boolean matches(@NotNull final ItemStack stack)
     {
-        return toolClass;
+        if (ItemStackUtils.isEmpty(stack))
+        {
+            return false;
+        }
+
+        if (!getToolClass().checkIsTool(stack)) {
```
Thodor12: This check is unnecessary because `ItemStackUtils#hasToolLevel` checks this as well.

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
```diff
@@ -0,0 +1,242 @@
...
+    public static RegistryObject<ToolTypeEntry> chestplate;
+    public static RegistryObject<ToolTypeEntry> boots;
+    public static RegistryObject<ToolTypeEntry> flint_and_steel;
+
+    static
+    {
+        none = register(NONE_ID,
+          builder -> builder.setDisplayName(Component.translatable(ToolTranslationConstants.TOOL_TYPE_NONE))
+                  .setIsTool(itemStack -> true)
+                  .setToolLevel(itemStack -> itemStack == null ? 0 : -1)
```
Thodor12: As said in the previous comment, this should always return -1, not conditionally. Even if the itemstack is null or not.

src/main/java/com/minecolonies/api/tools/registry/IToolTypeRegistry.java
```diff
@@ -0,0 +1,12 @@
+package com.minecolonies.api.tools.registry;
+
+import com.minecolonies.api.IMinecoloniesAPI;
+import net.minecraftforge.registries.IForgeRegistry;
+
+public class IToolTypeRegistry
```
Thodor12: You didn't strictly have to create this interface, because we're trying to actually consolidate logic, you can keep this method just fine within `ModToolTypes`, but regardless what you do, javadoc is required.

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,216 @@
...
+{
+    /**
+     * The registry identifier for this tool type.
+     */
+    private final ResourceLocation registryName;
+
+    /**
+     * The name of the tool type.
+     */
+    private final String name;
```
Thodor12: Name is obsolete and should not be used at all, that's what we have the registry name for.

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,216 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param name        The name of the tool type
+     * @param displayName The human readable name of the tool type
+     * @param isTool      A predicate for determining if an itemstack is the tool type
+     * @param itemLevel   A function to return the item level of an item stack
+     */
+    private ToolTypeEntry(final String name, final Component displayName, final Predicate<ItemStack> isTool, final Function<ItemStack, Integer> itemLevel, final ResourceLocation registryName)
```
Thodor12: Order of constructor is not identical to field order, would expect registryName in front (also missing javadoc for registryName)

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,216 @@
...
+    {
+        this.name = name;
+        this.displayName = displayName;
+        this.isTool = isTool;
+        this.itemLevel = itemLevel;
+        this.registryName = registryName;
+    }
+
+    public ResourceLocation getRegistryName()
```
Thodor12: Missing javadoc

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,216 @@
...
+    {
+        return registryName;
+    }
+
+    /**
+     * @return The tool type's name
+     */
+    public String getName()
```
Thodor12: Remove method

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,216 @@
...
+    {
+        return name;
+    }
+
+    /**
+     * @return the component for the human readable name.
```
Thodor12: Missing method description

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,216 @@
...
+    {
+        /**
+         * The registry identifier for this tool type.
+         */
+        private ResourceLocation registryName;
+
+        /**
+         * The name of the tool type.
+         */
+        private String name;
```
Thodor12: Name should be removed

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
```diff
@@ -0,0 +1,216 @@
...
+            return this;
+        }
+
+        /**
+         * Set the name for the new ToolTypeEntry
+         *
+         * @param name the new name
+         * @return this
+         */
+        public Builder setName(final String name)
```
Thodor12: Method should be removed

Nightenom (dismissed):
Looks decent to me now, just two small nitpicks

Thodor12 (changes requested):

Thodor12:
Also, given we changed the NBT data of several things changing from name -> registryName, this can cause some failures in attempting to reload the data.
For example the previous name used to be `pickaxe`, which is now not taken as an old string anymore, but read in as a resloc.
So what will happen is you're going to get a resulting resloc of `minecraft:pickaxe` instead of `minecolonies:pickaxe`, which cause issues if attempting to read from registry.

Raycoms:
> Also, given we changed the NBT data of several things changing from name -> registryName, this can cause some failures in attempting to reload the data. For example the previous name used to be `pickaxe`, which is now not taken as an old string anymore, but read in as a resloc. So what will happen is you're going to get a resulting resloc of `minecraft:pickaxe` instead of `minecolonies:pickaxe`, which cause issues if attempting to read from registry.

As they're all ours, we can make sure it's minecolonies:pickaxe if nothing else is defined

Thodor12:
Yeah but we need to have migration methods for this, because right now they're being read in directly.

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/registry/IToolTypeRegistry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
```diff
@@ -129,45 +160,36 @@ public static void serialize(final IFactoryController controller, final Friendly
      */
     public static Tool deserialize(final IFactoryController controller, final FriendlyByteBuf buffer)
     {
-        final IToolType type = ToolType.getToolType(buffer.readUtf(32767));
+        ResourceLocation resLoc = ToolTypeEntry.parseResourceLocation(buffer.readResourceLocation());
+        final ToolTypeEntry type = ModToolTypes.getRegistry().getValue(resLoc);
```
Thodor12: You do not have to do this for network buffer deserialization, only for NBT deserialization.
NBT is permanent storage on disk, network buffers can only be used during gameplay, so the data is already corrected on NBT load and can't be wrong right here.

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
```diff
@@ -0,0 +1,231 @@
...
+
+import java.util.Set;
+import java.util.function.Consumer;
+
+/**
+ * Class used for storing and registering any ToolTypes.
+ */
+public class ModToolTypes
+{
+    public final static DeferredRegister<ToolTypeEntry> DEFERRED_REGISTER = DeferredRegister.create(new ResourceLocation(Constants.MOD_ID, "tooltypes"), Constants.MOD_ID);
```
Thodor12: Static final, not final static

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
```diff
@@ -0,0 +1,231 @@
...
+    public final static RegistryObject<ToolTypeEntry> shield;
+    public final static RegistryObject<ToolTypeEntry> helmet;
+    public final static RegistryObject<ToolTypeEntry> leggings;
+    public final static RegistryObject<ToolTypeEntry> chestplate;
+    public final static RegistryObject<ToolTypeEntry> boots;
+    public final static RegistryObject<ToolTypeEntry> flint_and_steel;
+
+    static
+    {
+        none =
```
Thodor12: What happened here?

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * Get the tooltype registry.
+     * @return The tooltype registry
+     */
+    public static IForgeRegistry<ToolTypeEntry> getRegistry()
+    {
+        return IMinecoloniesAPI.getInstance().getToolTypeRegistry();
+    }
+
+    private static RegistryObject<ToolTypeEntry> register(final String id, final Consumer<ToolTypeEntry.Builder> consumer)
```
Thodor12: Javadoc

src/main/java/com/minecolonies/api/tools/ModToolTypes.java
```diff
@@ -0,0 +1,231 @@
...
+                  .setToolLevel((itemStack, toolType) -> ModToolTypes.armorLevel(itemStack))
+                  .build());
+
+        flint_and_steel = register("flintandsteel",
+          builder -> builder.setDisplayName(Component.translatable(ToolTranslationConstants.TOOL_TYPE_LIGHTER))
+                  .setIsTool((itemStack, toolType) -> itemStack.getItem() instanceof FlintAndSteelItem)
+                  .setToolLevel((itemStack, toolType) -> durabilityBasedLevel(itemStack, Items.FLINT_AND_STEEL.getMaxDamage()))
+                  .build());
+    }
+    /**
```
Thodor12: Spacing

src/main/java/com/minecolonies/api/util/ItemStackUtils.java
```diff
@@ -267,15 +264,15 @@ else if (!(entity instanceof MobEntity))
...
     {
         if (isEmpty(stack))
         {
             return false;
         }
 
-        final int level = Compatibility.isTinkersWeapon(stack) ? Compatibility.getToolLevel(stack) : getMiningLevel(stack, toolType);
-        return isTool(stack, toolType) && verifyToolLevel(stack, level, minimalLevel, maximumLevel);
+        final int level = Compatibility.isTinkersWeapon(stack) ? Compatibility.getToolLevel(stack) : toolType.getMiningLevel(stack);
+        return toolType.checkIsTool(stack) && verifyToolLevel(stack, level, minimalLevel, maximumLevel);
```
Thodor12: I asked this before, but could this method be reduced to
```java
return toolType.checkIsTool(stack) && verifyToolLevel(stack, toolType.getMiningLevel(stack), minimalLevel, maximumLevel);
```
Leave the compatibility stuff to the tool type logic if applicable.

Thodor12 (commented):
Some last things

src/main/java/com/minecolonies/api/tools/registry/ToolTypeEntry.java
Thodor12: Still applies

Thodor12 (dismissed):
Finally looks completely OK for me

Thodor12 (dismissed):

Raycoms (dismissed):

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
Wmaxlees, line 1: Changing this file to "Equipment" breaks existing colonies. I guess it'll have to be changed when 1.21 is released?

Raycoms: That' fine then, ye

Thodor12 (approved):

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/Tool.java
someaddons: it should actually just break the requests and should auto-cleanup

Wmaxlees: Is the expectation that the game crashes one time, cleans up the request, and then works as normal after that? My test world hard crashed when I changed the name but I didn't try to reload it afterwards.

Raycoms: It shouldn't crash with the newest version anymore. But it is also possible to add an RS reset checkpoint in to allow renaming it.

someaddons: if it crashes send us the log, then we're missing some resilience in some places when trying to load bad data

Wmaxlees: We're trying to create a class from a serialized name so it crashes when that class doesn't exist.

Here's the offending line; https://github.com/ldtteam/minecolonies/blob/ff5c2208be8002e0022f7b87ffcb39ccf155792a/src/main/java/com/minecolonies/api/colony/requestsystem/factory/standard/TypeTokenFactory.java#L56

Raycoms (approved):

someaddons (approved):

PR #10218, opened 2024-09-14
Fix/warehouse issue

### LINKED ISSUE #10208: [BUG] The warehouse is still taking items from other warehouses to fulfill requests.

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.660

### Structurize Version

1.20.1-1.0.742

### Current Behavior

https://youtu.be/xSeGljUcBzw
The warehouse is still requesting items from other warehouses to deliver them elsewhere.
(Please only watch the beginning and the end of the video.)

### Expected Behavior

Ensure that the warehouse containing the requested item fulfills the request directly.

### Reproduction Steps

Build two warehouses and create multiple requests to test the behavior?

### Logs

No crash

### Anything else?

Additionally, I'm not sure if I remember correctly, but I seem to recall seeing in a previous update that warehouses now distribute tasks on their own rather than assigning everything to the nearest one. However, I sometimes still see one warehouse overloaded with tasks while the couriers from the other warehouse are idling. I'm not sure if this is normal, so I'm just mentioning it here. Since building two warehouses, the workload usually isn't high enough to keep the couriers too busy.

### Footer

---
## Viewers

Raycoms:
Closes #10208
Closes #

- Fix warehouse accepting task it has no items for
-

## Testing

Review please


src/main/java/com/minecolonies/core/tileentities/TileEntityWareHouse.java
```diff
@@ -91,6 +91,32 @@ public int getCountInWarehouse(@NotNull final ItemStorage storage, int count)
         return totalCount;
     }
 
+    @Override
+    public int getCountInWarehouse(@NotNull final Predicate<ItemStack> predicate, int count)
```
someaddons: why is this within the warehouse tileentity? Since it just forwards to the building we should use the building directly, also alleviating the need to load/loaded check on the TE to begin with

someaddons: could probably even use pre-existing generic building get count inventoryutils functions instead

src/main/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -791,7 +787,7 @@ public static int hasBuildingEnoughElseCount(@NotNull final IBuilding provider,
                     totalCount += ((TileEntityRack) entity).getItemCount(stack);
                 }
 
-                if (totalCount > count)
+                if (totalCount >= count)
```
someaddons, line 790: nice find

someaddons (dismissed):

PR #10238, opened 2024-09-20
Add crops to JEI

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/473575384445878273/1284603515045679166)

- Fixes an error parsing certain 1.20+ loot tables (causing some items to be missing in JEI).
- Changes the crop loot generation to use definitions from the crops themselves rather than duplicating values.  The drop probabilities should be the same as before.
    - Removed rice from sugar cane and added it to seagrass and small dripleaf.
        - Added "river" to humid biomes so that seagrass is actually available.  (It's also still a temperate biome too.)
    - Added cabbage appearing in dungeon loot chests.
- Prevent dropping crops if using shears or a silk touch tool.
- Adds all growable crops to the existing JEI farmer page (with some display caveats).
    image
    - For the MineColonies crops, hovering the "Biomes" label will show the biomes where this crop can be grown (and found):
        image
        - I did have this one per line originally, which was easier to read, but this quickly overflowed the screen with some crops; that would only get worse when adding modded biomes.
- Adds a "Colony Crops" page that displays how to obtain the MineColonies crops:
    image
    - This only displays the blocks that do have MineColonies drops, but it does currently display *all* drops from that block, including vanilla and from other mods.  I did consider filtering the display but this seemed ok as is, at least in mostly-vanilla.
    - It also assumes that all source blocks do have an item form.  I did originally make it render the block directly but that didn't look as good.
    - The hoe is not **required** for any of these drops, but it's shown anyway as a hint that it improves your chances.  The chance percentages shown are assuming that you do use the hoe.
    - The biomes are not directly visible here, but you can click on any item to see it in the Farmer's tab.

## Testing

Review please (should port)


Thodor12 (dismissed):

someaddons:
Since rice is more a watery crop, could we add it to sth thematically better fitting than grass?

uecasm:
Ray suggested adding grass to enable the hoe bonus -- he didn't want that enabled on sugar cane because it's too easy to cheese.

Rice is still restricted to humid biomes only (e.g. jungle).  Bamboo is probably the only other likely candidate, and that has the same problem as sugar cane.  Perhaps worse since sugar cane requires water and bamboo doesn't.

uecasm:
I've changed the rules for rice (and the river biome) as discussed in Discord, and excluded shears/silk touch.

src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
```diff
@@ -171,4 +175,21 @@ public Block getPreferredFarmland()
...
+    public List<Block> getDroppedFrom()
+    {
+        return droppedFrom;
+    }
+
+    /**
+     * Get the preferred biome for this crop.
+     * @return the preferred biome, or null if not picky.
+     */
+    public @Nullable TagKey<Biome> getPreferredBiome()
```
Raycoms: Nullable should be above public afaik

src/main/java/com/minecolonies/core/event/EventHandler.java
```diff
@@ -114,6 +118,18 @@ public class EventHandler
      */
     private static Map<UUID, ChunkPos> playerPositions = new HashMap<>();
 
+    /**
+     * Cache of loot table -> crops.
+     */
+    private static Map<ResourceLocation, List<MinecoloniesCropBlock>> cropDrops;
+
+    // stolen from BlockLootSubProvider
+    private static final LootItemCondition.Builder HAS_SILK_TOUCH = MatchTool.toolMatches(ItemPredicate.Builder.item().hasEnchantment(new EnchantmentPredicate(Enchantments.SILK_TOUCH, MinMaxBounds.Ints.atLeast(1))));
```
Raycoms: I think we could have those in some utils or constants place. Seems in general useful even outside of this context. I feel we use some of these already somewhere

Raycoms (dismissed):

Raycoms (approved):

marchermans (approved):

PR #10239, opened 2024-09-20
Fix colony teams assignment behaviour

Thodor12:
- Consistently ensure citizens are registered to the correct team
- Use the colony team name instead of a per citizen team
- Ensure that upon deletion of a colony it's respective team is removed
- Fix the colony deletion event not triggering for command based deletion

## Testing

Review please


src/main/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
```diff
@@ -233,6 +235,18 @@ public boolean isNoAi()
         return false;
     }
 
+    @Override
+    @Nullable
+    protected PlayerTeam getAssignedTeam()
+    {
+        final ICitizenColonyHandler citizenColonyHandler = getCitizenColonyHandler();
+        if (citizenColonyHandler == null || citizenColonyHandler.getColony() == null)
```
Raycoms, line 243: No method for the client side?

src/main/java/com/minecolonies/api/entity/mobs/AbstractEntityRaiderMob.java
```diff
@@ -732,38 +733,15 @@ public void initStatsFor(final double baseHealth, final double difficulty, final
...
-     *
-     * @return
-     */
-    protected String getTeamName()
+    @Override
+    @Nullable
+    protected PlayerTeam getAssignedTeam()
     {
-        return RAID_TEAM;
+        return Colony.checkOrCreateTeam(level, RAID_TEAM);
```
Raycoms: Move to util, this shouldn't be in the colony then

src/main/java/com/minecolonies/core/colony/ColonyManager.java
```diff
@@ -134,46 +139,54 @@ public void deleteColonyByDimension(final int id, final boolean canDestroy, fina
...
 
         if (world == null)
         {
-            Log.getLogger().warn("Deleting Colony " + id + " errored: World is Null");
+            Log.getLogger().warn("Deleting Colony {} errored: World is Null", id);
+            return;
+        }
+
+        final IColonyManagerCapability cap = world.getCapability(COLONY_MANAGER_CAP, null).resolve().orElse(null);
+        if (cap == null)
```
Raycoms: why move this all the way up there?

src/main/java/com/minecolonies/core/network/messages/server/colony/ColonyDeleteOwnMessage.java
```diff
@@ -53,18 +51,6 @@ public void onExecute(final NetworkEvent.Context ctxIn, final boolean isLogicalS
         {
             IColonyManager.getInstance().deleteColonyByDimension(colony.getID(), false, colony.getDimension());
             MessageUtils.format(MESSAGE_INFO_COLONY_DESTROY_SUCCESS).sendTo(player);
-
-            if (isLogicalServer)
-            {
-                try
-                {
-                    MinecraftForge.EVENT_BUS.post(new ColonyDeletedEvent(colony));
```
Raycoms, line 61: ?

src/main/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
Thodor12: This logic happens to work identically for the client, so this works just fine.

src/main/java/com/minecolonies/core/colony/ColonyManager.java
Thodor12: Most other methods have a fast-fail if the colony manager capability is missing. It's a bit strange if we first go and bother to remove all data from the colony, to then realize "oh we actually can't delete the colony".

src/main/java/com/minecolonies/core/network/messages/server/colony/ColonyDeleteOwnMessage.java
Thodor12: The event was moved to the ColonyManager deleteColony method. I noticed this event was only fired upon deletion of your colony through the townhall GUI, not from the delete command.

So I moved that event call in there in favor of here.

src/main/java/com/minecolonies/api/entity/citizen/AbstractEntityCitizen.java
Raycoms: Did we recently rework the citizenColonyHandler? I was sure it only worked on the server

Raycoms: the registerWithColony call wasn't made for the client side, e.g. getCitizenmanager.registerCivilian

src/main/java/com/minecolonies/core/colony/ColonyManager.java
Raycoms: This is a "can't happen" thing anyway

src/main/java/com/minecolonies/core/network/messages/server/colony/ColonyDeleteOwnMessage.java
Raycoms: kk

src/main/java/com/minecolonies/core/colony/ColonyManager.java
```diff
@@ -185,31 +198,44 @@ private void deleteColony(@Nullable final IColony iColony, final boolean canDest
...
             {
-                Log.getLogger().warn(MISSING_WORLD_CAP_MESSAGE);
-                return;
+                world.getScoreboard().removePlayerTeam(team);
+            }
+
+            // Send the deletion event
+            try
+            {
+                MinecraftForge.EVENT_BUS.post(new ColonyDeletedEvent(colony));
```
someaddons: we probably should have a util doing this for all events, so we can just do ColonyEvents.send(event)

Raycoms: @Thodor12

src/main/java/com/minecolonies/core/colony/ColonyManager.java
Raycoms: please undo re-org otherwise it's a pain to backport

someaddons:
Is there btw a reason you did remove the friendly fire setting?

src/main/java/com/minecolonies/api/events/ColonyEvents.java
```diff
@@ -0,0 +1,38 @@
...
+ * Event manager for all forge events.
+ */
+public class ColonyEvents
+{
+    /**
+     * Event triggered when a colony is being deleted.
+     *
+     * @param colony the colony in question.
+     */
+    public static void deleteColony(final IColony colony)
```
Raycoms, line 19: maybe we could make this sendSave a default method on the event? new ColonyDeletedEvent(colony).trigger();

Thodor12: We'd need our own root event class in that case, the event it gets is just a forge event.

someaddons (approved):

src/main/java/com/minecolonies/api/events/ColonyEvents.java
Raycoms: Can just be an interface slapped on it.

uecasm:
#10399 probably related

PR #10242, opened 2024-09-20
Various fixes and Improvements

someaddons:
- Fix accuracy research to better fit its name, it now increases the guards chance to critically hit
- The guard sleep less research is now called coffee, and moved slightly down in the tree, its effect value also got fixed it incorrectly defaulted to 5 resulting in a negative chance.
- Research effect default level values are now 1 instead of 5 when no levels are declared. So value 1 = researched, 0 = not researched by default.
- Research Tree choices now display a tooltip to clarify that an "OR" really can only choose one of the two
<br/><br/>
- Sleeping guards no longer constantly move their head, nor look at nearby players anymore
- Sleeping guards now sleep for the intended duration, they no longer get woken up by unstuck
<br/><br/>
- Improve Itemlookup performance when searching for a stack within a list
- Improve citizen Interact goal performance, it now uses our already existing custom lookat goal
<br/><br/>
- Fix Ship raids no longer spawning, being always overtaken by underwater raiders, now it is 20% for underwater raiders chance(except for underwater spawnpoints ofc)
- Fix entity registration check, forge field is unreliable as it gets reset on player visual tracking radius
- Fix visitor NPE when they have no home building
- Fix some enchanter scrolls saving the colony name to nbt
- Improve backup loading
- Fix sleeping citizens spawning in bad locations
- Improved refreshdimension entity movement
- Improved Spawnlocation util

## Testing

Review please


Thodor12:
Is it smart to be changing research effect ids, these are stored inside the research manager NBT data aren't they?

someaddons:
ye I'm just exchanging the effect a research contains, not the research itself

src/main/java/com/minecolonies/api/research/util/ResearchConstants.java
```diff
@@ -146,6 +146,7 @@ public final class ResearchConstants
     public static final ResourceLocation CRUSHING_11       = new ResourceLocation(Constants.MOD_ID, "effects/crushing11unlock");
     public static final ResourceLocation ARCHER_USE_ARROWS = new ResourceLocation(Constants.MOD_ID, "effects/consumearrowsunlock");
     public static final ResourceLocation KNIGHT_TAUNT      = new ResourceLocation(Constants.MOD_ID, "effects/knighttauntmobsunlock");
+    public static final ResourceLocation GUARD_CRIT = new ResourceLocation(Constants.MOD_ID, "effects/guardcrit");
```
Raycoms, line 149: formatting

src/main/java/com/minecolonies/core/datalistener/ResearchListener.java
```diff
@@ -128,10 +128,10 @@ else if(effectJson.has(RESEARCH_SUBTITLE_PROP) && effectJson.get(RESEARCH_SUBTIT
                         }
                     }
                 }
-                // If no levels are defined, assume will go from zero to max level as a single action, and store on/off.
                 else
                 {
-                    category.add(5f);
+                    // By default no levels defined = 1 enabled, 0 disabled
+                    category.add(1);
```
Raycoms, line 134: wonder if anything else was affected by this?

Raycoms (changes requested):
Tiny things

src/main/java/com/minecolonies/core/datalistener/ResearchListener.java
someaddons: wondered too and spend 30mins checking the usages, seems we just do > 0 at other places

uecasm: The default strength being 5 was intended for the building unlocks, because the strength specifies what level of building is unlocked.  So unless you added levels to all of those, they'd all break.

someaddons: added the levels^^

Raycoms (approved):

PR #10254, opened 2024-09-22
Create StorageBlock registry

Wmaxlees:
Introduces a new Registry for StorageBlocks that allows addon authors to introduce new storage blocks besides Racks that are usable within the colony. This should be a no-op for vanilla Minecolonies.

With these new registry entries, you define an interface for how the colony interacts with a particular storage block, how storage blocks are mapped to the particular interface, and whether these storage blocks should be added automatically to buildings when in the schematic.

## Testing

Review please


src/main/java/com/minecolonies/api/MinecoloniesAPIProxy.java
```diff
@@ -235,4 +236,6 @@ public IForgeRegistry<EquipmentTypeEntry> getEquipmentTypeRegistry()
     {
         return apiInstance.getEquipmentTypeRegistry();
     }
+
+    public IForgeRegistry<StorageBlockEntry> getStorageBlockRegistry() { return apiInstance.getStorageBlockRegistry(); }
```
Thodor12: Formatting (no one-line methods)

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+import com.minecolonies.api.colony.buildings.IBuilding;
+import com.minecolonies.api.crafting.ItemStorage;
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.items.IItemHandlerModifiable;
+
+import java.util.Map;
+import java.util.function.Predicate;
+
+public interface IStorageBlockInterface
```
Thodor12: Missing javadoc on the interface itself.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return Whether the storageblock should be included in building containers automatically.
+     */
+    boolean shouldAutomaticallyAdd(final IBuilding building);
+
+    /**
+     * Sets whether the block is part of a warehouse.
+     *
+     * @param inWarehouse Whether it's in a warehouse
+     */
+    void setInWarehouse(final boolean inWarehouse);
```
Thodor12: I feel like there should be a better way to do this, rather than leaking unnecessary methods for blocks stored in the warehouse.
There's a very specific implementation behind these methods that all storage blocks would have to follow, which makes me think this is an ill-suited method to exist here.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Gets the current upgrade level of the storageblock
+     *
+     * @return The current level
+     */
+    int getUpgradeLevel();
```
Thodor12: Same as above

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Upgrades the size of the storage, if applicable.
+     */
+    void increaseUpgradeLevel();
```
Thodor12: Same as above

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return Whether the storageblock is empty
+     */
+    boolean isEmpty();
+
+    /**
+     * Get the modifiable ItemHandler for the given storageblock
+     *
+     * @return the itemhandler
+     */
+    IItemHandlerModifiable getInventory();
```
Thodor12: I thought we were clear on "no Item handler return", split into alternative methods that are actually used.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Return whether the storageblock contains a matching item stack
+     *
+     * @param stack        The item type to compare
+     * @param count        The amount that must be present
+     * @param ignoreDamage Whether the items should have matching damage values
+     * @return Whether the storageblock contains the match
+     */
+    boolean hasItemStack(final ItemStack stack, final int count, final boolean ignoreDamage);
```
Thodor12: I'd provide some more overrides, because we also have implementations that would filter on NBT.
Additionally, give these methods a default implementation that fall back to the predicate, that way implementations only have to fullfill the predicate implementation.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return Whether the storageblock has any matches
+     */
+    boolean hasItemStack(final Predicate<ItemStack> predicate);
+
+    /**
+     * Sets the block position of the building this storage belongs to
+     *
+     * @param pos The position of the building
+     */
+    void setBuildingPos(final BlockPos pos);
```
Thodor12: Again, I feel like this is ill-suited for the container implementation to manage. (Because this directly ties into the warehouse stuff as well).

src/main/java/com/minecolonies/api/tileentities/storageblocks/ModStorageBlocks.java
```diff
@@ -0,0 +1,69 @@
...
+import com.minecolonies.core.tileentities.storageblocks.AbstractRackStorageBlockInterface;
+import com.minecolonies.core.tileentities.storageblocks.RackStorageBlockInterface;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.RegistryObject;
+
+import java.util.Optional;
+
+public final class ModStorageBlocks
```
Thodor12, line 24: No javadoc

src/main/java/com/minecolonies/api/tileentities/storageblocks/ModStorageBlocks.java
```diff
@@ -0,0 +1,69 @@
...
+{
+    public static final DeferredRegister<StorageBlockEntry> DEFERRED_REGISTER = DeferredRegister.create(new ResourceLocation(Constants.MOD_ID, "storageblocks"), Constants.MOD_ID);
+
+    public static final RegistryObject<StorageBlockEntry> storageBlockRack;
+    public static final RegistryObject<StorageBlockEntry> storageBlockAbstractRack;
```
Thodor12: An abstract rack isn't a thing, it's a root class for racks.
And to be fair, this root class is frankly highly redundant.

src/main/java/com/minecolonies/api/tileentities/storageblocks/registry/StorageBlockEntry.java
```diff
@@ -0,0 +1,128 @@
+package com.minecolonies.api.tileentities.storageblocks.registry;
+
+import com.minecolonies.api.tileentities.storageblocks.IStorageBlockInterface;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.entity.BlockEntity;
+
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public final class StorageBlockEntry {
```
Thodor12: Missing javadoc

src/main/java/com/minecolonies/api/tileentities/storageblocks/registry/StorageBlockEntry.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Returns the registry name for this entry
+     *
+     * @return The registry name
+     */
+    public ResourceLocation getRegistryName() {
+        return registryName;
+    }
+
+    public static class Builder {
```
Thodor12, line 79: Missing javadoc

src/main/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -6,8 +6,9 @@
 import com.minecolonies.api.colony.buildings.IBuilding;
```
Thodor12, line 6: Please undo this file, again this file contains a bunch of formatting changes and reorders a load of methods, making the diff unnecessarily difficult to read.
Previous time I didn't make a big deal of it, told you, but now it happened again, so it's not ok anymore.

src/main/java/com/minecolonies/core/client/gui/WindowHutAllInventory.java
```diff
@@ -125,23 +126,28 @@ private void locate(final Button button)
 
         for (BlockPos blockPos : containerList)
         {
-            final BlockEntity rack = Minecraft.getInstance().level.getBlockEntity(blockPos);
-            if (rack instanceof TileEntityRack)
+            final BlockEntity entity = Minecraft.getInstance().level.getBlockEntity(blockPos);
+            Optional<IStorageBlockInterface> storageInterface = ModStorageBlocks.getStorageBlockInterface(entity);
```
Thodor12: If you want to use optionals, I'd recommend actually using the optional chaining. Otherwise we prefer using nullables.
So make this
```java
int count = ModStorageBlocks.getStorageBlockInterface(entity).map(storageBlock -> storageBlock.getCount(storage.getItemStack(), storage.ignoreDamageValue(), false)).orElse(0);
```

src/main/java/com/minecolonies/core/client/gui/WindowHutAllInventory.java
```diff
@@ -201,21 +207,23 @@ private void updateResources()
...
-            if (rack instanceof TileEntityRack)
-            {
-                final Map<ItemStorage, Integer> rackStorage = ((TileEntityRack) rack).getAllContent();
+            final BlockEntity entity = world.getBlockEntity(blockPos);
+            Optional<IStorageBlockInterface> storageInterface = ModStorageBlocks.getStorageBlockInterface(entity);
+            if (storageInterface.isEmpty()) {
+                continue;
+            }
+
+            final Map<ItemStorage, Integer> storage = storageInterface.get().getAllContent();
```
Thodor12: Same point as above, either use nullables or use the optional chaining

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingContainer.java
```diff
@@ -169,14 +169,15 @@ public void registerBlockPosition(@NotNull final Block block, @NotNull final Blo
...
-            final BlockEntity entity = world.getBlockEntity(pos);
-            if (entity instanceof TileEntityRack rackEntity)
-            {
-                rackEntity.setBuildingPos(this.getID());
+            BlockEntity entity = world.getBlockEntity(pos);
+            Optional<IStorageBlockInterface> storageInterface = ModStorageBlocks.getStorageBlockInterface(entity);
+            if (storageInterface.isEmpty()) {
+                return;
             }
+            addContainerPosition(pos);
```
Thodor12, line 216: I'd say the building needs to contain a list of storage blocks itself, not a list of positions.

Thodor12 (changes requested):
I've already seen enough so far. I don't like the current implementation, requiring something to have a tile entity probably isn't the way to go.
On top of that, putting faith in the custom storage block to implement update calls to warehouse logic is an absolute no-go for me.
This needs a big refactor to even become viable.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
```diff
@@ -0,0 +1,117 @@
...
+
+public interface IStorageBlockInterface
+{
+    /**
+     * Whether the storageblock should be included in building containers automatically.
+     *
+     * @param building The building that is being constructed.
+     * @return Whether the storageblock should be included in building containers automatically.
+     */
+    boolean shouldAutomaticallyAdd(final IBuilding building);
```
Thodor12: Also this method feels a bit weird, this implies there's a manual way to add storage blocks.
I'd leave this to the match function inside the registry entry.

src/main/java/com/minecolonies/api/MinecoloniesAPIProxy.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Wmaxlees: As far as I can tell. This only flips a boolean `inWarehouse` which is never used anywhere. Am I missing something?

src/main/java/com/minecolonies/api/tileentities/storageblocks/registry/StorageBlockEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/util/InventoryUtils.java
Wmaxlees: Cleaned it up. There's not formatting changes anymore. Just code changes.

src/main/java/com/minecolonies/core/client/gui/WindowHutAllInventory.java
Wmaxlees: Switched to nullable vs optional chaining where appropriate.

src/main/java/com/minecolonies/core/client/gui/WindowHutAllInventory.java
Wmaxlees: Done.

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingContainer.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Wmaxlees: My intention is to have manual ways to add storage blocks for my addon using a scepter

src/main/java/com/minecolonies/api/tileentities/storageblocks/ModStorageBlocks.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tileentities/storageblocks/ModStorageBlocks.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tileentities/storageblocks/registry/StorageBlockEntry.java
Wmaxlees: Done.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Wmaxlees: Yeah. I think I'll actually move all of the logic to update the warehouse out of the racks. It seems like we should instead have an onChange event that the warehouse subscribes to for all of it's containers. And then the warehouse takes care of updating itself when it receives those update events. And then the owners of the storage blocks just need to worry about firing the onChange event when appropriate. WDYT?

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Wmaxlees: Hmm, I'm not convinced on this one. The upgrades will be very storage block dependent. For example, does an upgrade for Iron Chests mean we add new rows to the chests, like we currently do with Racks? Or do we not upgrade the chests at all and actually just move from Iron Chests -> Gold Chests -> Diamond Chests (and only allow them at specific building levels). For Storage Drawers, it probably makes the most sense to just base the number of upgrades a drawer can have based on the building level. For AE2, it'd probably something like the size of disks that are allowed or the number of connections to the AE network.

It seems like it will have to be defined by the owner of the addon.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Thodor12: Not quite, the implementation that handles this is in AbstractTileEntityRack, and it basically causes an update to every deliverable item to update their respective requests.

In another comment I suggested writing this to an abstract class, and I'd put it in there with a final method that can't be overridden.

So you have a method `setStackInSlot` which is implemented to first update the data in the rack, return a boolean if successfull.
If successfull, trigger an update to the request system with the same input stack.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Thodor12: What I'd do instead (because this is actually needed on the block side to set an external level), is give this one a default implementation of nothing.
That way the majority of the blocks won't be forced to implement this with nothing inside of it, since it's not something that the majority of blocks will use.

Alternatively, define a separate interface `IStorageBlockUpgradeable` which contains this method, then the logic is opt-in.

src/main/java/com/minecolonies/api/tileentities/storageblocks/IStorageBlockInterface.java
Thodor12: Then I'd handle this inside your matches function for the storage block. Just return false for any block unless it's added in the allow list.
Additionally you can add containers after registration by accessing the container list in the building.

src/main/java/com/minecolonies/core/client/gui/WindowHutAllInventory.java
```diff
@@ -116,32 +117,36 @@ private void locate(final Button button)
     {
         final int row = stackList.getListElementIndexByPane(button);
         final ItemStorage storage = allItems.get(row);
-        final Set<BlockPos> containerList = new HashSet<>(building.getContainerList());
-        containerList.add(building.getID());
+        final Set<AbstractStorageBlockInterface> containerList = new HashSet<>(building.getContainerList());
```
Thodor12: This is probably not going to be functional anymore.
The list is wrapped in a set, however the block does not have an equality check anymore.
I'd probably make the building its container list be a set itself and not a list. And implement a proper equality check in the AbstractStorageBlock,

Raycoms:
Stale. Please get in contact with us in discord to discuss this if still interested

PR #10271, opened 2024-09-25
Add configuration and datapack support for the food system

IchHabeHunger54:
Closes ldtteam/minecolonies-features#851

- Transforms the list of acceptable/satisfying/happy foods for the citizens into tags, allowing mods or packs to easily add more entries to these tags. This change was made with mods that add other complex foods, such as Farmer's Delight, in mind.
- Adds four block tags for cold/temperate/humid/dry biome crops, to complement the existing biome tags.
- Adds a config option that, if set to false, allows the player to plant the crops themselves. If this is enabled, the limitation for the Minecolonies-added farmland blocks is also removed. The config option defaults to true, keeping the current behavior.
- On the technical side, moves all the logic out of `ItemFood`, instead adding the `FoodUtils#getFoodTier` method. All checks that previously were `stack.getItem() instanceof IMinecoloniesFoodItem` now use `FoodUtils.getFoodTier(stack) > 0` instead.

This PR explicitly does NOT change any default behavior. There will be no gameplay changes compared to before, unless you toggle the config option or add a datapack modifying the aforementioned tags.

A breaking API change is introduced by removing the `IMinecoloniesFoodItem` interface.

## Testing

Review please


src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
```diff
@@ -61,19 +58,17 @@ public class MinecoloniesCropBlock extends AbstractBlockMinecolonies<Minecolonie
...
 
     private final ResourceLocation blockId;
-    private final TagKey<Biome>    preferredBiome;
 
     /**
      * Constructor to create a block of this type.
      * @param blockName the block id.
      */
-    public MinecoloniesCropBlock(final String blockName, final Block preferredFarmland, @Nullable final TagKey<Biome> preferredBiome)
+    public MinecoloniesCropBlock(final String blockName, final Block preferredFarmland)
```
Thodor12, line 66: Do not remove this, we have several PRs making use of this.

src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
```diff
@@ -126,7 +121,10 @@ public void attemptGrow(BlockState state, ServerLevel level, BlockPos pos)
     @Override
     public boolean canSurvive(@NotNull BlockState state, LevelReader level, @NotNull BlockPos pos)
     {
-        return (level.getRawBrightness(pos, 0) >= 8 || level.canSeeSky(pos)) && super.canSurvive(state, level, pos) && level.getBlockState(pos.below()).getBlock() == preferredFarmland && (preferredBiome == null || level.getBiome(pos).is(preferredBiome));
+        return (level.getRawBrightness(pos, 0) >= 8 || level.canSeeSky(pos)) 
+               && super.canSurvive(state, level, pos) 
+               && (!MineColonies.getConfig().getServer().limitCropsToFarmers.get() || level.getBlockState(pos.below()).getBlock() == preferredFarmland)
```
Thodor12, line 126: Why has this config option been introduced here? The config option states it's used to check if the crops may be planted by farmers only. This doesn't seem to be related to that.

Thodor12 (changes requested):

src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
IchHabeHunger54: Is there really a point in leaving it in, even as a no-op? The PRs would need to adjust to this anyway by using the tags instead of the ctor arguments to have it work.

src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
IchHabeHunger54: If the player is supposed to be able to place them, they would expect them to be placeable on normal farmland, otherwise they would be effectively gated by a farmer as otherwise they wouldn't be able to get the Minecolonies farmland otherwise.

I can change this into a separate config option though, if desired.

src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
Thodor12: The tags make it too complex, if you want to merely know the preferred biome* you're going to have to iterate all 4 tags to see which one it occurs in.
Not to mention this option allows crops to appear in several biome tags at the same time, which is not what this system was designed with in mind. (I.e. something should not appear in humid and dry at the same time, that's illogical)

src/main/java/com/minecolonies/core/blocks/MinecoloniesCropBlock.java
Thodor12: Then I'd probably omit this completely, I see no reason players have to be able to cultivate these crops themselves (nobody is going to be able to do anything with them anyway as the recipes are crafter locked anyway)

Raycoms:
Will be irrelevant with the upcoming changes

PR #10281, opened 2024-09-26
Fix colony creation

Thodor12:
Closes [discord report](https://discord.com/channels/139070364159311872/1003402487417548801/1288943006946299936) Colony cannot be created

# Changes proposed in this pull request
- Fix a nullpointer during load because of missing colony center value
- Load all manager classes mid-constructor for consistency
- Remove needless secondary constructor
- Expand remaining constructor with more arguments, and remove logic from loadColony()

## Testing
- [X] Yes I tested this before submitting it.
- [ ] I also did a multiplayer test.

Review please

PR #10282, opened 2024-09-26
Hotfix request manager initialization

Thodor12:
See #10281

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/Colony.java
+++ b/src/main/java/com/minecolonies/core/colony/Colony.java
@@ -201,7 +201,7 @@ public class Colony implements IColony
     /**
      * The request manager assigned to the colony.
      */
-    private final IRequestManager requestManager;
+    private IRequestManager requestManager;
 
     /**
      * The request manager assigned to the colony.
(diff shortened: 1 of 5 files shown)


someaddons (dismissed):

Thodor12:
Wait, found another bug

src/main/java/com/minecolonies/core/colony/Colony.java
```diff
@@ -1349,6 +1348,10 @@ public ServerLevel getWorld()
     @Override
     public IRequestManager getRequestManager()
     {
+        if (requestManager == null)
```
Raycoms, line 1352: Also not sure how guaranteed this was.

Raycoms: Like if it's accessed anyway midclass directly. Probably saver with the previous setup with method to set?

someaddons: the reason why we need to delay it is that during colony loading from nbt the manager had to be set later, which was done originally through a specific set call. The usages of the getter all seem to expect nonnull though, no one checks for it so should be fine :)

Thodor12: I only wonder if we do this if we shouldn't remove all direct usages of `requestManager` as those can potentially be null right now.

PR #10286, opened 2024-09-29
Colony expedition system

Thodor12:
- Expeditions are virtualized events that can be started from a new visitor that will appear at the townhall. You may talk to them and begin preparing for an expedition.
You will have to give them a given amount of goods and assign one or more guards to travel alongside them.
- Current and past expeditions can be viewed from the townhall, you will be able to see what you provided them, which guards were sent along, who died (if any), and everything they managed to find.
- The system is highly modifiable through the means of datapacks, from the expedition types to the loot tables.

The process and some screenshots.

You start of by getting visitors spawn in your town hall, there is no research or any kind of requirement for this to start happening. These can be identified by the map icon above their heads.

2025-07-17 14_45_38-Window

Talking to them will result in them asking if you want to go on an expedition, you can accept or deny.

2025-07-17 14_45_42-Window

You will then receive the expedition guide scroll, from here you can manage the inventory for the expedition as well as which guards are assigned.

2025-07-17 14_47_35-Window

After providing them with all the stuff, you return to the expeditionary, and attempt to start the expedition, if any requirement isn't met they will tell you, otherwise they will prompt to start the expedition.

2025-07-17 14_47_48-Window
2025-07-17 14_47_53-Window

Initially the expeditionary will stick around for a bit whilst they start packing up (in future release this will involve walking to the gate). If you attempt to talk to them at this time you'll receive the following prompt (does nothing).

2025-07-17 14_47_58-Window

When they return you can claim the loot from the expedition, and you can then see the past 5 expeditions in the townhall expeditions overview.

2025-07-17 15_19_58-Window

Review please


Gerpuss:
Вы должны это добавить, это имба

Thodor12:
Please do not comment on pull requests unnecessarily, and use English.

src/main/java/com/minecolonies/api/colony/jobs/IJob.java
```diff
@@ -284,13 +284,23 @@ default void triggerActivityChangeAction(boolean newState)
...
     {
         return false;
     }
 
+    /**
+     * Whether the job is a guard that actively engages in combat
+     *
+     * @return true if so.
+     */
+    default boolean isCombatGuard()
```
Raycoms: That is "non druid" ?

src/main/java/com/minecolonies/api/colony/managers/interfaces/expeditions/ColonyExpedition.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Get the expeditionary visitor that is leading this expedition.
+     *
+     * @return the member instance.
+     */
+    @NotNull
+    public ExpeditionVisitorMember getLeader()
+    {
+        for (final IExpeditionMember<?> member : members.values())
```
Raycoms: Would store the leader in a seperate dara structure tbh. Leader + list of members

src/main/java/com/minecolonies/api/colony/jobs/IJob.java
Thodor12: Pretty much, I use this on the view and the guard assignment to check if the given guard is "any combat guard". Indeed to prevent hiring druids into the expedition.

Raycoms: As we had talked, I feel like we should have a specific "map of guard/weapon" instead somewhere where we add the jobs manually instead of these two global methods.

src/main/java/com/minecolonies/api/colony/jobs/IJobView.java
```diff
@@ -34,6 +36,22 @@ public interface IJobView
...
 
+    /**
+     * Get whether this job is a guard job.
+     */
+    boolean isGuard();
+
+    /**
+     * Get whether this job is a guard job that actively engages in combat.
+     */
+    boolean isCombatGuard();
```
Raycoms: Instead of all these 3 even

src/main/java/com/minecolonies/api/entity/ModEntities.java
```diff
@@ -26,7 +27,9 @@ public class ModEntities
 {
     public static EntityType<? extends AbstractEntityCitizen> CITIZEN;
 
-    public static EntityType<? extends AbstractEntityCitizen> VISITOR;
+    public static EntityType<? extends AbstractEntityVisitor> VISITOR;
+
+    public static EntityType<? extends AbstractEntityVisitor> EXPEDITIONARY;
```
Raycoms, line 34: Unsure about this name. Adventurer? Explorer?

src/main/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickRateStateMachine.java
```diff
@@ -140,6 +140,9 @@ public void setTickRate(final int tickRate)
     @Override
     public void setCurrentDelay(final int ticksToNext)
     {
-        executedTransition.setTicksToUpdate(ticksToNext);
+        if (executedTransition != null)
```
Raycoms, line 144: This shouldn't be nullable. Even an inactive citizen should just run into a dummy transition imo.

src/main/java/com/minecolonies/api/equipment/ModEquipmentTypes.java
```diff
@@ -232,4 +233,25 @@ public static boolean canPerformDefaultActions(ItemStack itemStack, Set<ToolActi
...
+    {
+        return new EquipmentTypeEntry[] {
+          sword.get(),
+          bow.get(),
+          axe.get(),
+        };
+    }
+
+    @NotNull
+    public static EquipmentTypeEntry[] getAllArmor()
```
Raycoms: Isn't this just vanilla armor slot enum or so then?

src/main/java/com/minecolonies/api/equipment/ModEquipmentTypes.java
```diff
@@ -232,4 +233,25 @@ public static boolean canPerformDefaultActions(ItemStack itemStack, Set<ToolActi
         }
         return true;
     }
+
+    @NotNull
+    public static EquipmentTypeEntry[] getAllWeapons()
```
Raycoms: What is this for?

src/main/java/com/minecolonies/core/client/gui/blockui/RotatingItemIcon.java
```diff
@@ -55,8 +55,13 @@ public void setItems(@NotNull final List<ItemStack> items)
             throw new IllegalArgumentException("Items list must contain at least one item.");
         }
 
+        final boolean hasChanged = !this.items.equals(items);
```
Raycoms, line 58: can't we trigger the hasChanged on change actually more easily instead of checking for change?

src/main/java/com/minecolonies/core/client/gui/modules/building/WindowBuilderResModule.java
```diff
@@ -1,10 +1,8 @@
 package com.minecolonies.core.client.gui.modules;
 
-import com.ldtteam.blockui.Color;
```
Raycoms, line 3: Generic Comment, but I like that this is properly extracted into its own class logic

src/main/java/com/minecolonies/core/client/gui/townhall/WindowTownHallExpeditions.java
```diff
@@ -0,0 +1,433 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+
+import static com.minecolonies.api.util.constant.ExpeditionConstants.*;
+
+/**
+ * Gui for viewing past expeditions.
+ */
+public class WindowTownHallExpeditions extends AbstractWindowSkeleton implements ButtonHandler
```
Raycoms, line 40: For a good review I think we need these UIs with example screenshots in the PR description.
The UI code is very simple, but I think it's important to see how it all looks like to give proper feedback.

src/main/java/com/minecolonies/core/colony/events/ColonyExpeditionEvent.java
```diff
@@ -0,0 +1,761 @@
...
+        if (scaleEncounterSize)
+        {
+            amount *= expeditionType.difficulty().getMobEncounterMultiplier();
+        }
+
+        // Determine the mob type
+        MobType mobType = MobType.UNDEFINED;
+        try
+        {
+            final Entity entity = encounter.entityType().create(colony.getWorld());
```
Raycoms: I still think we shouldn't actually create the entity for this. As this is just a simulation we can just play pretend on the health/strength/etc of the entities we encounter.

Raycoms: and then discount their health in a virtual obj that represents "generic entity #5"

src/main/java/com/minecolonies/core/colony/events/ColonyExpeditionEvent.java
```diff
@@ -0,0 +1,761 @@
...
+    }
+
+    /**
+     * Damage all armor in an armor list.
+     *
+     * @param list    the armor list.
+     * @param damage  the amount of damage dealt.
+     * @param onBreak function called when an armor slot breaks.
+     */
+    private void damageArmor(final @NotNull ArmorList list, final float damage, final Consumer<EquipmentSlot> onBreak)
```
Raycoms, line 694: Generic Comment: Have you tested complete client quit and restart mid expedition?

src/main/java/com/minecolonies/core/colony/expeditions/colony/requirements/ColonyExpeditionEquipmentRequirement.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        @Override
+        public Predicate<ItemStack> getItemPredicate()
+        {
+            return equipmentType::checkIsEquipment;
+        }
+
+        @Override
+        public ItemStack getDefaultItemStack()
```
Raycoms, line 120: This seems very bad performance wise. What is this for?

src/main/java/com/minecolonies/core/colony/expeditions/encounters/ExpeditionEncounter.java
```diff
@@ -0,0 +1,31 @@
...
+ * @param id               The id of the encounter.
+ * @param entityType       The entity type of the encounter.
+ * @param damage           The damage this encounter will deal.
+ * @param reflectingDamage The damage this encounter will reflect upon itself after dealing damage.
+ * @param health           The health this encounter has.
+ * @param armor            The armor level this encounter has.
+ * @param lootTable        The loot table killing this encounter will give.
+ * @param xp               The experience killing this encounter will give.
+ */
+public record ExpeditionEncounter(ResourceLocation id, EntityType<?> entityType, float damage, float reflectingDamage, double health, int armor, ResourceLocation lootTable,
```
Raycoms, line 19: Yeah, this is pretty much already what we need for the encounter, the entity type could just be display only

Raycoms (changes requested):
Okay, so I'm half way through. I hopefully will do some more tomorrow.
I left a bunch of comments on the way.
Generally, the PR description needs a lot more info:
a) UIs
b) Description of expected input cost of expeditions
c) Expected outcome (death/successrate/rewards)
d) Generic description how an expedition works and how the player workflow looks like.

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    @Override
+    public boolean isTravelling(final int citizenId)
+    {
+        return Optional.ofNullable(travelerDataMap.get(citizenId)).map(TravelerData::isTraveling).orElse(false);
+    }
+
+    @Override
+    public Optional<BlockPos> getTravellingTargetFor(final int citizenId)
```
Raycoms, line 47: The Optional wrappings are usually a pain for handling things. I'd rather use nulls and if null do x. It avoids a lot of garbage collection of the Optional wrappers and easier to debug when things go wrong imo.

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
```diff
@@ -0,0 +1,343 @@
...
+            Optional<AbstractEntityCitizen> optionalEntityCitizen = citizenData.getEntity();
+            if (optionalEntityCitizen.isEmpty())
+            {
+                Log.getLogger().warn(String.format("The traveller %d from colony #%d has returned very confused!", citizenData.getId(), colony.getID()));
+                citizenData.setNextRespawnPosition(EntityUtils.getSpawnPoint(colony.getWorld(), spawnHutPos));
+                citizenData.updateEntityIfNecessary();
+                optionalEntityCitizen = citizenData.getEntity();
+            }
+
+            optionalEntityCitizen.ifPresent(abstractEntityCitizen -> TeleportHelper.teleportCitizen(abstractEntityCitizen, colony.getWorld(), spawnHutPos));
```
Raycoms, line 101: can't we just do that in an else branch then with direct access?

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
```diff
@@ -0,0 +1,343 @@
...
+        return data;
+    }
+
+    @Override
+    public void deserializeNBT(final CompoundTag nbt)
+    {
+        final ListTag travelerData = nbt.getList(NbtTagConstants.TRAVELER_DATA, Tag.TAG_COMPOUND);
+        travelerDataMap.clear();
+
+        travelerData
```
Raycoms: We should avoid stream serialization. For once it's a lot more expensive and called quite regularily, but it's also impossible to debug if things go wrong.

Raycoms: deser and ser

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        /**
+         * Get the current percentage of travel time.
+         *
+         * @return a percentage.
+         */
+        public double getTravelPercentage()
+        {
+            return ((double) remainingTravelTime * 100) / (double) initialTravelTime;
```
Raycoms, line 234: should be enough just to do * 100.0, no?

src/main/java/com/minecolonies/core/entity/ai/visitor/EntityAIExpeditionary.java
```diff
@@ -0,0 +1,82 @@
...
+            EntityNavigationUtils.walkToRandomPos(this.visitor, 10, DEFAULT_SPEED);
+        }
+
+        return VisitorState.WANDERING;
+    }
+
+    /**
+     * States of the expeditionary AI.
+     */
+    public enum VisitorState implements IState
```
Raycoms: I think this should probably be then ExpeditionaryState not VisitorState

src/main/java/com/minecolonies/core/entity/visitor/ExpeditionaryVisitorType.java
```diff
@@ -0,0 +1,183 @@
...
+        {
+            visitor.getColony().getVisitorManager().removeCivilian(visitor);
+
+            if (expeditionStatus == ExpeditionStatus.CREATED || expeditionStatus == ExpeditionStatus.ACCEPTED)
+            {
+                visitor.getColony().getExpeditionManager().removeCreatedExpedition(visitor.getId());
+            }
+        }
+
+        if (expeditionStatus == ExpeditionStatus.FINISHED && entity.get().getInventoryCitizen().isEmpty() && StreamSupport.stream(entity.get()
```
Raycoms: Not a fan of this stream iteration over the inventory. This is called a bit more often. I think a loop should  be better here.

src/main/java/com/minecolonies/core/network/messages/server/colony/visitor/expeditionary/TransferItemsMessage.java
```diff
@@ -0,0 +1,125 @@
...
+        final ExpeditionSheetContainerManager expeditionSheetContainerManager = new ExpeditionSheetContainerManager(itemInHand);
+
+        final ColonyExpeditionType colonyExpeditionType = ColonyExpeditionTypeListener.getExpeditionType(expeditionTypeId);
+        if (colonyExpeditionType == null)
+        {
+            Log.getLogger().warn("Transferring items for expedition failed, expedition type '{}' does not exist on the server side.", expeditionTypeId);
+            return;
+        }
+
+        final Optional<ColonyExpeditionRequirement> requirement = colonyExpeditionType.requirements().stream()
```
Raycoms: Also loop

Raycoms (changes requested):
Okay second and final review (got fully through now).
Second part looks pretty okay as well, mostly a lot of streams we should avoid with small number of values and where things can break.

A thing that I think is missing is that we have to avoid citizens from being fired from their job while on an expedition (like you can't fire a guard when on expedition).

src/main/java/com/minecolonies/api/equipment/ModEquipmentTypes.java
Thodor12: I need this to fetch the weapon that an expeditionary himself is carrying. So I check their inventory for each possible equipment type that is classified as a weapon. Instead of duplicating that logic, I've defined it once here.

src/main/java/com/minecolonies/core/client/gui/blockui/RotatingItemIcon.java
Thodor12: No, the point of this is because `setItems` is invoked upon each update. If you pass the exact same list of items in several times over, what is going to happen is that is resets the rotation state.
This will essentially mean that the icon is permanently showing the first icon.

That's what this new hasChanged check is for, to ensure if you pass the same list over and over, it doesn't wipe the current rotation state.

src/main/java/com/minecolonies/core/colony/events/ColonyExpeditionEvent.java
Thodor12: The reason I did this is so enchants can actually have appropriate effects based on the mob type. To make it a bit more realistic. I just temporarily create it and instantly discard them just to fetch the mob type (blame mojank for not allowing to fetch that via entity type).

I use this to calculate the damage for the weapon attack, i.e. if something has smite, it'll do more damage against any undead mob.

I could remove this and just always set the mob type to undefined, but that'll kill off some of the potential extra bonuses.

src/main/java/com/minecolonies/core/colony/expeditions/colony/requirements/ColonyExpeditionEquipmentRequirement.java
Thodor12: This is only utilized upon transferring the items into the expedition sheet when in creative. It'll apply a default item stack to insert given the user will have nothing in their inventory at the time.
(This is also mentioned in the interface javadoc what it does)

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
Thodor12: Ignoring because traveling manager has been implemented by Orion now

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
Thodor12: Ignoring because traveling manager has been implemented by Orion now

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
Thodor12: Ignoring because traveling manager has been implemented by Orion now

src/main/java/com/minecolonies/core/colony/managers/TravelingManager.java
Thodor12: Ignoring because traveling manager has been implemented by Orion now

src/main/java/com/minecolonies/core/colony/events/ColonyExpeditionEvent.java
Thodor12: Yes, it's able to properly continue mid way as all necessary pieces of information are stored in NBT of the event.

src/main/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickRateStateMachine.java
Thodor12: I think I got a nullpointer here at some point, so I added this as a precaution.

src/main/java/com/minecolonies/api/entity/ModEntities.java
Thodor12: I can change the display name of them, but rather not change in code anymore (too far in)

src/main/java/com/minecolonies/api/entity/ai/statemachine/tickratestatemachine/TickRateStateMachine.java
Raycoms: We should probably log this then as this shouldn't happen

src/main/java/com/minecolonies/api/equipment/ModEquipmentTypes.java
Raycoms: Atm, we only allow swords and bows for weapons, though right.

src/main/java/com/minecolonies/core/client/gui/blockui/RotatingItemIcon.java
Raycoms: Honestly, I'd just make sure the index doesn't exceed list size and otherwise don't reset it on setItems?

src/main/java/com/minecolonies/core/colony/events/ColonyExpeditionEvent.java
Raycoms: I think for the time being it's probably fine to not consider smite as extra damage. Maybe we could do some generic "enchanted sword damage bonus" and "enchanted armor dmg resistance bonus". We had many times int he past issues with mods doing stuff with entities on creation but not added to world. So this becomes a lot more simple and performant.

Raycoms (commented):
some more comments

src/main/java/com/minecolonies/api/colony/interactionhandling/AbstractInteractionResponseHandler.java
```diff
@@ -89,7 +90,7 @@ public Component getResponseResult(final Component response)
     }
 
     @Override
-    public List<Component> getPossibleResponses()
+    public List<Component> getPossibleResponses(final ICitizen data)
```
someaddons: Since the interactions reside within the citizendata, I think the better way is to make that information available to the handler itself, either through constructor or setter

src/main/java/com/minecolonies/api/colony/jobs/ModJobs.java
```diff
@@ -126,4 +127,13 @@ public static List<ResourceLocation> getJobs()
     {
         return jobs;
     }
+
+    @NotNull
+    public static JobEntry[] getExpeditionMembersList()
```
someaddons: I'd move this into one of the expedition respective classes like the manager

src/main/java/com/minecolonies/api/util/constant/TranslationConstants.java
```diff
@@ -874,6 +874,13 @@ public final class TranslationConstants
 
     //</editor-fold>
 
+    //<editor-fold desc="Interaction texts">
+
+    @NonNls
+    public static final String INTERACTION_EXPEDITIONARY = "com.minecolonies.coremod.interaction.visitor.expeditionary";
```
someaddons: I'd put this together into the expeditions constants file, or is there any need to split it off?

src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
```diff
@@ -360,5 +363,8 @@ public static void init()
 
         InteractionValidatorRegistry.registerStandardPredicate(Component.translatable(QUARRY_MINER_FINISHED_QUARRY),
           citizen -> citizen.getJob() instanceof JobQuarrier &&  ((JobQuarrier) citizen.getJob()).findQuarry() != null && ((JobQuarrier) citizen.getJob()).findQuarry().getFirstModuleOccurance(QuarryModule.class).isFinished());
+
+        InteractionValidatorRegistry.registerStandardPredicate(Component.translatable(EXPEDITION_INTERACTION_VALIDATOR_ID),
```
someaddons: why is this needed? I'd expect to only ever trigger an expedition interaction on the right type, not on the wrong type and then needing a validator to clean it up

PR #10304, opened 2024-10-03
Use fake level for JEI

### LINKED ISSUE #10302: Memory Leak of Client Level

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.719

### Structurize Version

1.0.752

JEI - 19.19.3.226

### Current Behavior

To add JEI Compat, Minecolonies instantiate Citizens with the current real ClientLevel, causing it to hold forever.

image

### Expected Behavior

Maybe implement some sort of fake level or grab the level/instantiate lazily? dunno.

### Reproduction Steps

1. Load world with JEI in the mod list
2. Change dimensions
3. Check memory for active ClientLevels, you might see at least 2. the old and the current.

### Logs

check image

### Anything else?

_No response_

### Footer

---
## Viewers

uecasm:
Closes #10302 (once ported)

- Use a fake level when creating citizens for JEI

## Testing

Review please (should port)


Thodor12 (dismissed):

Nightenom (changes requested):
There is technically breaking change logic-wise in latest fakelevel backport, so please update
Use singleblockfakelevel, probably with blockstate of air 
Don't forget to override getRealLevel as redirect to mc.instance.level

Nightenom (dismissed):

Raycoms (dismissed):

Nightenom (dismissed):

src/main/java/com/minecolonies/core/compatibility/jei/JeiFakeLevel.java
```diff
@@ -0,0 +1,24 @@
+package com.minecolonies.core.compatibility.jei;
+
+import com.ldtteam.structurize.client.fakelevel.SingleBlockFakeLevel;
+import net.minecraft.client.Minecraft;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.Blocks;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+
+@OnlyIn(Dist.CLIENT)
```
someaddons, line 10: why onlyin? dont think we should use that in 1.21

pietro-lopes:
Please, check how you are instancing this LivingEntity if you would need to use a fake level here too.

https://github.com/ldtteam/minecolonies/blob/7066c5cc40296ced6aa53d24cdbe11f3040d1164/src/main/java/com/minecolonies/api/crafting/GenericRecipe.java#L96

This is where it starts:
https://github.com/ldtteam/minecolonies/blob/7066c5cc40296ced6aa53d24cdbe11f3040d1164/src/main/java/com/minecolonies/core/compatibility/jei/JEIPlugin.java#L131

src/main/java/com/minecolonies/core/compatibility/jei/JeiFakeLevel.java
uecasm: Paranoia, mostly.  Since it's using client-only classes it must not be loaded in the server jar.  Though that should be the case even without this, it seems safer to make it explicit.

uecasm:
> Please, check how you are instancing this LivingEntity if you would need to use a fake level here too.

Good catch, thanks.

src/main/resources/META-INF/accesstransformer.cfg
```diff
@@ -57,6 +57,7 @@ public net.minecraft.client.multiplayer.ClientAdvancements f_104390_ # advanceme
 public net.minecraft.world.level.block.DoorBlock m_245755_(Lnet/minecraft/world/entity/Entity;Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Z)V # playSound
 
 public net.minecraft.world.entity.Entity f_19816_ # eyeHeight
+public net.minecraft.world.entity.Entity m_284535_(Lnet/minecraft/world/level/Level;)V # setLevel
```
Nightenom: Plain no, unlike BE entity level is notnull, please find another way

Nightenom (dismissed):

src/main/resources/META-INF/accesstransformer.cfg
uecasm: There's no harm in it, I tested it.  The entity is never added to any world, it only exists for the renderer and it does get a level set while that happens.

Nightenom: Not even BE has this method nullable (few vanilla BEs), so at least try catch it. Plus I really don't like it, unlike BE entities simply have non-null level contract..

Nightenom (approved):

Nightenom:
Hmm, out of scope of this PR: but why do we pass instantiated entity in the first place?

uecasm:
Because (a) that's what the entity renderer wants in order to draw an entity, and (b) other places need to query various instance methods that only exist on the instantiated entity.

pietro-lopes:
Honestly, if possible, I prefer WeakReference instead of cached by *magic number 2 minutes*

uecasm:
Minecraft GCs all the time, so a WeakReference is almost identical to not caching it at all.  And entities are not sufficiently lightweight for that to seem justifiable.

Raycoms (approved):

someaddons (approved):

PR #10306, opened 2024-10-03
resilience

Raycoms:
Closes #
Closes #

- Remove manual extra rotation
- Add a (debatably very hacky) crash prevention for ATM packs where banner pattern weirdly disappear
- Log somethings in debug to prevent startup log spam

## Testing
- [ x ] Yes I tested this before submitting it.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingContainer.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingContainer.java
@@ -159,7 +159,7 @@ public abstract class AbstractBuildingContainer extends AbstractSchematicProvide
             if (entity instanceof TileEntityColonyBuilding buildingEntity)
             {
                 buildingEntity.setStructurePack(StructurePacks.getStructurePack(getStructurePack()));
-                buildingEntity.setRotationMirror(getRotationMirror());
+               
                 final IBuilding building = colony.getBuildingManager().getBuilding(pos);
                 if (building != null)
                 {
--- a/src/main/java/com/minecolonies/core/colony/crafting/ItemStorageFactory.java
+++ b/src/main/java/com/minecolonies/core/colony/crafting/ItemStorageFactory.java
@@ -9,6 +9,7 @@ import com.minecolonies.api.util.Utils;
 import com.minecolonies.api.util.constant.SerializationIdentifierConstants;
 import com.minecolonies.api.util.constant.TypeConstants;
 import net.minecraft.core.HolderLookup;
+import net.minecraft.core.component.DataComponents;
 import net.minecraft.network.RegistryFriendlyByteBuf;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.nbt.CompoundTag;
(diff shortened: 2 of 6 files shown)


src/main/java/com/minecolonies/core/colony/crafting/ItemStorageFactory.java
```diff
@@ -90,7 +91,31 @@ public ItemStorage deserialize(@NotNull final HolderLookup.Provider provider, @N
...
+        // Some mods do weird things with banner patterns atm, so we add some special backup functionality here.
+        // We might have to extend this to other types of items in the future.
+        if (input.getItemStack().has(DataComponents.BANNER_PATTERNS))
+        {
+            boolean success = false;
+            try
+            {
+                Utils.serializeCodecMess(ItemStack.OPTIONAL_CODEC, packetBuffer.registryAccess(), input.getItemStack());
+                success = true;
+                Utils.serializeCodecMess(packetBuffer, input.getItemStack());
```
someaddons: this seems odd, why are we doing two serializations incase it works?

Raycoms: one to nbt, just to check if the stack serializes, then to buf, because we can't undo buf.

someaddons: weird logic structure overall, why the boolean at all? If the actual serialize to buffer call fails then it wont write an empty stack either?

Raycoms: if the nbt one fails, the buf one will also fail, inversely if the nbt one passes the buf one should also pass. It's obviously not guaranteed, but pretty much the best we can do

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingContainer.java
```diff
@@ -159,7 +159,7 @@ public void registerBlockPosition(@NotNull final Block block, @NotNull final Blo
             if (entity instanceof TileEntityColonyBuilding buildingEntity)
             {
                 buildingEntity.setStructurePack(StructurePacks.getStructurePack(getStructurePack()));
-                buildingEntity.setRotationMirror(getRotationMirror());
```
Nightenom, line 162: you sure this works for now? :D

Raycoms: Until we do the full convoluted 3d chess checking 2 old blueprints kind of thing yes.

src/main/java/com/minecolonies/core/colony/crafting/ItemStorageFactory.java
MotionlessTrain: Could the empty stack serialisation not be moved into the catch () block instead?

Raycoms: I didn't want to write into it, if the buf partially wrote into it.

marchermans: You can store the write index before writing, and then if failed reset to that, and wrote something new, overwriting the broken data

someaddons: ^^ was about to say that too, save the writer index and reset to it

someaddons (approved):

PR #10331, opened 2024-10-13
Concrete mixer can do multiple things before going back to the hut

### LINKED ISSUE #10209: The concrete mixer's movements seem strange while working.

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.660

### Structurize Version

1.20.1-1.0.742

### Current Behavior

https://youtu.be/KQk7zUV4bZo
The concrete mixer, after placing or breaking a block, walks back to their hut block right at the last moment (I later confirmed this by setting Stamina to the minimum). This behavior makes their work extremely slow. In my style, they take about 30 seconds to produce one concrete, which means that on a server with no sleep and night acceleration, it takes roughly an hour to produce one batch of concrete.

I remember occasionally seeing them standing still while placing or breaking blocks normally, so this should be a bug. Around 99% of the time, they exhibit the strange behavior of walking back to their hut block.

### Expected Behavior

Placing or breaking blocks normally

### Reproduction Steps

Build a Concrete Mixer’s Hut and request some concrete.

### Logs

No crash

### Anything else?

_No response_

### Footer

---
## Viewers

Thodor12:
Closes #10209

- Introduced 2 new AI stages for placing and harvesting, so that the AI may be kept in a placing/harvesting stage until everything is done at once.

## Testing

Review please


src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingConcreteMixer.java
```diff
@@ -154,6 +146,21 @@ public int getMaxBuildingLevel()
...
+     * Get the max amount of concrete placed at once.
+     *
+     * @return the number of concrete.
+     */
+    public int getMaxConcretePlaced()
+    {
+        int size = 0;
+        for (List<BlockPos> positions : waterPos.values())
+        {
+            size += positions.size();
```
Raycoms, line 159: huh?

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingConcreteMixer.java
```diff
@@ -184,18 +192,18 @@ public BlockPos getBlockToMine()
...
     public BlockPos getBlockToPlace()
     {
-        for (int i = 1; i <= minWaterLevel; i++)
+        for (int i = 1; i <= WATER_DEPTH_SUPPORT; i++)
         {
             for (final BlockPos pos : waterPos.getOrDefault(i, Collections.emptyList()))
             {
                 final BlockState state = colony.getWorld().getBlockState(pos);
-                if (!state.getFluidState().isEmpty() && state.getBlock() == Blocks.WATER)
+                if (state.is(Blocks.WATER))
```
Raycoms, line 200: flowing water is not water though

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingConcreteMixer.java
Thodor12: The positions stored in the map contain the depth as the key, and the positions of the water at the given level.

By the way the code looks it's technically allowed to have several water streams at the same time, there's no practical limit on this.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingConcreteMixer.java
Thodor12: Well I tested this, and it is.

If a block contains flowing water the block is still marked as containing water, the fluid state however just lists a different water depth.
There's no need for a depth check because the waterPos already has determined the required water level should be good in the schematic.

Raycoms: Please double check this. Because at least in some cases as far as I am aware, if air gets some low level water into it, it's fluidstate not empty, but actual state is not water.

Raycoms: ping

src/main/java/com/minecolonies/core/entity/ai/workers/crafting/EntityAIConcreteMixer.java
```diff
@@ -42,136 +44,156 @@ public class EntityAIConcreteMixer extends AbstractEntityAICrafting<JobConcreteM
     public EntityAIConcreteMixer(@NotNull final JobConcreteMixer job)
     {
         super(job);
+        super.registerTargets(
+          new AITarget(CONCRETE_MIXER_PLACING, this::placePowder, TICKS_SECOND),
+          new AITarget(CONCRETE_MIXER_HARVESTING, this::harvestConcrete, TICKS_SECOND)
+        );
     }
 
-    @Override
```
Raycoms, line 47: the reogs make this class again very difficult to review

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingConcreteMixer.java
Thodor12: Even if that is the case with low level water, then it's a non issue, because we only support up to a max depth of 5, and it worked correctly for every depth we support.

src/main/java/com/minecolonies/core/entity/ai/workers/crafting/EntityAIConcreteMixer.java
Thodor12: It still may look a little messy, but there's actually no reorg going on. The methods just changed that much.

Thodor12: There are 3 new functions up top, `placePowder`, `harvestConcrete` and `getSlotWithPowder`.
The `decide` task was not changed, only it's commented out code removed.
The `craft` task was changed quite a bit, because it's logic was split up into the place and harvest methods up top.
One new method at the bottom, `performMixingWork`, and also added `getMaxActionsUntilDumping` here.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingConcreteMixer.java
Thodor12: For reference I had a debugger out, and every level it's blockstate was set to being the actual block of water, disregarding how high the water actually was

Thodor12: I've double checked it again and it works fine. On top of that we already calculate the actual positions during construction, so I need not worry about calculating the depth of the water because the position itself is calculated to have water in it.
So that means we know the position is valid, so as long as there's then some water in it again, we're fine.

Raycoms (approved):

someaddons (approved):

PR #10339, opened 2024-10-17
Remove free blocks from configuration (has become redundant)

Thodor12:
- Free blocks and positions become datapacks in favor of configuration values

## Testing

Review please


src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -144,16 +142,6 @@ public class CompatibilityManager implements ICompatibilityManager
      */
     private static ImmutableList<ItemStack> allItems = ImmutableList.of();
 
-    /**
```
Raycoms, line 147: Thinking about this, maybe we scrap this setting entirely? We already have a tag for the blocks, so the freeblock setting is actually redundant. And a global free pos setting seems kinda useless considering people can already set this per colony anyway.

Thodor12: Find by me, I can delete it all

Raycoms (approved):
-75 amazing =D

someaddons (approved):

PR #10345, opened 2024-10-19
Fix computation of the mean factor for happiness

### LINKED ISSUE #10320: New Citizens spawning with all 1 in stats (including happiness)

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

minecolonies-1.20.1-1.1.684-snapshot

### Structurize Version

structurize-1.20.1-1.0.760-snapshot

_No response_

### Current Behavior

Start colony, wait for citizens to spawn. The first one spawns in with assorted stats, the next three citizens all spawn in with 1's across the board.
I loaded up an old version of the mod to double check I wasn't misremembering how it worked, and in older versions all four initial citizens start with random stats

### Expected Behavior

All four citizens get random stats

### Reproduction Steps

Start colony, wait for citizens to spawn, see stats

### Logs

https://gist.github.com/Talyda/8fbd2c2f261491ecd3c5a3257287cb75

### Anything else?

Log's from old version when it did work as expected, I can grab a new one if you think it will show anything relevant

### Footer

---
## Viewers

Axl-Lvy:
Closes #10320 

- Fix the computation of the mean: a 1 - factors does influence the mean
- Add a small optimizeation that bypass the factor computation (complexe getter) if the weight is 0.0 (simple getter)

## Testing
- [x] Yes I tested this before submitting it. Happiness of the very first citizen is now 4.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenHappinessHandler.java
+++ b/src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenHappinessHandler.java
@@ -141,11 +141,12 @@ public class CitizenHappinessHandler implements ICitizenHappinessHandler
             double totalWeight = 0.0;
             for (final IHappinessModifier happinessModifier : happinessFactors.values())
             {
-                final double factor = happinessModifier.getFactor(citizenData);
-                if (factor == 1.0)
+                final double weight = happinessModifier.getWeight();
+                if (weight == 0.0)
                     continue;
-                total += factor * happinessModifier.getWeight();
-                totalWeight += happinessModifier.getWeight();
+                final double factor = happinessModifier.getFactor(citizenData);
+                total += factor * weight;
+                totalWeight += weight;
             }
 
             final double happinessResult = (total / totalWeight) * (1 + colony.getResearchManager().getResearchEffects().getEffectStrength(HAPPINESS));


Axl-Lvy:
I'm a new contributor here and I'm not at all familiar with modding, but I'm motivated to learn. I'm a bit surprised that there is no test suite. I believe simple unit tests could help prevent these kinds of regressions.

src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenHappinessHandler.java
```diff
@@ -141,11 +141,12 @@ public double getHappiness(final IColony colony, final ICitizenData citizenData)
             double totalWeight = 0.0;
             for (final IHappinessModifier happinessModifier : happinessFactors.values())
             {
-                final double factor = happinessModifier.getFactor(citizenData);
-                if (factor == 1.0)
+                final double weight = happinessModifier.getWeight();
+                if (weight == 0.0)
```
Raycoms, line 145: This was to filter out neutral factors (at 1.0) not to make it a mean. This is atm an average.

MotionlessTrain: Though a (weighted) factor of 1 should also be part of the average, otherwise the weights don't add up anymore in a good way

Axl-Lvy: (Sum(w\*f)+Sum(w1\*1))/(Sum(w)+Sum(w1)) is not equal to Sum(w\*f)/Sum(w)

Raycoms: > Though a (weighted) factor of 1 should also be part of the average, otherwise the weights don't add up anymore in a good way

We ignore them from the total weight as well, so it's fine. These are factors that "do currently not apply". They are removed intentionally. That could be something like "citizens are not unhappy with food because they are in a low level building" but this doesn't mean they're happy with it. They are just not unhappy with it.

MotionlessTrain: Are there enough factors initially to have a happiness at all? As currently, the very first citizens seem to spawn in with a happiness of 1. If everything starts out at default, then everything is stripped out until there is not much left to calculate happiness with, leaving it at a wrong value

Raycoms: Is this also a problem in 1.20.1

Raycoms: This is a recent problem, that was not a problem before.

Talyda: Yeah, I tested it in 1.20.1
Issue here https://github.com/ldtteam/minecolonies/issues/10320

Talyda: And yeah, it is a semi recent problem

Nightenom:
> I'm a new contributor here and I'm not at all familiar with modding, but I'm motivated to learn. I'm a bit surprised that there is no test suite. I believe simple unit tests could help prevent these kinds of regressions.

Regarding test suites, JUnits for non-minecraft related code were always doable, but units testing of vanilla related code is relatively new thing (cca year iirc), so that's why

src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenHappinessHandler.java
Axl-Lvy: Ignoring them is I think the problem. A factor that doesn't influence the total is not a 1.-factor, but a factor with a weight of 0. A factor of 1. does influence the weighted average to tend to 1.
I'll be happy to try implementing this new behavior if you also think it's a good pattern.

Raycoms: The way things were coded atm is that factors 1.0 only happens for neutral factors. E.g. mystical site non existent, school factors for non children, etc. Otherwise factors are always either > or < 1

If we reworked this, we'd add a special "don't apply" flag which is independent of factor and weight then.

PR #10357, opened 2024-10-23
Pathing debug additions

someaddons:
- Reworked debug tracking, now also allows tracking path jobs by type and can add players to a specific pathresult for direct tracking

## Testing
- [x ] Yes I tested this before submitting it.

Review please


src/main/java/com/minecolonies/core/commands/EntryPoint.java
```diff
@@ -92,7 +93,7 @@ public static void register(final CommandDispatcher<CommandSourceStack> dispatch
           .addNode(new CommandBackup().build())
           .addNode(new CommandResetPlayerSupplies().build())
           .addNode(new CommandHelp().build())
-          .addNode(new ScanCommand().build())
+                                               .addNode(ScanCommand.build())
           .addNode(new CommandPruneWorld().build());
```
Raycoms: We should probably adjust the formatter to not make those huge gaps, makes reading harder =D

someaddons: ye formatter is being stupid :D

Raycoms (approved):

PR #10367, opened 2024-10-27
Prevent the cookery from keeping all empty bottles

### LINKED ISSUE #10360: Cookery not requesting pickups for multiple stacks of empty bottles

- [ ] Single Player
- [X] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.696

### Structurize Version

1.20.1-1.0.760

Raid Server Competition

### Current Behavior

Currently, I have about six full stacks of the large bottles in circulation in the colony. Once they are filled with milk, they go to the cookery to make cheese for the min-stocked foods in the restaurant. Then the empty bottles just stay there in the cookery's racks. I have never seen the cookery request a pickup for them, they just sit until I notice and go get them and put them back in the warehouse. 

### Expected Behavior

I would hope that the cookery, once it had half a stack or more of empty bottles, would request a pickup so those bottles could be refilled. Glass is expensive, and while I note, and appreciate, that you guys have tried to make the mod-specific bottles cheaper, six stacks is still a lot of glass. And... honestly I'm not sure exactly how many empty stacks it would take to force a pickup request. 

### Reproduction Steps

Make glass bottles, couriers deliver them, they never leave cookery.

### Logs

https://gist.github.com/Estyria/d2ded8f437cb4bb20edf7f125e4d2ebb

### Anything else?

Thodor asked me to make an issue.

### Footer

---
## Viewers

Thodor12:
Closes #10360 

- The cookery was holding onto bottles because secondary outputs in crafters are never tracked, so they will remain in the inventory of the crafter forever
- ~~Extended pickup requests to supply an optional filter which the courier can utilize to see what items are allowed to be picked up.
This way crafters can tell couriers to pick up all secondary outputs of the crafter.~~
- At the end of a crafting request, a delivery request will be made with all the secondary outputs that were generated during the crafting.

## Testing

Review please


marchermans (dismissed):

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/deliveryman/Pickup.java
```diff
@@ -5,58 +5,137 @@
...
      * Set of type tokens belonging to this class.
      */
     private final static Set<TypeToken<?>>
       TYPE_TOKENS = ReflectionUtils.getSuperClasses(TypeToken.of(Pickup.class)).stream().filter(type -> !type.equals(TypeConstants.OBJECT)).collect(Collectors.toSet());
 
+    /**
+     * The itemstack to filter for this specific pickup.
+     */
+    @Nullable
+    private List<ItemStack> pickupFilter;
```
Raycoms: We should just initiate this with an empty list to simplify the rest of the logic.

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/deliveryman/Pickup.java
```diff
@@ -5,58 +5,137 @@
...
+     * @param priority the priority of the request.
+     * @param filter   a filter to apply to the pickup request.
      */
-    public Pickup(final int priority)
+    public Pickup(final int priority, final @Nullable List<ItemStack> filter)
     {
         super(priority);
+        if (filter != null)
+        {
+            this.pickupFilter = filter.stream().map(ItemStack::copy).collect(Collectors.toList());
```
Raycoms: we can just use empty list instead and then avoid the nullable. Why do we need a deep copy?

src/main/java/com/minecolonies/core/entity/ai/workers/crafting/AbstractEntityAICrafting.java
```diff
@@ -464,6 +464,11 @@ protected IAIState craft()
                         module.improveRecipe(currentRecipeStorage, job.getCraftCounter(), worker.getCitizenData());
                     }
 
+                    if (!currentRecipeStorage.getSecondaryOutputs().isEmpty())
+                    {
+                        building.createPickupRequest(10, currentRecipeStorage.getSecondaryOutputs());
```
Raycoms: This indicates that it's not a pickup filter, but instead it's sort of a specific pickup request".

src/main/java/com/minecolonies/core/entity/ai/workers/service/EntityAIWorkDeliveryman.java
```diff
@@ -241,20 +243,49 @@ private boolean pickupFromBuilding(@NotNull final IBuilding building)
...
+            final ListIterator<ItemStack> iterator = request.getRequest().getPickupFilter().listIterator();
+            while (iterator.hasNext())
+            {
+                final ItemStack filter = iterator.next();
+                if (ItemStackUtils.compareItemStacksIgnoreStackSize(stack, filter))
+                {
+                    final ItemStack activeStack = handler.extractItem(currentSlot, filter.getCount(), false);
+                    transferItemForPickup(activeStack);
+                    if (filter.getCount() == activeStack.getCount())
+                    {
```
Raycoms: Feels weird to be iterator based. Can't we just make this a if pickupList.contains(new ItemStorage(stackInCurrentSlot)?

Thodor12: The problem occurs when the items are split across several inventory slots, there's a low chance that the items are held in one and the same inventory slot. This one keeps iterating the slots until the request is completely fulfilled.

src/main/java/com/minecolonies/core/entity/ai/workers/crafting/AbstractEntityAICrafting.java
```diff
@@ -556,7 +562,18 @@ protected IAIState craft()
                     return START_WORKING;
                 }
                 recordCraftingBuildingStats(currentRequest, currentRecipeStorage);
+
                 currentRequest.addDelivery(currentRecipeStorage.getPrimaryOutput());
+
+                // If the primary output is empty, it is not returned as part of the result list
+                // Hence, if the primary output is empty, we can start iterating from index 0
+                // If it's not empty, it is added to the beginning of the stacks list, so we start iterating from index 1
+                for (int i = ItemStackUtils.isEmpty(currentRecipeStorage.getPrimaryOutput()) ? 0 : 1; i < addedStacks.size(); i++)
```
Raycoms: Why can't we just do currentRecipestorage.getSecondaryOutputs?

uecasm: That doesn't include the items generated by loot tables.

Thodor12: It was slightly finnicky, it were easier if I wouldn't have to again filter out the primary output, but this was the easiest way I could tell to do it

Raycoms: It is a bit hacky, yeah, what if we get a list of primary outputs or so, then it gets weird

Thodor12: We can't, it's always 1 item, and if it were to change we'd have to refactor much more

Raycoms: This specific method also makes things like buildtools, and general kitchen utensils and stuff return to the warehouse too right?

Thodor12: No, we specifically checked that. This method only grabs the primary output, secondary outputs and the result of the loot table. Tools and crafting tools are not included in here.

Raycoms: Can we instead just do a for each loop and if does not match primary output continue? I feel that would be cleaner

Thodor12: I figured that too, but that has the chance to catch false positives.

Let's say you had a crafting recipe that is guaranteed to make item A once, but like a 50% dice roll to produce the same item again.

That'd net you 2x an itemstack of A with 1 item, if I were to filter out all the items that match the primary output. It'd filter out potential extra drops from the secondary outputs (or loot table) too, which is not what it's supposed to do.

I only need to extract the first item (if it's there)

Raycoms: I feel this is a primary output then as well

Thodor12: I mean not really, that is chance based, which the primary output cannot suffice. You can only set a hard "this recipe produces X", that part is then tracked and sent as part of the delivery.

Any extras that it produces, via loot table or secondaries are not part of the initial delivery, even if it is the same item as the primary item stack.

Raycoms (dismissed):

src/main/java/com/minecolonies/core/colony/jobs/AbstractJobCrafter.java
```diff
@@ -51,6 +57,12 @@ public abstract class AbstractJobCrafter<AI extends AbstractEntityAIBasic<J, ? e
      */
     private int progress = 0;
 
+    /**
+     * The current map of secondary outputs that need to return to the warehouse after crafting is done.
+     */
+    @NotNull
+    private final Map<ItemStorage, Integer> secondaryOutputs = new HashMap<>();
```
someaddons: a Set<ItemStorage> should suffice, as item storage has an internal amount integer

Thodor12: I figured with a Set I couldn't really make use of the `.compute` logic that a map has. It'd be slightly more annoying to have to deal with merging items together into 1 stack.

Raycoms: can't you do set.put(set.getOrDefault()) ?

Thodor12: My sweet Ray, sets have no getters, that's the whole problem, can only iterate them

Raycoms: I thought I remembered that some hashset had some kind of get by hash, I'm disappointed

Raycoms: So yeah, a set has us iterate it for each element which is annoying, so a map is the best option. Although maybe we should use a dummy value for the value, so that we don't mix itemstorage count and map count up weirdly.

Thodor12: I mean I could make a `Map<ItemStorage, Boolean>` or something, just leave the value to false. But the problem is that it's difficult to modify the key, not the value.

I can't exactly store the `Item` itself either, since that ain't sufficient for things that need NBT data, nor `ItemStacks` because as far as I can tell they're not comparable (no equals implementation)

Raycoms: Then make it an object2Int map I would say, those are even more efficient for this particular usecase then.

someaddons: or you remove the existing from the set, addd the count to the new one and add it to the set again

Thodor12: @someaddons Set#remove returns a boolean if the item was found, not the original object, so a set is simply is not doable

src/main/java/com/minecolonies/core/colony/jobs/AbstractJobCrafter.java
```diff
@@ -61,7 +62,7 @@ public abstract class AbstractJobCrafter<AI extends AbstractEntityAIBasic<J, ? e
      * The current map of secondary outputs that need to return to the warehouse after crafting is done.
      */
     @NotNull
-    private final Map<ItemStorage, Integer> secondaryOutputs = new HashMap<>();
+    private final Map<ItemStorage, Integer> secondaryOutputs = new Object2IntArrayMap<>();
```
Raycoms, line 66: Object2IntHashMap<ItemStorage> = new Object2IntHashMap

Raycoms: then you can use the special sum API of it

someaddons (approved):

PR #10372, opened 2024-10-28
Some improvements in navigation

HexatomicRing:
Validated the facing of stairs while computing jumping cost.
Modified some path-finding cost to be closer to its real time cost.
Added path-finding cost for opening a door.

The modification to `gradle.properties` and `settings.gradle` is necessary, otherwise IDEA cannot import Minecraft dependencies!

## Testing

Review please


src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -1100,6 +1131,10 @@ else if (p.isOnRails() && points.length > pathLength + 1)
         }
 
         result.searchedNodes = totalNodesVisited;
+        Minecraft mc = Minecraft.getInstance();
+        for(Node n: points){
+            mc.level.addParticle(ParticleTypes.HAPPY_VILLAGER, n.x + 0.5, n.y + 1, n.z + 0.5, 0, 0, 0);
+        }
```
HexatomicRing, line 1137: Sorry, I left debugging code here. Please ignore this.

Thodor12:
The changes to gradle properties are unnecessary, we have our own system called https://github.com/ldtteam/OperaPublicaCreator which handles all of the setup logic for (Neo)Forge. So those changes are not required.

PR #10385, opened 2024-10-31
Update previous food tags

Thodor12:
- Upon request of Ray after accidental merge. Updated the food tags to just datagen, not be included in the ItemFood
- Undone food builder logic in ModItemsInitializer
- Brought list code more in line with what Ray already made for the cook rework
- Ran datagen, but no changes were made

## Testing

Review please


Raycoms (dismissed):
Much better

marchermans (dismissed):

src/main/java/com/minecolonies/api/items/ModItems.java
```diff
@@ -185,6 +183,12 @@ public static Item[] getAllIngredients()
     public static Item[] getAllFoods()
     {
         return new Item[] {
+          // Breads
+          milkyBread,
+          sugaryBread,
+          goldenBread,
+          chorusBread,
+
```
Raycoms: I think it's fine if we keep these out as currently of the getAllFoods. They are some "special items" for special uses cases and not regular food.

Thodor12: Alright, I wasn't sure if that was intentional or forgotten, I'll remove them

PR #10391, opened 2024-11-02
Remove reset config and just make available to managers

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/473575384445878273/1302053820800700437)

- Fixes the requestsystem reset commands blatantly lying about how much time they take.
- Allows all colony managers to run the colony-specific reset command, since that's usually one of the first troubleshooting steps anyway when things are misbehaving.
- Removes the `canplayeruseresetcommand` config setting, as it's no longer needed.
- Displays an explicit error message for any manager-only command run by a non-manager instead of silence.

## Testing

Review please (should port)

I considered adding an explicit TH permission setting for it, but I couldn't think of any particular reason why colony owners might want to restrict who could run it (there's a lot more dangerous things that managers can currently do anyway).

### DIFF
--- a/src/main/resources/assets/minecolonies/lang/manual_en_us.json
+++ b/src/main/resources/assets/minecolonies/lang/manual_en_us.json
@@ -259,8 +259,6 @@
   "minecolonies.config.delaybetweenretries.comment": "The amount of ticks between retries of the request system for retryable requests. Lower increases server load.",
   "minecolonies.config.creativeresolve": "Creatively Resolve Requests",
   "minecolonies.config.creativeresolve.comment": "Should the request system creatively resolve (if possible) when the player is required to resolve a request? This is a debugging tool and can take a very long time to resolve a request.",
-  "minecolonies.config.canplayeruseresetcommand": "Can Players Use the Request System Reset Command",
-  "minecolonies.config.canplayeruseresetcommand.comment": "Should players be allowed to use the /mc colony requestsystem-reset command?",
 
   "minecolonies.config.default.boolean": "[Default: %b]",
   "minecolonies.config.default.int": "[Default: %d, min: %d, max: %d]",
@@ -385,6 +383,7 @@
   "com.minecolonies.command.help.wiki": "Check out the wiki to learn more: ",
   "com.minecolonies.command.help.discord": "Join the community at  ",
   "com.minecolonies.command.notop": "Must be OP to use this command.",
+  "com.minecolonies.command.notofficer": "You do not have permission to use this command.",
   "com.minecolonies.command.whereami.nocolony": "No close colony.",
   "com.minecolonies.command.whereami.colonyclose": "You're not inside any colony. The closest colony is approx %.2f blocks away.",
   "com.minecolonies.command.whereami.incolony": "You're inside colony %s with ID %s. The colony center is approx %.2f blocks away.",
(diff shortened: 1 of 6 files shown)


Raycoms:
The reason we never made it that accessible for players is that we want them to report issues to us and not just constantly run RS reset (a lot of people already do that because they turn on the config). So I would keep that config for sure.

uecasm:
Yeah, but single-player cheats-enabled people can already run it regardless of the setting.  And "bad thing happened one time and went away after a reset" is not necessarily a helpful bug report, but "bad thing happens repeatedly even after reset" is more useful.

Raycoms:
Like, we have it disabled on our servers and here and there get actual useful things out of that, aside from that its maybe necessary in 1/100000 cases to run it. Running it too often can prejudice performance by a lot

Talyda:
We use it a lot as a troubleshooting step. If we sent every "problem" from discord to github before trying it, you'd get way more reports that wasted your time =P It's a valid step for us to see if it is just a temporary problem, or a persistent one.

And yeah, you have it disabled on the servers, but between builders that don't recognise they've been given items, them still asking for stuff that they wanted in the previous job, and the warehouse still listing things that were delivered ages ago, or waiting for requests that don't even show up on their request tab - we just sometimes really really want to be able to reset the request queue, and you guys can't be there to help out every time it futzes up (and pretty sure you wouldn't want us pinging you every time a request got stuck =P). It's incredibly frustrating to need to cancel/restart jobs, pause workers, wait for server restarts etc when it's just one of those long standing things that there is no fix for.

Raycoms:
> We use it a lot as a troubleshooting step. If we sent every "problem" from discord to github before trying it, you'd get way more reports that wasted your time =P It's a valid step for us to see if it is just a temporary problem, or a persistent one.
> 
> And yeah, you have it disabled on the servers, but between builders that don't recognise they've been given items, them still asking for stuff that they wanted in the previous job, and the warehouse still listing things that were delivered ages ago, or waiting for requests that don't even show up on their request tab - we just sometimes really really want to be able to reset the request queue, and you guys can't be there to help out every time it futzes up (and pretty sure you wouldn't want us pinging you every time a request got stuck =P). It's incredibly frustrating to need to cancel/restart jobs, pause workers, wait for server restarts etc when it's just one of those long standing things that there is no fix for.

yes, but if people reset this every time, us actually finding a reproduction step becomes slim to none then =D
Aside of course, the massive performance impact of recalculating all requests for larger colonies is insane

someaddons:
Debug command -> should not be part of gameplay

src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -421,8 +420,8 @@
   "com.minecolonies.command.loadbackup.success": "Successfully loaded backup.",
   "com.minecolonies.command.export.success": "Exported colony to zip: %s",
   "com.minecolonies.command.deleteable.success": "Changed deletable flag of colony ID %s. It is now set to %s.",
-  "com.minecolonies.command.rsreset.success": "The request system for colony %s has been restarted in 1.618 seconds.",
-  "com.minecolonies.command.rsresetall.success": "The request systems for all colonies have been restarted in 1.618 seconds.",
```
marchermans, line 425: ^Please leave this in, it is an inside joke. It is not ment to be a measure at all.

marchermans (changes requested):

PR #10397, opened 2024-11-04
Update all forge event code, attempt to become loader unaware

Thodor12:
- Introduce a new event handler in the IMinecoloniesAPI, which is backed by an interface containing basic methods for the events we want to send
- Event handler is implemented in such a way it could be made loader unaware in the future if needed.
- All logic for sending Forge specific events is ripped out in favor of the API event handler

## Testing

Review please


Raycoms (dismissed):

src/main/java/com/minecolonies/api/events/ForgeModEventHandler.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Override
+    public void colonyFlagChanged(final @NotNull IColony colony)
+    {
+        sendEvent(new ColonyFlagChangedEvent(colony));
+    }
+
+    @Override
+    public void buildingCompleted(final @NotNull IBuilding building, final @NotNull WorkOrderBuilding workOrder)
```
Raycoms: I chatted with sam about this. For once, having every single method in here, instead of just sendEvent(event) seems not very flexible. On the other hand we could just add loader independent guava events, doesn't necessariyl have to go over the forge bus

Raycoms: Then we're automatically loader independent and don't have to worry

Thodor12: Converted to our own simple implementation of an event bus, discussed with Ray in DM

marchermans (approved):

PR #10404, opened 2024-11-04
Fix builders requesting items again after already having them available

### LINKED ISSUE #10123: [BUG] <The builder is requesting materials that are already available.>

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.20.1

### MineColonies Version

1.20.1-1.1.629

### Structurize Version

1.20.1-1.0.742

### Current Behavior

Sometimes, after building for a while, the builder will request materials that are already satisfied in the material list. I've noticed that the materials requested by the builder are already in their inventory. Maybe they sometimes don't check if they already have enough materials in their inventory?

### Expected Behavior

Don't request materials that are already available.

### Reproduction Steps

1. Request to build a larger structure (preferably with more complex materials).
2. Fulfill their requests (sometimes I don’t click the + button directly but instead put the items in or give them to the builder directly; maybe this has an impact?)
3. Wait for them to build.

### Logs

No crash

### Anything else?

Thodor12:
Closes #10123

Some background in the problem:
When a builder calculates the resources for the build they'll be split into several buckets of items, to allow for gradual requesting of items as time progresses.
When the builder is requesting items they'll look for the current and the next bucket, to ask ahead for some items.

The current bucket was being checked against the building and worker inventory, to see if they have everything available.
However, the next bucket was only being checked against the building inventory, not the worker.
For some reason the builders are pulling items from any bucket into their inventory at times, not to mention that players could potentially be inserting items into their inventory directly.
Because of the old logic, you would see requests disappear, but because the next bucket is only verified against the building inventory, and not the worker, these requests would show up again and again, provided the items are not inside of the building.

So the easiest fix was to allow the next bucket to verify against the worker inventory as well, rather than somehow ensure the worker doesn't pull the items into their inventory.

- Builders will now check items in their inventory for the next bucket. Due to this change the last boolean of the method was always true, making it redundant, and thus removed.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingResourcesModule.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingResourcesModule.java
@@ -327,9 +327,8 @@ public class BuildingResourcesModule extends AbstractBuildingModule implements I
      *
      * @param requiredResources the bucket to check and request.
      * @param worker            the worker.
-     * @param workerInv         if the worker inv should be checked too.
      */
-    public void checkOrRequestBucket(@Nullable final BuilderBucket requiredResources, final ICitizenData worker, final boolean workerInv)
+    public void checkOrRequestBucket(@Nullable final BuilderBucket requiredResources, final ICitizenData worker)
     {
         if (requiredResources == null)
         {
@@ -353,13 +352,10 @@ public class BuildingResourcesModule extends AbstractBuildingModule implements I
                 continue;
             }
 
-            if (workerInv)
+            count += InventoryUtils.getItemCountInItemHandler(worker.getInventory(), stack -> ItemStackUtils.compareItemStacksIgnoreStackSize(stack, itemStack.getItemStack()));
+            if (count >= entry.getValue())
             {
-                count += InventoryUtils.getItemCountInItemHandler(worker.getInventory(), stack -> ItemStackUtils.compareItemStacksIgnoreStackSize(stack, itemStack.getItemStack()));
-                if (count >= entry.getValue())
-                {
-                    continue;
-                }
+                continue;
             }
 
             int requestCount = entry.getValue() - count;
--- a/src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingStructureBuilder.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingStructureBuilder.java
@@ -404,11 +404,10 @@ public abstract class AbstractBuildingStructureBuilder extends AbstractBuilding
      *
      * @param requiredResources the bucket to check and request.
      * @param worker            the worker.
-     * @param workerInv         if the worker inv should be checked too.
      */
-    public void checkOrRequestBucket(@Nullable final BuilderBucket requiredResources, final ICitizenData worker, final boolean workerInv)
+    public void checkOrRequestBucket(@Nullable final BuilderBucket requiredResources, final ICitizenData worker)
     {
-        getFirstModuleOccurance(BuildingResourcesModule.class).checkOrRequestBucket(requiredResources, worker, workerInv);
+        getFirstModuleOccurance(BuildingResourcesModule.class).checkOrRequestBucket(requiredResources, worker);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


Raycoms (commented):
Please test this extensively, like a couple of builders running for an hour in the background or so =D

MotionlessTrain:
Just in case, items in the builder's inventory can't be matched against both requests in the current bucket and in the next bucket?
(E.g. when a building needs multiple stacks of stone bricks, the current bucket has a request for stone bricks, and the next bucket has a request for stone bricks as well, then one stack of stone bricks in the builder's inventory wouldn't match the requests in both the buckets)

Thodor12:
That's not checked, no, but should theoretically end up fine, because if the current + next bucket does not end up requesting enough, as soon as the initial bucket is done, and they notice they don't have enough, they'll make another request to get the remaining missing items.

MotionlessTrain:
Fair enough. It would only just be a bit slower in that case

Shadizar:
What if any info would help nail it down?
I had this hit in my stream multiple times today <https://www.twitch.tv/videos/2293601288> , one at a bit after 0:35
I'll try to find the other timestamp
But I fear this likely doesn't really help you. Because the repro is literally just ... do buildings. Eventually it always happens, with resource scroll showing The Truth and clipboard requests popping up repeatedly

MotionlessTrain:
I have the feeling that this will help, if not solve it completely.
The background sketched here aligns with my own observations: Each of the materials asked multiple times this way were in the builder's inventory, and I actually managed to fulfill such requests a couple times by getting the items out of the builder's inventory and fulfilling the request again

Shadizar:
same vod 1:09:50 ish has  another instance (with brown carpet) fyi

Thodor12:
> What if any info would help nail it down? I had this hit in my stream multiple times today https://www.twitch.tv/videos/2293601288 , one at a bit after 0:35 I'll try to find the other timestamp But I fear this likely doesn't really help you. Because the repro is literally just ... do buildings. Eventually it always happens, with resource scroll showing The Truth and clipboard requests popping up repeatedly

I have it nailed down, it's just not reproducible the same way because the buckets of requests are randomized, no two times I can reproduce it the exact same way. Each time I have to look which items appear in the second bucket and use those items. That way I was able to reproduce the bug everytime.

Shadizar:
and my solve so far has been: if resource scroll shows they have it just keep canceling clipboard request until they stop asking

Raycoms:
> and my solve so far has been: if resource scroll shows they have it just keep canceling clipboard request until they stop asking

you shouldn't even need to cancel it, its an async request that doesn't even need fulfillment.

Shadizar:
Fair; I just do so I can sort what I need to supply :) with them still there it's sometimes easy to mistake.

Thodor12:
The only annoyance is that the couriers will automatically comply and deliver extra stuff you do not need

Shadizar:
yeah that part I have to handle by not teaching several types of things (iron or better armor, iron or better tools, anything that has any expense to it) until I have a boatload of it
Learned that after some of my guards once started trucking around in diamond gear that I had no intent of giving them

src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingResourcesModule.java
```diff
@@ -353,13 +352,10 @@ public void checkOrRequestBucket(@Nullable final BuilderBucket requiredResources
                 continue;
             }
 
-            if (workerInv)
+            count += InventoryUtils.getItemCountInItemHandler(worker.getInventory(), stack -> ItemStackUtils.compareItemStacksIgnoreStackSize(stack, itemStack.getItemStack()));
```
someaddons, line 355: while on it we should use the Building specific Inventory Utils hasBuildingEnoughElseCount

someaddons: nvm

Thodor12: Ray told me he was going to do further improvements himself.

someaddons (approved):

PR #10406, opened 2024-11-05
Several navigation fix

HexatomicRing:
- Path-finding can now identify the direction of stairs, avoiding jumping up stairs from the side.
- Path-finding can now identify doors. Citizens and intruders now prefer paths with fewer doors to open.
- Modified the drop cost in path-finding. 1 falling damage cost 25 in addition. Swimming enter cost is lowered so that citizens will drop in water to avoid falling damage. (In practical situations, it may fail because citizens sometimes move too fast to drop into the water)

* I am also preparing another PR to enable citizens to use lever to toggle rails at a T-cross railway.

## Testing

Review please


Talyda:
Don't make intruders prefer a path with fewer doors! That was such a good trap in the raid stream LOL!
(not a dev - just someone who watched the stream, and thought that was really clever!)

Thodor12:
> Don't make intruders prefer a path with fewer doors! That was such a good trap in the raid stream LOL! (not a dev - just someone who watched the stream, and thought that was really clever!)

You have to see path finding cost as "the path of least resistance", if there's no other option than doors, they'll use the doors anyway.

src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -917,15 +920,32 @@ protected double computeCost(
 
         if (!isDiving)
         {
-            if (dY != 0 && !(ladder && parent.isLadder()) && !(Math.abs(dY) == 1 && below.is(BlockTags.STAIRS)))
+            boolean correctlyOnStairs = false;
+            if(Math.abs(dY) == 1 && below.getBlock() instanceof StairBlock && below.getValue(StairBlock.HALF) == Half.BOTTOM)
```
Thodor12: Please reformat your code to ensure it's up to the defined code style (in this case, space between `if` and `(`

Thodor12 (changes requested):

Talyda:
> You have to see path finding cost as "the path of least resistance", if there's no other option than doors, they'll use the doors anyway.

Ah, fair point! =D

HexatomicRing:
> Don't make intruders prefer a path with fewer doors! That was such a good trap in the raid stream LOL! (not a dev - just someone who watched the stream, and thought that was really clever!)

While testing, citizens can open a door instantly. But they always take many seconds to open a door while playing, and they only need to walk two extra blocks to get around. I've really had enough. 😭

Just set the open door cost of intruders to 0 to trap them. 👿

src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
HexatomicRing: > Please reformat your code to ensure it's up to the defined code style (in this case, space between `if` and `(`

Done.

PR #10409, opened 2024-11-05
Serveral pathfinding fixes

someaddons:
Closes #
Closes #

- Add planks, slabs and carpets as pathblocks
- Pathblocks no longer increase pathing cost when they have a nonempty collision shape(carpets)
- Doors and pathblocks are now cheaper to traverse
<br></br>
- Re-visiting a pathfinding Node no longer counts towards the maximum allowed nodes to visit
- Pathfinding Nodes now got a visited counter, which limits how many times the cost of a node is allowed to update. Heuristic rebalancing allows an additional revisit
- Reevaluating the heuristic in pathfinding is now more accurate during search, using the distance to destination provided by the new IDestinationPathJob interface, and are also applied gradually instead of at once
- Fix a bug where corner nodes could connect to eachother
- Fix a bug where corner nodes that get their cost updated would not be recognized as such
<br></br>
- Add fast collision existance check to ShapeUtil
- PathfindingAIHelper#walkCloseToXNearY now runs the pathjob at least once, even if the entity may still already be in working range
- Stuckhandler now at minimum walks the entity 20 blocks away when trying to resolve stuck
-  PathJobFindWater now discourages pathfinding from going too far out of its search radius
- PathJobMoveCloseToXNearY got a slightly adjusted heuristic to prefer staying close to the near position
<br></br>
## Testing

Review please


src/main/java/com/minecolonies/api/util/EntityUtils.java
```diff
@@ -205,7 +206,7 @@ private static boolean checkValidSpawn(@NotNull final BlockGetter world, final B
         for (int dy = 0; dy < height; dy++)
         {
             final BlockState state = world.getBlockState(pos.above(dy));
-            if (!state.is(ModTags.validSpawn) && BlockUtils.isAnySolid(state))
+            if (!state.is(ModTags.validSpawn) && (PathfindingUtils.isLiquid(state) || ShapeUtil.hasCollision(world, pos.above(dy), state)))
```
Raycoms, line 209: isnt that an || in the beginning too

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/PathfindingAIHelper.java
```diff
@@ -23,19 +23,28 @@ public static boolean walkCloseToXNearY(
...
+                // Check distance once navigation is done, to let the entity walk
+                if (BlockPosUtil.dist(entity.blockPosition(), desiredPosition) < distToDesired)
+                {
+                    nav.stop();
+                    return false;
+                }
             }
 
-            PathJobMoveCloseToXNearY pathJob = new PathJobMoveCloseToXNearY(entity.level, desiredPosition, nearbyPosition, distToDesired, entity);
+            PathJobMoveCloseToXNearY pathJob = new PathJobMoveCloseToXNearY(entity.level, desiredPosition, nearbyPosition, 1, entity);
```
Raycoms, line 47: Intentionally only 1 dist?

src/main/java/com/minecolonies/api/util/EntityUtils.java
someaddons: nah because this check is specifically for allowing blocks which are not known as valid and if they are not valid tagged and have liquid or collisions they're excluded

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/PathfindingAIHelper.java
someaddons: ye to have the entity move as close as possible

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/PathfindingAIHelper.java
```diff
@@ -23,19 +23,28 @@ public static boolean walkCloseToXNearY(
...
-        if (nav.isDone() || (nav.getPathResult() != null
-                               && !(nav.getPathResult().getJob() instanceof PathJobMoveCloseToXNearY job
-                                      && job.nearbyPosition.equals(nearbyPosition)
-                                      && job.desiredPosition.equals(desiredPosition)
-                                      && job.distToDesired == distToDesired)))
+        // Three cases
+        // 1. Navigation Finished
+        // 2. Navigation is progressing towards a previous task
+        // 3. Navigation did not try once
+        boolean isOnRightTask = (nav.getPathResult() != null
```
Raycoms, line 30: final?

Raycoms (approved):

PR #10410, opened 2024-11-06
Enables citizens to use lever to toggle T-cross railway

HexatomicRing:
Citizens now are able to toggle the direction of a RailBlock using the Lever next to it. There only exists a small chance that more than one citizens are on the same rail block in one tick that they may go to wrong direction.
Let citizens use railways more efficiently!

## Testing

Review please

// Sorry that I deleted my fork by mistake, so I had to re-create this pull request😭.


Raycoms:
This should probably not be a goal, but should be handled by the pathfollow mechanism + should not need a lever even (just auto rotate the rail)

HexatomicRing:
> This should probably not be a goal, but should be handled by the pathfollow mechanism + should not need a lever even (just auto rotate the rail)

Yes, in future versions, due to the adjustable speed of the mining car, continuing to use Goal will have a certain probability of failure. I will modify my code later.
In addition, if two citizens heading in different directions pass through the same railway track in the same tick, one citizen will go in the wrong direction. Then he will walk the rest path instead of using a minecart. I haven't figured out how to treat this situation yet.

Raycoms:
> > This should probably not be a goal, but should be handled by the pathfollow mechanism + should not need a lever even (just auto rotate the rail)
> 
> Yes, in future versions, due to the adjustable speed of the mining car, continuing to use Goal will have a certain probability of failure. I will modify my code later. In addition, if two citizens heading in different directions pass through the same railway track in the same tick, one citizen will go in the wrong direction. Then he will walk the rest path instead of using a minecart. I haven't figured out how to treat this situation yet.

There is also the option of ignoring the rail and telling the railcart entity the junctions at which it would have to change direction.

someaddons:
better solution would be a T cross rail block :D

Raycoms:
> better solution would be a T cross rail block :D

Agreed, but it would still require at least to tell the cart which direction to go =D

PR #10411, opened 2024-11-06
Several navigation fix

HexatomicRing:
- Path-finding can now identify the direction of stairs, avoiding jumping up stairs from the side.
- Path-finding can now identify doors. Citizens and intruders now prefer paths with fewer doors to open.
- Modified the drop cost in path-finding. 1 falling damage cost 25 in addition. Swimming enter cost is lowered so that citizens will drop in water to avoid falling damage. (In practical situations, it may fail because citizens sometimes move too fast to drop into the water)

## Testing

Review please
// Sorry that I deleted my fork by mistake, so I had to re-create this pull request😭.


someaddons:
Sorry, but no you're lacking understanding of the system

HexatomicRing:
> Sorry, but no you're lacking understanding of the system

I may lack understanding, but the pathfinding provided by this system is indeed foolish in some situations, and only when the cost is close to the actual cost can a better path be given.

Raycoms:
A lot of things certainly can be further fine tuned. Doors are fine to traverse and actually indicate a proper path. Jumping should only be done im "emergencies" when can't reach something (still needs adjustments). So in water jumping shouldn't really be cheap.

Overall considering stair direction could be interesting, but can be done a lot easier.

HexatomicRing:
> A lot of things certainly can be further fine tuned. Doors are fine to traverse and actually indicate a proper path. Jumping should only be done im "emergencies" when can't reach something (still needs adjustments). So in water jumping shouldn't really be cheap.
> 
> Overall considering stair direction could be interesting, but can be done a lot easier.

The cost of dropping is also unreasonable. It equals to the cube of the falling distance, but clipped by a max cost of 25. This actually makes drops greater than or equal to 4 blocks equally expensive but not expensive enough. In many cases, a small detour can make citizens jump off cliffs and take damage, for example, from the second floor of the residential.

PR #10413, opened 2024-11-06
Raider improvements

someaddons:
Closes #
Closes #

- Raiders and guards now properly look at their target while fighting
- Raiders do now also properly display item use animations, spear/bow etc
- Archer raiders now also draw their bow in preparation of shooting
- Raiders no longer scale in attackspeed, as they already scale in damage with difficulty

## Testing

Review please


Raycoms (dismissed):
has conflicts

PR #10423, opened 2024-11-09
Sound fixes for block placing/breaking

Sindarin27:
This pull request makes slight adjustments to the sound effects created when citizens place or break blocks, to sound more in-line with the sound effects used in Vanilla.

- When placing, hitting, or breaking a block, the same pitch and volume calculations are used as in Vanilla functions. 
   - In practice: block sounds are lowered in pitch to match Vanilla sounds.
- When a builder places a solid substitution block, use the sound effect of the block that is substituted instead of the sound of the solid substitution block itself. 
   - In practice: builders placing dirt blocks to fill an area now makes dirt placing sounds instead of wood placing sounds.
- When hitting a block but not yet breaking it, use the block's hit sound instead of the break sound. 
   - In practice: a builder digging up e.g. grass blocks now makes a lot less noise.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/util/BuildingStructureHandler.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/util/BuildingStructureHandler.java
@@ -220,12 +220,14 @@ public class BuildingStructureHandler<J extends AbstractJobStructure<?, J>, B ex
             }
             
             BlockState blockStateForSound;
-            if (state.getBlock() == com.ldtteam.structurize.blocks.ModBlocks.blockSolidSubstitution.get()) {
+            if (state.getBlock() == com.ldtteam.structurize.blocks.ModBlocks.blockSolidSubstitution.get()) 
+            {
                 // If the builder is placing a substitution block, use the sound of the substituted block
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/core/entity/ai/workers/util/BuildingStructureHandler.java
```diff
@@ -218,8 +218,18 @@ public void triggerSuccess(final BlockPos pos, final List<ItemStack> list, final
                   .getStatisticsManager()
                   .increment(BLOCKS_PLACED, structureAI.getWorker().getCitizenColonyHandler().getColony().getDay());
             }
-
-            structureAI.getWorker().queueSound(state.getSoundType().getPlaceSound(), worldPos, 10, 0);
+            
+            BlockState blockStateForSound;
+            if (state.getBlock() == com.ldtteam.structurize.blocks.ModBlocks.blockSolidSubstitution.get()) {
```
Raycoms: Please apply our formatting rules here

Raycoms (changes requested):
Hello, thank you for your PR, I just left a minor comment for now

Raycoms (approved):

someaddons (approved):

PR #10439, opened 2024-11-12
No more teams

Raycoms:
Closes #
Closes #

- Remove teams in favor of our system that prevents friendly fire and guarantees the colors.
- Rename getColony method in colony handler to be more clear what it does and add a method that just does a simple fetch.

## Testing

Review please


src/main/java/com/minecolonies/core/entity/citizen/EntityCitizen.java
```diff
@@ -1980,4 +1982,10 @@ public void setMaxAir(final int maxAir)
     {
         this.maxAir = maxAir;
     }
+
+    @Override
+    public String getTeamName()
+    {
+        return level().dimension().hashCode() + "_" + citizenColonyHandler.getColonyId();
```
someaddons: why a string which impacts performance(creates a new string obj everytime to compare, needs to compare strings) instead of just the colony ID as int? just do int getTeam()

someaddons (approved):

PR #10449, opened 2024-11-15
Fix creative tab enumeration

uecasm:
- Fixes exception when doing ItemNbtCalculator datagen

## Testing

Review please (do not port)

### DIFF
--- a/src/main/java/com/minecolonies/core/generation/ItemNbtCalculator.java
+++ b/src/main/java/com/minecolonies/core/generation/ItemNbtCalculator.java
@@ -79,7 +79,7 @@ public class ItemNbtCalculator implements DataProvider
         // Force loading some tags, since the creative tabs don't enumerate properly without them
         return lookupProvider
             // Tag TagKey[minecraft:instrument / minecraft:goat_horns] can't be dereferenced during construction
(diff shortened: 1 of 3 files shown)


Raycoms (dismissed):

src/main/java/com/minecolonies/core/generation/ItemNbtCalculator.java
```diff
@@ -57,7 +74,18 @@ public String getName()
     @Override
     public CompletableFuture<?> run(@NotNull final CachedOutput cache)
     {
-        return lookupProvider.thenCompose(provider ->
+        final ResourceManager serverResources = loadServerData();
+
+        // Force loading some tags, since the creative tabs don't enumerate properly without them
+        return lookupProvider
```
someaddons, line 80: if it is a loading order thing, why dont we just use creative tabs later?

Raycoms: because this is the event to do handling and synching.

uecasm: It's not loading order, the tags never ordinarily get loaded during datagen at all.

Or rather, the tag keys get loaded (which is sufficient to reference the tags in other tags and recipes) but not the tag contents.

someaddons (approved):

PR #10468, opened 2024-11-19
Update all forge event code, attempt to become loader unaware

Thodor12:
- Introduce a new event handler in the IMinecoloniesAPI, which is backed by an interface containing basic methods for the events we want to send
- Event handler is implemented in such a way it could be made loader unaware in the future if needed.
- All logic for sending Forge specific events is ripped out in favor of the API event handler

## Testing

Review please


src/main/java/com/minecolonies/api/eventbus/DefaultEventBus.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    private final Map<TypeToken<?>, List<EventHandler<?>>> eventHandlersPerType = new HashMap<>();
+
+    @Override
+    public <T extends IModEvent> void subscribe(final @NotNull IModEventType<T> eventType, final @NotNull EventHandler<T> handler)
+    {
+        Log.getLogger().debug("Registering event handler for id {}.", eventType.getIdentifier());
+
+        eventHandlersPerType.putIfAbsent(eventType.getIdentifier(), new ArrayList<>());
+        eventHandlersPerType.get(eventType.getIdentifier()).add(handler);
```
Nightenom: seems like double hashing to me, computeIfAbsent(key).add(value)?

src/main/java/com/minecolonies/api/eventbus/DefaultEventBus.java
```diff
@@ -0,0 +1,54 @@
...
+
+        for (final EventHandler<?> handler : eventHandlers)
+        {
+            try
+            {
+                ((EventHandler<T>) handler).apply(event);
+            }
+            catch (Exception ex)
+            {
+                Log.getLogger().warn("Event id: {};'Error occurred during processing event handler:", event.getEventId(), ex);
```
Nightenom:
```suggestion
                Log.getLogger().warn("Error occurred during processing event '{}', handler '{}':", event.getEventId(), handler.toString() or handler#class#toString(), ex);
```

src/main/java/com/minecolonies/api/eventbus/events/colony/AbstractColonyEvent.java
```diff
@@ -1,37 +1,36 @@
...
+
+    /**
+     * Gets the colony related to the event.
+     */
+    @NotNull
+    public IColony getColony()
+    {
+        return colony;
+    }
+}
```
Nightenom: there are few classes apparently with cr/lf changes

src/main/java/com/minecolonies/apiimp/CommonMinecoloniesAPIImpl.java
```diff
@@ -52,34 +54,36 @@
...
+    private IForgeRegistry<ResearchEffectEntry>                        researchEffectRegistry;
+    private IForgeRegistry<ResearchCostType>                           researchCostRegistry;
+    private IForgeRegistry<RecipeTypeEntry>                            recipeTypeEntryRegistry;
+    private IForgeRegistry<CraftingType>                               craftingTypeRegistry;
+    private IForgeRegistry<QuestRegistries.ObjectiveEntry>             questObjectiveRegistry;
+    private IForgeRegistry<QuestRegistries.RewardEntry>                questRewardRegistry;
+    private IForgeRegistry<QuestRegistries.TriggerEntry>               questTriggerRegistry;
+    private IForgeRegistry<QuestRegistries.DialogueAnswerEntry>        questDialogueAnswerRegistry;
+    private IForgeRegistry<HappinessRegistry.HappinessFactorTypeEntry> happinessFactorTypeRegistry;
+    private IForgeRegistry<HappinessRegistry.HappinessFunctionEntry>   happinessFunctionRegistry;
```
Nightenom: I mean can we generally avoid unnecessary whitespace changes? _or ideally finally make proper autoformat_ :D

Nightenom:
Also (without looking at code if it's even possible) cannot we reuse existing forge eventbus for implementing the bus logic?

src/main/java/com/minecolonies/apiimp/CommonMinecoloniesAPIImpl.java
Thodor12: I mean we technically have a format guideline, it just wasn't applied to this file :D

Thodor12:
> Also (without looking at code if it's even possible) cannot we reuse existing forge eventbus for implementing the bus logic?

The point was from the beginning to not use Forge code where possible, so we quickly glanced over using a Guava event bus.
But I found that to be less than ideal since it's a third party library Mojang happened to insert, if that were to be dropped we'd have to fallback to shading it, which doesn't work with Guava.

For that reason I thought it was a better idea to make a small lightweight event bus tailored to our needs.

src/main/java/com/minecolonies/api/eventbus/events/colony/AbstractColonyEvent.java
Thodor12: These are just weird diffs by Github, it's because the file moved and slightly altered it doesn't know what to do with them.

someaddons:
Forge eventbus is an extion of guavas afaik

uecasm:
What's the point of avoiding (Neo)Forge APIs?  Unless you're anticipating a Fabric port in the future or something, but that can be handled with an abstraction layer rather than a different implementation.

src/main/java/com/minecolonies/api/eventbus/events/colony/AbstractColonyEvent.java
Raycoms: Yeah, but we shouldn't auto format all files we touch, because that makes it impossible to read the dif

Raycoms: Also applies to the other completely formatted files. Aside from porting madness problems =D

src/main/java/com/minecolonies/core/entity/citizen/EntityCitizen.java
```diff
@@ -1627,18 +1628,18 @@ public void die(@NotNull final DamageSource damageSource)
...
 
-            try
-            {
-                MinecraftForge.EVENT_BUS.post(new CitizenRemovedEvent(citizenData, damageSource));
-            }
-            catch (final Exception e)
-            {
-                Log.getLogger().error("Error during CitizenRemovedEvent", e);
-            }
+            IMinecoloniesAPI.getInstance().getEventBus().post(MinecoloniesEventTypes.CITIZEN_DIED, new com.minecolonies.api.eventbus.events.colony.citizens.CitizenDiedEvent(citizenData, damageSource));
```
Raycoms: too long import

src/main/java/com/minecolonies/core/colony/ColonyManager.java
```diff
@@ -202,7 +204,7 @@ private void deleteColony(@Nullable final IColony iColony, final boolean canDest
                 return;
             }
 
-            ColonyEvents.deleteColony(colony);
+            IMinecoloniesAPI.getInstance().getEventBus().post(MinecoloniesEventTypes.COLONY_DELETED, new ColonyDeletedEvent(colony));
```
Raycoms: Seems a bit long. Couldn't we
a) Do a singleton: DefaultEventBus.post
b) just post the new ColonyDeletedEvent, without needing the type constant?

Raycoms (changes requested):
Two smaller comments

Raycoms (approved):

src/main/java/com/minecolonies/api/eventbus/DefaultEventBus.java
```diff
@@ -0,0 +1,52 @@
...
+            return;
+        }
+
+        Log.getLogger().debug("Sending event '{}' for type '{}'. Sending to {} handlers.", event.getEventId(), event.getClass().getSimpleName(), eventHandlers.size());
+
+        for (final EventHandler<? extends IModEvent> handler : eventHandlers)
+        {
+            try
+            {
+                ((EventHandler<IModEvent>) handler).apply(event);
```
Nightenom, line 44: Mby change post also to `? extends Event` to avoid typing errors

Nightenom (commented):
Generally speaking I would avoid renaming event classes to ModEvent, just imho though

src/main/java/com/minecolonies/api/eventbus/DefaultEventBus.java
Thodor12: That doesn't matter, for one `? extends Event` is not valid for arguments, that's not valid syntax, only for type arguments.
Also the problem lies in the fact that Java cannot match the type argument for the key and value of the map to be identical. So even if I were to put a generic type on it, it won't be able to match it.
Lastly I have to do this cast because a method which takes an argument of `? extends Event` does not accept `Event` itself, because java dumb.

someaddons (approved):

PR #10470, opened 2024-11-19
avoid weird itemstacks chosen for ore

Raycoms:
Closes #
Closes #

- Fix weird duping issue
-

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
+++ b/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
@@ -694,7 +694,8 @@ public class CompatibilityManager implements ICompatibilityManager
             {
                 smeltableOres.add(new ItemStorage(stack));
             }
-            else if (stack.getItem() instanceof BlockItem)
+
+            if (stack.getItem() instanceof BlockItem)
             {
                 oreBlocks.add(((BlockItem) stack.getItem()).getBlock());
             }
--- a/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
+++ b/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
@@ -689,13 +689,14 @@ public class CompatibilityManager implements ICompatibilityManager
     {
         if (stack.is(Tags.Items.ORES) || stack.is(ModTags.breakable_ore) || stack.is(ModTags.raw_ore))
         {
-            if (stack.getItem() instanceof BlockItem)
-            {
-                oreBlocks.add(((BlockItem) stack.getItem()).getBlock());
-            }
-            if (!MinecoloniesAPIProxy.getInstance().getFurnaceRecipes().getSmeltingResult(stack).isEmpty())
+            final ItemStack smeltingResult = MinecoloniesAPIProxy.getInstance().getFurnaceRecipes().getSmeltingResult(stack);
+            if (smeltingResult != null && !ItemStackUtils.compareItemStacksIgnoreStackSize(smeltingResult, stack))
             {
                 smeltableOres.add(new ItemStorage(stack));
+            }
+            else if (stack.getItem() instanceof BlockItem)
+            {
+                oreBlocks.add(((BlockItem) stack.getItem()).getBlock());
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -689,13 +689,15 @@ private void discoverOres(final ItemStack stack)
     {
         if (stack.is(Tags.Items.ORES) || stack.is(ModTags.breakable_ore) || stack.is(ModTags.raw_ore))
         {
-            if (stack.getItem() instanceof BlockItem)
+            final ItemStack smeltingResult = MinecoloniesAPIProxy.getInstance().getFurnaceRecipes().getSmeltingResult(stack);
+            if (smeltingResult != null && !ItemStackUtils.compareItemStacksIgnoreStackSize(smeltingResult, stack))
```
uecasm: `smeltingResult` should never be null, it returns `ItemStack.EMPTY` on failure to find a recipe.

Other than that (which seems like a bug) this code seems equivalent to the prior version.

src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
```diff
@@ -689,13 +689,15 @@ private void discoverOres(final ItemStack stack)
...
             {
-                oreBlocks.add(((BlockItem) stack.getItem()).getBlock());
+                smeltableOres.add(new ItemStorage(stack));
             }
-            if (!MinecoloniesAPIProxy.getInstance().getFurnaceRecipes().getSmeltingResult(stack).isEmpty())
+
+            if (stack.getItem() instanceof BlockItem)
             {
-                smeltableOres.add(new ItemStorage(stack));
+                oreBlocks.add(((BlockItem) stack.getItem()).getBlock());
```
uecasm: Should use pattern variable.

src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
MotionlessTrain: The code is reversed, so it checks for a smelting recipe first, and marks it as being broken in other cases

Which solves the issue for raw redstone ore and ancient debris, but will probably give issues for regular ore blocks

Raycoms: It's just a sanity check, so we don't add sth that just becomes itself here as a smeltable item. The if is still checked for both items/blocks

src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -623,7 +623,7 @@ private List<ItemStack> insertCraftedItems(final @NotNull List<ItemStorage> inpu
         }
 
         if (loot == null && lootTable != null)
-        {
+        {
```
uecasm: 👀

src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
uecasm: > The code is reversed, so it checks for a smelting recipe first, and marks it as being broken in other cases

The reversed order makes no difference, because the two checks are not exclusive and operate on different output collections.

MotionlessTrain: I missed that the `else` was removed, which solved the issue for regular ores

uecasm: It wasn't removed; it was never there.  Hence why I'm saying that this code effectively did not change behaviour at all.

Raycoms: I added an else temporarily there, but removed it afterwards. This code change just avoids blocks with buggy smelting recipes from being added as a smeltable ore.

MotionlessTrain: I happened to see the version with the temporary `else`

someaddons (approved):

Raycoms:
closed in favor of https://github.com/ldtteam/minecolonies/pull/11243

PR #10471, opened 2024-11-20
Add JEI request search to crafting teaching window

uecasm:
Recreates #10438 (it was merged by accident and reverted)

- Adds a "requests" button to the crafting and furnace teaching GUIs.
    image
    - Button is hidden if you're silly enough to not have JEI installed.  (It wasn't _completely_ useless in that case, as it still showed requests, but it wasn't much of an improvement over the clipboard without JEI.)
    - Clicking the button shows the list of current clipboard item requests.
        - If a request accepts multiple items, all of them are shown.
        - It does not show DO-only blocks, since those are known to be handled elsewhere.
        - It does not otherwise filter the requests, since I couldn't think of a good way to do that (validity depends on input ingredients too, which aren't known yet).
    - Clicking "select" on a request will try to open JEI to a recipe for that item.
    - If you happen to be on the right page (or switch to it) for the underlying craft (i.e. crafting table or furnace) then you can click the Plus button to transfer the recipe to the teaching grid (as you could before).
- No button was added to the potion brewing GUI -- there's no technical reason this couldn't happen but it seemed less useful as it's less likely for there to be requests for potions outside of explicit player setup anyway.
- ~The button is also not shown on the 2x2 crafters (builder and miner) because there isn't a good way to limit it to only show 2x2-capable recipes, so it seemed more noisy than helpful.~

## Testing

Review please (do not port; this will need changes)


uecasm:
Bump

someaddons:
Can we add a pre-check for the item to add if it can be teached at this building at all? e.g. on the sawmill do not display latern recipe of a request, since laterns cannot be taught there anyways

uecasm:
> Can we add a pre-check for the item to add if it can be teached at this building at all? e.g. on the sawmill do not display latern recipe of a request, since laterns cannot be taught there anyways

As noted, not easily.  We could block based on product tags but we can't block based on ingredient tags.  It seemed tidier to me to not attempt to filter rather than to do only a half-hearted check (it's easier to document that way).

The current JEI API doesn't even let us filter to only show furnace or non-furnace recipes, though it's pretty obvious that those don't "fit" since the don't get the Plus button.  Though we **could** do that if we bump the min JEI version a bit.

someaddons (approved):

PR #10472, opened 2024-11-20
Use teleport permission for alliance

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/473575384445878273/1307643419354464286)

- Colonies are now considered as "allies" if the owners mutually have "teleport to colony" permission in the other colony.
    - By default, Friend and higher have this permission enabled, so now Friend is sufficient to enable teleportation; Manager is no longer required.
    - This permission didn't otherwise do anything at all in current versions.
    - Teleportation still requires the originating colony to be at least level 3 and the person doing the teleportation requires permission to access hut blocks (usually also Friend and above, but beware: it's possible to be unable to return if the destination colony is less than level 3 or you don't have permission).
    - Abandoned colonies still disable teleportation, even if you're officer/manager.
- Rescales teleport button text so that it's actually readable.

## Testing

Review please (could port)

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/ColonyView.java
+++ b/src/main/java/com/minecolonies/core/colony/ColonyView.java
@@ -8,6 +8,7 @@ import com.minecolonies.api.colony.buildings.views.IBuildingView;
 import com.minecolonies.api.colony.buildings.workerbuildings.ITownHallView;
 import com.minecolonies.api.colony.fields.IField;
 import com.minecolonies.api.colony.managers.interfaces.*;
+import com.minecolonies.api.colony.permissions.Action;
 import com.minecolonies.api.colony.permissions.ColonyPlayer;
 import com.minecolonies.api.colony.permissions.IPermissions;
 import com.minecolonies.api.colony.permissions.Rank;
@@ -31,6 +32,7 @@ import com.minecolonies.core.colony.buildings.views.AbstractBuildingView;
 import com.minecolonies.core.colony.buildings.workerbuildings.BuildingTownHall;
 import com.minecolonies.core.colony.managers.ResearchManager;
 import com.minecolonies.core.colony.managers.StatisticsManager;
+import com.minecolonies.core.colony.permissions.Permissions;
 import com.minecolonies.core.colony.permissions.PermissionsView;
 import com.minecolonies.core.colony.requestsystem.management.manager.StandardRequestManager;
 import com.minecolonies.core.colony.workorders.AbstractWorkOrder;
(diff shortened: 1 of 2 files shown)


uecasm:
Side note: when you actually click the button you get a confirmation message in chat that you have to click on, which seems more annoying than helpful.  Is there a good reason for that or can I kill it?

Raycoms:
I would leave the original ally requirement in place, as this was not created for the sake of teleportation alone and will be subject to some larger reworks in the future. I would target the problem from the other side. I would add a permission for "getting achievements" people can add to ranks instead.

uecasm:
> I would leave the original ally requirement in place, as this was not created for the sake of teleportation alone and will be subject to some larger reworks in the future.

But currently it's not used for anything else, and at such a time that the rework happens the teleportation can be reworked too.

Though "owners being mutually friends" does not sound like a bad initial definition for "these colonies are allies", regardless.

> I would target the problem from the other side. I would add a permission for "getting achievements" people can add to ranks instead.

That might be useful for other reasons, but would not solve this problem, since it would still limit teleportation to managers -- and managers have significantly more power in a colony.  **That** is what I am actually trying to fix, to reduce the number of colony managers.  That rank should only really be used for truly co-op colonies, not for "occasional helper" or "just wants to teleport".

And it would require UI rework to allow teleportation without Ally status, if that were left unmodified.

So this is both the minimal effective change for teleportation and also a good way to reduce managers.

Raycoms:
> > I would leave the original ally requirement in place, as this was not created for the sake of teleportation alone and will be subject to some larger reworks in the future.
> 
> But currently it's not used for anything else, and at such a time that the rework happens the teleportation can be reworked too.
> 
> Though "owners being mutually friends" does not sound like a bad initial definition for "these colonies are allies", regardless.
> 
> > I would target the problem from the other side. I would add a permission for "getting achievements" people can add to ranks instead.
> 
> That might be useful for other reasons, but would not solve this problem, since it would still limit teleportation to managers -- and managers have significantly more power in a colony. **That** is what I am actually trying to fix, to reduce the number of colony managers. That rank should only really be used for truly co-op colonies, not for "occasional helper" or "just wants to teleport".
> 
> And it would require UI rework to allow teleportation without Ally status, if that were left unmodified.
> 
> So this is both the minimal effective change for teleportation and also a good way to reduce managers.

yes, I understand your reasoning. But the teleport here was never intentioned for everyone in the world to super easily set up teleportation links, but rather for close allies to teleport to eachother. The future setup will require a bit more of a setup for this, so we don't want to make it a whole lot easier now just to "remove" this later in favor of the new system.

Raycoms:
Superceded by the soon to be coming colony connection (within next week)

PR #10509, opened 2024-12-07
Raid fixes

someaddons:
Closes #
Closes #

- Rework raid messages for raid ending, they now got an individual finished message followed by a general victory message for the whole raid
- Fix campfires getting spawned incorrectly, due to not loading a chunk when required to
- Reworked North/South directions util, it is now split into 8 sections a 45° to give proper ranges for North/Northeast etc
- Fix door break speed, research no longer strengthens normal doors and doors break slightly faster with more raiders hitting it.
- Improved despawn of few remaining raiders to not stall raids for longer than needed
- Full raid end is now handled by the raid manager
- Added more message variants to raid end messages
- Removed attackspeed scaling from archers, as those already scale in attack damage

## Testing

Review please


Raycoms (approved):

src/main/java/com/minecolonies/core/entity/ai/workers/guard/RangerCombatAI.java
```diff
@@ -233,9 +235,7 @@ protected double getAttackDistance()
     @Override
     protected int getAttackDelay()
     {
-        // TODO: Maybe better for balancing to not increase damage and speed, looks odd and drains arrows/bow durability
-        final int attackDelay = RANGED_ATTACK_DELAY_BASE - (user.getCitizenData().getCitizenSkillHandler().getLevel(Skill.Adaptability));
-        return Math.max(attackDelay, PHYSICAL_ATTACK_DELAY_MIN * 2);
+        return RANGED_ATTACK_DELAY_BASE;
```
Raycoms, line 238: Maybe we want to alter base slightly then?

someaddons: based 60 ticks seemed nice visually, so dont think we need to

PR #10525, opened 2024-12-11
Dontcrash

Raycoms:
Closes #
Closes #

- Dont crash with empty stack, but log error so we can find origin more easily. Better log spam than exceptions
-

## Testing

Review please


src/main/java/com/minecolonies/api/colony/requestsystem/requestable/crafting/AbstractCrafting.java
```diff
@@ -59,7 +60,7 @@ public AbstractCrafting(@NotNull final ItemStack stack, final int count, final i
 
         if (ItemStackUtils.isEmpty(stack))
         {
-            throw new IllegalArgumentException("Cannot deliver Empty Stack.");
+            Log.getLogger().error("Created Empty Stack", new Exception());
```
someaddons, line 63: should return here to prevent setting the size on the empty staack below

someaddons (approved):

PR #10557, opened 2024-12-26
Fix jmap integration

Thodor12:
Closes [discord report](https://discord.com/channels/139070364159311872/1003402487417548801/1321768200186236979) Journeymap no longer seems to be working

- Removed a SubscribeEvent annotation, this caused Neoforge to assume the method was a Neoforge event handler, throwing an exception because the parameter wasn't an Event class. This was a silent exception caused during the Jmap plugin load, causing the plugin not to start.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/compatibility/journeymap/EventListener.java
+++ b/src/main/java/com/minecolonies/core/compatibility/journeymap/EventListener.java
@@ -54,7 +54,6 @@ public class EventListener
         }
     }
 
-    @SubscribeEvent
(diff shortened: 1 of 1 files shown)


Raycoms:
But is this still properly called?

Thodor12:
Yeah, I booted it up and made sure it works, and it does. In 1.20 it works too so that ain't the problem.

uecasm:
It worked when we were using the shared event bus.  It still seems weird that we've split that off; it's likely to confuse addon mod makers by not following conventions.

PR #10599, opened 2025-01-15
Improve permission handling around interaction events

Thodor12:
- Fixed several warnings from appearing when attempting to open any building with only the ACCESS_HUT permission
- Fixed a bug that caused additional warnings to appear upon townhall opening due to the color list sending a color change event upon window open (unnecessary event send removed)
- Split the open inventory for buildings and citizens so individual permissions may be configured for each action (also should these be OPEN_CONTAINER, ACCESS_HUT or MANAGE_HUT? Currently MANAGE_HUT (potentially related, technically accessing the inventory should be ACCESS_HUT, but modifying the inventory should be MANAGE_HUT, can we make an inventoy readonly?)
- Improved event handling in the permission event handler
  - Right click block interactions don't chain through all possible options (previously, to open a hut block it requires ACCESS_HUT permission, if the block was a hut block, and person did have ACCESS_HUT permission, the method would continue instead of returning, causing other permissions to be checked unnecessarily, potentially throwing more warnings)
  - Trapdoors have been added to toggleabes
  - OPEN_CONTAINER doesn't check for any BaseEntityBlock, rather it checks if the tile entity implements Container
  - Right click items have been separated from right click blocks
  - Potion throwing permission checks for any item implementing ThrowablePotionItem, rather than PotionItem (so person may still drink self-consume potions)

## Testing

Review please


someaddons (dismissed):

src/main/java/com/minecolonies/core/client/gui/townhall/WindowMainPage.java
```diff
@@ -203,10 +171,7 @@ public String getLabel(final int index)
      */
     private void toggleTexture(final DropDownList dropDownList)
```
Raycoms, line 172: I remember we had those ifs so that it doesn't toggle on UI open already or so. (no need to toggle for same as selected)

Raycoms:
Did you test this very well for all the permission things you changed, both where you should have and shouldn't have permission? It touches a lot of stuff, so it should be well tested

src/main/java/com/minecolonies/core/client/gui/townhall/WindowMainPage.java
Thodor12: I fixed that by actually calling the methods in the correct order during setup.
-> First assign data provider
-> Set selected index
-> Set update handler

Thodor12: Also, the if statements had a negative side effect, because changing back to the initially selected item was not possible because the final values did not update, so that's another bug fixed

Thodor12:
I ensured that all events under every condition work as they should, I had a MP server open where I slowly granted the other player additional permissions to see if all the conditions worked as they should.

Note: There are some odd behavioural ones, but that's unfortunate due to how Mojang calls these different events.
For example you may not place blocks if you don't have right click permission, because the entityPlacement gets called first, and upon not having right click permission, you may not actually place the block, despite having placement permissions.

src/main/java/com/minecolonies/core/event/PermissionEventHandler.java
```diff
@@ -0,0 +1,667 @@
...
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public static void on(final BreakEvent event)
+    {
+        final LevelAccessor world = event.getLevel();
+        if (!(world instanceof Level level) || level.isClientSide)
+        {
+            return;
+        }
+
+        validatePrecondition(level, event.getPos(), (colony) -> {
```
Raycoms: I'm not a huge fan of this lambda wrapper. Can we do just Colony colony = validatePreconditionAndFetchColony?

src/main/java/com/minecolonies/core/network/messages/server/colony/InteractionClose.java
```diff
@@ -52,6 +54,13 @@ public InteractionClose(
         this.key = key;
     }
 
+    @Override
+    @Nullable
+    public Action permissionNeeded()
```
Raycoms, line 59: Shouldn't all interactions need the same condition?

Thodor12: So the reason I did this is because there is no reason for a message that closes a window to require any kind of permission level, on top of that the default is MANAGE_HUTS, which is wrong to begin with

src/main/java/com/minecolonies/core/event/PermissionEventHandler.java
Thodor12: So the intention here is to remove identical code (in this case, a bunch of extra if statements) on every method that utilizes this callback. Having it a callback means it's only invoked when the precondition is actually met.

I can change it if you desperately want to, but in order to reduce duplicate code I think this is better

src/main/java/com/minecolonies/core/network/messages/server/colony/InteractionClose.java
Raycoms: I mean, we should just have all interactions have the same and not have a new function in all of them for some edge case

Thodor12: This is an override, this method exists in the abstract message and is set by default to MANAGE_HUTS, I haven't added anything new, just lowered the permissions required for triggering this message

src/main/java/com/minecolonies/core/event/PermissionEventHandler.java
Raycoms: yeah, returns null colony is the condition is not met, so if null return

Raycoms (approved):
Before merge, please do another SMP test and test some fundamental interactions with 2 colonies existing in the same world.

PR #10651, opened 2025-02-09
feat: add minimum stock to the postbox

mcmanustfj:
## Testing

image

Structural notes that might need changed:

 - As noted in discord, I copied a bunch of code, which doesn't feel great.  `WindowPostBox extends AbstractWindowRequestTree extends AbstractWindowSkeleton`, but the existing code for rendering those tabs is in `AbstractModuleWindow extends AbstractWindowSkeleton`.  I copied the code to WindowPostBox because I don't see a good way of refactoring the inheritance to make it work without multiple inheritance (which isn't in Java)

 - The Minimum Stock module gets the amount of minimum stocks based on the hut level.  I added a check to use '1' if the building `getMaxBuildingLevel() == 0`.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/client/gui/AbstractModuleWindow.java
+++ b/src/main/java/com/minecolonies/core/client/gui/AbstractModuleWindow.java
@@ -6,6 +6,7 @@ import com.minecolonies.api.colony.buildings.modules.IBuildingModuleView;
 import com.minecolonies.api.colony.buildings.modules.IModuleWindow;
 import com.minecolonies.api.colony.buildings.views.IBuildingView;
 import com.minecolonies.core.colony.buildings.views.AbstractBuildingView;
+import net.minecraft.client.Minecraft;
 import net.minecraft.client.resources.sounds.SimpleSoundInstance;
 import net.minecraft.network.chat.Component;
 import net.minecraft.resources.ResourceLocation;
@@ -32,12 +33,19 @@ public abstract class AbstractModuleWindow extends AbstractWindowSkeleton implem
      * @param building   {@link AbstractBuildingView}.
      * @param res        the resource String.
      */
+
     public AbstractModuleWindow(final IBuildingView building, final String res)
     {
         super(res);
         this.buildingView = building;
+        placeWindowModuleViews(this, building, -20, this.mc);
+
+    }
+    // need mc as its own arg because it's protected in Pane
+    static void placeWindowModuleViews(AbstractWindowSkeleton window, final IBuildingView building, int xOffset, Minecraft mc) {
         final Random random = new Random(building.getID().hashCode());
         int offset = 0;
+        final int iconXOffset = xOffset+5;
 
         boolean anyVisible = false;
 
@@ -50,26 +58,26 @@ public abstract class AbstractModuleWindow extends AbstractWindowSkeleton implem
             }
         }
 
-            //todo We have to move this to 0 as soon as we're finished with modularization and remove the switch views in favor of a sidenav xml.
-        if (building.getAllModuleViews().size() > 0 && anyVisible)
+        //todo We have to move this to 0 as soon as we're finished with modularization and remove the switch views in favor of a sidenav xml.
+        if (!building.getAllModuleViews().isEmpty() && anyVisible)
         {
             final ButtonImage image = new ButtonImage();
             image.setImage(new ResourceLocation("minecolonies:textures/gui/modules/tab_side" + (random.nextInt(3) + 1) + ".png"), false);
-            image.setPosition(-20, 10 + offset);
+            image.setPosition(xOffset, 10 + offset);
             image.setSize(32, 26);
             image.setHandler(button -> building.getWindow().open());
 
             final ButtonImage iconImage = new ButtonImage();
             iconImage.setImage(new ResourceLocation("minecolonies:textures/gui/modules/main.png"), false);
             iconImage.setID("main");
-            iconImage.setPosition(-15, 13 + offset);
+            iconImage.setPosition(iconXOffset, 13 + offset);
             iconImage.setSize(20, 20);
             iconImage.setHandler(button -> building.getWindow().open());
 
             offset += image.getHeight() + 2;
 
-            this.addChild(image);
-            this.addChild(iconImage);
+            window.addChild(image);
+            window.addChild(iconImage);
 
             PaneBuilders.tooltipBuilder().hoverPane(iconImage).build().setText(Component.translatable(LABEL_MAIN_TAB_NAME));
         }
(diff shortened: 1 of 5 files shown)


Thodor12:
> The Minimum Stock module gets the amount of minimum stocks based on the hut level. I added a check to use '1' if the building getMaxBuildingLevel() == 0.

I wouldn't touch this behavior. Otherwise level 0 huts could start having min stocks which is not allowed.

I'd rather create a specific implementation of the min stock module and have a separate module producer.

Use this new implementation for the postbox only, and give a way to change the implementation for getting the limit.

src/main/java/com/minecolonies/core/colony/buildings/modules/MinimumStockModule.java
```diff
@@ -57,8 +57,9 @@ public class MinimumStockModule extends AbstractBuildingModule implements IMinim
     private int minimumStockSize()
     {
         final double increase = 1 + building.getColony().getResearchManager().getResearchEffects().getEffectStrength(MINIMUM_STOCK);
+        final int buildingLevelOr1 = building.getMaxBuildingLevel() != 0 ? building.getBuildingLevel() : 1;
 
-        return (int) (building.getBuildingLevel() * STOCK_PER_LEVEL * increase);
+        return (int) (buildingLevelOr1 * STOCK_PER_LEVEL * increase);
```
Raycoms: Should maybe be a buildingLevel or maxlevel = 5 (postbox/stash should be maxlevel by default)

src/main/java/com/minecolonies/core/client/gui/WindowPostBox.java
```diff
@@ -97,6 +100,75 @@ public WindowPostBox(final AbstractBuildingView buildingView)
...
+        {
+            if (view.isPageVisible())
+            {
+                anyVisible = true;
+                break;
+            }
+        }
+
+
+        if (buildingView.getAllModuleViews().size() > 0 && anyVisible)
```
Raycoms: If that code is exactly the same as in the other class, we could make this a utility method and call the utility method from both places

mcmanustfj:
> > The Minimum Stock module gets the amount of minimum stocks based on the hut level. I added a check to use '1' if the building getMaxBuildingLevel() == 0.
> 
> I wouldn't touch this behavior. Otherwise level 0 huts could start having min stocks which is not allowed.
> 
> I'd rather create a specific implementation of the min stock module and have a separate module producer.
> 
> Use this new implementation for the postbox only, and give a way to change the implementation for getting the limit.

This only uses 1 if the *max* hut level = 0, which AFAIK is only stash and postbox? not that stash would get minimum stock

src/main/java/com/minecolonies/core/colony/buildings/modules/MinimumStockModule.java
Thodor12: This is why I said make a specific override of the min stock module, create a method that returns the stock level, and stock per level, and let the postbox override these with different values (not building level)

src/main/java/com/minecolonies/core/client/gui/WindowPostBox.java
mcmanustfj: the offsets are slightly different, but I can add that as a parameter

src/main/java/com/minecolonies/core/colony/buildings/modules/MinimumStockModule.java
mcmanustfj: I think that a special case is allowed to be special when there's only one of it, but if there's another special case then it's worth refactoring + generalizing.  As it is, there's the general case of hut blocks with levels, and the special case of hut blocks without levels (just postbox), so I think it's fine to leave as is.  

If you decide you want to make it affected differently by research or something, then you should extend the min stock module and move stuff around.

src/main/java/com/minecolonies/core/client/gui/AbstractModuleWindow.java
```diff
@@ -32,12 +33,19 @@ public abstract class AbstractModuleWindow extends AbstractWindowSkeleton implem
...
         super(res);
         this.buildingView = building;
+        placeWindowModuleViews(this, building, -20, this.mc);
+
+    }
+    // need mc as its own arg because it's protected in Pane
+    static void placeWindowModuleViews(AbstractWindowSkeleton window, final IBuildingView building, int xOffset, Minecraft mc) {
         final Random random = new Random(building.getID().hashCode());
         int offset = 0;
+        final int iconXOffset = xOffset+5;
```
Thodor12, line 48: Formatting

src/main/java/com/minecolonies/core/client/gui/AbstractModuleWindow.java
```diff
@@ -32,12 +33,19 @@ public abstract class AbstractModuleWindow extends AbstractWindowSkeleton implem
      * @param building   {@link AbstractBuildingView}.
      * @param res        the resource String.
      */
+
```
Thodor12, line 36: Remove newline (docs should be attached to class)

src/main/java/com/minecolonies/core/client/gui/AbstractModuleWindow.java
```diff
@@ -32,12 +33,19 @@ public abstract class AbstractModuleWindow extends AbstractWindowSkeleton implem
...
     public AbstractModuleWindow(final IBuildingView building, final String res)
     {
         super(res);
         this.buildingView = building;
+        placeWindowModuleViews(this, building, -20, this.mc);
+
```
Thodor12, line 42: Remove newline

src/main/java/com/minecolonies/core/client/gui/AbstractModuleWindow.java
```diff
@@ -32,12 +33,19 @@ public abstract class AbstractModuleWindow extends AbstractWindowSkeleton implem
...
+    }
+    // need mc as its own arg because it's protected in Pane
```
Thodor12, line 44: Add newline between methods

src/main/java/com/minecolonies/core/client/gui/AbstractModuleWindow.java
```diff
@@ -32,12 +33,19 @@ public abstract class AbstractModuleWindow extends AbstractWindowSkeleton implem
...
+    static void placeWindowModuleViews(AbstractWindowSkeleton window, final IBuildingView building, int xOffset, Minecraft mc) {
```
Thodor12, line 45: Give methods the correct access modifier, if this is intended to be called by subclasses only, it should be protected

src/main/java/com/minecolonies/core/colony/buildings/modules/MinimumStockModule.java
```diff
@@ -57,8 +57,9 @@ public class MinimumStockModule extends AbstractBuildingModule implements IMinim
     private int minimumStockSize()
     {
         final double increase = 1 + building.getColony().getResearchManager().getResearchEffects().getEffectStrength(MINIMUM_STOCK);
+        final int buildingLevelOr5 = building.getMaxBuildingLevel() != 0 ? building.getBuildingLevel() : 5;
 
-        return (int) (building.getBuildingLevel() * STOCK_PER_LEVEL * increase);
+        return (int) (buildingLevelOr5 * STOCK_PER_LEVEL * increase);
```
Thodor12, line 62: Again, please remove this change and override `getBuildingLevel()` of the postbox instead

Thodor12 (changes requested):

mcmanustfj:
I think after discussion in discord, it was decided that y'all don't want this implemented without changing the layout and adding an upgradable tier for the postbox or similar, which I'm not willing to do at this time.  I welcome someone else picking this up, though

PR #10666, opened 2025-02-17
Style Fixes

Moobien:
Closes #
Closes #

1: removed pagoda alt citizen from warped style
2: rescanned supplyship/nethership/ship/camp/supplycamp
3: added nethership alongside supply ship

## Testing

Review please


MotionlessTrain:
I added the warped nether ship in #10648 
Not sure why that is not reflected here (it would be an update rather than an addition)

Raycoms:
> I added the warped nether ship in #10648 Not sure why that is not reflected here (it would be an update rather than an addition)

It doesn't reflect because this branch is based on an outdated branch (that's why there are conflicts too).
Please, remove things like the reload anchor and some other special things for the overworld version of the ship then.

Raycoms:
Closing, in favor of new one with fixes =D

PR #10677, opened 2025-02-23
Fix unnecessary item handler invalidation

### LINKED ISSUE #10670: Warehouse item capability is invalidated every tick leading to compatibility issues

Thodor12:
Closes #10671 
Closes #10670

- Move container removal to onColonyTick in the building
- Item handler capability will only be recalculated if the underlying container list has changed
- Removed obsolete code in CombinedItemHandler (the nameable has no effect, nor any usages)

## Testing

Review please


someaddons (dismissed):

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingContainer.java
```diff
@@ -71,6 +71,29 @@ public AbstractBuildingContainer(final BlockPos pos, final IColony colony)
...
 
+    @Override
+    public void onColonyTick(final IColony colony)
+    {
+        super.onColonyTick(colony);
+
+        // Clean containers that are no longer detected as racks
+        for (final BlockPos pos : new ArrayList<>(containerList))
+        {
+            if (WorldUtil.isBlockLoaded(colony.getWorld(), pos) && !pos.equals(getPosition()))
```
Raycoms: I guess we should do this in the tileentity where we're ticking anyway?

Thodor12:
You mean like this @Raycoms https://github.com/ldtteam/minecolonies/pull/10677/commits/e317346127403fd873a48470fbee3af6f6cc3f42

TrevorCow:
Forgive me if this is not helpful, I am not super familiar with this codebase, and I am probably doing close to the most unsupported thing (Trying to fix a bug in the infamous Compatibility addon for Minecolonies), but I digress. I am running a local compiled version of this patch, backported to `version/main` for 1.20.1, and think that we might be forgetting to add the colony building itself when rebuilding the `combinedInv`/`capability`. I had numerous problems with couriers not being able to deliver to guard towers and other "colony block only" buildings. Adding 

## TileEntityColinyBuilding
```diff
if (combinedInv == null)
{
    final List<IItemHandlerModifiable> handlers = new ArrayList<>();
    final Set<BlockPos> newPositions = new HashSet<>();

    final Level world = colony.getWorld();

    for (final BlockPos pos : building.getContainers())
    {
        if (WorldUtil.isBlockLoaded(world, pos) && !pos.equals(this.worldPosition))
        {
            final BlockEntity te = world.getBlockEntity(pos);
            if (te != null)
            {
                if (te instanceof final AbstractTileEntityRack rack)
                {
                    handlers.add(rack.getInventory());
                    newPositions.add(pos);
                }
            }
        }
    }

+    handlers.add(this.getInventory());
+    newPositions.add(this.getPosition());

    combinedInv = LazyOptional.of(() -> new CombinedItemHandler(handlers));
    currentInvPositions = newPositions;
``` 
seemed to fix that. I don't know if this is the best option but thought it might be of help.

someaddons:
We actually went about this the wrong way a bit, we should change it to add an invalidatin listener to a rack, which when a rack becomes unavailable triggers the reset/invalidation of the combined inventory capability and that way forces a new one to be created

Nightenom (commented):
Just blind-shot: could we use/copy/extend https://github.com/neoforged/NeoForge/blob/c01798ea3a4a17e8b1c5bf917444e9990d3b85c3/src/main/java/net/neoforged/neoforge/items/wrapper/CombinedInvWrapper.java

PR #10695, opened 2025-03-08
Disease Durations

LetterShapedGlyphs:
initial config and default values for disease duration

- Sets a maximum number of days a citizen can be sick before natural recovery
- Sets the default for that to 5 days
- Makes that value configurable

## Testing

First time touching anything config related. I tried to mirror how the surrounding code was set up but I might have missed things.
Since a citizen sleeping in a hospital will get better in less than one minecraft day (on average) even without a cure, I started with a default of 100 minutes (5 days). Long enough to be noticeable but close to the existing timescale things operate on.

In an ideal world this duration would be set on a disease level with a config based multiplier.

### DIFF
--- a/src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
+++ b/src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
@@ -30,6 +30,7 @@ public class ServerConfiguration extends AbstractConfiguration
     public final ForgeConfigSpec.IntValue     minThLevelToTeleport;
     public final ForgeConfigSpec.DoubleValue  foodModifier;
     public final ForgeConfigSpec.IntValue     diseaseModifier;
+    public final ForgeConfigSpec.IntValue     diseaseDuration;
     public final ForgeConfigSpec.BooleanValue forceLoadColony;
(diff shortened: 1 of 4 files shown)


someaddons:
Since diseases are defined via datapacks this should also be part of the datapack, which would also allow setting a time per disease. Default values should be large, ~ 100 days as citizens currently are expected to be sick for long until cured or the tiny random chance cures them

LetterShapedGlyphs:
> Default values should be large, ~ 100 days as citizens currently are expected to be sick for long until cured or the tiny random chance cures them

That "tiny random chance" is that a citizen will spontaneously recover even without treatment in less than one day as long as they have access to a hospital bed.

That said I agree that it would be better to have per disease recovery times but I've already gotten so much pushback on this I'm not keen on spending the effort doing that unless I get any sort of indication that would actually be merged.

Raycoms:
I'd be in favor of having different disease durations as one of the datapack options per disease. Allows us to play with "citizen has the flu for a day" kind of things. @someaddons what do you think?

someaddons:
yes as datapack

Raycoms:
Okay, I will close this in favor of a "future PR with datapack modifiable duration values per disease"

PR #10705, opened 2025-03-10
Allow all researches to be undone + moving some researches

Thodor12:
Closes [discord report](https://discord.com/channels/139070364159311872/1284909613401247785/1348685271390224395)
Closes any other report of identical nature

- Ensured that all researches may be undone, assuming they not immutable or auto starting, or they have any child that is already researched
- Moved the nimble research as a direct child of keen, allowing rails to be undone if so wished
  - This additionally caused nimble and all researches below to move down 1 level
- Moved Minimum order quantity to technology, as a sibling of warehouse master, rather than a child of rails (which makes no sense)
- Ensured that citizen inventory researches are immutable, so that inventory upgrades cannot be undone (which could result in item loss)

## Testing

Review please


someaddons:
I'm not sure if we really want to move a level 6 research to level 5, level 6 reseachres were meants as a unique, choose one only

Thodor12:
> I'm not sure if we really want to move a level 6 research to level 5, level 6 reseachres were meants as a unique, choose one only

I did say this but didn't really get an appropriate answer to it, I was wondering if it's a good idea to introduce another tier of the speed researches

someaddons:
can't we just connect a tier 4 to a tier 6?

Thodor12:
Datagen builders technically do not allow this as they increment the parents always by a + 1, but I also don't know if the research tree supports jumping like that. cc @Raycoms

Raycoms (approved):

PR #10725, opened 2025-03-22
add colony map item & improve map data

Raycoms:
Closes #
Closes #

- Add Colony Map Item (needs to be registered with townhall)
- Add complain text of interactions to map
- Add overlay render of how many citizens have an urgent interaction
- Add guard tower claim range to map on hover
- Add sick/hungry overlay to map

## Testing

Review please


someaddons (commented):
looks fine just needs the better icon first

PR #10742, opened 2025-03-31
Cleanup registry logic and API instances

Thodor12:
- Added javadoc to the API instances
- Use shared reslocs for the deferred registers similar to 1.21

## Testing

Review please


src/main/java/com/minecolonies/api/MinecoloniesAPIProxy.java
```diff
@@ -100,93 +109,93 @@ public IJobDataManager getJobDataManager()
...
-    public IInteractionResponseHandlerDataManager getInteractionResponseHandlerDataManager()
+    public IForgeRegistry<GuardType> getGuardTypeRegistry()
     {
-        return apiInstance.getInteractionResponseHandlerDataManager();
+        return apiInstance.getGuardTypeRegistry();
     }
 
     @Override
-    public IGlobalResearchTree getGlobalResearchTree()
+    public IForgeRegistry<InteractionResponseHandlerEntry> getInteractionResponseHandlerRegistry()
```
Raycoms, line 160: Please don't re-order these, that's always a nightmare for difs

src/main/java/com/minecolonies/apiimp/CommonMinecoloniesAPIImpl.java
```diff
@@ -382,27 +408,41 @@ public IForgeRegistry<QuestRegistries.DialogueAnswerEntry> getQuestDialogueAnswe
...
     @Override
-    public EventBus getEventBus()
+    public IModelTypeRegistry getModelTypeRegistry()
     {
-        return eventBus;
+        return null;
+    }
+
+    @Override
+    public IGlobalResearchTree getGlobalResearchTree()
```
Raycoms, line 437: Not really possible to review what changed here because of the re-ordering too

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingKitchen.java
```diff
@@ -88,9 +88,7 @@ public boolean isRecipeCompatible(@NotNull final IGenericRecipe recipe)
             if (isRecipeAllowed.isPresent()) return isRecipeAllowed.get();
 
             final ItemStack output = recipe.getPrimaryOutput();
-            return FoodUtils.EDIBLE.test(output)
-                || FoodUtils.EDIBLE.test(FurnaceRecipes.getInstance()
-                    .getSmeltingResult(output));
+            return FoodUtils.EDIBLE.test(output) || FoodUtils.EDIBLE.test(IMinecoloniesAPI.getInstance().getFurnaceRecipes().getSmeltingResult(output));
```
Raycoms: I think it was nicer if we could query this through the class proxy which goes itself to api.getInstance, makes it more readable

Thodor12: Can do, but best case I'm making a method in the IFurnaceRecipes interface. There isn't much reason to have this appear in the IMinecoloniesAPI if it's not going to be used and people use the implementation directly.

So either it'll have to use through the IMinecoloniesAPI (either directly like this or through an interface static method), or we rip this thing out of the API completely and just let it be it's own static class. I'm good with either.

Raycoms: Yes, the interface works perfectly for this

src/main/java/com/minecolonies/apiimp/CommonMinecoloniesAPIImpl.java
Thodor12: Nothing changed, the PR does exactly what it means to do, it's a cleanup. I arranged all the different classes along together (data managers, forge registries, custom registries, other classes, etc).

The only other change that's been done there is that the names of the registries were put into reslocs which are shared across the API and the deferred registers, similar to what's been done in 1.21

Raycoms (dismissed):

someaddons (approved):

Raycoms:
@Thodor12 need 1.21 version

PR #10743, opened 2025-03-31
Format gui xmls

Thodor12:
- Added spotless using prettier to format XML files
- Format all XML GUI files

## Testing

Review please

### DIFF
--- a/src/main/resources/assets/minecolonies/gui/citizen/family.xml
+++ b/src/main/resources/assets/minecolonies/gui/citizen/family.xml
@@ -1,28 +1,29 @@
-<window xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" size="210 244"
-        xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/ldtteam/BlockUI/version/1.20.1/src/main/resources/assets/blockui/gui/block_ui.xsd">
+<window
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/ldtteam/BlockUI/version/1.20.1/src/main/resources/assets/blockui/gui/block_ui.xsd"
+    size="210 244">
     <layout source="minecolonies:gui/citizen/nav.xml"/>
 
-    <text pos="80 15" size="110 11" color="black" label="§n$(com.minecolonies.coremod.gui.citizen.family.title):§r"/>
+    <text color="black" label="§n$(com.minecolonies.coremod.gui.citizen.family.title):§r" pos="80 15" size="110 11"/>
 
-    <text size="110 11" pos="90 40" color="black" label="§n$(com.minecolonies.coremod.gui.citizen.family.parents):§r"/>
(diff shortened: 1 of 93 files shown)


src/main/resources/assets/minecolonies/gui/citizen/family.xml
```diff
@@ -1,27 +1,29 @@
...
-    
-    <text size="110 11" pos="90 40" color="black"  label="§n$(com.minecolonies.coremod.gui.citizen.family.parents):§r"/>
-    <text id="parentA" size="130 11" pos="50 51" color="black" textalign="MIDDLE"/>
-    <text id="parentB" size="130 11" pos="50 62" color="black" textalign="MIDDLE"/>
+    <text color="black" label="§n$(com.minecolonies.coremod.gui.citizen.family.title):§r" pos="80 15" size="110 11"/>
 
-    <text size="110 11" pos="90 80" color="black"  label="§n$(com.minecolonies.coremod.gui.citizen.family.partner):§r"/>
-    <text id="partner" size="130 11" pos="50 95" color="black" textalign="MIDDLE"/>
+    <text color="black" label="§n$(com.minecolonies.coremod.gui.citizen.family.parents):§r" pos="90 40" size="110 11"/>
+    <text color="black" id="parentA" pos="50 51" size="130 11" textalign="MIDDLE"/>
```
Nightenom, line 10: can we have more logical sorting for these? like id, pos, size going first (id -> size -> pos seems to be the most used one)

Raycoms: yeah, Id like id/pos/size by alphabet

Thodor12: I can't configure it that much since prettier can only do it fully by alphabet. I did look at this myself but I'm not sure if it's doable

Thodor12: I can try and do this, draft this PR for now, whilst I modify the prettier plugin in the source to allow for such an option

Thodor12:
Waiting on https://github.com/prettier/plugin-xml/pull/860

src/main/resources/assets/minecolonies/gui/citizen/family.xml
Thodor12: @Raycoms @Nightenom Is this something we could merge without this change, the linked pull request I made on the prettier plugin has been denied. So I'm not going to be able to force these properties to appear at the front.

Keep it as it is now or scrap the idea?

(Alternative is I make a fork of the plugin, and introduce the option in my own version of the plugin, but I'd rather not have to do so)

Raycoms: Could we use the intellij format rules to better format those things locally just, instead of relying on plugins?

Thodor12: Intellij doesn't really have any of the formatting rules for XML that the plugin would have, not to mention that it wouldn't work on Eclipse for those that would like to use it (if we have any).

Thodor12:
Will be opening a PR as this one is very out of date by this point

PR #10746, opened 2025-04-01
Fix visitor spawning conditions

### LINKED ISSUE #10745: Tavern will not spawn Colonist

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.1.848

### Structurize Version

1.0.768

1. Forge version 47.4.0
2. BlockUI 1.0.190
3. TownTalk 1.1.0
4. Multi-Piston 2.43-RELEASE
5. Domum Ornamentum 1.0.285

### Current Behavior

After building a level 1 Tavern and waiting colonist don't spawn. If you down grade to MineColonies version 1.1.845, and load back in to saved game the tavern will spawn colonist.

### Expected Behavior

After building a Tavern I was expecting colonist to spawn so I can find a better builder. I normally always use the starting colonist as guards because they complain, and then wait for the Tavern to spawn new colonist with better stats for builders, and others.

### Reproduction Steps

Update to version 1.1.848 of MineColonies, build a Tavern, and wait to see if they spawn. If they don't down grade to MineColonies 1.1.845, load back into saved game, and Tavern should start spawning colonist.

### Logs

https://gist.github.com/Ashlockheart/393ee242257b22367821cbfc770ef557

### Anything else?

Me, and another player think it has to do with the update in 1.1.846
Rework recruitment items to become datapacks ([#10342](https://github.com/ldtteam/minecolonies/pull/10342)) ([commit](https://github.com/ldtteam/minecolonies/commit/3d18b750543b197a233b484b93b9d675705343ab))

I removed Continuity, and other fabric mods from my pack. So if you see fabric in the log I don't know why its showing when the mods were removed. Something is still left over, but I don't know how to find it to remove it.
### Footer

---
## Viewers

Thodor12:
Closes #10745

- Ensure that rarities are matched the correct way around, not in reverse
- Ensure that all recruitment items are looked up in a null safe way, in case there are no available recruitment items present, or the rarities would match no recruitment items

## Testing

Review please


Raycoms (dismissed):

src/main/java/com/minecolonies/core/colony/buildings/modules/TavernBuildingModule.java
```diff
@@ -185,10 +186,15 @@ private void spawnVisitorInternal()
     /**
      * Spawns a visitor citizen that can be recruited.
      */
+    @Nullable
     public IVisitorData spawnVisitor()
     {
         final int recruitLevel = building.getColony().getWorld().random.nextInt(10 * building.getBuildingLevel()) + 15;
         final RecruitCost cost = RecruitmentItemsListener.getRandomRecruitCost(building.getColony().getWorld().getRandom(), recruitLevel);
+        if (cost == null)
```
someaddons, line 180: why would the recuitcost be null?

Thodor12: At the off-chance there are no data files with recruit costs or the filters return nothing, we don't want NPE's happening

someaddons: We need to log the cause of this then, recruit cost should never be null

Thodor12: Added a log statement in the listener class

src/main/java/com/minecolonies/core/datalistener/RecruitmentItemsListener.java
```diff
@@ -95,8 +98,13 @@ else if (level > IRON_SKILL_LEVEL)
         {
             minimumRarity = 0;
         }
-        final List<RecruitCost> recruitCostStream = RECRUIT_COSTS.stream().filter(f -> f.rarity <= minimumRarity).toList();
-        return recruitCostStream.get(source.nextInt(recruitCostStream.size()));
+        final List<RecruitCost> recruitCosts = RECRUIT_COSTS.stream().filter(f -> f.rarity >= minimumRarity).toList();
+        if (recruitCosts.isEmpty())
+        {
+            Log.getLogger().warn("No recruitment items found, please ensure valid recruitment items exist in the datapacks.");
```
someaddons: This should be during loading(#apply) :) So you know which datapack file and which part of the datapack is wrong, and what you need to add/change in the pack

Thodor12: It doesn't necessarily have to be during load, if one does add a single low rarity item and has no higher rarity items, it can result in no items in the filter if a higher level citizen has to spawn, hence why I put it here instead

someaddons: but the list does not change after loading, so can check the full list once and warn once about it
aka valdiate data on input

Thodor12: But then still, if I have no data on load, then what? It still can be null

someaddons: I'm not telling you to remove the null check, I'm telling you to on data load verify the items loaded and log an error if there is a problem within

Raycoms (commented):
Btw, does this solve the spawning? Because I don't see retries here if spawning fails

Thodor12:
> Btw, does this solve the spawning? Because I don't see retries here if spawning fails

There's no actual problem with the spawning itself, that code has always stayed the same, it was just unable to spawn due to a NPE occuring

Raycoms:
> > Btw, does this solve the spawning? Because I don't see retries here if spawning fails
> 
> There's no actual problem with the spawning itself, that code has always stayed the same, it was just unable to spawn due to a NPE occuring

Yes, but now we just "return on finding a null value" and don't spawn anything either, no?

Thodor12:
Yeah, if I don't have a recruit cost I can't spawn anything, nothing I can do about that unfortunately

Raycoms:
> Yeah, if I don't have a recruit cost I can't spawn anything, nothing I can do about that unfortunately

I'm still confused. This can happen if we pick a visitor that has a recruitcost that is only going to appear at higher levels.
But this should be picked by random, and even if it causes an NPE, the next attempt should allow spawning, no?

Thodor12:
> > Yeah, if I don't have a recruit cost I can't spawn anything, nothing I can do about that unfortunately
> 
> I'm still confused. This can happen if we pick a visitor that has a recruitcost that is only going to appear at higher levels. But this should be picked by random, and even if it causes an NPE, the next attempt should allow spawning, no?

Each recruit cost has a rarity attached to it, higher levels of visitors (according to the boot levels) should only have higher rarity recruit costs.
If a person would only provide datapacks for lower rarities, higher level rarities will be unable to be found (assuming all our added datapacks are removed).
In this case the visitors will continuously be unable to spawn because they will not be able to have a recruit cost, unless a lower level visitor spawns again.

If there's no recruit costs at all, no visitors can spawn at all.

This is similar to diseases, if no disease datapacks exist, no one can get sick.

Raycoms:
> > > Yeah, if I don't have a recruit cost I can't spawn anything, nothing I can do about that unfortunately
> > 
> > 
> > I'm still confused. This can happen if we pick a visitor that has a recruitcost that is only going to appear at higher levels. But this should be picked by random, and even if it causes an NPE, the next attempt should allow spawning, no?
> 
> Each recruit cost has a rarity attached to it, higher levels of visitors (according to the boot levels) should only have higher rarity recruit costs. If a person would only provide datapacks for lower rarities, higher level rarities will be unable to be found (assuming all our added datapacks are removed). In this case the visitors will continuously be unable to spawn because they will not be able to have a recruit cost, unless a lower level visitor spawns again.
> 
> If there's no recruit costs at all, no visitors can spawn at all.
> 
> This is similar to diseases, if no disease datapacks exist, no one can get sick.

Yes, I understand this. But in that case, we should retry with a vistior of lower rarity. Maybe the class that holds the data should already make sure that we only generate combinations that make sense.

Thodor12:
Best I can do is remove the spawn interval if no one was able to spawn, that wasy it's retried on the next colony tick rather than X minutes later.
I don't want to cause some infinite loop instead that will permanently attempt to respawn a visitor.

src/main/java/com/minecolonies/core/datalistener/RecruitmentItemsListener.java
```diff
@@ -41,71 +39,71 @@ public class RecruitmentItemsListener extends SimpleJsonResourceReloadListener
...
      *
-     * @param source the random source.
-     * @param level  the recruitment level.
+     * @param source        the random source.
+     * @param buildingLevel the building level.
      * @return a random recruit cost.
      */
-    public static RecruitCost getRandomRecruitCost(final RandomSource source, final int level)
+    @Nullable
+    public static RecruitCostResult getRandomRecruitCost(final RandomSource source, final int buildingLevel)
```
someaddons: use ColonyConstants.rand instead of passing/using vanilla's random sources

someaddons:
I think the better idea is to error with improper data input on loading, aka do not allow empty/missing entries on datapack load.
Could e.g. iterater from 15 to 65(min to max recruitment level) and check if we have something present that fits

Raycoms (approved):

someaddons (approved):

PR #10773, opened 2025-04-06
Workorder rework

someaddons:
Workorders are now split into 3 different interface types: 
- IWorkorder: Now is a shared type for both serverside workorders and clientside workorder views
- IServerWorkOrder contains all functions relating to saving
- IBuilderWorkOrder has builder specific functionality

Workorders now also store their associated blueprint as central access to them
- getblueprintfuture became loadBlueprint which allows easier handling and avoids sidedness/threading issues
- jobs no longer store the blueprint, and instead access/load the blueprint into the workorder

Workorders now also remember their AABB which allows checking their area, esp useful for decorations  which unlike buildings do not remember corners(also is synced to the client for rendering purposes)

Also fixes drop cost calc for pathfinding

## Testing
- [ x] I also did a multiplayer test.

Review please


Thodor12:
> jobs no longer store the blueprint, and instead access/load the blueprint into the workorder

Does this mean that having a builder die will no longer void the progress of a building?

someaddons:
> > jobs no longer store the blueprint, and instead access/load the blueprint into the workorder
> 
> Does this mean that having a builder die will no longer void the progress of a building?

nah the builder progress is stored seperatedly, but could be sth to pull into the workorder aswell

uecasm:
> nah the builder progress is stored seperatedly, but could be sth to pull into the workorder aswell

Being able to deliberately reset the progress by fire/rehire is occasionally useful.

Thodor12:
> > nah the builder progress is stored seperatedly, but could be sth to pull into the workorder aswell
> 
> Being able to deliberately reset the progress by fire/rehire is occasionally useful.

This is something we could add on a button in the workorder menu too, shouldn't require some weird trick like firing and rehiring (not to mention if people are told to this due to something unrelated, that'd be annoying as well)

src/main/java/com/minecolonies/core/client/gui/WindowBuildBuilding.java
```diff
@@ -314,7 +312,6 @@ private void updateResources()
                 return;
             }
 
-            blueprint.rotateWithMirror(BlockPosUtil.getRotationFromRotations(building.getRotation()), building.isMirrored() ? Mirror.FRONT_BACK : Mirror.NONE, world);
             StructurePlacer placer = new StructurePlacer(new LoadOnlyStructureHandler(Minecraft.getInstance().level, building.getPosition(), blueprint, new PlacementSettings(), true));
```
Raycoms: don't we need to rotate it properly here already? or it this handled in the placer already correctly? Does this mean currently this is bugged?

src/main/java/com/minecolonies/core/colony/workorders/WorkManager.java
```diff
@@ -183,7 +187,16 @@ public Map<Integer, IWorkOrder> getWorkOrders()
     public void clearWorkForCitizen(@NotNull final ICitizenData citizen)
     {
         dirty = true;
-        workOrders.values().stream().filter(o -> o != null && o.isClaimedBy(citizen)).forEach(IWorkOrder::clearClaimedBy);
+        if (citizen.getWorkBuilding() != null)
+        {
+            for (final IWorkOrder workOrder : workOrders.values())
+            {
+                if (workOrder.getClaimedBy() == citizen.getWorkBuilding().getPosition())
```
Raycoms: Can we == this?

src/main/java/com/minecolonies/core/client/gui/WindowBuildBuilding.java
someaddons: ah I removed it since its just used for itemstack resource need calculation

src/main/java/com/minecolonies/core/colony/workorders/WorkManager.java
someaddons: oh no, thats an error

src/main/java/com/minecolonies/core/client/gui/WindowBuildBuilding.java
Raycoms: but it should still be correctly rotated to calculate world offset, no?

someaddons: hm whatfor? shouldnt matter if its rotated if we just iterate for resources

Raycoms: Don't we compare there with what is there in the real world already too?

someaddons: oh true :D

Raycoms (approved):

PR #10791, opened 2025-04-13
Fix enchanted book requests 1.21

someaddons:
Closes #
Closes #

- Fix enchanted book requests on postbox
- CompatibilityManager is now side specific, due to Itemstacks now being side sensitive
- Fix enchanted books now showing up on sync due to item comparisons failing as CHECKED_NBT_KEYS were loaded too late
- Switched compat manager sync to player login, as it has to happen after registries got loaded

## Testing
- [x ] I also did a multiplayer test.

Review please


src/main/java/com/minecolonies/core/event/DataPackSyncEventHandler.java
```diff
@@ -121,6 +118,21 @@ else if (event.getPlayer().getGameProfile() != owner)
             }
         }
 
+        /**
+         * Sends compat manager data on login, has to be after datapack sync registries are loaded
+         *
+         * @param event
+         */
+        @SubscribeEvent(priority = EventPriority.LOWEST)
+        public static void sendOnLogin(final PlayerEvent.PlayerLoggedInEvent event)
```
Raycoms, line 126: Is this always 100% triggered after datapack sync? Or could there be a race condition?

uecasm: PlayerLoggedIn is one of the last events to fire in the sequence.

I'm not sure I understand why DataPackSync was not sufficient, though.  That does trigger before recipes/tags (otherwise we wouldn't need to serialize as much stuff ourselves), but it should still be after registries are synched.  Unless it's a neoforge bug?

uecasm:
If `ItemStack`s are side-sensitive, there might need to be changes to `FurnaceRecipes` as well.  Currently this is populated on server side and synced to client only if not single-player.

src/main/java/com/minecolonies/core/event/DataPackSyncEventHandler.java
someaddons: it could be a neoforge bug not sure, but the resulting stacks during deserialize on client side are using the old registries which are getting replaced after the datapack load finishes and then error when the client tries to serialize them with the updated registries.
And no there is no race condition, the events are all on the server thread not run concurrently

someaddons:
good call, currently it seems to mostly affect enchanted books as their data components store a holder to a registry object but should fix the others too

src/main/java/com/minecolonies/core/event/DataPackSyncEventHandler.java
uecasm: You might need to check what happens on `/reload` as well then.  The events get fired in different orders for that case.

someaddons:
Btw I updated this and added the furnace recipes

Raycoms (approved):

PR #10841, opened 2025-05-06
Prevent crash on loottable serialization

someaddons:
Closes #
Closes #

- Prevent crash and log error instead

## Testing

Review please


uecasm:
Why are there item stacks that can't serialize?

someaddons:
its the registry sensitive stacks again, gotta find out where the stack came from to fix it

Raycoms (approved):

PR #10848, opened 2025-05-11
Port travelling manager and use it in the Netherworker.

marchermans:
- Port Travelling Manager from the old Archeologist branch as a preparation for the new Archeologist
- Make the Netherworker use the Travelling Manager. This means that the huts don't need a vault location anymore.

## Testing

Review please


src/main/java/com/minecolonies/core/colony/Colony.java
```diff
@@ -61,6 +61,7 @@
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.TickEvent;
 import net.minecraftforge.registries.ForgeRegistries;
+import org.checkerframework.checker.units.qual.A;
```
Raycoms: weird import

src/main/java/com/minecolonies/core/colony/CitizenData.java
```diff
@@ -916,9 +916,33 @@ public void updateEntityIfNecessary()
             }
         }
 
+        //Check if we are traveling, we don't spawn an entity if we are traveling.
+        if (getColony().getTravelingManager().isTravelling(this)) {
```
Raycoms: All the formatting is off, I think you need to important the minecolonies formatter =D

src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
```diff
@@ -201,6 +207,10 @@ protected ServerConfiguration(final ForgeConfigSpec.Builder builder)
 
         creativeResolve = defineBoolean(builder, "creativeresolve", false);
 
+        swapToCategory(builder, "debugging");
+
+        netherWorkerTakesDamage = defineBoolean(builder, "netherworkertakesdamage", true);
```
Raycoms, line 214: I wouldn't add that as a config, the nether worker should just take damage. Otherwise people should datapack the netherworker stuff

src/main/java/com/minecolonies/api/colony/managers/interfaces/ITravellingManager.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.resources.ResourceKey;
+import net.minecraftforge.common.util.INBTSerializable;
+
+import java.util.Optional;
+import java.util.logging.Level;
+
+/**
+ * Manages the traveling system for a given colony.
+ */
+public interface ITravellingManager
```
Raycoms, line 12: formatting and javadoc

src/main/java/com/minecolonies/core/colony/ColonyView.java
```diff
@@ -1505,6 +1511,11 @@ public IColonyPackageManager getPackageManager()
         return null;
     }
 
+    @Override
+    public ITravellingManager getTravelingManager() {
```
Raycoms: formatting

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
```diff
@@ -0,0 +1,213 @@
...
+
+    private final IColony                    colony;
+    private final Map<Integer, TravelerData> travelerDataMap = Maps.newHashMap();
+
+    public TravellingManager(final IColony colony) {this.colony = colony;}
+
+    @Override
+    public boolean isTravelling(final int citizenId)
+    {
+        return Optional.ofNullable(travelerDataMap.get(citizenId)).map(TravelerData::isTraveling).orElse(false);
```
Raycoms, line 42: That seems like a weird unpacking chain that could be done more simple without packing it in an optional first

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
```diff
@@ -0,0 +1,213 @@
...
+        {
+            travelerDataMap.remove(citizenId);
+            colony.markDirty();
+        }
+    }
+
+    @Override
+    public void recallAllTravellingCitizens()
+    {
+        final List<ICitizenData> travellersToRecall = this.travelerDataMap
```
Raycoms: If we iterate over it anyway afterwards, I don't think we need to do this mapping before here?

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
```diff
@@ -0,0 +1,213 @@
...
+        return data;
+    }
+
+    @Override
+    public void deserializeNBT(final CompoundTag nbt)
+    {
+        final ListTag travelerData = nbt.getList(NbtTagConstants.TRAVELER_DATA, Tag.TAG_COMPOUND);
+        travelerDataMap.clear();
+
+        travelerData
```
Raycoms: We had gotten rid of most streams in serialize/deserialize logic as they are nasty to debug when things go wrong and meh on log output too.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
```diff
@@ -0,0 +1,213 @@
...
+          .stream()
+          .filter(CompoundTag.class::isInstance)
+          .map(CompoundTag.class::cast)
+          .map(TravelerData::new)
+          .forEach(data -> travelerDataMap.put(data.getCitizenId(), data));
+    }
+
+    private static final class TravelerData implements INBTSerializable<CompoundTag>
```
Raycoms, line 139: I would almost say tha tthis should be its own file with doc

src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkNether.java
```diff
@@ -415,204 +373,191 @@ protected IAIState leaveForNether()
...
         final ExpeditionLog expeditionLog = building.getFirstModuleOccurance(ExpeditionLogModule.class).getLog();
 
         //This is the adventure loop. 
         if (!job.getCraftedResults().isEmpty())
         {
-            ItemStack currStack = job.getCraftedResults().poll();
-            if (currStack.getItem() instanceof ItemAdventureToken)
-            {
-                if (currStack.hasTag())
+            for (ItemStack currStack : job.getCraftedResults()) {
```
Raycoms: lot's of formatting issues here

Raycoms (dismissed):
Some nitpicks & formatting

src/main/java/com/minecolonies/api/colony/managers/interfaces/ITravellingManager.java
Raycoms: still missing javadoc on the methods

src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
Raycoms: still applies

src/main/java/com/minecolonies/core/colony/CitizenData.java
```diff
@@ -916,9 +917,36 @@ public void updateEntityIfNecessary()
...
+        }
+
         if (nextRespawnPos != null)
         {
-            colony.getCitizenManager().spawnOrCreateCivilian(this, colony.getWorld(), nextRespawnPos, true);
+            ICitizenData data = colony.getCitizenManager().spawnOrCreateCivilian(this, colony.getWorld(), nextRespawnPos, true);
+            getEntity().get().getCitizenJobHandler().setModelDependingOnJob(data.getJob());
+            if (!spawnVisible)
+            {
+                data.getEntity().ifPresent(entity -> {
```
Raycoms: we already do a getEntity().get() before. So maybe we could just cache the entity, != null and then use it directly

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: still

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: still

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: still

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: still

src/main/java/com/minecolonies/core/colony/CitizenData.java
marchermans: Applied the formatter

src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
marchermans: This is for debugging only.
There is no reason to have this not as a config option.

What is your argument to remove it? Because this is very helpful debugging stuff with the traveller.

src/main/java/com/minecolonies/core/colony/ColonyView.java
marchermans: Applied the formatter

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: In what way is this weird?

The optional is unpacked as soon as we know if we have a traveller and are travelling?

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: We need this....
We can't iterate over the traveler data map, and we need the citizen data for all entries.......
So this is perfectly a okey.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: Addressed.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: Why? this class is completely internally and not exposed anywhere in this classes API nor in the super interface type.

So I see no reason to do this.

src/main/java/com/minecolonies/core/colony/CitizenData.java
marchermans: Refactored so that the returned ICitizenData can actually modify the entity.

src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
Raycoms: Yeah, I would remote it and make it a flag in the class we can set via code. Modpacks are not smart with config options.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: I would have it just spelled out without the optional wrapping tbh.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: I mean we can just do a for each over all citizens and early return if not a traveler instead of putting it in a list first. Would be smaller, and more performant.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: If it's not used anywhere outside, that's probably fine then. But I could imagine future code wanting to query this at some point.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: Does it matter? Because we just had a discussion on formatting..... And you agreed that this is nitpicking to far.

src/main/java/com/minecolonies/api/colony/managers/interfaces/ITravellingManager.java
marchermans: Done

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: Individually it doesn't matter. It makes it a harder to debug if there is a problem in there and a bit harder to read compared to written out. And, if we do this unpacking in a lot of places it creates more for the GC to handle.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: The datamap is already cleared!
There is no information anymore when the loop runs.
So there is no way to check.

Hence me capturing the ids first, and since i need to capture the citizen data anyway I do that directly in the list.

src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
marchermans: I would like to keep it a config for now. This should help diagnose issues people might have with the new travelling system.

marchermans: But we can remove it later if it turns out it is not needed.
We have many of these debugging flags, things like creative resolve etc.

So I don't see this as a particular problem.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
Raycoms: Why can't we just clear this after the for each loop?
Also for each in loops allow hot reload compared to lambda if we ever want to debug this.

src/main/java/com/minecolonies/api/configuration/ServerConfiguration.java
Raycoms: The problem is less about it being needed, the problem is that tomorrow ATM might just turn on this config randomly.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: Done.

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
```diff
@@ -0,0 +1,216 @@
...
+        {
+            travelerDataMap.remove(citizenId);
+            colony.markDirty();
+        }
+    }
+
+    @Override
+    public void recallAllTravellingCitizens()
+    {
+        for (final Integer citizenId : travelerDataMap.keySet())
```
Raycoms, line 71: <3

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
```diff
@@ -0,0 +1,216 @@
...
+import net.minecraft.nbt.ListTag;
+import net.minecraft.nbt.NbtUtils;
+import net.minecraft.nbt.Tag;
+import net.minecraftforge.common.util.INBTSerializable;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class TravellingManager implements ITravellingManager, INBTSerializable<CompoundTag>
```
someaddons, line 31: docs are missing in this file

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
```diff
@@ -0,0 +1,216 @@
...
+            colony.markDirty();
+        }
+    }
+
+    @Override
+    public void recallAllTravellingCitizens()
+    {
+        for (final Integer citizenId : travelerDataMap.keySet())
+        {
+            final ICitizenData citizenData = this.colony.getCitizenManager().getCivilian(citizenId);
```
someaddons, line 73: does the citizen still exists if it somehow dies during the travel?  I recall the nether worker can take damage and die. If not probably need a null check here

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: We have several other managers without docs.  This is not a particular difficult implementation, its API interface has the docs. Is this still needed here?

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
marchermans: The entity and it's AI does moet exist during travelling. See the worker implementation. When the citizen completes the travel, they spawn invisible and then process the results and die in the over world / at the portal

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
someaddons: ah you dont need to put doc on the override methods, but the general class desc would be nice for its purpose and on the methods that arent part of the interface

src/main/java/com/minecolonies/core/colony/managers/TravellingManager.java
someaddons: ah then it could only crash on loading backups via cmd

Raycoms (approved):

someaddons (approved):

PR #10868, opened 2025-05-15
bunch of food related fixes

Raycoms:
Closes #
Closes #

- Verify interaction handlers before adding them (simplifies code) & reduces short lived popups.
- Correctly report the last eaten index (not just return on the first index which could be -1)
- Check if there is better food in the building we could serve to the citizen (and pick up if so)
- Remove double saturation increase for 1.21 (oopsy) this resulted in them walking around with leftover food etc

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java
+++ b/src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java
@@ -73,7 +73,7 @@ public class CitizenFoodHandler implements ICitizenFoodHandler
         if (lastEatenFoods.size() >= FOOD_QUEUE_SIZE)
         {
             citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_DIVERSITY), ChatPriority.IMPORTANT));
-            citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_DIVERSITY), ChatPriority.IMPORTANT));
+            citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_QUALITY), ChatPriority.IMPORTANT));
             citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_QUALITY + URGENT), ChatPriority.BLOCKING));
             citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_QUALITY + URGENT), ChatPriority.BLOCKING));
         }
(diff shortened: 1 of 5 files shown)


someaddons (dismissed):

src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java
```diff
@@ -73,7 +73,7 @@ public void addLastEaten(final Item item)
         if (lastEatenFoods.size() >= FOOD_QUEUE_SIZE)
         {
             citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_DIVERSITY), ChatPriority.IMPORTANT));
-            citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_DIVERSITY), ChatPriority.IMPORTANT));
+            citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_QUALITY), ChatPriority.IMPORTANT));
             citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_QUALITY + URGENT), ChatPriority.BLOCKING));
             citizenData.triggerInteraction(new StandardInteraction(Component.translatable(NO + FOOD_QUALITY + URGENT), ChatPriority.BLOCKING));
```
uecasm, line 78: Isn't this line also wrong?  Otherwise duplicate.

src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java
```diff
@@ -88,16 +88,18 @@ public Item getLastEaten()
...
         for (final Item foodItem : lastEatenFoods)
         {
+            index++;
             if (foodItem == item)
             {
-                return index;
+                foundIndex = index;
             }
-            index++;
         }
```
uecasm, line 101: Wouldn't it be sufficient to just move the `index++` to the beginning but keep the early return?  This is now returning the last index found instead of the first.

src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java
Raycoms: you're correct

src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java
Raycoms: Yes, we want the last index (the most recent time the food was consumed) not the longest ago it was consumed

src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java
Raycoms: fixed

uecasm (approved):

addlp:
Bravo administration, excellent work

PR #10923, opened 2025-06-07
Better handling of entities trying to move too far

someaddons:
Closes #
Closes #

- Increased the distance at which we handle a too far pathfinding task
- Citizen now teleport to their home when too far
- Other entities teleport to their destination

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
+++ b/src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
@@ -264,13 +264,25 @@ public class MinecoloniesAdvancedPathNavigate extends AbstractAdvancedPathNaviga
 
         if (dest != null && !dest.equals(BlockPos.ZERO))
         {
-            if (job.getStart().distSqr(dest) > 500 * 500)
+            if (job.getStart().distSqr(dest) > 900 * 900)
             {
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
```diff
@@ -264,13 +264,25 @@ public <T extends AbstractPathJob> PathResult<T> setPathJob(
...
+                if (!dest.equals(BlockPos.ZERO))
+                {
+                    if (ourEntity instanceof AbstractEntityCitizen citizen)
+                    {
+                        final BlockPos tpPos = citizen.getCitizenData().getHomePosition();
+                        ourEntity.moveTo(tpPos.getX(), tpPos.getY(), tpPos.getZ());
+                        return null;
+                    }
+
+                    ourEntity.moveTo(dest.getX(), dest.getY(), dest.getZ());
```
Raycoms, line 284: Won't return null here still cause issues with some of the AI? I read some null exceptions because of this. Or, okay because recover afterwards?

someaddons: we've been returning null here since forever :D the function is intentionally annotated as nullable

Raycoms (approved):

PR #10925, opened 2025-06-08
Fix Disappearing Lava Bucket (Bug 9728)

### LINKED ISSUE #9728: [BUG] Builder consumes bucket when placing lava

- [X] Single Player
- [X] Multi Player

### Minecraft Version

1.20.1

### MineColonies Version

1.20.1-1.1.425-BETA

### Structurize Version

1.20.1-1.0.680-BETA

_No response_

### Current Behavior

When lava is a part of a schematic, the builder requests a bucket of lava.  Upon placing the lava, the bucket is consumed. No empty bucket is returned to the builders inventory.

If there is flowing lava, the builder will replace the flowing lava with a lava source block, and not consume the lava bucket at all.

Video of the builder placing the buckets, but only getting the blocks replaced back, no buckets. 
https://github.com/ldtteam/minecolonies/assets/85356635/22fdd266-d898-4c97-98a5-e8f43ad36aab

### Expected Behavior

It used to be in the past, if I remember correctly, that the builders would get an empty bucket back when they placed the lava block.

### Reproduction Steps

Place a schematic that requires lava assigned to a builder.  Give the builder buckets of lava, and watch them place the lava.

### Logs

None

### Anything else?

armele:
Closes [9728](https://github.com/ldtteam/minecolonies/issues/9728)

- During the build step where a lava bucket (or any fluid bucket) is used, when the full bucket is used, replace it with an empty bucket.

## Testing
- [No] I also did a multiplayer test.

Review please


Raycoms (commented):
I think I would handle this a bit upstream in:

 @Override
    public void consume(final List<ItemStack> requiredItems)
    {
        if (this.getInventory() != null)
        {
            for (final ItemStack tempStack : requiredItems)
            {
                if (!ItemStackUtils.isEmpty(tempStack))
                {
                    InventoryUtils.reduceStackInItemHandler(this.getInventory(), tempStack);
                }
            }
        }
    }
    
    The reduceStackInItemHandler is a bit more generic and used in different places as well. This could also be a new utility method that is called here by consume and then calls the reduceStackInItemHandler and then adds a bucket back if it was a full bucket.

armele:
I have refactored this to use a "bucket aware" helper, and call that from consume() as recommended.  I did notice during testing that the last item placed by the builder still shows in the main hand after the build is complete (even if that item is not in the inventory any more), but wasn't sure if this was intentional or not, and so did not attempt to change that behavior.

src/main/java/com/minecolonies/api/util/InventoryUtils.java
```diff
@@ -2901,6 +2903,40 @@ public static void reduceStackInItemHandler(final IItemHandler invWrapper, final
...
+     * This is "bucket-aware", in that if a full bucket is used by the decrease, an empty
+     * bucket replaces it.
+     *
+     * @param invWrapper the inventory item handler.
+     * @param itemStack  the itemStack to decrease.
+     * @param quantity   the quantity.
+     */
+    public static void reduceBucketAwareStackInItemHandler(final IItemHandler invWrapper, final ItemStack itemStack, final int quantity)
+    {
+        if  (attemptReduceStackInItemHandler(invWrapper, itemStack, quantity))
```
Raycoms, line 2930: I'm not 100% sure, but there is a chance of the itemStack becoming air here when its reduced. Did you test this?

armele: Not entirely sure I understand. Would the worker have "air" in their inventory?  For testing I watched the worker place all the blocks in my small decorative structure that included lava.  I observed the dirt blocks going away, and the lava buckets converting to empty buckets. I didn't see anything that looked like "air" where the lava buckets were (they looked like empty inventory slots), but I am not sure what I would have seen if they were air...

armele: Just FYI, this is the same structure I used to reproduce the issue originally - I observed the lava buckets disappearing without an empty bucket replacing them.

Raycoms: yeah, the stack you reduce sometimes (depending on the underlying implementation) becomes empty (empty slot, empty stack). So that you can't do the bucket check anymore. You could do a final boolean = bucketcheck and use that later to avoid these things causing a bug in the future

armele: Makes sense... updated and re-tested successfully.

Raycoms (approved):

someaddons (approved):

PR #10927, opened 2025-06-08
Stats for Archery Training

armele:
Includes minor refactoring of StatsUtil methods for clarity on use.

Does not close any existing defects.

- Added statistics module for the archery training facility.
-Tracks arrows shot and arrows hit.  (Building only - nothing worthy of inclusion at the town hall.)

## Testing
- [ No] I also did a multiplayer test.

Review please


src/main/java/com/minecolonies/api/util/StatsUtil.java
```diff
@@ -105,20 +106,46 @@ public static void trackStat(IBuilding building, String statIdentifier, String d
...
+            Log.getLogger().warn("Attempted to track stat '{}' with null building: ", statIdentifier);
+            return;
+        }
+
+        BuildingStatisticsModule statsModule = building.getModule(STATS_MODULE);
+        
+        if (statsModule != null) 
+        {
+            statsModule.incrementBy(statIdentifier, count);
+        } else {
```
Raycoms: formatting 

}
else
{

Raycoms (commented):
Hmm, maybe we can somehow also track here how successful this has been in leveling them up?

armele:
> Hmm, maybe we can somehow also track here how successful this has been in leveling them up?

I like the idea! Not sure of the best way to implement.  Is there a "leveling" that happens other than the primary/secondary stat increasing? 

I'm not terribly familiar with CitizenExperienceHandler, but it looks like that's it.  If so, I suppose we could track stat levels increased as a result of training?   Perhaps just a before and after check right there in EntityAIArcherTraining.checkShot() that increments a stat if the primary or secondary skill level increased...

Very open to other ideas and suggestions!

Raycoms:
> > Hmm, maybe we can somehow also track here how successful this has been in leveling them up?
> 
> I like the idea! Not sure of the best way to implement. Is there a "leveling" that happens other than the primary/secondary stat increasing?
> 
> I'm not terribly familiar with CitizenExperienceHandler, but it looks like that's it. If so, I suppose we could track stat levels increased as a result of training? Perhaps just a before and after check right there in EntityAIArcherTraining.checkShot() that increments a stat if the primary or secondary skill level increased...
> 
> Very open to other ideas and suggestions!

Yeah, that would work probably.

armele:
Level training stat has been added (and formatting fixed).  This is ready for a re-review.

src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -2534,7 +2534,9 @@
   "com.minecolonies.coremod.gui.townhall.stats.num_diseases_treated": "Treated Diseases: %d",
   "com.minecolonies.coremod.gui.townhall.stats.diseases_treated": "Treated %d %s",
   "com.minecolonies.coremod.gui.townhall.stats.items_cooked": "Cooked %d %s",
-
+  "com.minecolonies.coremod.gui.townhall.stats.arrows_fired": "Arrows Fired: %d",
+  "com.minecolonies.coremod.gui.townhall.stats.arrows_hit": "Arrows Hit: %d",
+  "com.minecolonies.coremod.gui.townhall.stats.levels_trained": "Levels Trained: %d",
```
Raycoms: Neat, we just discussed a bit and I think we could do it separate for primary/secondary. Like (I don't remember the precise skills used for this job atm) but "strength gained" or "stamina gained".

armele: Will-do!  I'll do that tomorrow. Changes should be easy, as I'm pretty familiar with that area of the code. Rather than hard-code the string to the job I'll probably use the trackStatsByName variant and make formatting %d %s so it could be used in the future by other buildings for which you decide to track leveling. (I could see it for probably the library and the school...?)

Raycoms (dismissed):

armele:
I have updated the stats to be recorded in a skill-specific way. Note that I used a string format of "Levels Gained: X Skill" rather than "Skill Gained: X" because reversing the numeric component and the string component would require changes in WindowsStatsModule that would have broad testing implications (and didn't seem worth the value).

armele:
Converting to "draft" until version/main is merged successfully and this PR is re-tested.

armele:
Merged and re-tested.

someaddons (approved):

PR #10928, opened 2025-06-08
Colony match updated and gradle properties

Aaron9174:
Colony match needs to take dimension ID for context. Colony IDs in different dimensions start at 1, so if you have two colonies in different dimensions trying to be allies, they won't show up in the list since the colony ID is not unique. Colony ID + dimension ID is unique and that's what this change involves.

Also, gradle properties setup to run the required version of gradle. Perhaps I have misconstrued the README directions, but I couldn't get this to run with JVM21, and instead got this working with JVM17 + a different gradle version. Possible that the gradle version change can be removed and only the ColonyView change is needed. Has been tested locally and on a server on version 1.20.1 at the most recent tag.

Closes #
Closes #

- Updating colony ID check for populating ally list to include dimension ID as well
- Gradle version for build distribution

## Testing
- [ x ] Yes I tested this before submitting it.
- [ x ] I also did a multiplayer test.

Review please


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.1.1-bin.zip
```
Nightenom: Whenever you change version of gradle wrapper please run twice `gradlew wrapper --gradle-version [version]`

uecasm:
JVM21 is only for the 1.21 branch.  The main branch is still 1.20 with JVM17.

Aaron9174:
> JVM21 is only for the 1.21 branch. The main branch is still 1.20 with JVM17.

This should be updated in the 1.20.1 tagged README then. I will update the README for this branch, so in the next tagged 1.20.1 patch update it will be more accurate.

gradle/wrapper/gradle-wrapper.properties
Aaron9174: Ah that makes sense, I uh, definitely just wrote this in by hand which isn't the correct way to do it lol
edit: All done

Aaron9174:
All comments and merge conflicts addressed, looking for a re-review here

src/main/java/com/minecolonies/core/colony/ColonyView.java
```diff
@@ -377,7 +377,8 @@ public static void serializeNetworkData(@NotNull Colony colony, @NotNull Friendl
             {
                 for (final ColonyPlayer owner : colony.getPermissions().getPlayersByRank(colony.getPermissions().getRankOwner()))
                 {
-                    if (col.getPermissions().getRank(owner.getID()).isColonyManager() && col.getID() != colony.getID())
+                    boolean colonyMatch = (col.getID() == colony.getID()) && (col.getDimension() == colony.getDimension());
```
Raycoms: I would probably inline this as && (id != id || dim != dim), no?

Aaron9174: You got it chief, don't mind making the adjustment

Aaron9174: The change is in, but after re-visiting this, I think it was better before I made the change. As it was more readable via a variable name indicating there was a colonyMatch. Further, the boolean algebra was just simpler to understand before, even with context of these changes, I still had to whip out a truth table to make sure this logic was identical (which it is).

I don't readily see from this statement
col.getID() != colony.getID() || col.getDimID() != colony.getDimID() 
that the current colony of the loop doesn't match the ID of the colony sent in the network request.

Raycoms (dismissed):

Aaron9174:
@Nightenom Got Raycoms approval, mind taking a look at this when you get a chance?

Nightenom (dismissed):

Aaron9174:
Was building locally wrong, just using the ./gradlew (without build), was just missing an parenthesis and was confused why it worked locally. Committing now.

someaddons (approved):

PR #10930, opened 2025-06-08
Allow florist to use shears on tall grass and ferns

### LINKED ISSUE #10019: [BUG] Florist not harvesting ferns and grass

- [X] Single Player
- [ ] Multi Player

### Minecraft Version

1.20.1

### MineColonies Version

1.20.1-1.1.605-BETA

### Structurize Version

1.20.1-1.0.742-RELEASE

Forge: 47.3.5
Domum Ornamentum: 1.20.1-1.0.188-BETA

### Current Behavior

Florist just breaks ferns and grass without using shears, not generating actual items in their inventory.

### Expected Behavior

When florist finds grown ferns or grass, they should check if shear is in their inventory and use it on those shearable plants to obtain them properly, not destroying them by hand.

### Reproduction Steps

1. Build a flower shop which is above level three.
2. Set the other flowers except fern, large fern, and tall grass to off at the filter setting of the hut.
2. Assign a worker and supply a bunch of composts.
3. Set pick up priority to never so you can properly check its inventory contents.
3. Now you can see the florist breaks plants when they grow, but nothing is generated in their inventory nor in their hut inventory.

### Logs

https://gist.github.com/jinwoo829/e3db508f7c69a831a3af40dce5e0b60d

### Anything else?

I used Medieval Spruce style for the flower shop.

Composter was set to not use ferns or grass while the test so this should not be the cause.

I recreated the same problem on the clean instance only jei, minecolonies and its dependencies are installed. Attached log is from this instance.

A bug report about 2 years ago (https://github.com/ldtteam/minecolonies/issues/8304), which is marked as solved in #8305, looks similar to this problem.

pop1040:
Closes #10019

Added as setting rather than hard requirement, in case the player doesn't need tall grass or ferns and would rather not waste iron on shears (or would explicitly like the byproducts of breaking grass without shearing)

- Adds setting to flower shop to use shears
- Modifies Florist AI to request shears only if use shears is enabled
- Adds check to automatically dismiss requests for shears if the setting has been turned back off

## Testing

Review please


src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingFlorist.java
```diff
@@ -99,6 +108,28 @@ public void registerBlockPosition(@NotNull final Block block, @NotNull final Blo
         }
     }
 
+    @Override
+    public void onColonyTick(IColony colony) {
```
Raycoms, line 112: formatting here to
{

src/main/java/com/minecolonies/core/entity/ai/workers/production/agriculture/EntityAIWorkFlorist.java
```diff
@@ -151,6 +153,14 @@ private IAIState decide()
...
 
+        if(building.getSetting(AbstractBuilding.USE_SHEARS).getValue())
+        {
+            if(checkForToolOrWeapon(ModEquipmentTypes.shears.get()))
+            {
+                return IDLE;
+            }
+        }
+
         worker.setItemInHand(InteractionHand.MAIN_HAND, ItemStack.EMPTY);
```
Raycoms, line 164: Wouldn't this have to switch to shears in hand if available then?

pop1040: I was just debugging this and found that while in development they did, on a new world they didn't, so I'm gonna have to debug this.

Raycoms: I think it's also fine to make it a fixed requirement. It makes sense thematically to use shears on flowers too and it's a mid-late game worker anyway.

pop1040: So after further testing the invocation of
```java
        if (!mineBlock(harvestPosition))
````
is responsible for automatically selecting the correct tool as well as breaking the block, as used in the `EntityAIWorkLumberjack.java` and lots of other things, however the reason it wasn't working in further testing, is because it will correctly select shears only for 1 block tall tall grass and ferns, but not 2 block tall ferns and tall grass. I will work on fixing the tool selection for double tall grass and ferns.

pop1040: Interestingly the `mineBlock` call always requests the right tool, but doesn't mark it as required for grass and ferns, so shears aren't requested. It also determines if shears are the right tool by checking if the block implements `IForgeShearable` (but notably it _doesn't_ call `isShearable`, just checks if it's an instance of `IForgeShearable`) in [`WorkerUtil` in the `getBestToolForBlock` method (line 143)](https://github.com/ldtteam/minecolonies/blob/a5f52bf1bcd3df0556b4f33d898c09077e71556a/src/main/java/com/minecolonies/core/util/WorkerUtil.java#L142), and while ferns and grass implement `IForgeShearable`, double tall ferns and grass do _not_ implement it, they are merely instances of `DoublePlantBlock` which doesn't implement the interface. Instead they handle shearing with a custom loot table `forge/data/minecraft/loot_tables/blocks/large_fern.json`

pop1040: Not sure if it's a good idea/ok to add a manual check there like `(state.getBlock() instanceof IForgeShearable || state.getBlock() == Blocks.LARGE_FERN || state.getBlock() == Blocks.TALL_GRASS)` or if I need to delve into the loot table mechanic and check for custom rules on shearing (or if I should change the florist to not use mineBlock and instead do something custom) but I found that even when using shears, they still don't trigger the custom loot table for mining with shears, will have to debug further.

Raycoms:
I'd say we add a custom tag for it

someaddons:
I actually want the florist to use shears for everything, its his working tool and as such contributes to his working cost/requirements. Its a small cost to pay for flowers which you normally cannot grow at all in vanilla

Shadizar:
Several flowers (not all, true) can be bonemealed in vanilla fyi

Raycoms:
Hii =) Could you adjust this PR to remove the setting and just have the florist always use the shears instead?

Thodor12:
Code style is also incorrect throughout the PR

Raycoms:
Done in another PR

PR #10944, opened 2025-06-12
Lumberjack handling of productive bees

armele:
Closes - None

- Generalizes the previously existing Mangrove pattern to all base trees (allowing a family of logs to be recognized as part of the same tree), and adds the productive bee nests as a a recognized part of the tree family.
- As a result bee nests of the same base type as the logs in their tree will be harvested, but a hand-placed nest in a different tree (say, an oak nest in a spruce tree) will be left behind intentionally.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
@@ -748,21 +748,21 @@ public class Tree
     /**
      * Check if this is a log in the same tree type.
      *
-     * @param existingBlock the current block in the tree.
-     * @param newBlock      block to check.
+     * @param checkBlock the current block in the tree being evaluated.
+     * @param stumpBlock the block to check against.
      * @return true if this is the same type of tree; false if it's something different.
      */
     private boolean isBlockPartOfSameTree(
-      @NotNull final BlockState existingBlock,
-      @NotNull final BlockState newBlock)
+      @NotNull final BlockState checkBlock,
+      @NotNull final BlockState stumpBlock)
     {
-        if (existingBlock.is(ModTags.mangroveTree))
+        if (checkBlock.is(ModTags.mangroveTree))
         {
-            return newBlock.is(ModTags.mangroveTree);
+            return stumpBlock.is(ModTags.mangroveTree);
         }
 
-        return existingBlock.getBlock().equals(newBlock.getBlock()) || newBlock.is(ModTags.extraTree);
-    }
+        return checkBlock.getBlock().equals(stumpBlock.getBlock()) || checkBlock.is(ModTags.extraTree);
+	}
 
     /**
      * Adds a leaf and searches for further leaves.
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
@@ -767,13 +767,6 @@ public class Tree
      * @param newBlock      block to check.
      * @return true if this is the same type of tree; false if it's something different.
      */
-    /**
-     * Check if this is a log in the same tree type.
-     *
-     * @param existingBlock the current block in the tree.
-     * @param newBlock      block to check.
-     * @return true if this is the same type of tree; false if it's something different.
-     */
     private boolean isBlockPartOfSameTree(
       @NotNull final BlockState existingBlock,
       @NotNull final BlockState newBlock)
@@ -782,7 +775,6 @@ public class Tree
             return true;
         }
 
-        // TODO: Implement and Test this
         for (TagKey<Block> treeFamilyTag : TREE_FAMILY_TAGS) {
             if (existingBlock.is(treeFamilyTag) && newBlock.is(treeFamilyTag)) {
                 return true;
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
@@ -771,12 +771,15 @@ public class Tree
       @NotNull final BlockState existingBlock,
       @NotNull final BlockState newBlock)
     {
-        if (existingBlock.getBlock().equals(newBlock.getBlock())) {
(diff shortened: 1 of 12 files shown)


src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
```diff
@@ -750,12 +771,17 @@ private boolean isBlockPartOfSameTree(
       @NotNull final BlockState existingBlock,
       @NotNull final BlockState newBlock)
     {
-        if (existingBlock.is(ModTags.mangroveTree))
-        {
-            return newBlock.is(ModTags.mangroveTree);
+        if (existingBlock.getBlock().equals(newBlock.getBlock())) {
+            return true;
```
uecasm: Don't forget to keep the braces on new lines 😄

src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
```diff
@@ -125,6 +131,21 @@ public class Tree
...
+    private static final List<TagKey<Block>> TREE_FAMILY_TAGS = List.of(
+        ModTags.mangroveTree,
+        ModTags.oakTree,
+        ModTags.spruceTree,
+        ModTags.birchTree,
+        ModTags.darkOakTree,
+        ModTags.jungleTree,
+        ModTags.acaciaTree,
+        ModTags.cherryTree
+    );
```
uecasm: I sort of like this in theory as it preserves the ability to try to avoid treating an adjacent different-tree-type as the same tree (which can mess up sapling placement).  But it's not extensible to additional tree types added by other mods.

If you look at the `minecolonies:tree` tag there's already an entry in there that is supposed to cover all these cases -- the missing link was that the `isBlockPartOfSameTree` check defeated it.  I wonder if we should just relax that check to any #tree and then we don't need these additional tags.  Players shouldn't really be placing different tree types close enough to get mixed up anyway.

src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
armele: (how embarrassing...)

src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
armele: Definitely would have been easier. I thought there might be a valid use case of people hand-placing wood nests that should be respected... also respecting log type "felt right" to me :)

uecasm: Possibly a middle ground might be to make a single new tag for the generic bee_nests tag (move it out of #trees), and then have `isBlockPartOfSameTree` check for "is it the same log block OR in that new tag" (something like #extra_tree_blocks?).

uecasm: Or actually put the new tag into #tree in its place, rather than removed entirely.  It does still need to be there for some other checks.

armele: If I'm understanding correctly, would this result in #extra_tree_blocks qualifying as a log for any type of tree?  (So birch nest would get chopped in oak tree, etc?) Or am I misunderstanding?

uecasm: Technically yes, but that should be mostly harmless.  The wrong ones shouldn't really appear directly adjacent to a different tree anyway, especially due to how Productive Bees replaces a log block rather than adjoining it.

The main thing that method was trying to do was to avoid treating a spruce log as part of an oak tree (which is what can lead to sapling problems), and that would still work with this idea (and wouldn't have worked with what I initially suggested).

armele: Converting to draft while I re-test...

Raycoms: two small inputs we can think about here as well:
a) I like the idea of the "extra-tree-blocks" tags that would just fine and people can datapack add anything they need
b) I kinda do like the idea of a datadriven way to identify individual trees. We could load that map from a datapack to be able to individually separate vanilla trees, modpacks can adjust if necessary (if its datapack we can add some optional tags for some modded trees) and we can have some sort of "tree_any" fallback as well?

uecasm: Yeah, we could make a tree registry, but that feels like overkill for the problem.

src/main/java/com/minecolonies/core/generation/defaults/DefaultBlockTagsProvider.java
```diff
@@ -170,6 +170,9 @@ protected void addTags(final HolderLookup.Provider holder)
                 .add(Blocks.MANGROVE_LOG)
                 .add(Blocks.MANGROVE_ROOTS);
 
+        tag(ModTags.extraTree)
+                .addOptionalTag(new ResourceLocation("productivebees", "nests/wood_nests"));
+
```
uecasm, line 175: Also put `extraTree` into `tree` instead of the `wood_nests` it had.

src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
Raycoms: Yeah, that wouldn't be to solve this problem, but rather solve the problem of connecting trees together.

Raycoms: also doesn't have to be part of this PR, just maybe food for thought.

armele:
Ok, retested with the simpler implementation and it is working as expected per above conversation. Nests will be taken down (even if it is an oak nest in a spruce tree).  Regarding Raycom's comments - I think this gives us a good way to accomplish (a) now, and (b) I have mentally filed away but not addressed in this PR.  (Seems pretty straightforward, but also not needed for this specific use case of this PR.)

someaddons (approved):

uecasm (approved):

PR #10951, opened 2025-06-14
separate postbox/stash from blockhut

Raycoms:
Closes #
Closes #

- Refactor to remove postbox/stash from the blueprint subclasses. This makes it easier as they're now not normal anchors anymore. + Allows us to detect if they indicate a child block that we might need to mirror for structurize changes
-

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/blocks/AbstractBaseBlockHut.java
+++ b/src/main/java/com/minecolonies/api/blocks/AbstractBaseBlockHut.java
@@ -1,0 +1,348 @@
+package com.minecolonies.api.blocks;
+
+import com.minecolonies.api.MinecoloniesAPIProxy;
+import com.minecolonies.api.blocks.interfaces.ITickableBlockMinecolonies;
+import com.minecolonies.api.colony.IColony;
+import com.minecolonies.api.colony.IColonyManager;
+import com.minecolonies.api.colony.buildings.IBuilding;
+import com.minecolonies.api.colony.buildings.registry.BuildingEntry;
+import com.minecolonies.api.colony.buildings.views.IBuildingView;
+import com.minecolonies.api.colony.permissions.Action;
+import com.minecolonies.api.entity.ai.workers.util.IBuilderUndestroyable;
+import com.minecolonies.api.items.ItemBlockHut;
+import com.minecolonies.api.tileentities.MinecoloniesTileEntities;
+import com.minecolonies.core.tileentities.TileEntityColonyBuilding;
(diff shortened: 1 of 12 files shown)


src/main/java/com/minecolonies/api/blocks/AbstractBaseBlockHut.java
```diff
@@ -0,0 +1,348 @@
...
+
+/**
+ * Abstract class for all minecolonies blocks.
+ * <p>
+ * The method {@link AbstractBaseBlockHut#getName()} is abstract.
+ * <p>
+ * All AbstractBlockHut[something] should extend this class.
+ */
+@SuppressWarnings("PMD.ExcessiveImports")
+public abstract class AbstractBaseBlockHut<B extends AbstractBaseBlockHut<B>> extends AbstractBlockMinecolonies<B> implements IBuilderUndestroyable, ITickableBlockMinecolonies
```
someaddons: AbstractSchematicBlockHut? So its named similarly to our building schematic class

MotionlessTrain: It is not a schematic anchor, though. So I don’t think Schematic should be in the name

Raycoms: This one is the "non blueprint" version of it though.

someaddons: it has a function called "getBlueprintName()"

someaddons: btw in addition to a nongeneric name also add proper doc on what this class is used for, not a generic desc of its technical function

src/main/java/com/minecolonies/api/blocks/AbstractColonyBlock.java
```diff
@@ -0,0 +1,297 @@
...
+import static com.minecolonies.api.util.constant.TranslationConstants.*;
+
+/**
+ * Abstract class for all minecolonies colony blocks.
+ * <p>
+ * The method {@link AbstractColonyBlock#getName()} is abstract.
+ * <p>
+ * All AbstractBlockHut[something] should extend this class.
+ */
+@SuppressWarnings("PMD.ExcessiveImports")
```
someaddons, line 52: bad doc still, should say what the purpose of the class is, not that is an abstract class

someaddons (approved):

Issue #10959, opened 2025-06-20
Farmer doesn't work in underwater colony


VeryWet-Duck:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Are you using the latest MineColonies Version?

- [x] I am running the latest **beta/release** version of MineColonies for my Minecraft version.
I am also running the latest versions of other mods that are part of my problem.

### Did you check on the Wiki? or ask on Discord?

- [x] I checked the MineColonies Wiki and made sure my issue is not covered there. Or I was sent from discord to open an issue here.

### What were you playing at the time? Were you able to reproduce it in both settings?

- [ ] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.997-1.21.1

### Structurize Version

1.0.776-1.21.1

### Related Mods and their Versions

AllTheMods10

### Current Behavior

Farmers just stands around waiting for "something" or the courier. They do not work, plant or till, even if they have all necessary resources.
There was another ticket/issue written, but it doesn't seem to have been fixed as of right now. This is in an underwater base. The farmer stands and the voiceline for requesting items keeps repeating, but when interacting with the farmer, they do not request anything

### Expected Behavior

I of course expect the farmer to till and plant seeds/vegetables when given those instructions

### Reproduction Steps

1. Make a colony
2. Make a Farm 
3. Hire a farmer 
4. Give them an eternal hoe (i've tried gold, stone and iron in lvl 2)
4. Tell the to plant Garlic (the thing i'm currently trying to grow)
5. Watch

### Logs

[...]

Issue #10911, opened 2025-06-01
Farmers "Waiting for something"


pr0jektBL4CK:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Are you using the latest MineColonies Version?

- [x] I am running the latest **beta/release** version of MineColonies for my Minecraft version.
I am also running the latest versions of other mods that are part of my problem.

### Did you check on the Wiki? or ask on Discord?

- [x] I checked the MineColonies Wiki and made sure my issue is not covered there. Or I was sent from discord to open an issue here.

### What were you playing at the time? Were you able to reproduce it in both settings?

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.905

### Structurize Version

1.20.1-1.0.772

### Current Behavior

Farmers plant/harvest for one cycle then go idle.  When you interact they say "I'm waiting for something" "Where's the courier", but have no open requests.  All items they need are in their inventories.  They only restart work after a re log, and then break after one cycle again.

### Expected Behavior

Farmers should plant/harvest continuously.

### Reproduction Steps

Start game
Watch farmer for plant/harvest cycle.
Farmer stops working

### Logs

https://gist.github.com/pr0jektBL4CK/dd862af76160ed0364e23dd82d9e37ec

### Footer

---
## Viewers
* Add a thumbs-up to the bug report if you are also affected. This helps the bug report become more visible to the team and doesn't clutter the comments.
* Add a comment if you have any insights or background information that isn't already part of the conversation.

Bidgam:
Ideas: It could be compost they are waiting for, or possibly a hoe if they ran out of tools. Basic regular-use items should be checked on occasion as the colonists don't always state they need them.

DarkCh0colate:
> Ideas: It could be compost they are waiting for, or possibly a hoe if they ran out of tools. Basic regular-use items should be checked on occasion as the colonists don't always state they need them.

So I've been running into the same issue, they have tools and I've turned off requesting for fertilizer. Sometimes I do give them some fertilizer if I have and they still do the same thing.

RedRockBandit:
> Ideas: It could be compost they are waiting for, or possibly a hoe if they ran out of tools. Basic regular-use items should be checked on occasion as the colonists don't always state they need them.

[...]

PR #10960, opened 2025-06-21
Add field stage to fields module window

armele:
Related to #10959 and #10911, but does not close them.

- Add the field stage to the fields module window in the hopes that this information will help provide insights as to either the root cause or method of reproduction for the "farmer not working" problem.
-

## Testing

Review please


src/main/java/com/minecolonies/core/client/gui/modules/FarmFieldsModuleWindow.java
```diff
@@ -171,6 +176,7 @@ public void updateElement(final int index, @NotNull final Pane rowPane)
                 if (field instanceof FarmField farmField && !farmField.getSeed().isEmpty())
                 {
                     rowPane.findPaneOfTypeByID(TAG_ICON, ItemIcon.class).setItem(farmField.getSeed());
+                    rowPane.findPaneOfTypeByID(TAG_STAGE, Text.class).setText(Component.literal(farmField.getFieldStage().toString()));
```
Thodor12: I'd recommend an actual translation key for this, over just dumping the field state enum directly in view (we've been trying to move away for that with other things)

uecasm: I was thinking that it might be nice to have an icon representing the current stage instead (e.g. hoe).

Raycoms:
Could do hoe, seed, wheat for stage indication

armele:
> Could do hoe, seed, wheat for stage indication

Do these map as follows, in your mind?
        EMPTY, = Hoe (basically meaning "next step is to hoe", I suppose)
        HOED,  = Seed (meaning "next step is to seed")
        PLANTED = Wheat (meaning "next step is to harvest")

I find it  a bit counter-intuitive, but probably because I am looking at it more as troubleshooting data than as player info. From a player info perspective, seems fine as long as it can be explained clearly.  I will make this change later today (after housecleaning ;) )

Raycoms:
I would map it to next step only, then we don't need empty. Just hoe, seed, wheat (next step is hoe, plant, harvest)

armele:
I've struggled a bit on getting rid of the default tooltip from the icon. Since this might be something others know how to do quickly, I thought I'd stop spinning my wheels and ask for help.  Any guidance would be appreciated!  
image

Thodor12:
> I've struggled a bit on getting rid of the default tooltip from the icon. Since this might be something others know how to do quickly, I thought I'd stop spinning my wheels and ask for help. Any guidance would be appreciated! image

I've done this once, but I wouldn't be able to tell you where anymore. Alternatively you can use a regular image (not itemicon) and use the texture path to the item.

armele:
Ok, I've switched to Image (thanks, Thodor12), tweaked the sizing a bit and re-tested and this is ready to go.

someaddons (approved):

PR #10964, opened 2025-06-23
Stats for Barracks, Blacksmith, Builder, Chickenherder

### LINKED ISSUE #10961: Chicken Herder stats say "Used Raw Chicken" instead of "Got Raw Chicken"

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.1.939-snapshot

### Structurize Version

1.0.773-snapshot

_No response_

### Current Behavior

Image
The same behavior happens with eggs and feathers too.

### Expected Behavior

The raw chicken and the other obtained items should be "Got", not "Used"

### Reproduction Steps

1. Build a chicken herder's hut
2. Assign a chicken herder to the hut
3. Bring two chickens inside the hut
4. Give the herder some seeds to breed the chickens
5. Watch the stat change after the herder has killed one of the chicken, listing `Used x seeds` and `Used x raw chickens`

### Logs

-

### Anything else?

The issue might be the way [AbstractEntityAIHerder](https://github.com/ldtteam/minecolonies/blob/version/main/src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/AbstractEntityAIHerder.java) tracks the number of used items.

### Footer

---
## Viewers

armele:
Also includes introduction of "hide zero values" button on the stats window.

Closes #10961 

- Builder stats: block placed by detail, as well as building built, repaired, upgraded, removed (the latter 4 were already in the town hall stats)
- Blacksmith stats: Items crafted
- Barracks stats: Simply collects the tower-level stats that already existed for the towers of the barracks.
- Chickenherder: Changed "Used" to "Got" for collected items, added breeding, refactored existing stats to use helper method for consistency.
- Because some buildings have long stat lists and by default any key ever used will show in every interval, created a filter on the stat window to allow zero-value stats to be filtered out.

## Testing

Review please


src/main/java/com/minecolonies/core/client/gui/modules/WindowStatsModule.java
```diff
@@ -52,6 +80,18 @@ public class WindowStatsModule extends AbstractModuleWindow
      */
     private static final String HUT_RESOURCE_SUFFIX = ":gui/layouthuts/layoutstatsmodule.xml";
 
+    /*
+     * Module view
+     */
+    private BuildingStatisticsModuleView moduleView = null;
```
Raycoms, line 86: If we cache this here we have to make sure that we're not under the hood replacing the stats module. (I.e. on deserialize we don't make a new instance).

src/main/java/com/minecolonies/core/entity/ai/workers/util/BuildingStructureHandler.java
```diff
@@ -195,6 +197,7 @@ public void triggerSuccess(final BlockPos pos, final List<ItemStack> list, final
 
             for (final ItemStack stack : list)
             {
+                StatsUtil.trackStatByName( structureAI.getWorker().getCitizenData().getWorkBuilding(), BLOCKS_PLACED_DETAIL, stack.getHoverName(), 1);
```
Raycoms: I would not add all individual blocks here, this would become a lot and we have to sync all of this data for all the builders to the clients etc.

Raycoms (commented):
I like the barracks overview stats, makes the barracks hut block more useful.
I would not individually track all builder blocks as this would become too much data.
And if we don't do that, we probably can probably skip the stat filtering.

armele:
Good point on the builders - I will change this to just be blocks placed as well as the four building stage counts. I do think that with some of the crafters (carpenter and stonemason come to mind) it will be useful to keep the ability to filter the stats. How strongly do you feel about wanting that out?

someaddons (approved):

PR #10965, opened 2025-06-24
Allow players to pick up their own items thrown over a colony border

uecasm:
Closes [discord report](https://discord.com/channels/472875599422291968/473575384445878273/1386918044739309602)

- Allow players to pick up their own items dropped, thrown, or otherwise travelled into a colony where they do not have "pick up items" permission.

## Testing

Review please (should port)


Raycoms (approved):
I would next line the return

uecasm:
I like single lines for guard clauses that just immediately exit at the top of a method.  Otherwise style rules make them take five lines instead of one.

Not saying I won't change it if you really want me to, just saying I'd prefer to keep it as is.

someaddons (approved):

PR #10968, opened 2025-06-24
Fix assistant hammer clearing

### LINKED ISSUE #10957: Assistant Hammer's area clear is too restrictive

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.924-snapshot

### Structurize Version

1.20.1-1.0.773-snapshot

Forge 47.4.0
DO 1.20.1-1.0.285-snapshot

### Current Behavior

Using the Assistant Hammer (any material, Iron Gold or Diamond all the same) works fine on tier 1 buildings; however once the building is complete and you go to upgrade, the hammer becomes effectively useless. Every swing returns the error "You need to wait until the Builder finishes clearing the area" - even if you right-click the exact block/area the builder is currently working in, and even if it is actually clear.

### Expected Behavior

I'm not sure what the best behavior is here and I'll leave that up to others.

### Reproduction Steps

1. Start constructing any tier 1 structure
2. Equip an Assistant Hammer
3. Make sure area is clear
4. Help builder, it should work
5. Once building is complete, upgrade and repeat step 4, it should not work

### Logs

n/a

### Anything else?

Overall the check for "is the area clear enough to allow the hammer to work" is a bit too restrictive; it seems (I haven't looked at the code and probably couldn't if I tried) but it seems to check for any foreign blocks not matching the schematic inside the build zone and if it finds even one it says no.

### Footer

---
## Viewers

someaddons:
Closes #10957
Closes #

- Fix the assistant hammer requesting to wait for the builder clearing for leveling up buildings
- The enum for the building progress is now named BuildingProgressStage and got moved into its own class

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/util/BuildingStructureHandler.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/util/BuildingStructureHandler.java
@@ -60,7 +60,7 @@ public class BuildingStructureHandler<J extends AbstractJobStructure<?, J>, B ex
     /**
      * The total number of stages.
      */
-    private final Stage[] stages;
+    private final BuildingProgressStage[] stages;
 
     /**
      * The building associated with this placement.
@@ -70,7 +70,7 @@ public class BuildingStructureHandler<J extends AbstractJobStructure<?, J>, B ex
     /**
      * The current structure stage.
      */
-    private int stage;
+    private int stage = 0;
 
     /**
      * The respective workorder used for placement
@@ -88,7 +88,7 @@ public class BuildingStructureHandler<J extends AbstractJobStructure<?, J>, B ex
       final Level world,
         final IWorkOrder workOrder,
       final AbstractEntityAIStructure<J, B> entityAIStructure,
-      final Stage[] stages)
+        final BuildingProgressStage[] stages)
     {
         super(world,
             workOrder.getLocation(),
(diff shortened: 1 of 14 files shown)


src/main/java/com/minecolonies/core/entity/ai/workers/util/BuildingStructureHandler.java
```diff
@@ -98,7 +98,16 @@ public BuildingStructureHandler(
...
         this.structureAI = entityAIStructure;
         this.stages = stages;
-        this.stage = workOrder.getStage();
+
+        for (int i = 0; i < stages.length; i++)
+        {
+            final BuildingProgressStage stage = stages[i];
+            if (stage == workOrder.getStage())
+            {
+                this.stage = i;
```
Raycoms, line 107: Why do we need to loop match here?

someaddons: because the stage index we're saving is not the enum ordinal, which is what originally created the bug.
The stages are an arbitrary array which do not have all stages in them and could even have stages in a custom order for which we save the index.

Raycoms: Can't we just save the stage here as well than as the enum value?

someaddons: I dont think so since the array determines the stage order, not the enum. But it is your code, ask yourself? :D

Raycoms: No, I mean, the loop you do to assign a stage, we could just store the stage in the class instead of the int version of it.

someaddons: the int we store
is NOT the int version of the stage
it is the index of the array of stages, it may point to whatever stage

PR #10953, opened 2025-06-16
Bug 10952

armele:
Closes #10952

# Changes proposed in this pull request
- Minor change to assignment logic (defect caused by reversed logic on 
- Refactoring to allow logging of root cause in valid situations where assignment should not succeed.

## Testing
- [X] Yes I tested this before submitting it.
- [ ] I also did a multiplayer test.

Review please

PR #10969, opened 2025-06-25
Manual builder assignment tweaks.

armele:
Provides in-game messages for situations that prevent manual assignment, and allows the distance limit to be disregarded when manually assigning a job.

Closes #10953 further discussion.

- Translatable warning messages sent to player when manual build assignment fails.
- Added canBuildIgnoringDistance to IBuilderWorkOrder interface.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/colony/workorders/IBuilderWorkOrder.java
+++ b/src/main/java/com/minecolonies/api/colony/workorders/IBuilderWorkOrder.java
@@ -2,6 +2,9 @@ package com.minecolonies.api.colony.workorders;
 
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.IColony;
+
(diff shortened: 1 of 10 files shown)


Raycoms (commented):
I thought about it a bit and I'd approach this differently. 
a) There is already a code-path for manual assignment when you select in the buildtool window a builder specifically for this. We should probably follow the same code path here.
b) We should disable the button with a tooltip if any of the wrong conditions apply (e.g. citizen null)

armele:
I took some time to analyze this... (a) seems like a pretty significant refactor. The code paths right now are entirely separate.  At my experience level I feel more likely to break something than to improve simplicity.  (b) Looks pretty straightforward to do on the client side.  Any interest in only (b) without the refactoring?

armele:
A (b)-only implementation is ready.
image

someaddons (approved):

PR #10970, opened 2025-06-25
Fix herder item stack request corruption

### LINKED ISSUE #10884: Cowhand requesting a lot of air.

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.894

### Structurize Version

1.20.1-1.0.772

_No response_

### Current Behavior

I have my cowhand requesting air. They should not.

Image
Their current minimum stock.

Image

Deleting the air just makes it recreate itself, and they aren't requesting things or doing much.

### Expected Behavior

They should not ask for air.

### Reproduction Steps

I don't know what caused this or how to reproduce it, alas.

### Logs

https://gist.github.com/Nuku/24f16573db3d46dcdb3bfaee40183912

### Anything else?

_No response_

### Footer

---
## Viewers

rivon0507:
Closes #10884 

- Prevent item stack corruption in herder AI by requesting single-item stacks alongside the count parameter instead of an oversized stack that get corrupted during network transmission. (Inspired by the postbox)

- Use placeholder values in error logging for deserialization issues.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/AbstractEntityAIHerder.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/AbstractEntityAIHerder.java
@@ -314,7 +314,7 @@ public abstract class AbstractEntityAIHerder<J extends AbstractJob<?, J>, B exte
 
         for (final ItemStack item : getExtraItemsNeeded())
         {
-            checkIfRequestForItemExistOrCreateAsync(item);
+            checkIfRequestForItemExistOrCreateAsync(item.copyWithCount(1), item.getCount(), item.getCount());
         }
 
         return DECIDE;
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/AbstractEntityAIHerder.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/AbstractEntityAIHerder.java
@@ -1,5 +1,6 @@
 package com.minecolonies.core.entity.ai.workers.production.herders;
 
+import com.minecolonies.api.crafting.ItemStorage;
 import com.minecolonies.api.entity.ai.statemachine.AITarget;
 import com.minecolonies.api.entity.ai.statemachine.states.IAIState;
 import com.minecolonies.api.entity.citizen.VisibleCitizenStatus;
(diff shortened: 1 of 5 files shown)


src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/AbstractEntityAIHerder.java
```diff
@@ -314,7 +314,7 @@ private IAIState prepareForHerding()
 
         for (final ItemStack item : getExtraItemsNeeded())
         {
-            checkIfRequestForItemExistOrCreateAsync(item);
+            checkIfRequestForItemExistOrCreateAsync(item.copyWithCount(1), item.getCount(), item.getCount());
```
Raycoms: Ohhh, that's why it's running into the problem. I think we should probably
a) Inside the request creation do this, and log if the itemstack does exceed the 64 with stacktrace where this was created.
b) Make getExtraItemsNeeded an actual ItemStorage(ItemStack, qty). To avoid creating bad stacks to begin with here.

rivon0507: Does the item count in the `ItemStack` matter in a request, in the request system? If not (which is probably the case) the constructor of `com.minecolonies.api.colony.requestsystem.requestable.Stack` can be altered so that the `ItemStack` will always have a count of `1`, since that data is already available in `Stack.getCount()`.

And I'll make getExtraItemsNeeded return ItemStorage now.

Raycoms: I'm not 100% sure if it matters. But overall we should never make too large stacks and log if we accidentially do from somewhere.

rivon0507: In `com.minecolonies.api.colony.requestsystem.requestable.Stack`, after looking up usages of `theStack` and ensuring no logic depends on `ItemStack.getCount()`, `theStack`'s count is set to `1` in the constructor of `Stack`. That way, we'll never have too large `ItemStack`s in requests.

And `getExtraItemsNeeded()` now returns an `ItemStorage`.

Raycoms (approved):

someaddons (approved):

PR #10973, opened 2025-06-27
Allow external building module icon

armele:
Closes None

- Slight refactoring of AbstractModuleWindow allows add-on mods to specify their own icon for building modules while remaining backwards-compatible to current usage.
-

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/colony/buildings/modules/IBuildingModuleView.java
+++ b/src/main/java/com/minecolonies/api/colony/buildings/modules/IBuildingModuleView.java
@@ -4,9 +4,14 @@ import com.ldtteam.blockui.views.BOWindow;
 import com.minecolonies.api.colony.IColonyView;
 import com.minecolonies.api.colony.buildings.registry.BuildingEntry;
 import com.minecolonies.api.colony.buildings.views.IBuildingView;
+import com.minecolonies.api.util.constant.Constants;
+
 import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.resources.ResourceLocation;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.api.distmarker.OnlyIn;
+
+import org.apache.http.annotation.Obsolete;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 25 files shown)


someaddons (commented):
I think the more proper way would be to return a resourcelocastion on view.getIcon(), so the custom module can return its own texture there

armele:
Unless I misunderstand, this would require refactoring of all the existing modules. Tat seems potentially too impactful for such a minor benefit. Would you be open to that refactoring? I am happy to do it if so...

uecasm:
What you actually should do is to keep `getIcon()` as is but mark it as `@Obsolete`, then introduce a new `default` method directly in the interface that returns a `ResourceLocation` by calling `getIcon()` and applying the prefix, so the behaviour remains the same for all existing modules.  Then change all callers to call the new method instead.

This allows existing external modules to still link properly and still work (if they worked before), but gives new modules the option to override the new method and return a different location directly.  Eventually (in some later MC version) the old method can be removed.

someaddons:
ye refactoring all modules a bit to return a resource location is what I meant, we could do what mirality suggested, though not sure if that is worth it, I'm not aware of any addons adding modules

Raycoms:
I agree with Sam here, just making the existing icon a resourcelocation instead of a string seems to be the best course of action.

uecasm:
> I'm not aware of any addons adding modules

[armele wrote one](<https://github.com/armele/mctradepost-retro/tree/main/src/main/java/com/deathfrog/mctradepost/api/colony/buildings/moduleviews>) 😛

armele:
I have updated this pull request to follow the feedback of @uecasm, and retested. In one spot a getIcon() implementation returned null and in another an empty string. Both are in cases where the icon doesn't currently get rendered, but I chose to return the "custom" icon ResourceLocation rather than null or empty as safer in case that ever changes.

Raycoms (approved):

someaddons (approved):

armele:
I believe this needs a 1.21 port!  Let me know if you'd like me to tackle it.

PR #10976, opened 2025-06-28
Simplify providing DO items to builder

Raycoms:
Closes #
Closes #

- Builder only requires the "base" do block type for do blocks with many different types
- Builder only obtains this "base" block on mining/replacing them.

This makes it a) Easier for players to craft the items
b) Reduces inventory management overhead for players and workers
c) Less garbage in warehouse

## Testing

Review please

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -29,7 +29,7 @@ mappingsVersion=1.20.1
 dataGeneratorsVersion=1.19.3-0.1.54-ALPHA
 blockUI_version=1.20.1-1.0.190-snapshot
 structurize_version=1.20.1-1.0.771-snapshot
-domumOrnamentumVersion=1.20.1-1.0.184-BETA
+domumOrnamentumVersion=1.20.1-1.0.288-snapshot
 multiPistonVersion=1.20-1.2.30-ALPHA
 
 jei_mcversion=1.20.1
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/AbstractEntityAIInteract.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/AbstractEntityAIInteract.java
@@ -1,5 +1,6 @@
 package com.minecolonies.core.entity.ai.workers;
 
+import com.ldtteam.domumornamentum.item.interfaces.IDoItem;
 import com.minecolonies.api.entity.ai.workers.util.IBuilderUndestroyable;
 import com.minecolonies.api.util.BlockPosUtil;
 import com.minecolonies.api.util.InventoryUtils;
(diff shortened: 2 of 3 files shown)


uecasm:
What does "base item" mean here?  Unless I'm misunderstanding, doesn't this defeat all the DO crafting (and later recycling) options we already have?

Raycoms:
> What does "base item" mean here? Unless I'm misunderstanding, doesn't this defeat all the DO crafting (and later recycling) options we already have?

There are many different panels. The builder only wants the full panel now and can place the right one from this.
The current recipe is "all types all at once". So it's actually equal to that.

uecasm:
But then why are they still requiring the exact matching door type, and not the others?

Also isn't this going to mess up people who want to place/break blocks manually to help their builders?  I guess for placing the assistant hammer will be fine, but not for breaking, or if they do it directly.

Raycoms:
> But then why are they still requiring the exact matching door type, and not the others?
> 
> Also isn't this going to mess up people who want to place/break blocks manually to help their builders? I guess for placing the assistant hammer will be fine, but not for breaking, or if they do it directly.

Placing is fine, breaking not.

someaddons (approved):

uecasm:
This change appears to be causing [a problem](https://discord.com/channels/139070364159311872/1003402487417548801/1388992963589312673).

The underlying cause appears to be that when a DO block is already in-world, and is being replaced with a different block from the schematic, the `handleRemoval` of the **new** (non-DO) block is called on the existing DO block, not the handler for the DO block itself.  As a result, it generates the "real" drops, not the overridden drops.

This is arguably a bug in Structurize, but until that's fixed this should probably be reverted.

Knaxia:
With all the testing I've done, here are the results from where it's working fine, and where it's broken. I've been checking only with Framed Blocks, the default version of the block requested and received by the builder is the "Framed" Type. I will use that one as default block.

-If the builder breaks the block with a tool = Working
Example: I place a bunch of DO blocks (Framed Paper Extra, Oak Planks) down of different variants (left right, right left, crossed, plain), and then use the Shape Tool to replace everything with anything else. The builder will destroy all of those correctly with an Axe, and receive Framed Paper Extra , Oak Planks (Framed).

-If the builder replaces an existing DO block with another DO Block = Working
Example : I placed down a medieval Forester Hut level 3 (not alt), the little house on top has oak/oak framed blocks (Plain) - I made the builder upgrade the building, those blocks were replaced by DO Cobblestone/oak. The builder ended up with the oak/oak (Framed) in their inventory.

-If the builder replaces an existing DO block with any other block = Broken
Example : I placed down a Medieval Builder Hut level 2 (not alt). The left side wall is all paper extra/oak DO. I made the builder upgrade that building to level 3. That wall now is replaced by Cobblestone blocks. The builder ended up with the actual DO variant blocks instead of the Framed version.

-If the builder replaces and existing DO block with air  = Broken
Example, same as above. The Builder Hut level 2 has 3 crates made with DO blocks , oak/oak crossed. When the builder is done placing all the blocks, then did the gravity affected blocks, then goes into the cleanup phase which is replacing all extra blocks with air, the builder will end up with the DO crossed variant blocks in his inventory instead of the framed default version.

I feel like it will also be the same if a DO blocks is being replaced by a torch, carpet or door on the phase after, but I couldn't find an immediate building to test it with.

Knaxia:
I can confirm the update to Structurize has fixed the issue.

Thank you very much :)

PR #10986, opened 2025-06-30
Fix planter nonstop complaining

### LINKED ISSUE #10985: Plantation/Planter doesn't stop complaining

- [ ] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1009-1.21.1

### Structurize Version

1.1.1009-1.21.1

ATM10 Modpack

### Current Behavior

So I am playing with the Desert Oasis style pack and wanted to use the plantation to make some paper/books/sugar and noticed, that the planter kept complaining that he cannot work due to missing items. I checked the wiki on what he could need, but noticed he only really needs the sugar cane and the plantation field to work. He planted sugar cane himself on the given field, but still complains he cannot work due to missing items. He does not show me what he is missing, but now complains he couldn't work for weeks, even though he collects materials. Am I doing something wrong or is this straight up a bug? I hope the attached images work, never used github before. Can also dm me on discord: godrelion if further information is needed.

Image

Image

Image

Image

Image

### Expected Behavior

He works on the plantation and does not complain about missing items, therefore bringing down the happiness of the colony. (Idling workers)

### Reproduction Steps

Desert Oasis plantation (I tried level 1 to 5, he complains on all) in a Windswept Desert biome. He has all 3 basegame plants to plant, yet he complains. 

### Logs

https://gist.github.com/Godrelion/5e1c7576a693b52abeb7e9851771a6a5

### Anything else?

_No response_

### Footer

---
## Viewers

rivon0507:
Closes #10985 

- Similar to #10956 

## Testing

Review please


Thodor12:
I don't see how this would change anything. `setIdleAtJob` only sets a boolean field on the AI. This doesn't modify any behaviour except for removing 1 extra invocation of the method call.

Please explain why this would fix it.

rivon0507:
Before, the planter is set to `idleAtJob=true` at the start of `pickField()` and only set to `idleAtJob=false` when it finds a field to work on (`fieldToWork` is not `null`), but the planter has the same mechanics (?) as the farmer, working a field once per day. So if `fieldTowork` is `null`, the planter will stay `idleAtJob` for the rest of the day and will complain.

src/main/java/com/minecolonies/core/entity/ai/workers/production/agriculture/EntityAIWorkPlanter.java
```diff
@@ -142,6 +139,20 @@ private IAIState pickField()
...
     }
 
+    /**
+     * Ensures that we have at least one field assigned.
+     * Also sets the worker to be idle at the job if we have no assigned field.
+     *
+     * @param module the extensions assigned to the planter's hut
+     * @return false if we have at least one assigned plantation field
+     */
+    public boolean checkForField(@NotNull BuildingExtensionsModule module)
```
Thodor12: In that case I'd probably just move the method calls around and not introduce a new method for it, not worth it.
So remove this method.

Thodor12 (changes requested):

armele:
That boolean seems to source the predicate for the dayRollOverPredicate in TimeBasedHappinessModifier.dayEnd(), unless my code trace went astray.   I stopped short of tracing the code path in debug mode since @rivon0507 has got this one. :)

So my theory was that any day that ends with isIdleAtJob = true, when TimeBasedHappinessModifier.dayEnd is called then the days of idleness are incremented (even if work was done earlier in the day).  And reviewing the planter AI supported that theory.  (Again, not tested or traced - just reviewed through code.)

  HappinessRegistry.idleatjobFunction = DEFERRED_REGISTER_HAPPINESS_FUNCTION.register(IDLEATJOB_FUNCTION.getPath(), () -> new HappinessFunctionEntry(data -> data.isIdleAtJob() ? 0.5 : 1.0));

Thodor12:
Ok actually hold on this is not a bug at all. You're saying yourself "if the worker can't find a field to work on", that definition constitutes being idle. So he should be complaining if there's nothing to do.

Rather, figure out why fieldToWorkOn is returning null.

rivon0507:
fieldToWorkOn is returning null when all field have been worked for the day. So yes, the citizen can be idle, but should not complain.

Thodor12:
Other side of the spectrum, if you have a field, but nothing is able to grow due to whatever reason. You do want to hear about it that he's unable to work.

Thodor12:
The proposed solution is too straight forward and inhibits legitimate cases where a complaint should be thrown, so please come u with a different solution.

Raycoms:
We've been using idle at job as an unexpected idle at job so far. We'd need a new flag for expected idle

rivon0507:
> The proposed solution is too straight forward and inhibits legitimate cases where a complaint should be thrown, so please come u with a different solution.

I’d like to understand better — could you list the legitimate cases that might be inhibited by the current solution? That way, I can make sure we’re not accidentally suppressing valid complaints.



> Other side of the spectrum, if you have a field, but nothing is able to grow due to whatever reason. You do want to hear about it that he's unable to work.

That scenario wasn’t handled before either, so I don’t think this fix inhibits it. I agree it’s a valid case, but it seems more like a feature request than a bug regression

armele:
If it's helpful, here's what I was thinking: https://github.com/ldtteam/minecolonies/compare/version/main...armele:minecolonies:complaining_planters

Basically the valid condition is the NO_FREE_FIELDS situation. Instead of starting pickField() pessimistically (by setting idle to true) we move that inside the two spots where it returns an early IDLE.  The first I suspect shouldn't be possible , but is included for completeness.  The second should correspond to the situation where nothing has been assigned to the planter.

src/main/java/com/minecolonies/core/entity/ai/workers/production/agriculture/EntityAIWorkPlanter.java
Raycoms: this seems the cleanest solution for me tbh

Raycoms:
Would you be willing to adjust the setIdle to "setJobStatus" with an enum that is stuck,idle,working, and we then properly set stuck/working everywhere for now and for the complaints/happiness we check for stuck?

In a future PR we could then also set idle and use this for some nice things like going to the tavern or so =D

uecasm:
`STUCK` or `BORED`?  To me `STUCK` implies a pathfinding issue.

MotionlessTrain:
In cases where setIdleAtJob is currently used, it is actually STUCK

BORED could be an alternate name for IDLE (nothing to do at the moment, but there is nothing wrong either at this moment)

uecasm:
Idk, I would think "idle" means "I did things today but have nothing to do right this moment" (good) while "bored" means "I haven't had anything to do today" (bad).  (I also have a lovely bikeshed.)

src/main/java/com/minecolonies/core/entity/ai/workers/production/agriculture/EntityAIWorkPlanter.java
rivon0507: > In that case I'd probably just move the method calls around and not introduce a new method for it, not worth it. So remove this method.

Done :)

rivon0507:
So now we have a jobStatus (IDLE, WORKING, STUCK) enum instead of a idleAtJob boolean.

src/main/java/com/minecolonies/api/colony/jobs/IJob.java
```diff
@@ -123,7 +124,7 @@ default void onLevelUp()
      */
     default void initEntityValues(AbstractEntityCitizen citizen)
     {
-        citizen.getCitizenData().setIdleAtJob(false);
+        citizen.getCitizenData().setJobStatus(JobStatus.IDLE);
```
Raycoms, line 127: Initially this was WORK, right

src/main/java/com/minecolonies/core/colony/CitizenData.java
```diff
@@ -1491,7 +1492,15 @@ public void deserializeNBT(final CompoundTag nbtTagCompound)
             }
         }
 
-        this.idle = nbtTagCompound.getBoolean(TAG_IDLE);
+        final String jobStateString = nbtTagCompound.getString(TAG_JOB_STATUS);
+        if (!jobStateString.isEmpty())
+        {
+            this.jobStatus = JobStatus.valueOf(jobStateString);
```
Raycoms: I'd go by ordinal for this

src/main/java/com/minecolonies/api/colony/jobs/IJob.java
Raycoms: But I guess IDLE is fine

src/main/java/com/minecolonies/core/colony/CitizenData.java
Raycoms: Stil l missing the ordinal change here

rivon0507: Yeah, sorry I missed the comment. I'm working on it :)

rivon0507: Done

src/main/java/com/minecolonies/core/colony/CitizenData.java
```diff
@@ -233,9 +234,9 @@ public class CitizenData implements ICitizenData
     protected final Map<Component, IInteractionResponseHandler> citizenChatOptions = new HashMap<>();
 
     /**
-     * If idle at job.
+     * The current status of the citizen's working (?)
      */
-    private boolean idle;
+    private JobStatus jobStatus;
```
Raycoms: oh, we can just initialize this here with idle, so that it can't be null. simplifies reading then.

someaddons (dismissed):

PR #11023, opened 2025-07-20
Introduce the initial infrastructure for the Archeologist.

marchermans:
- Adds initial support implementation for the Archeologist
  - Building
  - Worker
  - Job
  - Crafting results etc.

## Testing

Review please


Raycoms:
I think it's too much partial infrastructure already. It has a lot of AI but not finished and a lot of loottable sbut not finished.
So we can't properly review this now because "will change later".

I think just building/worker/job/ai but all empty and no loottables would be fine for the PR. Otherwise we just ask the builders to use an emerald placeholder block to build the building for the time being.

PR #11028, opened 2025-07-23
Lumberjack chopping of mixed log/wood

armele:
Closes Discord discussion: https://discord.com/channels/472875599422291968/1395820683811885176 
- Allow chopping of trees that contain a mix of log and wood by using existing tags.

## Testing

Review please


src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
```diff
@@ -745,6 +745,37 @@ private void addAndSearch(@NotNull final Level world, @NotNull final BlockPos lo
...
+     * @param checkBlock The block state to check.
+     * @param stumpBlock The block state of the stump to compare against.
+     * @return true if both block states are of the same type of log, false otherwise.
+     */
+    private boolean isSameTypeOfLog(@NotNull final BlockState checkBlock, @NotNull final BlockState stumpBlock)
+    {
+        boolean same = false;
+
+        if (
+            (checkBlock.is(ModTags.mangroveTree) && stumpBlock.is(ModTags.mangroveTree))
```
Raycoms: I'm wondering if there is any more elegant solution to this

armele: I think an "adequate" amount of checking of sameness of log is important, and with that the simpler solution of just adding the extra_tree tag to logs is not available.

src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
```diff
@@ -761,7 +775,7 @@ private boolean isBlockPartOfSameTree(
             return stumpBlock.is(ModTags.mangroveTree);
         }
 
-        return checkBlock.getBlock().equals(stumpBlock.getBlock()) || checkBlock.is(ModTags.extraTree);
+        return checkBlock.getBlock().equals(stumpBlock.getBlock()) || checkBlock.is(ModTags.extraTree) || (logPrefix(checkBlock).equals(logPrefix(stumpBlock)));
```
Raycoms: We can speed this up by doing a == check on the blocks, they're singletons.

src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
```diff
@@ -745,6 +747,18 @@ private void addAndSearch(@NotNull final Level world, @NotNull final BlockPos lo
...
     }
 
+    /**
+     * Get the prefix of a log block, which is the path of the block in the
+     * registries, with "_log" or "_wood" removed from the end.
+     *
+     * @param block the block to get the prefix from.
+     * @return the prefix of the log block.
+     */
+    private String logPrefix(BlockState block) {
```
Raycoms: { on next line, otherwise this looks good. Did you test it?

armele: I did test it - let me test it again after the == change!

armele:
Addressed code review feedback and re-tested.

Raycoms (dismissed):

src/main/java/com/minecolonies/core/entity/ai/workers/util/Tree.java
```diff
@@ -745,6 +747,19 @@ private void addAndSearch(@NotNull final Level world, @NotNull final BlockPos lo
...
+     * Get the prefix of a log block, which is the path of the block in the
+     * registries, with "_log" or "_wood" removed from the end.
+     *
+     * @param block the block to get the prefix from.
+     * @return the prefix of the log block.
+     */
+    private String logPrefix(BlockState block) 
+    {
+        String path = ForgeRegistries.BLOCKS.getKey(block.getBlock()).getPath();
+        return path.replaceFirst("(_log|_wood)$", "");
```
uecasm: This is better than being too explicit like it was before (because that wouldn't support modded wood nicely).

This version doesn't quite handle mangrove or `_stem` / `_hyphae` blocks properly though.

armele: I will add in the _stem and _hyphae... For mangrove there is already some custom logic for that in the caller. I was thinking it didn't need to be addressed here outside the _log _wood scope... Let me know if you disagree and I will research what goes into those mangrove tags.

armele:
Updated and retested.

someaddons (approved):

PR #11032, opened 2025-07-25
Safe removal of add-on jobs and buildings.

armele:
Closes None

- Adjust the error handling of restoring a saved game such that if an add-on mod is removed (say... Trade Post) the colony doesn't become corrupted by the unrecognized buildings and jobs that it included.  This will log the errors, but just leave the citizens unemployed instead of corrupting the load process.
-

## Testing

Review please


src/main/java/com/minecolonies/core/colony/buildings/registry/BuildingDataManager.java
```diff
@@ -22,6 +22,8 @@
 import static com.minecolonies.api.util.constant.NbtTagConstants.TAG_BUILDING_TYPE;
 import static com.minecolonies.api.util.constant.NbtTagConstants.TAG_LOCATION;
 
+import java.util.Optional;
```
Raycoms: unused

someaddons (approved):

PR #11036, opened 2025-07-26
feat(commands): introduce /mc query command with initial 'idlebuilders' query

mherzogjr:
This PR introduces a new `/mc query` command subtree designed to support queries for players and admins.

Please note - this is my first PR ever.  Add all the caveats that might go along with that.  I have no real JAVA experience (but have considerable experience in other languages) and I wanted to set a goal to learn how to develop a mod and I thought contributing something I thought would be useful in one of my favorite modpacks would be a good place to start.  

I did my best to match your structure and style so that there isn't any pain for you there, but definitely welcome to any feedback you might have on my approach.  Thanks.

**New command:**
`/mc query idlebuilders`
Reports a list of builder citizens who currently have no active work orders and are therefore idle. Each entry includes a clickable teleport link to the citizen’s builder hut.  Justification:  I always do a lot of unassigned build commands, but many times a building will only be close enough for a single builder to ever get assigned this build automatically. As a result you might end up with a builder or 2 just hiding in their huts because there is nothing to assign to them (even though there are lots of things set to build).

**Design Notes:**
Adds a new class: `CommandQueryIdleBuilders`

Uses `getClosestColony(...)` and `isCoordinateInAnyColony(...)` to determine the relevant colony context based on the player’s current position.

Duplicates the `styleWithTeleport() `method (originally in `CommandCitizenInfo.java`) for generating clickable `/tp` links.  I thought about just creating a public version, but that seemed beyond scope of this PR, but this could be made a shared utility in a future refactor.

**Future extensions could include:** (but I didn't want to go here without your interest).  
I am sure others may have more creative ideas than me, but a couple that I quickly though might be useful:
-  `/mc query vacancies`: Report houses which have an open bed
-  `/mc query workdistance <blockdistance>` : Identify citizens whose home/work distance exceeds a threshold.

## Testing

Review please
Appreciate any comments or feedback you might be able to provide.

### DIFF
--- a/src/main/java/com/minecolonies/core/commands/EntryPoint.java
+++ b/src/main/java/com/minecolonies/core/commands/EntryPoint.java
@@ -3,6 +3,7 @@ package com.minecolonies.core.commands;
 import com.minecolonies.api.util.constant.Constants;
 import com.minecolonies.core.commands.citizencommands.*;
 import com.minecolonies.core.commands.colonycommands.*;
+import com.minecolonies.core.commands.querycommands.*;
 import com.minecolonies.core.commands.colonycommands.requestsystem.CommandRSReset;
 import com.minecolonies.core.commands.colonycommands.requestsystem.CommandRSResetAll;
 import com.minecolonies.core.commands.generalcommands.*;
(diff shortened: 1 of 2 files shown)


someaddons:
Hi and thanks for your contribution!
I think the better approach to your problem would be showing idle builders on the colony map with some icon on them(similar to how a hunger icon already gets rendered on a citizen). In our design commands do not belong into gameplay, but serve debugging or admin purposes.

Raycoms:
I think a cool option would be also an "idle builder scrolls" that the enchanter can craft that teleports you to an idle builder.

PR #11053, opened 2025-07-30
Craft-only mode for lumberjack

armele:
Closes Discord suggestion at [Strip-only forester would be awesome](https://discord.com/channels/472875599422291968/1312908084254867657) 

- Added a setting for "Craft Only" mode on the Lumberjack that has them skip their tree searching and chopping states.

## Testing

Review please


someaddons:
I'm actually not a fan of this change as it may confuse ppl, they may toggle this setting and forget, or unknowingly click it and wonder why their forester is not working. Does not seem worth it for increase efficiency esp since the crafting is not slow either

Shadizar:
The crafting is actually very slow...
I followed my forester for hours. Both with an empty zone and with all saplings off, they spent 50 to 60 percent of the time loitering in tree areas (btw, no trees present). Often, stripped wood requests would not be serviced for very long periods because of them walking around zone and back. It is bad enough that the logs for an order often get delivered 2 or 3 times because the crafting cycle takes so long it loses track

Shadizar:
968 bld btw

Raycoms:
The forester should probably not start pathfinding to search for trees if there are no saplings turned on in the list.

Shadizar:
Would another option be to move this crafting to the sawmill and not have foresters craft?

MotionlessTrain:
No, that takes away recipe slots from the sawmill for all stripping recipes. So you can’t teach it recipes anymore until it is level 3 or 4 or so (depending on the amount of trees in the modpack)

Shadizar:
True..I wonder if it would be nice for crafterrecipes to be their own thing for that reason

someaddons:
I think we just make the foresters AI more responsive to crafting tasks then, that should solve it

Shadizar:
Wouldn't that negatively impact other players, that use them for both cutting and crafting though? Especially in styles that use a whole lot of stripped wood

Shadizar:
By which I mean, some players would then ge getting a lot less logs

armele:
Ok, I updated this pull requests to take the approach of having the forester check for crafting tasks between trees.  So they will finish their tree, check for crafting tasks - if none, next tree (as before).  If crafting tasks are there, they reset to START_WORKING (same strategy used in a few other places in the lumberjack AI previously) and handle them before resuming cutting.

src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
```diff
@@ -339,7 +339,7 @@ private IAIState gathering2()
 
         // Reset everything, maybe there are new crafting requests
         resetGatheringItems();
-        if (ColonyConstants.rand.nextInt(100) <= 10)
+        if ((ColonyConstants.rand.nextInt(100) <= 10))
```
Raycoms: too many parenthesis

Raycoms:
> Ok, I updated this pull requests to take the approach of having the forester check for crafting tasks between trees. So they will finish their tree, check for crafting tasks - if none, next tree (as before). If crafting tasks are there, they reset to START_WORKING (same strategy used in a few other places in the lumberjack AI previously) and handle them before resuming cutting.

I think you can add a check to not search for trees if there are not trees they're allowed to cut

armele:
> > Ok, I updated this pull requests to take the approach of having the forester check for crafting tasks between trees. So they will finish their tree, check for crafting tasks - if none, next tree (as before). If crafting tasks are there, they reset to START_WORKING (same strategy used in a few other places in the lumberjack AI previously) and handle them before resuming cutting.
> 
> I think you can add a check to not search for trees if there are not trees they're allowed to cut

Not allowed based on the sapling configuration (no saplings turned on), or based on having a 1x1x1 zone set?  (Or something else I might have missed...)

Raycoms:
> > > Ok, I updated this pull requests to take the approach of having the forester check for crafting tasks between trees. So they will finish their tree, check for crafting tasks - if none, next tree (as before). If crafting tasks are there, they reset to START_WORKING (same strategy used in a few other places in the lumberjack AI previously) and handle them before resuming cutting.
> > 
> > 
> > I think you can add a check to not search for trees if there are not trees they're allowed to cut
> 
> Not allowed based on the sapling configuration (no saplings turned on), or based on having a 1x1x1 zone set? (Or something else I might have missed...)

No saplings turned on

armele:
Added check to not search for trees if none allowed as configured in sapling menu. Retested.

src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
```diff
@@ -240,17 +240,20 @@ protected IAIState decide()
             return getState();
         }
 
-        // This got moved downwards compared to the AICrafting-implementation,
-        // because in this case waiting for dumping is more important
-        // than restarting chopping
-        if (job.getTaskQueue().isEmpty())
+        if (building.getModuleMatching(ItemListModule.class, m -> m.getId().equals(SAPLINGS_LIST)).getList().size() > 0)
```
someaddons: building.getModule(BuildingModules.ITEMLIST_SAPLING)

armele: Much cleaner! Updated and retested.

someaddons: ye the searching module lookups are mostly old code and are only actually needed when you do not know what the building is

someaddons (approved):

PR #11080, opened 2025-08-07
Fix decoration controllers not showing upgrade/remove buttons

### LINKED ISSUE #11062: Deco Controller weirdness

### Are you using the latest Structurize Version?

- [x] I am running the latest **beta/release** version of Structurize for my Minecraft version.

- [x] I checked the MineColonies/Structurize [Wiki](https://wiki.minecolonies.ldtteam.com/) or I asked on [discord](https://discord.minecolonies.com/).

### Minecraft Version

1.21

### Structurize Version

1.0.779-1.21.1

### MineColonies Version (if related bug)

1.1.1009-1.21.1

Neoforge 21.1.141

### Current Behavior

https://cdn.discordapp.com/attachments/724810039281451028/1387678927346073681/image.png?ex=685e3836&is=685ce6b6&hm=875f980d066a8a1191ffa4b0d30e1692b52f05b52a06c526cc8329e89a6605d5&

Incorrect information stored in deco controllers after a builder has built it.  In the attached image, The upper one is a builder built decoration. the lower one is the same decoration constructed pasted in. the builder constructed one has no upgrade/repair button but the pasted one does.

### Expected Behavior

well...would expect it to work

### Reproduction Steps

make a random upgradable deco of a couple of levels
constructed paste it into a colony
have a builder also build it
compare the block data on the 2 controllers

### Logs

n/a

### Anything else?

_No response_

### Footer

---
## Viewers

Thodor12:
Closes #11062 

- Remove old schematic migration code causing a failure for loading new structures.
- This code is dated and can be removed anyway, but the new structures were improperly implemented, causing all structures to go through the old method, which no longer worked.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/jobs/AbstractJobStructure.java
+++ b/src/main/java/com/minecolonies/core/colony/jobs/AbstractJobStructure.java
@@ -2,13 +2,11 @@ package com.minecolonies.core.colony.jobs;
 
 import com.ldtteam.structurize.blockentities.interfaces.IBlueprintDataProviderBE;
 import com.ldtteam.structurize.blueprints.v1.Blueprint;
-import com.ldtteam.structurize.storage.StructurePacks;
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.buildings.IBuilding;
 import com.minecolonies.api.colony.workorders.IBuilderWorkOrder;
 import com.minecolonies.api.colony.workorders.IWorkOrder;
 import com.minecolonies.api.util.Log;
-import com.minecolonies.api.util.Utils;
 import com.minecolonies.api.util.constant.NbtTagConstants;
 import com.minecolonies.core.colony.buildings.AbstractBuildingStructureBuilder;
 import com.minecolonies.core.entity.ai.workers.AbstractAISkeleton;
@@ -19,7 +17,6 @@ import net.minecraft.world.level.block.entity.BlockEntity;
 import org.jetbrains.annotations.Nullable;
 
 import static com.ldtteam.structurize.blockentities.interfaces.IBlueprintDataProviderBE.TAG_BLUEPRINTDATA;
-import static com.minecolonies.api.util.constant.NbtTagConstants.TAG_NAME;
 
 /**
  * Common job object for all structure AIs.
@@ -118,10 +115,6 @@ public abstract class AbstractJobStructure<AI extends AbstractAISkeleton<J>, J e
                             if (te instanceof IBlueprintDataProviderBE)
                             {
                                 final CompoundTag tagData = compoundNBT.getCompound(TAG_BLUEPRINTDATA);
-                                final String schematicPath = tagData.getString(TAG_NAME);
-                                final String location = StructurePacks.getStructurePack(blueprint.getPackName()).getSubPath(Utils.resolvePath(blueprint.getFilePath(), schematicPath));
-
-                                tagData.putString(TAG_NAME, location);
                                 tagData.putString(NbtTagConstants.TAG_PACK, blueprint.getPackName());
 
                                 try
@@ -130,7 +123,7 @@ public abstract class AbstractJobStructure<AI extends AbstractAISkeleton<J>, J e
                                 }
                                 catch (final Exception e)
                                 {
-                                    Log.getLogger().warn("Broken deco-controller at: " + x + " " + y + " " + z);
+                                    Log.getLogger().warn("Broken deco-controller at: {} {} {}", x, y, z);
                                 }
                                 ((ServerLevel) getColony().getWorld()).getChunkSource().blockChanged(tePos);
                                 te.setChanged();
--- a/src/main/java/com/minecolonies/core/colony/jobs/AbstractJobStructure.java
+++ b/src/main/java/com/minecolonies/core/colony/jobs/AbstractJobStructure.java
@@ -2,11 +2,13 @@ package com.minecolonies.core.colony.jobs;
 
 import com.ldtteam.structurize.blockentities.interfaces.IBlueprintDataProviderBE;
 import com.ldtteam.structurize.blueprints.v1.Blueprint;
+import com.ldtteam.structurize.storage.StructurePacks;
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.buildings.IBuilding;
 import com.minecolonies.api.colony.workorders.IBuilderWorkOrder;
 import com.minecolonies.api.colony.workorders.IWorkOrder;
 import com.minecolonies.api.util.Log;
+import com.minecolonies.api.util.Utils;
 import com.minecolonies.api.util.constant.NbtTagConstants;
 import com.minecolonies.core.colony.buildings.AbstractBuildingStructureBuilder;
 import com.minecolonies.core.entity.ai.workers.AbstractAISkeleton;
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/core/tileentities/TileEntityDecorationController.java
```diff
@@ -148,68 +144,9 @@ public void readSchematicDataFromNBT(CompoundTag compound)
...
-        {
-            this.packName = blueprintDataProvider.getString(TAG_PACK);
-        }
 
-        if (this.packName == null)
-        {
-            this.packName = DEFAULT_STYLE;
-        }
+        this.packName = blueprintDataProvider.getString(TAG_PACK);
+        this.schematicPath = StructurePacks.getStructurePack(this.packName).getSubPath(StructurePacks.findBlueprint(this.packName, schematicName));
```
Raycoms: find blueprint should never be the default path. This is a backup mechanism. This is super costly

src/main/java/com/minecolonies/core/tileentities/TileEntityDecorationController.java
```diff
@@ -148,68 +144,9 @@ public void readSchematicDataFromNBT(CompoundTag compound)
     {
         IBlueprintDataProviderBE.super.readSchematicDataFromNBT(compound);
         final CompoundTag blueprintDataProvider = compound.getCompound(TAG_BLUEPRINTDATA);
-        if (compound.contains(TAG_PACK)) // New structure
-        {
-            // path is the folder containing the schematic
-            final String path = blueprintDataProvider.getString(TAG_NAME);
-            this.schematicPath = path + File.separator + this.schematicName + ".blueprint";
```
Raycoms, line 155: What is wrong with this default?

Thodor12: The path here became empty, because I removed the TAG_NAME

src/main/java/com/minecolonies/core/tileentities/TileEntityDecorationController.java
Thodor12: This was always called in the other file to determine the location, and pass it into the TAG_NAME data. So nothing changes, it's just moved here now.

Raycoms: It was always only supposed to be called as a backup if the other lookup mechanism doesn't work. If it was always called, that was a huge bug then.

Thodor12: The issue is, there's zero other data to determine the path from. It's not present in the blueprint data itself, not present in the blueprintDataProvider, nowhere, this is the only place to fetch the correct path from.

Raycoms: The path is set when placing it down. Either the construct paste, or the builder entity. This is only written if its the primary anchor.
We need to persist this path then in the blockentity, because that's what makes child decorations possible then.

MotionlessTrain: This mechanism actually gave issues with Fortress, where walls and paths have similar names (but are in different folders). One of them use the schematics of the other while upgrading, as this fallback mechanism found the wrong path

For schematics that have been scanned in 1.19+, the path is indeed stored in the nbt of the decoration controller, and the previous code used it from there

MotionlessTrain: Note that even if this is the correct way, previously it stored it in the tile entity, and calculated it once (when the builder built it). Now it calculates it every time the tile entity is loaded instead

Wouldn’t the right fix be to move the code that sets TAG_NAME to the constructed placement handler as well, instead of removing it?

MotionlessTrain:
Does this mean everyone needs to break all decoration controllers in their colony, and replace the decorations? (To get the TAG_PATH set properly)
And I assume you still need to implement the construction paste case?

src/main/java/com/minecolonies/core/colony/jobs/AbstractJobStructure.java
```diff
@@ -110,27 +107,28 @@ public void complete()
                 {
                     for (short z = 0; z < blueprint.getSizeZ(); z++)
                     {
+                        final BlockPos offset = new BlockPos(x, y, z);
```
Raycoms: I'd move this deeper inside the if closer to where it's needed

Raycoms:
We should test this thoroughly as this could also affect buildings the way it is atm. So we should paste/builder place a couple of different newer/older buildings and see if it still works properly

Thodor12:
> Does this mean everyone needs to break all decoration controllers in their colony, and replace the decorations? (To get the TAG_PATH set properly) And I assume you still need to implement the construction paste case?

Check the linked issue, constructed paste works fine.
If people actually want to have their decorations upgradeable, yes, there is no way to fix this anymore.

MotionlessTrain:
> Check the linked issue, constructed paste works fine 

Does it still work fine after those changes? Did it already set TAG_PATH, or does it work in a different way?

Thodor12:
Constructed paste never reaches this specific code, this code is only for builders

src/main/java/com/minecolonies/core/tileentities/TileEntityDecorationController.java
```diff
@@ -148,68 +143,9 @@ public void readSchematicDataFromNBT(CompoundTag compound)
...
-        {
-            this.packName = blueprintDataProvider.getString(TAG_PACK);
-        }
 
-        if (this.packName == null)
-        {
-            this.packName = DEFAULT_STYLE;
-        }
+        this.packName = blueprintDataProvider.getString(TAG_PACK);
+        this.schematicPath = blueprintDataProvider.getString(TAG_PATH);
```
MotionlessTrain, line 148: I mean, am I misunderstanding something, or is this only set by the builder at the moment?

Thodor12: Afaik this is only called by the builder, I've done several constructed pastes but I've not once hit a debug trigger

src/main/java/com/minecolonies/core/colony/jobs/AbstractJobStructure.java
```diff
@@ -113,24 +113,25 @@ public void complete()
...
-                            if (te instanceof IBlueprintDataProviderBE)
+                            if (te instanceof IBlueprintDataProviderBE blueprintDataProviderBE)
                             {
                                 final CompoundTag tagData = compoundNBT.getCompound(TAG_BLUEPRINTDATA);
-                                final String schematicPath = tagData.getString(TAG_NAME);
-                                final String location = StructurePacks.getStructurePack(blueprint.getPackName()).getSubPath(Utils.resolvePath(blueprint.getFilePath(), schematicPath));
-
-                                tagData.putString(TAG_NAME, location);
                                 tagData.putString(NbtTagConstants.TAG_PACK, blueprint.getPackName());
+                                if (blueprint.getPrimaryBlockOffset().equals(offset))
```
someaddons, line 123: this looks weird, the offset is a relative position to the primary block offset, why are we equalling those two?

Thodor12: Offset is counting from 0,0,0, which is the same as which the primary block offset is determined from

someaddons (approved):

Moobien:
Just merge it already. nobody can upgrade any upgradable deco, and hasn't been able to do it in MONTHS. this is game breaking

Moobien:
> > Just merge it already. nobody can upgrade any upgradable deco, and hasn't been able to do it in MONTHS. this is game breaking
> 
> I still have to do local testing to make sure this doesn't break anything fundamental.

Isnt this a fix to fix a major game breaking bug?

Raycoms:
> > > Just merge it already. nobody can upgrade any upgradable deco, and hasn't been able to do it in MONTHS. this is game breaking
> > 
> > 
> > I still have to do local testing to make sure this doesn't break anything fundamental.
> 
> Isnt this a fix to fix a major game breaking bug?

It touches so many things, what if it breaks buildings? Colonies? Crashes? Servers?

PR #11085, opened 2025-08-08
Goats

uecasm:
Closes [discord request](https://discord.com/channels/139070364159311872/240841248750043137/1400992041365798932)

- The cowhand now knows how to corral and milk goats.
    - You **can** mix cows and goats (and mooshrooms) in the same pen, but you're encouraged to use dedicated pens (i.e. separate buildings) for each animal.
    - This is still a rather challenging endeavor, because goats can jump ridiculously high if they feel thwarted by your pitiful fences.
    - It can also be a rather fruitless endeavor, because they don't drop anything when butchered (but the cowhand will still do so if not disabled, to keep the population in check).
    - You might sometimes get some goat horns out of it, though.  Also bruises.  And did I mention the jumping?
- The player now knows how to milk goats into a large empty bottle.
- The cowhand has lost their ninja privileges and makes noise when milking now.
- All herders have had their telekinesis license revoked and now actually have to walk over to the animals they're herding instead of taking a single step in their general direction before finishing their task.

## Testing

Review please (should port)


src/main/java/com/minecolonies/core/items/ItemLargeBottle.java
```diff
@@ -50,6 +51,16 @@ public InteractionResult interactLivingEntity(@NotNull final ItemStack stack, @N
             stack.shrink(1);
             return InteractionResult.SUCCESS;
         }
+        else if (entity instanceof final Goat goat && !entity.isBaby())
```
someaddons, line 54: there is actually a bug here with the cows aswell, we need to set the milking cooldown of the entity aswell as checking it, would be nice if you could fix that for both :)

uecasm: Cows don't have any milking cooldown in vanilla either.

someaddons (approved):

Raycoms (approved):

PR #11101, opened 2025-08-11
Idle&foresterfix

### LINKED ISSUE #11096: Forester doesn't do their job

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.1006-snapshot

### Structurize Version

1.20.1-1.0.780-snapshot

_No response_

### Current Behavior

Forester is constantly stuck at in a state of "START_WORKING", and just hangs around the forester hut block. Downgrading to .995 fixes the issue, so it suggests something between then and this version did something.

### Expected Behavior

He's supposed to be cutting down trees and doing work.

### Reproduction Steps

1. Get a forester (Mine is level 3) and plant some trees around their hut block.
2. Watch as they stare very very hard at their forester hut block. And then do nothing.

### Logs

https://pastebin.com/XU4YLfSa

### Anything else?

Like said above, downgrading to .995 fixes the issue for now.

### Footer

---
## Viewers

Raycoms:
Closes #11096 
Closes #

- Forester fix
- Only leisure if not super busy right now

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
+++ b/src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
@@ -240,7 +240,8 @@ public class EntityAIWorkLumberjack extends AbstractEntityAICrafting<JobLumberja
             return getState();
         }
 
-        if (building.getModule(BuildingModules.ITEMLIST_SAPLING).getList().size() > 0)
+        if (building.getModule(BuildingModules.ITEMLIST_SAPLING).getList().size()
+            < IColonyManager.getInstance().getCompatibilityManager().getNumberOfSaplings())
         {
             // This got moved downwards compared to the AICrafting-implementation,
             // because in this case waiting for dumping is more important
--- a/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
+++ b/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
@@ -786,4 +786,10 @@ public class CompatibilityManager implements ICompatibilityManager
     {
         return furnaceRecipes;
     }
+
+    @Override
+    public int getNumberOfSaplings()
+    {
+        return saplings.size();
+    }
 }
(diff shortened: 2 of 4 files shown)


someaddons (approved):

src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
armele, line 1: I'd love to better understand this change so I can avoid creating a similar problem in the future!

src/main/java/com/minecolonies/core/entity/ai/workers/production/EntityAIWorkLumberjack.java
```diff
@@ -240,7 +240,8 @@ protected IAIState decide()
             return getState();
         }
 
-        if (building.getModule(BuildingModules.ITEMLIST_SAPLING).getList().size() > 0)
+        if (building.getModule(BuildingModules.ITEMLIST_SAPLING).getList().size()
+            < IColonyManager.getInstance().getCompatibilityManager().getNumberOfSaplings())
```
Thodor12, line 244: Little late to the party, but isn't this always true @Raycoms?

Unless you're counting for a mod removal which might remove the amount of saplings in the compat manager, you can never select more saplings in the list then there exist, no?

The only thing this could check is if you have every single sapling selected I guess

MotionlessTrain: I have the feeling the list works inverted. The list within the ITEMLIST_SAPLING module likely contains the saplings that have been turned off, rather than turned on
Note that the menu starts with all enabled, which would mean an empty list if my feeling is correct, and so everything disabled, means that all available items are in the list, i.e. that the list sizes between the module and the compatibility manager are equal

uecasm: Yep, it's a blacklist.  Empty list => all saplings on, full list => all saplings off.

PR #11109, opened 2025-08-14
Tavern wasn't spawning anything but gold and iron tier visitors

Thodor12:
Closes [discord report](https://discord.com/channels/139070364159311872/1284909613401247785/1405481982686658581)

- In the previous method, let's say we had a max tavern, limit is 10. The gaussian method was incapable of ever providing a good enough random. What happened is you have a mean of 5, a deviation of 0.5, so you can only get numbers between 4.5 and 5.5. This means that at max level, every visitor is either gold or iron boots. Diamond and netherite are simply impossible.
- New situation uses the same limit as before, but uses a regular nextInt call to grab a random rarity between 1-9, this way we're back to the old situation where any rarity between 1-[limit] is possible at any level of the tavern, so you get more variety.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/datalistener/RecruitmentItemsListener.java
+++ b/src/main/java/com/minecolonies/core/datalistener/RecruitmentItemsListener.java
@@ -5,9 +5,7 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.minecolonies.api.util.Log;
-import com.minecolonies.api.util.MathUtils;
 import com.minecolonies.api.util.constant.ColonyConstants;
-import com.minecolonies.api.util.constant.Constants;
 import net.minecraft.core.registries.BuiltInRegistries;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.packs.resources.ResourceManager;
@@ -92,17 +90,7 @@ public class RecruitmentItemsListener extends SimpleJsonResourceReloadListener
     @Nullable
     public static RecruitCost getRandomRecruitCost(final int buildingLevel)
     {
-        // Number between 1-9
-        final int limit = 3 * buildingLevel + 1;
-        int rarity = (int) MathUtils.RANDOM.nextGaussian(limit/2.0,0.5);
-        if (rarity <= 0)
-        {
-            rarity = 1;
-        }
-        else if (rarity >= limit)
-        {
-            rarity = limit - 1;
-        }
+        final int rarity = ColonyConstants.rand.nextInt(1, 3 * buildingLevel + 1);
         final List<RecruitCost> recruitCostsAtTier = RECRUIT_COSTS.get(rarity);
         return recruitCostsAtTier.get(ColonyConstants.rand.nextInt(recruitCostsAtTier.size()));
     }


someaddons (approved):

Raycoms (changes requested):
Then just make it a bigger gaussian variation, I didn't want all tiers to appear at the same likelyhood. 1.0 or 1.5 for example

SpecialT131:
isn't 1.0 a little bit to rare?

1.5 ~ 0.88% (1 of 100) rare but possible (Rarity 9)
1.0 ~ 0.02% (1 of 5000) extrem rare (Rarity 9)

Raycoms:
Can even go a bit beyond 1.5 then

SpecialT131:
@Raycoms Yes bit higher would be better. Done rundown of Diviation from 1.5 to 2.5.
| Boot Type     | stdDev 1.5 | stdDev 2.0 | stdDev 2.5 |
|------------------|------------|------------|------------|
| Leather (1-2)     | 4.7%         | 10.6%       | 15.8%      |
| Golden (3-4)     | 32.1%       | 29.6%       | 26.4%      |
| Iron (5-6)          | 47.2%       | 37.0%       | 30.5%      |
| Diamond (7-8)  | 15.1%       | 18.8%       | 19.1%      |
| Netherite (9)     | 0.9%         | 3.9%         | 8.2%       |

2.0 or 2.5 Seems the Sweetspot for rarity

The only problem is that it messes up the rarities of tavern level 1 and 2 because of the clamping effect:

Tavern lv1
| Rarity | Boot Type | stdDev 1.5 | stdDev 2.0 | stdDev 2.5 |
|--------|-----------|------------|------------|------------|
| **1**  | Leather   | 36.8%      | 40.2%      | 42.0%      |
| **2**  | Leather   | 26.1%      | 19.9%      | 15.8%      |
| **3**  | Golden    | 37.1%      | 39.8%      | 42.3%      |

Raycoms:
2.0 looks pretty good

MotionlessTrain:
Maybe a different deviation based on the tavern level?
I think it makes sense for a level 1 tavern to have lower deviations, as the options are closer together, so maybe 0.5 or 1 for a level 1 tavern, 1 or 1.5 for a level 2 tavern and 2 or 2.5 for a level 3 tavern?

SpecialT131:
@MotionlessTrain That would also be a viable option.

with the static deviation of 2.0 on tavern level 2, it looks "OK".
Anything above will result in higher "lows" and "highs" because of the clamping effect:

| Rarity | Boot Type | stdDev 1.5 | stdDev 2.0 | stdDev 2.5 |
|--------|-----------|------------|------------|------------|
| **1**  | Leather   | 8.9%        | 15.8%       | 21.2%      |
| **2**  | Leather   | 16.2%      | 15.2%       | 13.2%      |
| **3**  | Golden    | 24.7%      | 19.1%      | 15.4%      |
| **4**  | Golden    | 24.8%      | 19.3%      | 15.7%      |
| **5**  | Iron        | 16.3%      | 14.9%       | 13.3%      |
| **6**  | Iron        | 9.2%       | 15.7%        | 21.2%      |

The tavern lavel Scaling diviation could look like this:

```
/**
 * Get the standard deviation for Gaussian distribution based on building level.
 * Uses adaptive scaling: Level 1 = 1.2, Level 2 = 1.6, Level 3 = 2.0
 *
 * @param buildingLevel the tavern building level
 * @return the standard deviation to use
 */
private static double getStandardDeviation(int buildingLevel) 
{
    return 0.8 + (buildingLevel * 0.4);
}
```

and 
```
    // Calculate limit and adaptive standard deviation
    final int limit = 3 * buildingLevel + 1;
    final double stdDev = getStandardDeviation(buildingLevel);
    
    // Generate rarity using Gaussian distribution
    int rarity = (int) MathUtils.RANDOM.nextGaussian(limit/2.0, stdDev);
```
After this the distribution is ok if we look at tavern lvl 2:

| Rarity | Boot Type | Percentage |
|--------|-----------|------------|
| **1**  | Leather   | 10.4%      |
| **2**  | Leather   | 16.1%      |
| **3**  | Golden    | 23.5%      |
| **4**  | Golden    | 23.3%      |
| **5**  | Iron      | 16.2%      |
| **6**  | Iron      | 10.6%      |

Raycoms:
O would keep it at static two, we want to keep the variants not too spread out

PR #11112, opened 2025-08-14
Fix nether miner not going on expeditions [1.21]

### LINKED ISSUE #11111: Nether Miner Not Working (Not asking anything but food)

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1087

### Structurize Version

1.0.786

_No response_

### Current Behavior

My nether Miner is **not** going to nether anymore.
I think the problem is here since I downloaded **Minecolonies 1.1.1072** (the last update version on the server the problem happened first). It asks for nothing but food and when I give it, he only goes to the portal, if the portal is activated he just stands in the middle and **nothing happens** even in the hut GUI.

Image

 and if the portal is deactivated he just comes back to his hut and **ask for food again** (even if he already has more than a stack).

Image

Image

The two screenshots are from the server where you can see the worker is not the same as the last one who went to nether who died, but it's been real life days since he died, so the new one never went to nether, even if I fully stuffed him manually

As a mod asked on Discord, I tested it on a new instanse with Only Minecolonies and dependencies and same problem

Image

With this one I noticed he asks for absolutely nothing, no gear, no tools and nothing for his expedition.

### Expected Behavior

Nether Miner Asks for stuff and recipe blocks then goes to nether

### Reproduction Steps

1. Have a nether mine
2. Have a Nether Miner
3. He doesn't work

### Logs

https://gist.github.com/ykretzcode/77b9520900ec47c97e50bbe0cbc8efe5

### Anything else?

_No response_

### Footer

---
## Viewers

Thodor12:
Closes #11111

- Leisure system updates caused nether miner not to go on expeditions anymore
- ~~Nether miner was completely broken after the traveling manager changes, reverted for now until we can properly implement an event based system similar to colony expeditions~~
- Increased the traveling time to twenty seconds per item to make sure that the expeditions still take plenty of time, as before

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingNetherWorker.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingNetherWorker.java
@@ -14,9 +14,7 @@ import com.minecolonies.core.colony.buildings.modules.MinimumStockModule;
 import com.minecolonies.core.colony.buildings.modules.settings.BoolSetting;
 import com.minecolonies.core.colony.buildings.modules.settings.SettingKey;
 import net.minecraft.core.BlockPos;
-import net.minecraft.core.Holder;
 import net.minecraft.core.HolderLookup;
-import net.minecraft.core.registries.BuiltInRegistries;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.util.Tuple;
@@ -251,6 +249,16 @@ public class BuildingNetherWorker extends AbstractBuilding
             return portalLocation.above();
         }
         return null;
+    }
+
+    /**
+     * Get the tagged location where the worker can hide while "away" in the nether
+     *
+     * @return the tagged location, null if not available.
+     */
+    public BlockPos getVaultLocation()
+    {
+        return getFirstLocationFromTag("vault");
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Raycoms:
wdym with completely broken?

Thodor12:
Like I said during internal discussions, the AI of any worker doesn't run if they're out traveling and their entity is removed.
The way it works now is that:
- He goes to the nether for all of 10 seconds
- Returns, start randomly walking around the place
- Generates the loot table after returning
- You visually see them get xp (the particles)
- They walk back to the portal and close it
- Then walk back to the hut block to drop the loot

This looks completely silly, and it's why we need a similar system as colony expeditions to let the loot generation be ran in the background, and not part of the AI code. So the traveling manager can't be used just yet.

I discussed with Sam, and we basically had 2 ways to fix this, leave it as is and accept it looks silly. Or we revert the traveling manager for now and it looks ok. In either case we have to come up with a long term fix to make it proper.

Raycoms:
> Like I said during internal discussions, the AI of any worker doesn't run if they're out traveling and their entity is removed. The way it works now is that:
> 
> * He goes to the nether for all of 10 seconds
> * Returns, start randomly walking around the place
> * Generates the loot table after returning
> * You visually see them get xp (the particles)
> * They walk back to the portal and close it
> * Then walk back to the hut block to drop the loot
> 

So, we need to transition them to:
a) Take a lot longer
b) Tick the loottable stuff in the building, not in the entity AI
c) In the building then drop them out of the traveling if mission finished/death.

Tbh, this should be like this anyway, as I'm sure the old way was buggy with night, restarts, unload vault, etc.

Thodor12:
> So, we need to transition them to:
> a) Take a lot longer
> b) Tick the loottable stuff in the building, not in the entity AI
> c) In the building then drop them out of the traveling if mission finished/death.
> 
> Tbh, this should be like this anyway, as I'm sure the old way was buggy with night, restarts, unload vault, etc.

Yes, but because this is a bit larger of a change, and not the intended the bugfix, I wasn't planning on doing so in this PR

someaddons:
already ported

PR #11114, opened 2025-08-15
Set up standardized formatting for JSON and XML files

Thodor12:
- Configure Intellij formatting for XML and JSON files, and run a full format on the assets and data folder
- Datagen is explicitly excluded

## Testing

Review please


src/main/resources/assets/minecolonies/blockstates/blockconstructiontape.json
```diff
@@ -62,8 +58,7 @@
         "model": "minecolonies:block/blockconstructiontape",
         "uvlock": true
       }
-    },
-    {
+    }, {
```
someaddons, line 61: those look odd, think I prefer different lines

Thodor12: Unfortunately that seems to be one limitation of the Intellij formatter

PR #11120, opened 2025-08-16
Enhanced Building Info to Map

armele:
Closes [Discord Conversation](https://discord.com/channels/449079260070674443/548426496956432403/1394077229033717891)

- Remove minimum stock requests from the clipboard.
- Enhance map to show a building status icon where the building needs something that is player-resolved.
- Enhance map to highlight housing that has the capacity for new citizens.
- Minor adjustment to map to show custom building name if a building is renamed.

## Testing
- [ ] 
PlayerResolvedBuildingIndicator
ShowCustomBuildingNames
HighlightBox
HousingCapacity

Review please


uecasm:
> Remove minimum stock requests from the clipboard.

Why?  I could perhaps understand hiding them unless you switch to showing unimportant requests, but they're still useful because they indicate a supply chain failure.

Thodor12:
> > Remove minimum stock requests from the clipboard.
> 
> Why? I could perhaps understand hiding them unless you switch to showing unimportant requests, but they're still useful because they indicate a supply chain failure.

Isn't this what the important requests toggle also does?

MotionlessTrain:
It only filters out the unimportant things. Or it shows everything.
There is no toggle to only show important ones, or only show non-important ones (under which the minimum stock ones would fall)

Raycoms:
I think we could instead make the clipboard non sticky, and maybe we need some better wording than "player resolved request".

armele:
When you say "make the clipboard non-sticky" help me understand what that means?

Regarding minimum stock requests, in the discord chat the problem this was aimed at solving was people being overwhelmed by food requests in their clipboard.  From there the suggestion was made that min stock be excluded.

How about we make min stock requests part of what the "hide" toggle affects? This allows the player to control it.

armele:
Instead of "Player resolved requests" how about "Help needed"?

Raycoms:
Could just be needed then ye. I mean make min stock part of what you can hide and persist hide setting in item nbt

armele:
I didn't see an existing server message which might be used for this, but wanted to double-check before I add a new one to persist the clipboard state. Is there an existing one I should use?

armele:
Made minimum stack exclusion part of the "hide" toggle, and made the toggle state persistent, and visually clear.

MotionlessTrain:
That will help a lot! Thank you!

someaddons (approved):

Raycoms (approved):

PR #11124, opened 2025-08-20
Update study items datalistener

Thodor12:
- Created a BaseDataListener class that handles all the common stuff for processing the `neoforge:conditions`, `remove` and `replace` functionality.
- Due to the fact that this BaseDataListener now requires a ConditionContext, the setup around the data listeners themselves had to change marginally.
  - The DataListenerManager is basically a container for registering all the data listeners, this gets executed by the event where the data listeners are registered, and this is where the ConditionContext originates from.
  - The manager class has a dedicated method for safely getting the entries of these data listeners, by class, which is null safe. So in case the instantation of the listeners hasn't happened yet, they shouldn't be throwing NPE's but instead returning empty maps of entries.
  - The individual listeners no longer keep a static list of their entries, so the whole `StudyItemsListener.getStudyItems()` no longer exists. Instead you must now go through the manager like so: ` DataListenerManager.getInstance().getEntries(StudyItemListener.class)`. It's a bit longer, but it works fine and is completely null safe.

## Preview of the new options at work

So, by default the study items have 2 items, paper and books, very basic. I've added a custom datapack to my local world which will start to add new entries like so:
```json
{
  "remove": [
    "minecolonies:paper"
  ]
}
```
This one will remove the paper item from the list of items, and it's gone...
2025-08-20 18_21_02-minecolonies – BaseDataListener java  minecolonies main

I can also replace everything completely, as per the Neoforge documentation, like so:
```json
{
  "replace": true
}
```
And surprise...
2025-08-20 18_22_10-minecolonies – BaseDataListener java  minecolonies main

Lastly there's the conditions, if I modify one of the existing ones, let's say paper:
```json
{
  "neoforge:conditions": [
    {
      "type": "neoforge:false"
    }
  ],
  "item": "minecraft:paper",
  "skill_increase_chance": 40,
  "break_chance": 100
}
```
This condition is a forceful condition to just plainly always ignore the given item, but other conditions exist as well, so...
2025-08-20 18_25_19-minecolonies – BaseDataListener java  minecolonies main

The intention is to port this to all other listeners, but I'm keeping this to study items first for initial review.

## Testing

Review please


src/main/java/com/minecolonies/core/datalistener/BaseDataListener.java
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Mapping result class for returning the entry + warnings.
+     *
+     * @param success whether the entry was successfully parsed or not.
+     * @param item    the entry, if successfully parsed.
+     * @param reason  the reason when failed.
+     * @param <T>     the type of entry.
+     */
+    public record MappingResult<T>(
```
uecasm: Don't codecs already have a `DataResult` that does this?

src/main/java/com/minecolonies/core/datalistener/BaseDataListener.java
```diff
@@ -0,0 +1,221 @@
...
+            }
+            final JsonObject value = entry.getValue().getAsJsonObject();
+
+            try
+            {
+                final boolean replace = GsonHelper.getAsBoolean(value, KEY_REPLACE, false);
+                if (replace)
+                {
+                    newEntries.clear();
+                }
```
uecasm: This seems inconsistent with vanilla, unless I'm misunderstanding.

It looks like this will work across files and ignore all previously parsed files -- which is bad, because files are parsed in unspecified order, which means there's basically no safe or sensible way to use this.

In vanilla (see e.g. tags) the "replace" only affects whether multiple files with the **identical** ResourceLocation (from separate packs) are merged or not.  It has no effect between different ResourceLocations.

Thodor12: "replace: A boolean that will clear the data map before adding the values of this file. This should never be shipped by mods, and only be used by pack developers that want to overwrite this map for their own purposes."

https://docs.neoforged.net/docs/resources/server/datamaps/

And from Neo source:
```java
    private <T, R> Map<ResourceKey<R>, T> buildDataMap(Registry<R> registry, DataMapType<R, T> attachment, List<DataMapFile<T, R>> entries) {
        record WithSource<T, R>(T attachment, Either<TagKey<R>, ResourceKey<R>> source) {}
        final Map<ResourceKey<R>, WithSource<T, R>> result = new IdentityHashMap<>();
        final BiConsumer<Either<TagKey<R>, ResourceKey<R>>, Consumer<Holder<R>>> valueResolver = (key, cons) -> key.ifLeft(tag -> registry.getTagOrEmpty(tag).forEach(cons)).ifRight(k -> cons.accept(registry.getHolderOrThrow(k)));
        final DataMapValueMerger<R, T> merger = attachment instanceof AdvancedDataMapType<R, T, ?> adv ? adv.merger() : DataMapValueMerger.defaultMerger();
        entries.forEach(entry -> {
            if (entry.replace()) {
                result.clear();
            }
```

src/main/java/com/minecolonies/core/datalistener/BaseDataListener.java
Thodor12: Given that Ray wants to stay away as much as possible from Codecs, I opted not to use the full codec way that I initially posted in Discord (it wasn't possible to not use codecs with that way)

src/main/java/com/minecolonies/core/datalistener/BaseDataListener.java
uecasm: A data map is a single file with multiple entries, like a tag.  It is not separate files that affect the same data structure.  Separate files with separate resource locations are completely independent (like tags).

The study items are not even slightly a data map.  I don't think any of our other listeners are either, though I haven't checked too closely.

Thodor12: The point is that data maps also may be overridden by multiple other datapacks after one another, and that's what the replace option is meant for.
This is actually the exact same what tags do, however since all tags live within the same folders, there it makes sense to only wipe the given tag you intend to wipe. So it technically does the same thing.

Let's say you have a specific tag that is populated by vanilla, and item B is added by another datapack. You don't want any of this data present and want to bring everything yourself. You use `replace` and bring a completely new list of items for that given tag.

This implementation does the same, we ship a list of study items, another mod may still add to it, but a server owner wants none of it, he replaces everything in a locally installed datapack using `replace` and brings a complete set of custom study items.

uecasm: i.e. if you wanted to make the study items into a data map, it would be a single json file minecolonies:study_items that would list each element, and someone could add onto or replace by making their own study_items.json file (and no other filename would work).

To be clear: there's nothing wrong with the current structure of a directory with individual files.  But it does not conform to the data map (or tag) design and so the replace semantics are meaningless.

The remove semantics are kind of pointless too -- in theory all you need to do to remove a file added by another mod is to put in a blank file with the same ResourceLocation.  At least that's how it works for recipes etc; you don't need an explicit "remove" value (though it's not a bad thing to be explicit).

Thodor12: It exactly conforms to the system, compare my code with the code that data maps do, the implementation is exactly scaffolded based on that...

uecasm: No, it doesn't.  You're missing that the data maps operate within a single ResourceLocation only, whereas yours operates across multiple ResourceLocations.

Thodor12: Omg that's not true Mirality, go check the data maps implementation...

uecasm: I have.  You are not correct.  Read the docs more carefully.

someaddons (approved):

Raycoms (approved):

PR #11125, opened 2025-08-22
Smart colony id command arguments

uecasm:
Closes [discord request](https://discord.com/channels/139070364159311872/1407552866096251004)

- All commands that take a colony id now accept additional values in its place:
    - The colony id, as before (but now it suggests colony ids that you've been close to recently, i.e. are still in the client-side colony manager; you can still type other ids)
    - The name or id of a player, which resolves to their owned colony (and it suggests known player names, but not ids)
    - `@here`, which means the colony where you're running the command
    - `@mine`, which means your own colony
- `IColonyManager.getColonies()` now works on client side too.

## Testing

Review please (should port)

One caveat of note is that a lot of commands still require a player to execute them (used to check permissions), which in turn means that they still don't work in command blocks as desired in the linked thread.  But some now will.

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/ColonyManager.java
+++ b/src/main/java/com/minecolonies/core/colony/ColonyManager.java
@@ -327,22 +327,13 @@ public final class ColonyManager implements IColonyManager
     @NotNull
     public List<IColony> getColonies(@NotNull final Level w)
     {
-        if (w.isClientSide())
+        final IColonyManagerCapability cap = w.getCapability(COLONY_MANAGER_CAP, null).resolve().orElse(null);
+        if (cap == null)
         {
-            // this might be a subset of colonies since it's only those known to the player right now
-            final ColonyList<IColonyView> colonies = colonyViews.get(w.dimension());
-            return colonies == null ? List.of() : new ArrayList<>(colonies.getCopyAsList());
+            Log.getLogger().warn(MISSING_WORLD_CAP_MESSAGE);
+            return Collections.emptyList();
         }
-        else
-        {
-            final IColonyManagerCapability cap = w.getCapability(COLONY_MANAGER_CAP, null).resolve().orElse(null);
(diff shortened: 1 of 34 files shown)


src/main/java/com/minecolonies/core/colony/ColonyManager.java
```diff
@@ -327,13 +327,22 @@ public boolean isFarEnoughFromColonies(@NotNull final Level w, @NotNull final Bl
     @NotNull
     public List<IColony> getColonies(@NotNull final Level w)
     {
-        final IColonyManagerCapability cap = w.getCapability(COLONY_MANAGER_CAP, null).resolve().orElse(null);
-        if (cap == null)
+        if (w.isClientSide())
```
someaddons: we have a specific function to get an IColony already, should make a new function in the same style instead. Else it'd be confusing that this one does return client colonies but all the other calls do not

uecasm: This is that function though.  It just stopped working on client side at some point.

tbf the whole colony API is messy, with `IColonyManager` having some single-sided methods and some both-sided methods, and `IColony` being both-sided some of the time and not all of the time.  But this is where we are right now.

someaddons: the specific both sided function is called
 `public IColony getIColony(@NotNull final Level w, @NotNull final BlockPos pos)` 
 Instead of modifying this serverside used one, potentially introducing hard to track sidededness bugs I want an extra new method instead called 
 public List<IColony> getIColonies(@NotNull final Level w)
 which follows the same style

uecasm: Sounds messy, but k.

someaddons (approved):

Raycoms:
Please port =)

PR #11154, opened 2025-09-01
Fix research problems and improve code

### LINKED ISSUE #11151: University Research not counting multiple structures for research requirements

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1105

### Structurize Version

1.0.787

_No response_

### Current Behavior

University Research is only counting individual structures levels for requirments

### Expected Behavior

University Research should be totalling all related structure levels within requirements

### Reproduction Steps

1. start new colony
2. build several structures that unlock research (eg residence or guard towers) along with a university
3. check research tab

### Logs

https://pastebin.com/m92jjBFB

### Anything else?

_No response_

### Footer

---
 ## Viewers

Thodor12:
Closes #11151

- Ensure that the requirements parse the ResourceLocation during NBT/JSON parse, not during validation
- Requirement translation key should've never been translated by mod id
- Instead of relying on "block.\<namespace\>.blockhut.\<path\>", just fetch the building entry it's translation key
- Removed "mandatory building" requirement class, as it's virtually identical to the building requirement itself, changed parsing inside the building requirement to accomodate for mandatory buildings
  - This is done because apparently the research tree window doesn't actually accomodate for mandatory buildings when showing the conditions, rather than copying an enormous block of code to make this work, and leaving a near duplicate class standing, I opted to merge the 2
- Changed several datagen things
  - Actually make use of mandatory buildings for at least the townhall, that way this can be validated that it actually functions. (The townhall can only exist once anyway)
  - Also include alternate buildings, used for some specific researches
  - Changed some other researches to include the correct building requirement (rich veins needing a miner, not blacksmith)
- Fixed datagen code for alternate buildings ~~(unused, so can't be validated)~~
- Modified the devilish `IColony#hasBuilding` method to no longer accept a string building name, rather a ResourceLocation
- Improved the way the research tree window displays alternate buildings, by using a rotating icon to highlight the different buildings, similar to tag costs. Also removed the red separator
- Removed the restriction on only having a single alternate building requirement

## Breaking changes
- JSON structures of mandatory and alternate building research requirements have been modified due to this, existing datapacks that utilize either of these will probably break

## Testing

Review please


uecasm:
> Removed "mandatory building" requirement class

I never remember which is which, but there's one requirement for "N levels total across all buildings of this type", another for "N levels max in a single building of this type", and we still need both of these (even though the built-in pack usually only uses the first).

(There's also another which is similar but "any one of these buildings" rather than "all of these buildings", which is the norm.)

So I don't think removing this is safe.

Thodor12:
> Removed "mandatory building" requirement class, as it's virtually identical to the building requirement itself, changed parsing inside the building requirement to accomodate for mandatory buildings

I've removed the class itself, not the requirement, both requirements now parse through the same class:
```java
    /**
     * Create a building research requirement.
     *
     * @param json the json containing the relevant data.
     */
    public BuildingResearchRequirement(final JsonObject json)
    {
        // TODO: 1.22: Change to GsonHelper.getAsResourceLocation(json, RESEARCH_REQUIREMENT_BUILDING_PROP);
        building = parseFallbackBuildingKey(GsonHelper.getAsString(json, RESEARCH_REQUIREMENT_BUILDING_PROP));
        buildingLevel = GsonHelper.getAsInt(json, RESEARCH_REQUIREMENT_BUILDING_LEVEL_PROP);
        singleBuilding = GsonHelper.getAsResourceLocation(json, RESEARCH_REQUIREMENT_TYPE_PROP).equals(BUILDING_MANDATORY_RESEARCH_REQ_ID);
    }
```

```java
        buildingResearchRequirement = create(BUILDING_RESEARCH_REQ_ID, BuildingResearchRequirement::new, BuildingResearchRequirement::new);
        buildingAlternatesResearchRequirement = create(BUILDING_ALTERNATES_RESEARCH_REQ_ID, BuildingAlternatesResearchRequirement::new, BuildingAlternatesResearchRequirement::new);
        buildingMandatoryResearchRequirement = create(BUILDING_MANDATORY_RESEARCH_REQ_ID, BuildingResearchRequirement::new, BuildingResearchRequirement::new);
```

> I never remember which is which, but there's one requirement for "N levels total across all buildings of this type", another for "N levels max in a single building of this type", and we still need both of these (even though the built-in pack usually only uses the first).

Regular building requirements > "Must have building totalling at least X levels"
Mandatory buildings requirements > "Must have one building of at least X level"

src/main/java/com/minecolonies/core/event/EventHandler.java
```diff
@@ -826,7 +827,7 @@ public static void onEntityConverted(@NotNull final LivingConversionEvent.Pre ev
         {
             final Level world = entity.getCommandSenderWorld();
             final IColony colony = IColonyManager.getInstance().getIColony(world, entity.blockPosition());
-            if (colony != null && colony.hasBuilding("tavern", 1, false))
+            if (colony != null && colony.hasBuilding(new ResourceLocation(Constants.MOD_ID, ModBuildings.TAVERN_ID), 1, false))
```
Raycoms: Can't we get this from the ModBuildings.TAVERN.getResLoc() or so?

Thodor12: ModBuildings in 1.20 only lists strings still, compared to 1.21 where we finally tackled this (I think).
(Hence why all the datagen code has to make new reslocs for each building requirement)

Raycoms (dismissed):

src/main/java/com/minecolonies/core/event/EventHandler.java
someaddons: ModBuildings.tavern.get().getRegistryName() does exist in 1.20

someaddons (approved):

armele:
Thanks for the save!

PR #11173, opened 2025-09-07
Fixed enchanter not doing any work [1.21]

Thodor12:
See #11172


armele:
Any chance the Alchemist needs the same thing?  A few people have reported on discord they're not making mistletoe, and I've observed the same thing when playing (but have not debugged/investigated code yet).

PR #11186, opened 2025-09-15
Cardinal direction messages

someaddons:
Closes #
Closes #

- Adjust builder building messages and citizen death messages to display the cardinal direction instead of coordinates. The coordinates are shown on hover
- Townhall now displays its level
- Adjusted sickness icon a bit to look more like high temp
- Enable citizen AI history tracking when citizen info command is used

## Testing

Review please


src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -361,15 +361,16 @@
   "entity.worker.inventoryfullchestfull": "Please make space in my chests/racks. I am not able to dump my inventory.",
 
   "entity.builder.messagebuildstart": "I've started work on %s!",
-  "entity.builder.messagebuildcomplete": "I've finished building the %s at %s %s %s!",
-  "entity.builder.messagerepaircomplete": "I've finished repairing the %s at %s %s %s!",
-  "entity.builder.messageremovalcomplete": "I've finished deconstructing building %s! §6To move the building while retaining its current level, use the 'Pick Up' button on the building's 'Build Options' GUI.",
+  "entity.builder.messagebuildcomplete": "The %s is complete on the %s side of the colony.",
```
Raycoms: Completed the repair of %s on the ....
Completed building %s on...
Completed deconstructing %s on ...

Btw, on the top where you construct the message, I think there was some "important" flag on the message builder or so which makes it green too.

someaddons: I can make it yellow with important type

Raycoms: Green is better then, you're right

PR #11202, opened 2025-09-22
allow teleport back and tag support and bug fix

Raycoms:
Closes #
Closes #

- Fix an accepting request bug which didn't invert order
- Allow opening tp window for non perm players
- Allow tp if permission is set
- Teleport to gate tag pos.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/blocks/huts/BlockHutGateHouse.java
+++ b/src/main/java/com/minecolonies/core/blocks/huts/BlockHutGateHouse.java
@@ -1,6 +1,7 @@
 package com.minecolonies.core.blocks.huts;
 
 import com.minecolonies.api.blocks.AbstractBlockHut;
+import com.minecolonies.api.colony.IColony;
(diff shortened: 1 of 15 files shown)


src/main/java/com/minecolonies/core/blocks/huts/BlockHutGateHouse.java
```diff
@@ -31,4 +44,43 @@ public BuildingEntry getBuildingEntry()
...
+        final Level worldIn,
+        final BlockPos pos,
+        final Player player,
+        final InteractionHand hand,
+        final BlockHitResult ray)
+    {
+       /*
+        If the world is client, open the gui of the building
+         */
+        if (worldIn.isClientSide)
```
someaddons, line 68: need to check permissions on serverside here then and deny interaction since canRightClickWithoutPermissions  makes it skip the existing ones

someaddons (approved):

PR #11203, opened 2025-09-23
Some necessary de-fuckery of the window logic

Thodor12:
- General cleanup of some of the module window code (note: no actual modifications were made)
- Firstly, all of the windows now are required to pass a proper ResourceLocation as the window identifier, no more strings!
- All of the page identifiers were ripped out of constants, and all moved into the respective windows directly
- Constructors of windows have all been made that the `IBuildingModuleView` or whatever implementation must be passed into the window as the only argument. Building views are obsolete as these can be obtained via the module view themselves anyway
- Unhinged the root building window logic, there used to be 2 classes, the `AbstractWindowModuleBuilding` and the `AbstractModuleWindow`. The setup behind this was a little confusing. The `AbstractWindowModuleBuilding` was handling the "main" page of buildings, but was needing a module, which had to just be passed as null. I shuffled the logic around a bit more easily separated. There are now 3 classes: `AbstractBuildingWindow` which stores the building view, `AbstractModuleWindow`, which extends from the building window, and stores the module view, and `AbstractBuildingMainWindow` which defines all the "main" page code
- All of the doubly storing of the building/module views has been removed, and are now using the protected fields from the superclasses
- Made support for potentially rendering tab icons on either side of the page, via new methods in the superclass

## Testing

Review please


src/main/java/com/minecolonies/core/client/gui/AbstractBuildingMainWindow.java
```diff
@@ -54,90 +54,74 @@ public AbstractWindowModuleBuilding(final B building, final String resource)
...
+            new BuildRequestMessage(buildingView, BuildRequestMessage.Mode.REMOVE, BlockPos.ZERO).sendToServer();
         }
         else
         {
-            new WindowBuildBuilding(building.getColony(), building).open();
+            new WindowBuildBuilding(buildingView.getColony(), buildingView).open();
         }
     }
 
+    /**
```
Raycoms: Please don't re-arrange methods in PRs, it's a nightmare to review =D

src/main/java/com/minecolonies/core/client/gui/WindowClipBoard.java
```diff
@@ -128,11 +122,6 @@ public ImmutableList<IRequest<?>> getOpenRequestsFromBuilding(final IBuildingVie
 
         final IRequestManager requestManager = colony.getRequestManager();
 
-        if (requestManager == null)
```
Raycoms, line 131: This was intentionally there

src/main/java/com/minecolonies/core/client/gui/modules/ItemListModuleWindow.java
```diff
@@ -119,15 +105,14 @@ private void switchClicked(@NotNull final Button button)
         final ItemStorage item = currentDisplayedList.get(row);
         final boolean on = button.getText().equals(Component.translatableEscape(ON));
         final boolean add = (on && isInverted) || (!on && !isInverted);
-        final IItemListModuleView module = building.getModuleViewMatching(ItemListModuleView.class, view -> view.getId().equals(id));
 
         if (add)
         {
-            module.addItem(item);
+            moduleView.addItem(item);
```
Raycoms, line 111: Do we deserialize modules into the same one, or do we have to re-query them from the building?

src/main/java/com/minecolonies/core/client/gui/townhall/AbstractWindowTownHall.java
```diff
@@ -51,15 +52,4 @@ public AbstractWindowTownHall(final BuildingTownHall.View townHall, final String
...
      */
     protected abstract String getWindowId();
-
-    /**
-     * Returns the name of a building.
-     *
-     * @return Name of a building.
-     */
-    @Override
-    public String getBuildingName()
```
Raycoms, line 61: Also intentional

Thodor12: This method wasn't being used, so I just removed it

src/main/java/com/minecolonies/core/client/gui/modules/ItemListModuleWindow.java
Thodor12: Looks like they get deserialized into the same one

Thodor12: Modules get deserialized from the `AbstractBuildingView`, which get deserialized from the `ColonyView` under this method:
```java
    @Override
    public void handleColonyBuildingViewMessage(final BlockPos buildingId, @NotNull final RegistryFriendlyByteBuf buf)
    {
        if (buildings.containsKey(buildingId))
        {
            //Read the string first to set up the buffer.
            buf.readUtf(32767);
            buildings.get(buildingId).deserialize(buf);
        }
        else
        {
            @Nullable final IBuildingView building = IBuildingDataManager.getInstance().createViewFrom(this, buildingId, buf);
            if (building != null)
            {
                buildings.put(building.getID(), building);

                if (building instanceof BuildingTownHall.View)
                {
                    townHall = (ITownHallView) building;
                }
            }
        }
    }
```

it checks for existing instances of buildings then deserialized into it, or creates a new building view if it doesn't exist on the client yet

Raycoms (dismissed):

someaddons (dismissed):

Raycoms:
conflicts

PR #11205, opened 2025-09-23
Reusable module templates

Thodor12:
- Dependent on #11203 
- Introduces a template system that allows a player to configure templates or presets of configuration on certain modules per type of building, this works on the following modules:
  - Minimum stock
  - Item lists
  - Entity lists
  - Restaurant menu
  - Settings
- Templates can be accessed by clicking a new tab on the right of the window, this tab will only appear for modules that implement the new `ITemplateModule`
- A player can create new templates by filling the input and clicking the add icon, which will create a new template based on the current configuration of the module
- Other buildings can then opt to "use" the template, which will overwrite it's configuration based on the template
- Any building may overwrite a template by clicking on save, which will automatically synchronize the configuration across all tracking buildings
- Templates can be removed by pressing the X

2025-09-23 18_51_55-Window
2025-09-23 18_50_57-Window

## Implementation details
- Two new interfaces must be implemented on both the SERVER and CLIENT: `ITemplateModule` and `ITemplateModuleView`
- This only requires both sides to implement the `getTemplateStorageId()` method, **the values that the server and client return must be identical!!**
- The module must already implement `IPersistentModule`, if not you must implement the (de)serialize methods
- Nothing else is required for starters, however it can be further customized
  - The server can (and should) implement the `resetTemplateAssignment()` method on places where the configuration is modified, this is so that if the building is actively tracking a template, upon modification of data it'll automatically untrack the template, meaning that if another module syncs, the changes in this building won't be wiped again
  - The client should overwrite the `getDescriptionText(CompoundTag data)` method, this is to implement a more appropriate description text for the current module, otherwise it'll say "No description" on the second line in the window
- General note: This implementation only works on modules where configuration must be non unique

## Testing

Review please


KrzesloSzatan:
Can i vote for this ? Or i dont know, like ask why its not merged for a year now ? This is something i thing should be available by default in the mod.

PR #11219, opened 2025-09-29
Cavalry Prework (utilities and refactoring)

armele:
Closes None

PR 1 of 4

- Minor refactoring, utilities and language components that support upcoming pull requests for a CavalryHorseEntity, a Stablemaster (job and building) to care for them, and a Cavalry guard unit to supplement the colony military force.
- Notable in this PR is the introduction of the IMinecoloniesPather interface and some helper methods in EntityNavigationUtils that take this instead of AbstractFastMinecoloniesEntity.  This will allow the CavalryHorseEntity to extend Horse but still take advantage of minecolonies navigation utils and the MinecoloniesAdvancedPathing class.
- There are also helper functions that allow the Cavalry guards to patrol exterior corners of buildings (instead of building hut blocks) since they are restricted from going in doors.

## Testing

Review please


src/main/java/com/minecolonies/core/entity/pathfinding/PathingOptions.java
```diff
@@ -86,9 +86,15 @@ public class PathingOptions
...
     /**
-     * Allowed to enter doors?
+     * Allowed to enter doors (includes gates).
      */
     private boolean enterDoors       = false;
+
+    /**
+     * Allowed to enter gates?
+     */
+    private boolean enterGates       = false;
```
someaddons: this should default to true, else no entity would walk through gates if we merge it. Btw raiders are intended to walk through those

src/main/java/com/minecolonies/api/colony/requestsystem/requestable/StackList.java
```diff
@@ -431,4 +441,48 @@ public Set<TypeToken<?>> getSuperClasses()
...
+    {
+        final Registry<Item> itemReg = registryAccess.registryOrThrow(Registries.ITEM);
+        final Optional<HolderSet.Named<Item>> opt = itemReg.getTag(tag);
+
+        if (opt.isEmpty())
+        {
+            return List.of();
+        }
+
+        return opt.get()
```
Raycoms: As this is probably called with higher frequency we should probably make this a normal for loop and directly do it in the copy constructor.

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/EntityNavigationUtils.java
```diff
@@ -155,6 +161,38 @@ public static boolean walkToPos(
...
+     * @param nav the navigation to use
+     * @param entity the entity to move
+     * @param desiredPosition the position to move to
+     * @param distToDesired the minimum distance to the desired position
+     * @param safeDestination if the destination is safe and should be set
+     * @param speedFactor the speed to move at
+     * @return true if the navigation has been updated, false otherwise
+     */
+    protected static boolean walkToPosHelper(
+        MinecoloniesAdvancedPathNavigate nav, Entity entity, final BlockPos desiredPosition,
```
someaddons: I dont think we should have the navigator as parameter, it should always be the given entity's navigator.
For the horse setup I'd make this:
The worker does the pathfinding still, the difference is in path follow.
When the worker finished its pathfinding and the navigator processed the result we add a vehicle check and when they are riding a horse then the path is passed along to the horse's navigator.
Same with pathfollow, if the worker's navigator is riding a horse(vehicle) it just does nothing/does not move the entity.

That way the whole movement AI/decisions stay with the worker, only difference is that the horse does execute the intended movement

src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -1379,6 +1379,7 @@ protected boolean isPassable(@NotNull final BlockState block, final int x, final
                 else
                 {
                     return pathingOptions.canEnterDoors() && (block.getBlock() instanceof DoorBlock || block.getBlock() instanceof FenceGateBlock)
+                             || (pathingOptions.canEnterGates() &&  block.getBlock() instanceof FenceGateBlock)
```
Raycoms: odd whitespace after &&

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/EntityNavigationUtils.java
```diff
@@ -290,4 +354,129 @@ public static boolean walkToRandomPosAround(final AbstractFastMinecoloniesEntity
...
+        int y = level.getHeight(Heightmap.Types.MOTION_BLOCKING_NO_LEAVES, x, z);
+        BlockPos pos = new BlockPos(x, y, z);
+
+        // Fast path
+        if (isStandable(level, pos)) 
+        {
+            return pos;
+        }
+
+        // Tiny local search up/down to land on something standable
```
Raycoms: I thought we had a utility for this somewhere already

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/EntityNavigationUtils.java
```diff
@@ -290,4 +354,129 @@ public static boolean walkToRandomPosAround(final AbstractFastMinecoloniesEntity
...
     }
+
+    /**
+     * Check if an entity can stand at the given BlockPos without clipping into any blocks or fluids or falling through the ground.
+     *
+     * @param level the level to check
+     * @param feet the position to check
+     * @return true if the entity can stand at the given position
+     */
+    public static boolean isStandable(Level level, BlockPos feet)
```
someaddons: pretty sure our existing "SurfaceType" we use in pathfinding does cover this

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/EntityNavigationUtils.java
```diff
@@ -290,4 +354,129 @@ public static boolean walkToRandomPosAround(final AbstractFastMinecoloniesEntity
...
+
+    /**
+     * Find the closest valid surface position outside the given building that is a corner of its footprint.
+     * If the building does not have a valid corner, return the first corner of its footprint.
+     * Example use case: Colony patrolling (which cannot enter buildings)
+     * @param building the building to find a corner for
+     * @param referencePos the position to find the closest corner to
+     * @return the closest valid surface position outside the building that is a corner of its footprint
+     */
+    public static BlockPos closestOutsideCornerofBuilding(@Nonnull IBuilding building, @Nonnull BlockPos referencePos)
```
Raycoms: This is quite expensive. I wonder if there is a better way we can handle this. For example by a building tag.

someaddons: I see your intention here to find a guarding spot outside a building, but manually searching a "standable" position here doesnt really work well, as it doesnt take reachability into account. E.g. the standing space you find may be fully enclosed.
This kind of "search" is better done by a pathfinding job which tries to walk close to the given position instead, as searching many blocks like this is also expensive performance wise

Raycoms: A pathjob from hut block to "outside of building boundaries". But it can't enter doors once it's outside the boundary? =D

someaddons: You could also simply just use one of the lower building corners and have a pathjob try to "walk close" instead

Raycoms: So, I think the main use case is to have cavalry to not ride into building by horse. So it's probably easiest to do a general "walk close to building boundary and accept the first pos in building boundary" but don't allow entering doors.

armele: This all seems reasonable - how about I use the existing PathJobMoveCloseToXNearY path job - effectively, just use MinecoloniesAdvancedPathNavigate.walkCloseToXNearY()?

I would randomly pick a lower building corner as the target point.

armele: I do really like the building tag idea as well, though... perhaps we allow schematic designers to designate patrol points with a building tag, and fall back to the lower building corner for buildings which do not have these tags (which of course will be all of them initially).

armele: This strategy seems quite viable, after initial testing. Most of the changes will be to the Cavalry AI, and all the questionable stuff in EntityNavigationUtils will go away ;)

armele:
Anything else you'd like to see changed here?  I don't want to submit 2 of 4 while this one is outstanding, since they build on each other. 2 of 4 will be the horse entity itself, and a handful of associated tweaks to reference it.  No gameplay impact until 3 of 4 (the stable) and then 4 of 4 will round it out (with the Cavalry job).

src/main/java/com/minecolonies/core/entity/ai/workers/guard/AbstractEntityAIGuard.java
```diff
@@ -493,6 +493,16 @@ protected IAIState startWorkingAtOwnBuilding()
...
+    /**
+     * Provides a random patrol point from all buildings in the colony when the guard is set to automatic patrol mode.
+     * @return a BlockPos of the patrol point.
+     */
+    protected BlockPos randomPatrolPoint()
+    {
+        return buildingGuards.getColony().getBuildingManager().getRandomBuilding(b -> true);
+    }
+
+
```
Raycoms: extra whitespace

src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -1379,6 +1379,7 @@ protected boolean isPassable(@NotNull final BlockState block, final int x, final
                 else
                 {
                     return pathingOptions.canEnterDoors() && (block.getBlock() instanceof DoorBlock || block.getBlock() instanceof FenceGateBlock)
```
Raycoms: I think it's fine to have only canEnterGates apply to this and not canEnterDoors.
Also, I think we should with this new pathingOption also check where else we currently allow gates and also check the option there & make sure it's set properly for raiders

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/IMinecoloniesPather.java
```diff
@@ -0,0 +1,11 @@
+package com.minecolonies.core.entity.pathfinding.navigation;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.entity.ai.navigation.PathNavigation;
+
+public interface IMinecoloniesPather
+{
+    // This interface should be applied only to classes whose navigation can be safely cast to MinecoloniesAdvancedPathNavigate
+    public PathNavigation getNavigation();
+    public BlockPos blockPosition();
```
Raycoms: javadoc

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/IMinecoloniesPather.java
```diff
@@ -0,0 +1,11 @@
+package com.minecolonies.core.entity.pathfinding.navigation;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.entity.ai.navigation.PathNavigation;
+
+public interface IMinecoloniesPather
+{
+    // This interface should be applied only to classes whose navigation can be safely cast to MinecoloniesAdvancedPathNavigate
+    public PathNavigation getNavigation();
```
Raycoms: this comment should move above the interface declaration with the /** javadoc

src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
armele: I've explicitly separated gates and doors, and added gate access in wherever door access was previously allowed.

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/IMinecoloniesPather.java
```diff
@@ -0,0 +1,13 @@
...
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.world.entity.ai.navigation.PathNavigation;
+
+/**
+ * This interface should be applied only to classes whose navigation can be safely cast to MinecoloniesAdvancedPathNavigate
+ */
+public interface IMinecoloniesPather
+{
+    public PathNavigation getNavigation();
```
someaddons: since vanilla already declares this method it'll have problems with the unmapped method name in 1.20.1 forge, 
since the intention here is to get the minecolonies navigator why not change this to a custom
getMcolNavigator()
function? which the entities then implement.
Same issues with the blockpos, though you can just remove that and instead use a combined type for the Utilities param, e.g. T <T extends Mob & IThreatTableEntity>
so you automatically get the Interface & normal entity methods

armele: Current minecolonies entities use getNavigation to return the advanced minecolonies navigation, and I introduced this interface to allow non-Citizens to use some of the same helper functions as the citizens. My interpretation of this feedback is that I'd end up changing the method name on the citizens as well.  I wanted to double-check my understanding, since that feels more impactful that might be desired.

someaddons: the getNavigation() used is from vanilla, and we just cast it all the time, because we replaced the navigator object it returns with our custom one. You could do the same or add a custom call via this interface which returns the right typed navigator for convenience.
Its just you cannot make an interface method with the exact same name as the vanilla call

armele: I went the route of T <T extends Mob> (should have done that from the beginning - it was exactly my intention with the interface). 

I retested things in my merged master branch, and this 1 of 4 branch.

Raycoms (approved):

someaddons (approved):

Raycoms:
Please port

PR #11223, opened 2025-09-30
feat: Make name style changes apply instantly to all citizens

haggbart:
- Name style changes now apply instantly to all existing citizens and visitors
- Added ColonyNameStyleChangedModEvent

Previously, when changing the colony name pack (e.g., from Spanish to Norse), existing citizens would keep their old names. This created an inconsistency where players could change citizen appearance instantly with texture styles, but names only updated for newly spawned citizens.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/Colony.java
+++ b/src/main/java/com/minecolonies/core/colony/Colony.java
@@ -1939,6 +1939,14 @@ public class Colony implements IColony
     public void setNameStyle(final String style)
     {
         this.nameStyle = style;
+        for (final ICitizenData citizen : citizenManager.getCitizens())
+        {
+            citizen.regenerateName();
+        }
+        for (final ICivilianData visitor : visitorManager.getCivilianDataMap().values())
+        {
+            visitor.regenerateName();
+        }
         this.markDirty();
     }
 
--- a/src/main/java/com/minecolonies/api/colony/ICivilianData.java
+++ b/src/main/java/com/minecolonies/api/colony/ICivilianData.java
@@ -213,4 +213,19 @@ public interface ICivilianData extends ICitizen, INBTSerializable<CompoundTag>
      * @return uuid
      */
     UUID getUUID();
+
+    /**
+     * Set the name with optional preserved flag.
+     *
+     * @param name the name to set.
+     * @param isPreserved true to mark this name as preserved (from nametag or special visitor) and prevent regeneration.
+     */
+    void setName(String name, boolean isPreserved);
+
+    /**
+     * Set whether this civilian has a preserved name.
+     *
+     * @param hasPreservedName true if the name should be preserved.
+     */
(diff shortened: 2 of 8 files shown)


src/main/java/com/minecolonies/core/colony/Colony.java
```diff
@@ -1939,6 +1939,14 @@ public String getTextureStyleId()
     public void setNameStyle(final String style)
     {
         this.nameStyle = style;
+        for (final ICitizenData citizen : citizenManager.getCitizens())
```
Raycoms, line 1942: I think we need a flag for special visitors and nametag changed citizen to not change their names

Shadizar:
Is this opt in?
I'd hope existing citizens would never change automatically! I may intentionally be making a mixture colony over time and once they're present i certainly would hate them changing!

haggbart:
> Is this opt in? I'd hope existing citizens would never change automatically! I may intentionally be making a mixture colony over time and once they're present i certainly would hate them changing!

I think that's a valid concern. Makes sense to be opt-in via config to not cause any surprises.

Shadizar:
This might be a good alternative?  https://github.com/ldtteam/minecolonies-features/issues/740

So you don't get starter "wrong" names but don't also change existing ones unexpectedly

Raycoms:
Instead of opt in it could also be a button option.

Thodor12:
Yeah I'm not sure a config option is going to work for this as this is going to vary from player to player whether they want this or not.
Personally I'd say, leave this be, don't touch it, as if you really care that much you can always nametag replace the names

Raycoms:
> Yeah I'm not sure a config option is going to work for this as this is going to vary from player to player whether they want this or not. Personally I'd say, leave this be, don't touch it, as if you really care that much you can always nametag replace the names

I think a  button + manually named citizens don't change works well.

haggbart:
> This might be a good alternative? [ldtteam/minecolonies-features#740](https://github.com/ldtteam/minecolonies-features/issues/740)
> 
> So you don't get starter "wrong" names but don't also change existing ones unexpectedly

A button like @Raycoms mentioned would solve this too, as you can just change it after you start your colony :)

src/main/java/com/minecolonies/core/colony/Colony.java
haggbart: Added a TAG_PRESERVED_NAME, tested with both special visitors and nametag

haggbart:
Screenshot 2025-10-01 at 01 12 26

Something like this

Shadizar:
Very cool
Thanks for considering the options, greatly appreciated

Raycoms:
Why closed?

Thodor12:
Fork was deleted, they're recreating it?

haggbart:
No sleep, mistakes were made. I'll make a new PR referencing this one in it

PR #11225, opened 2025-10-01
feat: Make name style changes apply instantly to all citizens

haggbart:
- Name style changes now apply instantly to all existing citizens and visitors
- Added ColonyNameStyleChangedModEvent

Previously, when changing the colony name pack (e.g., from Spanish to Norse), existing citizens would keep their old names. This created an inconsistency where players could change citizen appearance instantly with texture styles, but names only updated for newly spawned citizens.

## Testing

Review please 

Referencing #11223 🤡

## To be decided
- [ ] Button vs config option (if we want that)

A button could work something like this:
image

This would cover feature request https://github.com/ldtteam/minecolonies-features/issues/740


Raycoms (commented):
I've been thinking about this. I think we should store in the citizen-data the name-pack, so that only citizens are renamed when the namepack doesn't match. If on-load this is empty (and not special-name), we can query it from the colony.

Then we can only display the button if there are citizens with non matching names for the pack.

hasPreservedName should be maybe something around "specialName".

src/main/java/com/minecolonies/core/colony/CitizenData.java
```diff
@@ -1159,6 +1180,21 @@ public void setName(final String name)
...
+        this.name = name;
+        this.hasPreservedName = isPreserved;
+        markDirty(0);
+    }
+
+    @Override
+    public void setHasPreservedName(final boolean hasPreservedName)
+    {
+        this.hasPreservedName = hasPreservedName;
+        markDirty(0);
```
Raycoms: This can be 60, doesn't need to be synched to the client quickly

haggbart:
> I've been thinking about this. I think we should store in the citizen-data the name-pack, so that only citizens are renamed when the namepack doesn't match. If on-load this is empty (and not special-name), we can query it from the colony.
> 
> Then we can only display the button if there are citizens with non matching names for the pack.
> 
> hasPreservedName should be maybe something around "specialName".

Feels a bit overengineered tracking all this just to hide a button, but yeah it's cleaner UX. 

I realigned the buttons with consistent spacing and increased the dropdown height so you can see more name packs at once.
   
Screenshot 2025-10-18 at 17 56 48

someaddons:
We should add a nametag cost to the button, as that hooks nicely into how you rename things in vanilla. 
1 Nametag per 25 citizens sounds reasonable(minimum 1)
That'll need the citizen sync their nametagpack to the client then, then you also dont need to send the boolean on the colony for hasmismatched, instead you can do that logic on the client directly(think you dont need to do it for visitors, those are short-lived anyways)

haggbart:
> We should add a nametag cost to the button, as that hooks nicely into how you rename things in vanilla. 1 Nametag per 25 citizens sounds reasonable(minimum 1) That'll need the citizen sync their nametagpack to the client then, then you also dont need to send the boolean on the colony for hasmismatched, instead you can do that logic on the client directly(think you dont need to do it for visitors, those are short-lived anyways)

Interesting idea, I made a quick and dirty draft (#11306). Something like this you had in mind?

Thodor12:
Please don't open entirely new PR's with the same code unless you explicitly make them drafts and state they are a hard dependency on another PR.
Closing in favor of #11306

PR #11231, opened 2025-10-04
Multi-level drafting

uecasm:
- Fix some weirdness with goggle/hammer blueprint rendering
- Blueprint multi-level drafting
    - Composite multiple blueprint levels together by checking the previous-level blueprint wherever there's a tan placeholder in the current level.
    - In particular, this is applied when deconstructing a building back to level 0 (to remove more blocks).
    - It's also applied when rebuilding a deconstructed building back to its normal level, and is especially helpful for buildings that placeholder large areas in higher levels, such as barracks and farms with integrated fields.
    - This does **not** affect the build tool preview before placement, but does affect the resource list preview, builder resource list, and what's actually built.  It will also be visible in the builder goggles and assistant hammer previews.
    - This does **not** affect pasting, neither schematic nor constructed.
- Places waterlogged blocks at the decoration stage, so that they're less likely to spill water everywhere.

## Testing

Review please (maybe port?)


Raycoms (dismissed):

src/main/java/com/minecolonies/api/util/WorldUtil.java
```diff
@@ -1,6 +1,7 @@
 package com.minecolonies.api.util;
 
 import com.minecolonies.api.colony.buildings.IBuilding;
+import net.minecraft.client.Minecraft;
```
someaddons, line 4: pretty sure this will cause issues on serverside

uecasm: It doesn't, I tested it.  At least not in a dev runServer instance.

src/main/java/com/minecolonies/core/colony/workorders/AbstractWorkOrder.java
```diff
@@ -416,23 +416,27 @@ public String getStructurePack()
...
-                },
-                e -> afterLoad.accept(null));
+            future = ColonyUtils.queueBlueprintLoad(world, getStructurePack(), getStructurePath(), getCurrentLevel(), getTargetLevel(),
+                blueprint -> setBlueprint(blueprint, world),
+                e -> {});
+            return future;
         }
         else
         {
-            afterLoad.accept(null);
```
someaddons, line 439: I recall this was special behaviour needed for the miner's fallback for styles which do not have their own mineshaft schematics, make sure you test that fallback, I recall the acacia style does not have mineshaft schematics

uecasm: I haven't specifically tested that case, but if there are no loadable blueprints then the future will complete with null, and `thenAccept(afterLoad)` will call `afterLoad(null)`, so that behaviour should be the same as before in the normal case.

There *is* one potential change in that previously if you called `loadBlueprint` while the future was still pending with different `afterLoad` values then it would call will null on the second and subsequent handlers before eventually calling only the first handler with the actual blueprint.  In the new version anyone calling `loadBlueprintFuture` will get the same pending state, and anyone calling `loadBlueprint` will have their `afterLoad` eventually called with the actual blueprint instead of immediately called with null.  But I consider this a bugfix.

One quirk that remains with both implementations (and I couldn't decide if it's a bug or intended) is that if the load has previously failed then it will still try async loading again instead of remembering the failure.

uecasm: One new quirk of this implementation is that if _some_ of the blueprints are loadable then it will still report a blueprint -- e.g. if it's trying to load levels 1-3 and level 2 is missing, then you'll get a composite of levels 1 & 3.  I think this is acceptable because you'll still detect the missing blueprint in the majority case where you're upgrading one level at a time.

uecasm: Tested with True Dwarven, it works fine as expected.  (Acacia does have custom shafts.)

Raycoms (dismissed):

uecasm:
ping

Raycoms:
Does the LoadOnly handler properly account for this?

uecasm:
> Does the LoadOnly handler properly account for this?

It just sees the merged blueprint, since it's intercepted at blueprint load time.  So there's nothing to account for, it's just a normal blueprint like any other.

Shadizar:
Will waterlogged blocks all still guarantee being placed before coral blocks, fans, etc.? We often relied on placement of waterlogged blocks to keep coral alive

someaddons (approved):

Raycoms:
please port

Shadizar:
I also wonder how this will work for huts that have explicit blocks at lvl 1 (like to clear to air for player spaces) and then placeholder them at higher levels...will they now nuke all player placed stuff in places that are placeholdered after level 1? If so, how can we as schematic  makers make sure to preserve player changes where we want to (via placeholder usage)?

uecasm:
> I also wonder how this will work for huts that have explicit blocks at lvl 1 (like to clear to air for player spaces) and then placeholder them at higher levels...will they now nuke all player placed stuff in places that are placeholdered after level 1? If so, how can we as schematic makers make sure to preserve player changes where we want to (via placeholder usage)?

@Shadizar Yes and no.  There is no change to normal upgrade behaviour, so if someone builds a level 1, puts their own things in places that are placeholdered, then upgrades to level 2, it will behave as before and their changes will be preserved.

What this PR does is that if they then deconstruct the level-3-upgraded building and then move it somewhere else and rebuild directly as level 3, it will now properly clear that placeholdered space (as if they'd rapid-fire-built the levels 1-2-3, except it doesn't place the superseded blocks) and then stop.  At that point the player is free to make their own further changes which will then be preserved by the level 3 to 4 upgrade as the designer intended.

This may still misbehave for some styles that use levels to deliberately pause the build to let things settle (e.g. to help clear water in the underwater styles), but not any worse than it did before, and hopefully better.

Shadizar:
Thanks for the info! Any data on the coral question above?

uecasm:
> Will waterlogged blocks all still guarantee being placed before coral blocks, fans, etc.? We often relied on placement of waterlogged blocks to keep coral alive

No, the waterlogged blocks are now placed at the same phase as water blocks, which is also the same priority as coral blocks etc.  So it will depend on the position / iteration order which is placed first.

If the water is placed in a previous building level and coral added in a later level then it will be fine.  Since coral is a high-level resource this should usually be how it's designed anyway.

Shadizar:
> > Will waterlogged blocks all still guarantee being placed before coral blocks, fans, etc.? We often relied on placement of waterlogged blocks to keep coral alive
> 
> No, the waterlogged blocks are now placed at the same phase as water blocks, which is also the same priority as coral blocks etc. So it will depend on the position / iteration order which is placed first.
> 
> If the water is placed in a previous building level and coral added in a later level then it will be fine. Since coral is a high-level resource this should usually be how it's designed anyway.

The waterlogged blocks are placed only to keep the coral alive,  they wouldn't make sense in prior levels
I suspect other styles may encounter this (maybe) too, especially in water features that get added at lvl 4
Is this part separable by chance?

PR #11252, opened 2025-10-06
Fix certain blocks not dropping inventories [1.21]

Thodor12:
See #11251

### DIFF
--- a/src/main/java/com/minecolonies/core/blocks/BlockMinecoloniesRack.java
+++ b/src/main/java/com/minecolonies/core/blocks/BlockMinecoloniesRack.java
@@ -7,7 +7,6 @@ import com.minecolonies.api.blocks.types.RackType;
 import com.minecolonies.api.colony.IColony;
 import com.minecolonies.api.colony.IColonyManager;
 import com.minecolonies.api.colony.permissions.Action;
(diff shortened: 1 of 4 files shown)


Raycoms (approved):

src/main/java/com/minecolonies/core/blocks/BlockMinecoloniesRack.java
```diff
@@ -227,18 +218,6 @@ public BlockState updateShape(
         return super.updateShape(state, dir, neighbourState, level, pos, neighbourPos);
     }
 
-    @Override
-    public void spawnAfterBreak(final BlockState state, final ServerLevel worldIn, final BlockPos pos, final ItemStack stack, final boolean p_222953_)
```
someaddons, line 231: Is this unused? Your newly added call does not cover the racks, they are not buildings

Thodor12: I'm pretty sure I tested all these methods, and absolutely none of them worked. The reason buildings were dropping items was due to the logic in AbstractBuilding.

The only method that I could get to actually do anything at all was `onRemove`, others would never even get hit by breakpoints.
I'm pretty sure I tried spawnAfterBreak and destroy.

someaddons (approved):

PR #11253, opened 2025-10-06
fix: GregTech and modded tool support for recipes

haggbart:
Fix GregTech and modded tool recipes by properly handling tools that store durability in NBT.

- Set `ignoreNBT=true` for damageable items so tools match regardless of durability changes
- Fix comparison flags to respect `ignoreDamageValue`/`ignoreNBT` 
- Filter tools from secondaryOutputs to prevent duplication

Tools with NBT-stored durability (GregTech saws, etc.) weren't recognized after first use, preventing recipe crafting.

**Needs more testing**
Multiplayer, various modded tools, enchanted tools.
I don't have full understanding of the whole picture and probably missed something / a better solution. Feel free to add commits

## Testing

## Todo
- [ ] More testing
- [ ] Clean up (eg. temp draft dependency)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,1 +1,15 @@
 apply from: 'https://raw.githubusercontent.com/ldtteam/OperaPublicaCreator/main/gradle/mod.gradle'
+
+repositories {
+    maven {
+        name = 'GTCEu Maven'
+        url = 'https://maven.gtceu.com'
+        content {
+            includeGroup 'com.gregtechceu.gtceu'
+        }
+    }
+}
+
+dependencies {
+    runtimeOnly fg.deobf("com.gregtechceu.gtceu:gtceu-1.20.1:7.2.0")
+}
--- a/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -161,7 +161,7 @@ public class RecipeStorage implements IRecipeStorage
             this.input = recipe.getInput();
             this.primaryOutput = recipe.getPrimaryOutput();
             this.alternateOutputs = recipe.getAlternateOutputs();
-            this.secondaryOutputs = recipe.getSecondaryOutputs();
+            this.secondaryOutputs = recipe.getCraftingToolsAndSecondaryOutputs();
             this.intermediate = recipe.getIntermediate();
             this.gridSize = recipe.getGridSize();
             this.token = null;
--- a/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
+++ b/src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
@@ -161,7 +161,7 @@ public class RecipeStorage implements IRecipeStorage
             this.input = recipe.getInput();
             this.primaryOutput = recipe.getPrimaryOutput();
             this.alternateOutputs = recipe.getAlternateOutputs();
-            this.secondaryOutputs = recipe.getCraftingToolsAndSecondaryOutputs();
+            this.secondaryOutputs = recipe.getSecondaryOutputs();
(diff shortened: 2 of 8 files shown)


build.gradle
```diff
@@ -1 +1,15 @@
-apply from: 'https://raw.githubusercontent.com/ldtteam/OperaPublicaCreator/main/gradle/mod.gradle'
\ No newline at end of file
+apply from: 'https://raw.githubusercontent.com/ldtteam/OperaPublicaCreator/main/gradle/mod.gradle'
+
+repositories {
```
Thodor12, line 3: Remove all of this, we're not adding random third party extensions anymore. Create an addon mod instead.

Thodor12 (changes requested):

build.gradle
Thodor12: I.e. if their repository breaks we can't compile our own code anymore, we don't want these hard dependencies

MotionlessTrain: It should also be possible to add a mod in the mods folder within the run folder, if you want to test with that mod installed

src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -161,7 +161,7 @@ public Builder(@NotNull final IRecipeStorage recipe)
             this.input = recipe.getInput();
             this.primaryOutput = recipe.getPrimaryOutput();
             this.alternateOutputs = recipe.getAlternateOutputs();
-            this.secondaryOutputs = recipe.getCraftingToolsAndSecondaryOutputs();
+            this.secondaryOutputs = recipe.getSecondaryOutputs();
```
uecasm: This is incorrect, it needs to fetch the full list because the RecipeStorage itself splits them again later once the builder completes.

build.gradle
haggbart: It's convenience for the draft, it'll be removed from the final PR. :)

src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
haggbart: Fixed, thanks!

haggbart:
~~There's one issue, multiple requests treat tools as consumable; 2x 64 Oak Planks fails to find the saw (when there's only 1 saw available), but 1x 64 works.~~

src/main/java/com/minecolonies/api/crafting/RecipeStorage.java
```diff
@@ -378,17 +378,33 @@ private void processInputsAndTools(@Nullable final List<ItemStack> secOutputs)
...
             }
 
+            boolean isToolDetected = false;
             for (ItemStack result : this.secondaryOutputs)
             {
                 if (ItemStackUtils.compareItemStacksIgnoreStackSize(inputItem.getItemStack(), result, false, true))
                 {
-                    inputItem = new ItemStorage(inputItem.getItemStack(), inputItem.getAmount(), true, inputItem.shouldIgnoreNBTValue);
+                    // For damageable tools (eg. GregTech), also set ignoreNBT=true since durability is stored in NBT
+                    final boolean ignoreNBT = inputItem.shouldIgnoreNBTValue || inputItem.getItemStack().isDamageableItem();
```
someaddons, line 387: doesnt this cause issues with e.g. enchanted pickaxes?

haggbart:
I'm having second thoughts about this approach. I don't think it makes sense for workers to need the actual tool items delivered to them.

Workers should be using the "machine recipes" where the tool is part of the workshop itself, not a consumable/reusable input. A sawmill worker doesn't need a saw item in their inventory, the sawmill just has a saw.

Closing this PR since it doesn't align with how workshops should work. Reopen if you disagree.

PR #11263, opened 2025-10-11
Fix/a bunch

### LINKED ISSUE #11254: Cancelling a queued build resets the builder

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.21.1-1.1.1145

### Structurize Version

1.0.790-1.21.1

_No response_

### Current Behavior

When cancelling a queued build on a specific builder, that builder resets and has to go back to their hut block to continue on their current build.

### Expected Behavior

The builder just continuing the current build without having to go to the hut block after the cancel of a queued build.

### Reproduction Steps

1. Make a colony
2. Have a builder
3. Set the builder to build
4. Set the next build to that specific builder
5. Cancel the queued build (not the one the builder is currently doing)

### Logs

https://gist.github.com/gbdalen00/8561d55fef41ea68086259ad57205ce0

### Anything else?

_No response_

### Footer

---
 ## Viewers

### LINKED ISSUE #11274: AI log upload

### Minecraft Version

1.20

### MineColonies Version

1.1.1065 (per someone reviewing the versions, I'm told that the two more recent versions do not contain AI fixes)

### Structurize Version

1.0.784

Unknown if other dependencies have anything to do with this. Just in case:
BlockUI 1.0.194
DO 1.0.291

### Reproduction Steps

1. Update Minecolonies.
2. Play.
3. See your builder with a red gear.
4. Right-click on them. They will complain their AI is broken. 

### Logs

To Be Delivered

### Anything else?

I have no context. I have not yet babysat a builder until it happens because it seems to be so sporadic. This is the first time I've noticed the problem since the server I play on was updated to 1.1.1065. In the past, the builder _sometimes_ continued to build through the AI problem, so I may have missed important correlations.

### Footer

---
 ## Viewers
  *  ⚠️ Only add a comment if you have new insights or background information not already mentioned. Off-topic or "+1" comments will be deleted to keep the discussion focused.``

Raycoms:
Closes #11254
Closes #11274 

- Fix walking on water lillies
- Remove min stock from restaurant
- Fix couple of builder issues by moving work order storage to building instead of job
- Citizens eat more flexibly outside of the restaurant
- Crafting now considers min count

- Rename:
-> Restaurant to Dining Hall
-> Cookery to Chef's Kitchen
-> Stonesmeltery -> Brick yard
-> Cook to Waiter

## Testing

Review please


src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingStructureBuilder.java
```diff
@@ -69,6 +76,11 @@ public abstract class AbstractBuildingStructureBuilder extends AbstractBuilding
      */
     private Map<Integer, List<BlockPos>> fluidsToRemove = new LinkedHashMap<>();
 
+    /**
+     * The id of the current workOrder.
+     */
+    private int workOrderId;
```
someaddons, line 74: I think we can actually simplify it more by removing this, dont think the building needs to save the ID actually to assign the workorder to itself. Should be able to just have the workorder manager which is loaded after buildings to assign the building the workorder on load? That also keeps all the saving/loading in the manager and simplifies the logic needed in here

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingStructureBuilder.java
```diff
@@ -443,20 +466,152 @@ public BuilderBucket getNextBucket()
...
         this.setProgressPos(null, null);
         this.cancelAllRequestsOfCitizenOrBuilding(null);
     }
+
+    /**
+     * Get the Work Order ID for this Job.
+     *
+     * @return UUID of the Work Order claimed by this Job, or null
+     */
+    public int getWorkOrderId()
```
someaddons: we actually do not need to expose the internal ID used for saving, all the other places can get it from the workorder

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingStructureBuilder.java
```diff
@@ -443,20 +466,152 @@ public BuilderBucket getNextBucket()
...
+    }
+
+    /**
+     * Do final completion when the Job's current work is complete.
+     */
+    public void complete(ICitizenData citizen)
+    {
+        getWorkOrder().onCompleted(colony, citizen);
+
+        final Blueprint blueprint = getWorkOrder().getBlueprint();
```
someaddons: I think we can move this whole logic into the workorders oncomplete? doesnt look like its related to this building at all

src/main/java/com/minecolonies/core/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -1059,6 +1060,34 @@ public String getId()
         {
             return MODULE_SMELTING;
         }
+
+        @Override
+        public IRecipeStorage getFirstFulfillableRecipe(final Predicate<ItemStack> stackPredicate, final int count, final boolean considerReservation)
+        {
+            boolean hasFuel = false;
+            final ImmutableList<ItemStorage> fuelList = building.getModuleMatching(ItemListModule.class, m -> m.getId().equals(FUEL_LIST)).getList();
```
someaddons: wrong module lookup

src/main/java/com/minecolonies/core/colony/buildings/AbstractBuildingStructureBuilder.java
Raycoms: The order and everything can change in the manager. It's best to store here the "active" one.

someaddons: Why is it best to store the same information twice creating a conflict in data and requiring data sanitization to ensure both data sources are in sync?

Raycoms: We're not storing it twice. There is a "claim" which just means, only this builder will build this. And then there is the "active construction" which is only stored here.

someaddons: Wouldnt it be easier to store that in the workorder aswell then? As it already does store claimed which is similar

Raycoms: Ish. We would have to store everything in the workorder including progress and requests and resource lists then, otherwise the builder should at least some kind of reference what they're currently building directly. The workorder list itself can be aborted, reshuffled, etc.

someaddons: I was actually thinking simpler, store & save the builders building pos in the workorder, and inject the actual workorder, not the ID into the Building like you did here, so you just dont have to handle ids/saving within the building itself and just get a full workorder reference which gets set on load from the manager/workorder

Raycoms: Yeah, we already have a claim storage, but the workorder needs to be connected to the progress directly and not only loosely.

someaddons (approved):

PR #11287, opened 2025-10-23
Raid commands update with optional raider amount and position arguments

Thodor12:
- Updated raid command to include new amount and position arguments
- Removed raid-all command in favor of selecting `/mc colony raid all`
- Renamed `@here` and `@mine` to `here` and `mine` respectively, to match the naming of `all`
- Made a `MultiColonyIdArgument` and updated the `ColonyIdArgument` to have identical logic when it comes to choosing colonies. The only addition that `MultiColonyIdArgument` has is the addition of the option to choose `all`
- Created a base class for arguments that allows easier switching over several types of different values, you can provide multiple `AllowedOption` instances which match against the inputted argument, and see if the argument somehow matches the provided input. The argument then finds the first `AllowedOption` that matches the argument value, and stores this one as the argument result.
- Removed extraneous overloads of the `raiderEvent` and `setRaidTonight` by using a container class to pass different settings

## Testing

Review please


someaddons:
Actually I think we can simplify it to one single command, if we had a colony id or "all" as the first parameter

Thodor12:
> Actually I think we can simplify it to one single command, if we had a colony id or "all" as the first parameter

You know, good call, I'm going to change that, can remove some of the extra command overhead like the base class and the chain builder then.

There is only one difference, and that is that raid-all is not in the colony context:

Raid: `/mc colony raid <id>`
Raid-all: `/mc raid-all`

I guess it'd be ok if it sounded like so though:

Raid: `/mc colony raid <id>`
Raid-all: `/mc colony raid all`

someaddons:
ye think that'd be just fine

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,8 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="ClassCanBeRecord" enabled="true" level="WEAK WARNING" enabled_by_default="true">
+      <option name="myConversionStrategy" value="DO_NOT_SUGGEST" />
+    </inspection_tool>
+  </profile>
```
Raycoms, line 7: ?

src/main/java/com/minecolonies/core/commands/arguments/MultipleOptionsArgument.java
```diff
@@ -0,0 +1,90 @@
...
+import net.minecraft.client.multiplayer.ClientSuggestionProvider;
+import net.minecraft.commands.CommandSourceStack;
+import net.minecraft.commands.SharedSuggestionProvider;
+import net.minecraft.commands.arguments.selector.EntitySelectorParser;
+import net.minecraft.world.level.Level;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public abstract class MultipleOptionsArgument<TValue> implements ArgumentType<MultipleOptionsArgument.OptionContainer<TValue>>
```
Raycoms, line 36: this class needs doc

.idea/inspectionProfiles/Project_Default.xml
Thodor12: This is an inspection profile file for Intellij. Intellij has this warning on classes to convert them into records, however at times it asks this, but then comes up with a warning stating it will "make the members more visible", because private fields without a getter imperatively become public in a record.

I've disabled this setting, so it only asks for record conversion if all members are already public, hence why this inspection profile file got made.

Raycoms (dismissed):

src/main/java/com/minecolonies/api/colony/colonyEvents/registry/ColonyEventTypeRegistryEntry.java
```diff
@@ -57,10 +78,21 @@ public IColonyEvent deserializeEvent(@Nonnull final IColony colony, @NotNull fin
...
     {
         return registryName;
     }
+
+    /**
+     * Check whether this is a raid event or not.
+     *
+     * @return true if so.
+     */
+    public boolean isRaidEvent()
```
someaddons, line 94: whatfor is this?

Thodor12: I know for a fact that my colony expeditions are event types too. And we don't want to be able to trigger a raid event with a colony expedition event ID. So I needed something quick to filter this out.

I've made a more elaborate solution in a branch I was working on based on this one, but this serves as a quick fix for this.

someaddons: How would a non raid event id end up as a raid event? This should be a unique 1:1 mapping? huh?

Raycoms: I couldnt' find where this is called in this PR

Thodor12: The system is designed for "colony events", currently this only consists of raids, but is not limited to raids. (As with the colony expeditions).
The raid command however queries the entire colony events registry to check for any possible colony event, including events that wouldn't be raids, that's the intention behind said filter.

Thodor12: > I couldnt' find where this is called in this PR

I might've accidently made it unused after an update, let me check

someaddons: so you want to include possible third party raid events into the raid command, not sure if thats needed at all, but if you do I think a Tag system similarly to vanillas makes more sense. So you'd have a Tag<ColonyEventTypeRegistryEntry> IS_RAID e.g.

Thodor12: > so you want to include possible third party raid events into the raid command, not sure if thats needed at all, but if you do I think a Tag system similarly to vanillas makes more sense. So you'd have a Tag IS_RAID e.g.

No you're missing the point, this is the `ColonyEventType` registry. It's not the `RaidEventType` registry, this is a shared registry that holds more things (under which the colony expeditions).

In the larger rework I've introduced a special subclass of the registry entry specific for raid events that also introduce extra methods that have to be filled, which is a better solution over this boolean flag. But for now, this will have to do.

someaddons: no you missed my point, I know that this is the colony event registry, I made the colony event system xD
I'm saying if you do not need dynamically added raid events to use from the registry, then just define a static list containg all the raid event registry entries to use for the command.

Thodor12: I mean I'm not going to hardcode stuff, that's a surefire way to have stuff forgotten. And I'm not going to make a tag since that's even more extra work which serves no point. This getter is really fine, it's part of the detail information that you'd expect to find in a registry entry.

Raycoms: This kinda invites every category of event creating a new boolean getter. Can't we just do an enum of "event_type" and pre-fill it with a single one RAID, atm, and then check if event.getType() == RAID ?

Thodor12: I mean I did say I already have another solution for the next branch I was working in. Instead of getters, I'm using subclasses of the registry entry. This is because they need to define additional logic regardless of what. So that new code is just using instanceof checks.

someaddons (approved):

Raycoms:
Please backport

uecasm:
> Renamed @here and @mine to here and mine respectively, to match the naming of all

Wait, why this?  This makes them inconsistent with Minecraft's selector syntax (and potentially confusable with player names).  I would regard this as a bug and we should change everything back to include the `@` prefix, including `@all` if that bugs you.

PR #11306, opened 2025-10-30
feat: Add instant name regeneration with nametag cost

### LINKED ISSUE #740: Allow name style selection prior to colony creation
### Prerequisites 

### Context 

- Minecraft Version: 1.19.2
- MineColonies Version: 1.0.1160-ALPHA
- Related Mods and their Versions: N/A

Sometimes I am not quick enough to 'create  colony, turn off kids being born, change naming style, turn kids back on', and am annoyed when my first few citizens have different naming style.

Ability to set name style in townhall prior to 'create colony ' to make it easier to have custom (elven, Chinese, mage, etc) for your first colonists 

### Notes or related things

---
## Viewers

haggbart:
Closes https://github.com/ldtteam/minecolonies-features/issues/740 (sort of)
Closes #11225

Following feedback on the previous PR (#11225), this adds a nametag cost to the name regeneration feature and moves the mismatch detection logic to the client.

The regenerate names button now requires nametags to use, 1 nametag per 25 citizens (minimum 1). Citizens with custom names from nametags are excluded from both the cost calculation and regeneration. The button shows a nametag icon with the cost and displays different tooltips depending on whether you have enough.

Instead of syncing a boolean from the server about mismatched names, each citizen's assigned name pack is now sent to the client. The client checks this directly to determine if the button should appear.  Visitors are excluded.

The server validates the nametag cost before regenerating.

## Screenshots

With enough nametags in inventory:
Screenshot 2025-10-30 at 17 12 14

With lacking nametags:
Screenshot 2025-10-30 at 17 12 55

## Testing

## Todo 
- [ ] Go through code and clean up if needed
- [ ] Translation plural handling / rephrase


src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -2222,6 +2222,9 @@
   "com.minecolonies.coremod.gui.workerhuts.miner.repair.enqueued": "Added the repair request to the queue of the miner. Miner will pick it up once the current task is finished.",
   "com.minecolonies.coremod.gui.townhall.picktexturestyle": "Citizen Style:",
   "com.minecolonies.coremod.gui.townhall.picknamestyle": "Name Pack:",
+  "com.minecolonies.coremod.gui.townhall.regeneratenames": "Rename all",
+  "com.minecolonies.coremod.gui.townhall.regeneratenames.tooltip": "Updates citizen names to match the colony's current name style. Requires nametags (1 per 25 citizens, minimum 1).",
```
Thodor12: I'd probably not type out "nametags" yourself, (it's also typed wrong). Rather type something like "Requires one or more %s" and then use the translation key of the name tag item itself.

src/main/java/com/minecolonies/core/network/messages/server/colony/ColonyNameStyleMessage.java
```diff
@@ -39,6 +41,7 @@ public ColonyNameStyleMessage(final IColony colony, final String style)
     protected void onExecute(NetworkEvent.Context ctxIn, boolean isLogicalServer, IColony colony)
     {
         colony.setNameStyle(style);
+        IMinecoloniesAPI.getInstance().getEventBus().post(new ColonyNameStyleChangedModEvent(colony));
```
Thodor12: Place this inside `setNameStyle` in case this is called externally, not inside the message.

src/main/java/com/minecolonies/core/colony/Colony.java
```diff
@@ -1942,6 +1948,42 @@ public void setNameStyle(final String style)
...
     }
 
+    /**
+     * Check if a civilian has a mismatched name pack.
+     *
+     * @param civilian the civilian to check.
+     * @param currentNamePack the current colony name pack.
+     * @return true if the name should be regenerated.
+     */
+    private boolean hasMismatchedNamePack(final ICivilianData civilian, final String currentNamePack)
```
Thodor12: I'm highly against placing this in the Colony, this is logic that should be present in the CitizenManager. (Goes for all these 3 methods)

src/main/java/com/minecolonies/core/colony/Colony.java
```diff
@@ -736,6 +737,11 @@ public void read(@NotNull final CompoundTag compound)
         mercenaryLastUse = compound.getLong(TAG_MERCENARY_TIME);
         additionalChildTime = compound.getInt(TAG_CHILD_TIME);
 
+        if (compound.contains(TAG_COL_NAME_STYLE))
```
Thodor12: Remove, this is already done on line 851

src/main/java/com/minecolonies/core/colony/CitizenDataView.java
```diff
@@ -672,4 +684,24 @@ public boolean isSick()
     {
         return this.isSick;
     }
+
+    /**
+     * Get the name pack that was used to generate this citizen's name.
+     *
+     * @return The assigned name pack, or empty string if none/special name.
+     */
+    public String getAssignedNamePack()
```
Thodor12, line 693: Does the client really need this information? (both methods)

src/main/java/com/minecolonies/core/client/gui/townhall/WindowMainPage.java
```diff
@@ -258,7 +265,55 @@ public void onUpdate()
...
+                Items.NAME_TAG);
+
+            regenerateNamesButton.show();
+            costIcon.show();
+
+            final ItemStack nametagStack = new ItemStack(Items.NAME_TAG, requiredNametags);
+            costIcon.setItem(nametagStack);
+
+            // Enable button only when player has enough nametags or is in creative mode
+            if (Minecraft.getInstance().player.isCreative() || availableNametags >= requiredNametags)
```
Thodor12: You are needlessly checking the player inventory even if they're in creative. This logic can be improved.

src/main/java/com/minecolonies/core/client/gui/townhall/WindowMainPage.java
```diff
@@ -258,7 +265,55 @@ public void onUpdate()
         final Pane textPane = findPaneByID(DROPDOWN_TEXT_ID);
         final Pane namePane = findPaneByID(DROPDOWN_NAME_ID);
         final Pane resetButton = findPaneByID(BUTTON_RESET_TEXTURE);
+        final ButtonImage regenerateNamesButton = findPaneOfTypeByID(BUTTON_REGENERATE_NAMES, ButtonImage.class);
+        final ItemIcon costIcon = findPaneOfTypeByID("regenerateNamesCostIcon", ItemIcon.class);
         final boolean isOwner = building.getColony().getPermissions().getOwner().equals(Minecraft.getInstance().player.getUUID());
+
+        if (building.getColony().hasCitizensWithMismatchedNamePack())
+        {
+            // Calculate nametag cost: 1 per 25 citizens without special names, minimum 1
```
Thodor12: See if there's a way you can combine this logic through a method in the CitizenManager, because this is duplicate in the network message.

src/main/java/com/minecolonies/api/colony/ICivilianData.java
```diff
@@ -55,6 +55,11 @@ public interface ICivilianData extends ICitizen, INBTSerializable<CompoundTag>
      */
     void setGenderAndGenerateName(boolean isFemale);
 
+    /**
```
Thodor12: This shouldn't be required to implement by ICivilianData, that means also visitors have to inherit this logic, to whom this does not apply at all. (Goes for all the new methods here)

src/main/java/com/minecolonies/api/colony/IColony.java
```diff
@@ -459,7 +459,20 @@ default List<BlockPos> getWayPoints(@NotNull BlockPos position, @NotNull BlockPo
     void setNameStyle(final String style);
 
     /**
-     * Get the matching citizen name file of the colony .
+     * Check if any civilians have names from a different name pack.
```
Thodor12, line 462: As explained in the implementation, this shouldn't be part of the Colony, but the CitizenManager

Thodor12 (changes requested):

src/main/java/com/minecolonies/core/colony/CitizenDataView.java
haggbart: Yes, the client needs this data to show/hide the "Rename All" button (only show when names are mismatched and calculate the "nametag" cost displayed in the UI

This follows the original PR feedback which suggested syncing individual citizen name pack data to the client instead of syncing a redundant colony-level boolean from the server.

src/main/java/com/minecolonies/core/colony/Colony.java
haggbart: I moved the implementations to `CitizenManager` as suggested, `Colony` now just delegates to it. However, I kept the methods in `IColony` because `WindowMainPage` (client-side) calls it via `building.getColony().hasCitizensWithMismatchedNamePack()` and `building.getColony().calculateNameRegenerationCost()`.

Since `ColonyView` doesn't have a real `CitizenManager` structure on the client, callers can't use `getCitizenManager()` directly.

Should I keep the current delegation pattern (IColony -> CitizenManager on server, IColonyView implements inline on client) or restructure it differently? If so, how should client-side code access this? Thanks!

src/main/java/com/minecolonies/core/colony/CitizenDataView.java
Thodor12: I just don't think this needs to be on the view, we don't need this data for "display" purposes. The CitizenManager on the client should get a boolean from the server if name change should be allowed to take place, which is based off of the server only, who holds this kind of internal data.

src/main/java/com/minecolonies/api/colony/ICitizenData.java
```diff
@@ -463,4 +463,38 @@ default boolean hasCustomTexture()
      * @return the leftover leisure time.
      */
     int getLeisureTime();
+
+    /**
+     * Regenerates the name using the current gender and colony name pack.
+     */
+    void regenerateName();
```
Thodor12, line 470: Have you also considered the legacy code we already use, under which `generateName` in this interface?

src/main/java/com/minecolonies/api/colony/IColony.java
```diff
@@ -471,6 +471,13 @@ default List<BlockPos> getWayPoints(@NotNull BlockPos position, @NotNull BlockPo
      */
     void regenerateAllNames();
 
+    /**
+     * Calculate the nametag cost for regenerating citizen names.
+     *
+     * @return The number of nametags required (1 per 25 citizens without special names, minimum 1).
+     */
+    int calculateNameRegenerationCost();
```
Thodor12, line 479: This one needs to move as well

src/main/java/com/minecolonies/core/colony/Colony.java
```diff
@@ -1946,42 +1943,25 @@ public void setNameStyle(final String style)
...
+        IMinecoloniesAPI.getInstance().getEventBus().post(new ColonyNameStyleChangedModEvent(this));
     }
 
     @Override
     public boolean hasCitizensWithMismatchedNamePack()
     {
-        final String currentNamePack = this.getNameStyle();
-        return citizenManager.getCitizens().stream()
-            .anyMatch(civilian -> hasMismatchedNamePack(civilian, currentNamePack));
+        return citizenManager.hasCitizensWithMismatchedNamePack();
```
Thodor12, line 1952: My original comments didn't mean to say just pass through the methods. No. Remove this entirely from the Colony, only the citizen manager should be allowed to know about the existence of these methods. People can fetch the citizen manager externally and shouldn't need passthrough via the colony.

src/main/java/com/minecolonies/core/colony/managers/CitizenManager.java
```diff
@@ -709,4 +709,49 @@ public void afterBuildingLoad()
...
+
+        colony.markDirty();
+    }
+
+    @Override
+    public int calculateNameRegenerationCost()
+    {
+        final long citizensWithoutSpecialNames = citizens.values().stream()
+            .filter(citizen -> !citizen.hasSpecialName())
+            .count();
```
Thodor12, line 754: This would technically mean I'm paying for literally everyone in the colony even though only a single citizen could have the wrong name. Not particularly fair.

Also this is double iteration, we already perform an `anyMatch` inside `hasCitizensWithMismatchedNamePack`, this method is ran first, and then you iterate everyone again to calculate the cost.

This is quite inefficient since this is essentially a 2N loop each time the player opens the townhall UI.
The reason it's 2N because this whole system is a non-ideal flow, this BARELY happens and is a convenience for when it does happen.

I'd highly recommend we go back to the drawing board and find a performant solution that is able to cache this information.
For example, the citizen manager should store whether the names are (probably) out of date. This is quite easy because if a namepack change is done, we know for a fact that everyone currently in the colony is going to be wrong.

So, realistically, the condition we need to check for that regeneration should be allowed:
- Namepack was just changed, and there is more than 0 colonists present in the colony (if no one is around, no names have to be changed either)

You can still keep the originally calculated namepack present in the CitizenData if you want to prevent overwriting their name needlessly, in case for example someone joined the colony after the name change, but before the reset.

Thodor12 (changes requested):

haggbart:
Appreciate the detailed feedback, but this PR is still a draft/concept to explore whether we even want this feature. It's not ready for review yet, just wanted to get the basic idea out there to see if it's something we'd want to include.

Once we decide if we want to move forward with it, I'll mark it ready for review and can implement these improvements then.

Raycoms:
I think it's a nice feature and with the nametag cost it also feels more immersive in the "minecraft realm of possibilities".

Raycoms:
You don't have to close the draft PR if you're still wanting general feedback.

PR #11308, opened 2025-10-30
Fix to update shields with banner when changed

### LINKED ISSUE #11105: Shield / Banner mismatch

- [ ] Single Player
- [x] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.978

### Structurize Version

1.20.1-1.0.775

_No response_

### Current Behavior

The guard shields are not showing the colony banner properly, and changing the colony banner does not change the shields. (yes, I know it's not latest-latest, but I haven't seen this issue reported before.)

Original banner:
Image

Shield:
Image

Shield item info:
Image

Changed colony banner: 
Image

Banner itself showing correctly:
Image

### Expected Behavior

I expect the shield to match the colony banner

### Reproduction Steps

1. Create colony banner with tinyflowers
2. layers below tinyflowers don't show on shield
3. change banner in TH
4. shield doesn't change

### Logs

https://gist.github.com/AnyankaT/98125ad18060bd1271759b6044848a2d

### Anything else?

_No response_

### Footer

---
## Viewers

buffdelta:
Fixes #11105.

- Fixed guard shield banners to switch to the new colony flag whenever a ColonyFlagMessage is executed.
- There are still a few issues with the guard shield banners where if the shield is picked up by or placed into their inventory, the shield banner will not change to the colony flag. Also, if you take the shield out of their inventory the guard will still have it equipped, and the shield will not update to the correct state. I figured these could be fixed in a subsequent PR.

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/jobs/JobKnight.java
+++ b/src/main/java/com/minecolonies/core/colony/jobs/JobKnight.java
@@ -1,5 +1,6 @@
 package com.minecolonies.core.colony.jobs;
 
+import com.minecolonies.api.colony.jobs.IJobWithColonyFlag;
 import com.minecolonies.core.util.citizenutils.CitizenItemUtils;
 import net.minecraft.resources.ResourceLocation;
 import com.minecolonies.api.client.render.modeltype.ModModelTypes;
@@ -28,7 +29,7 @@ import static com.minecolonies.api.util.constant.NbtTagConstants.TAG_BANNER_PATT
  *
  * @author Asherslab
  */
-public class JobKnight extends AbstractJobGuard<JobKnight>
+public class JobKnight extends AbstractJobGuard<JobKnight> implements IJobWithColonyFlag
 {
     /**
      * Desc of knight job.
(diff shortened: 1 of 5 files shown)


src/main/java/com/minecolonies/core/network/messages/server/colony/ColonyFlagChangeMessage.java
```diff
@@ -44,6 +52,18 @@ public ColonyFlagChangeMessage(IColony colony, ListTag patternList)
     protected void onExecute(NetworkEvent.Context ctxIn, boolean isLogicalServer, IColony colony)
     {
         colony.setColonyFlag(patterns);
+        for (ICitizenData citizen : colony.getCitizenManager().getCitizens())
```
Thodor12: In the eye of "separation of concern", this is the wrong place to put this. This should realistically live inside `colony.setFlag` and preferably should be more dynamic over just doing "instanceof JobKnight"

Raycoms: I agree, I would add this in the set flag of the colony.
Additionally, the entity might not be loaded, so you need an isPresent check.

Thodor12: When you realize that you did the exact same thing and now commenting on someone else (who copied you) to fix it 😎

buffdelta:
Is this more like what you are looking for? I modeled it after the [ExternalWorkStations](https://github.com/ldtteam/minecolonies/blob/version/main/src/main/java/com/minecolonies/api/colony/jobs/IJobWithExternalWorkStations.java) interface as I _think_ they do similar things. But, I don't know if this what you had in mind, or if the citizen manager function goes there, or should even exists as a function.

src/main/java/com/minecolonies/core/colony/jobs/JobKnight.java
```diff
@@ -98,4 +99,15 @@ public boolean ignoresDamage(@NotNull final DamageSource damageSource)
         }
         return super.ignoresDamage(damageSource);
     }
+
+    @Override
+    public void onColonyFlagChanged()
+    {
+        final AbstractEntityCitizen worker = this.getCitizen().getEntity().get();
+        CitizenItemUtils.setHeldItem(worker, InteractionHand.OFF_HAND, InventoryUtils.findFirstSlotInItemHandlerWith(this.getCitizen().getInventory(), Items.SHIELD));
```
Raycoms: the worker could still be null here due to chunk loading. Still needs ispresent check

Thodor12 (approved):

someaddons (approved):

someaddons:
nvm

PR #11310, opened 2025-11-01
Fix crusher and fletcher things

### LINKED ISSUE #11300: Stack constructor error in logs after requesting item

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.1079

### Structurize Version

1.20.1-1.0.784

_No response_

### Current Behavior

After requesting flint via a postbox an error appears in the console.

"Stack count mismatch (stack=3 gravel, count=384). ItemStack's count will be ignored. [...] java.lang.Exception: Stack constructor [...]"
 
https://pastebin.com/raw/6qvTwY6G

### Expected Behavior

Flint should be delivered with no error.

### Reproduction Steps

1. Have a fletcher
2. Use a postbox
3. Have no flint in warehouse
4. Request 64 flint

### Logs

https://pastebin.com/raw/6qvTwY6G

### Anything else?

_No response_

### Footer

---
 ## Viewers

uecasm:
Closes #11300

- Makes the crusher produce flint from crushing gravel, not crushing cobble.
- Simplifies AI request logic and fix sync logic not accounting for existing items correctly.
    - Probably not thoroughly tested, but seems sane.
    - In particular this replaces a check for "has any amount of the item -> request none" with "has some amount -> request enough to top off".
- Ensure persisted and otherwise `ItemStorage` uses unit stacks to avoid request warnings.

## Testing

Review please (needs selective porting)

### DIFF
--- a/src/main/java/com/minecolonies/api/crafting/ItemStorage.java
+++ b/src/main/java/com/minecolonies/api/crafting/ItemStorage.java
@@ -47,7 +47,7 @@ public class ItemStorage
      */
     public ItemStorage(@NotNull final ItemStack stack, final int amount, final boolean ignoreDamageValue)
     {
-        this.stack = stack.copyWithCount(1);
+        this.stack = stack;
         this.shouldIgnoreDamageValue = ignoreDamageValue;
         this.shouldIgnoreNBTValue = ignoreDamageValue;
         this.amount = amount;
@@ -63,7 +63,7 @@ public class ItemStorage
      */
     public ItemStorage(@NotNull final ItemStack stack, final int amount, final boolean ignoreDamageValue, final boolean ignoreNBTValue)
     {
-        this.stack = stack.copyWithCount(1);
+        this.stack = stack;
         this.shouldIgnoreDamageValue = ignoreDamageValue;
         this.shouldIgnoreNBTValue = ignoreNBTValue;
         this.amount = amount;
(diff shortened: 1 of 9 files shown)


Raycoms (dismissed):

src/main/java/com/minecolonies/api/crafting/ItemStorage.java
```diff
@@ -46,7 +47,7 @@ public class ItemStorage
      */
     public ItemStorage(@NotNull final ItemStack stack, final int amount, final boolean ignoreDamageValue)
     {
-        this.stack = stack;
+        this.stack = stack.copyWithCount(1);
```
someaddons: we probably should clean up the callsites which are making unnecessary copies then

Raycoms: I had overlooked this, I'm not so sure we should do a copy on new ItemStorage always. We often do a quick wrap to get a hashable object in a loop so this would be terrible performance wise.

uecasm: It does seem safer to do a copy regardless, to ensure it's disconnected from an inventory stack that might be modified later.  Though if we're not worried about that I could make it only do this if the count > 1.

Raycoms: Sometimes we use this just for the hashing, without permanent storage. E.g.
if map contains (new ItemStorage(stack)), count doesn't matter here, as we just are doing a temp contains check. Don't need to copy and set to 1 for this.

uecasm: Hmm, ok, I've removed those.  Haven't tested it since my test colony doesn't have the problematic recipe persisted any more, but I believe it should still work.

someaddons (approved):

Raycoms:
Please make port PR

PR #11311, opened 2025-11-01
Fixes command selector arguments missing leading @

uecasm:
Fixes [inconsistency](https://github.com/ldtteam/minecolonies/pull/11287#issuecomment-3475445391)

- Renames colony selector arguments back to `@here` and `@mine`, and changes `all` to `@all` to match for consistency with Minecraft conventions and existing Discord tricks.

## Testing

Review please (should port)

### DIFF
--- a/src/main/java/com/minecolonies/core/commands/arguments/MultipleOptionsArgument.java
+++ b/src/main/java/com/minecolonies/core/commands/arguments/MultipleOptionsArgument.java
@@ -84,7 +84,16 @@ public abstract class MultipleOptionsArgument<TValue> implements ArgumentType<Mu
     {
         if (reader.canRead())
         {
-            final String argumentValue = reader.readString();
+            final String argumentValue;
+            if (reader.peek() == '@')
+            {
+                reader.skip();
+                argumentValue = "@" + reader.readString();
+            }
+            else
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/core/commands/arguments/MultipleOptionsArgument.java
```diff
@@ -84,7 +84,16 @@ public final OptionContainer<TValue> parse(final StringReader reader) throws Com
     {
         if (reader.canRead())
         {
-            final String argumentValue = reader.readString();
+            final String argumentValue;
```
Thodor12, line 87: This whole peeking behavior shouldn't be necessary as readString will just include it

uecasm: It doesn't, the `@` is not treated as a valid character by `readString()` so it just returns `""`.

`readString()` requires `peek()` to return one of `-.+"_0-9A-Za-z` only.  (And if not a leading quote, the whole string must only consist of the other non-quote characters.)

Raycoms (approved):

someaddons (approved):

PR #11326, opened 2025-11-03
Large block and item registry cleanup

Thodor12:
Technical debt begone! Lots of places touched, not too many lines compared to other major refactors.

This is not going to be backported because it relies on registry features from Neoforge.

No implementation changes, only refactors! I've double checked every if condition that was touched to make sure that there weren't any accidential inversions.

- Updated the block and item registries to use `DeferredRegisters` rather than the `RegistryEvent`.
- Removed the `ModBlockInitializer` and `ModItemsInitializer`, instead favoring the single class that both defines and assigns the registry entries at once.
- All block/item properties are now defined in the registry, so you can get an easier complete overview of all blocks and their properties. (Maybe we can fix the resistance and hardness values of some of the blocks later, so they're easier to break)
- Removed all the abstract class/interface layers from the blocks and items, opting for a more minimal approach by directly implementing HorizontalDirectionalBlock instead of the Minecolonies variant.
- Removed all the abstract classes to the decorational blocks (barrels/graves/racks/decocontrollers/etc), these abstract classes only had a single underlying actual implementation, so I just hoisted the logic over to the actual implementation instead.
- Modified all usages of `ModBlocks` and `ModItems` to use new methods made available by the `DeferredBlock/Item` classes, such as `ModItems.item.toStack()` rather than `new ItemStack(ModItems.item.get())`, as well as `.is()` checks rather than equals checks where possible.

## Testing

Review please


Raycoms:
I think this is waaay too many refactors in a single PR and a lot of these refactors should only happen inbetween version changes when for a short time we only support a single version

PR #11334, opened 2025-11-04
Fix Clipboard item decorator render layer

YoungOnionMC:
Before:
image
After:
image

Since there wasn't a tracking issue, relevant discord message: https://discord.com/channels/139070364159311872/372217492547829762/1435373950422093945

- Moves the pose stack translation back by 300 so that it does not conflict with the tooltip render layer

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/client/render/ClipBoardDecorator.java
+++ b/src/main/java/com/minecolonies/core/client/render/ClipBoardDecorator.java
@@ -83,12 +83,12 @@ public class ClipBoardDecorator implements IItemDecorator
                             {
                                 final PoseStack ps = graphics.pose();
                                 ps.pushPose();
-                                ps.translate(0, 0, 500);
+                                ps.translate(0, 0, 200);
                                 graphics.drawCenteredString(font,
                                   Component.literal(count + ""),
                                   xOffset + 15,
                                   yOffset - 2,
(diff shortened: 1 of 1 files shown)


src/main/java/com/minecolonies/core/client/render/ClipBoardDecorator.java
```diff
@@ -83,12 +83,12 @@ public boolean render(GuiGraphics graphics, Font font, ItemStack stack, int xOff
                             {
                                 final PoseStack ps = graphics.pose();
                                 ps.pushPose();
-                                ps.translate(0, 0, 500);
+                                ps.translate(0, 0, 200);
```
Thodor12, line 86: Are there any constants for this, so we have to rely less on magic numbers?

YoungOnionMC: not really? it just needs to be less than the tooltip internal render layer but greater than the item stack render layer, I tried both 400 and 300 and they were still above the tooltip layer, and even in the forge internal class it uses a 400 magic number for tooltip zed.

Thodor12 (approved):

Raycoms (approved):

someaddons (approved):

PR #11338, opened 2025-11-05
Fix rendering of module tabs (for real this time)

### LINKED ISSUE #11331: Cookery main tab not visible in UI

- [ ] Single Player
- [x] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1-1.1.1093

### Structurize Version

1.20.1-1.0.784

_No response_

### Current Behavior

Main tab is missing for cookery.

Image

### Expected Behavior

Main tab should not be missing for cookery.

### Reproduction Steps

1. Go to cookery
2. Click on the building block

### Logs

not required

### Anything else?

_No response_

### Footer

---
 ## Viewers

Thodor12:
Closes #11331

After quite a bit of tinkering, I've come up with a decent, but still not perfect solution. The major problem we have with the windows is our inheritance structure. The request tree is a bunch of reusable code across 3 different places:
- Clipboard
- Citizen requests tab
- Postbox

This makes it difficult to adhere to a standardized implementation, because each of these places have different parent windows.
I've come up with a way we can mitigate this, window capabilities (name can change if you don't like it).

It's a special interface that can be implemented which holds identical methods to the BOWindow itself (such as onOpened, onUpdate, onClosed, etc).
The AbstractWindowSkeleton can register one or more window capabilities, and the skeleton window will forward all methods to the relevant methods on each of the registered window capabilities.

This basically follows the composition over inheritance way of working, rather than using layers of layers of inheritance, which will eventually end up in problems (like now), we register side objects that can do a part of reusable logic instead.

- Introduce a new window capability system, allowing for reusable code parts across different window types.
  - Turned AbstractWindowRequestTree into a window capability, with some abstract parts for the different requirements of each specific view
  - Turned the tabs into a window capability, so the postbox can overrule the tabs in it's own way, rather than following the default building implementation
- Layout capabilities can automatically also insert a specific layout into the parenting view, so that it can be guaranteed that the capability has the correct pieces of UI present, the only thing you have to define is the position where to place the layout within the parenting window
- Ensure that default tab rendering in building modules can be turned off (i.e. for the townhall)
- Fixed the Postbox, Clipboard and Citizen requests windows to utilize the new request tree window capability

# What does pull request not do

- I've not gone after the entire inheritance structure for all GUI's, but this is something that should be taken into consideration going further.
- I've limited window capabilities to ONLY the request tree and the tabs, obviously this can be applied to more places, but that's a future improvement (like the resource list for the builder/workorders/etc, a similar solution exists for that to make it easier reusable, but can opt into this system instead)
- The citizen sidenav has not been touched to reduce necessary changes, but can be converted to use the tabs window capability as well.

## Testing

Review please


Raycoms (commented):
If the problem is the request tree, why don't make the request tree a module that they get?

someaddons:
isnt capabilities the same thing as we already have with modules just added ontop?

Thodor12:
> isnt capabilities the same thing as we already have with modules just added ontop?

Modules are for the buildings, not for windows. And this problem goes beyond just buildings, as this is used on the clipboard, postbox and citizen GUI's. So this isn't something that can be solved with a building module.

Thodor12:
> If the problem is the request tree, why don't make the request tree a module that they get?

And how am I going to add this to the clipboard or the citizen GUI?

src/main/java/com/minecolonies/core/client/gui/AbstractBuildingWindow.java
```diff
@@ -75,66 +51,47 @@ public AbstractBuildingWindow(final BOWindow parent, final B buildingView, final
     {
         super(parent, resource);
         this.buildingView = buildingView;
-        this.iconRandom = new Random(buildingView.getID().hashCode());
 
-        int nextTabIndex = 0;
+        final TabsWindowCapability tabsWindowCapability = registerModule(TabsWindowCapability::new, new Random(buildingView.getID().hashCode()));
```
Raycoms: Still called capability.
Also I don't think we should need to register this on each constructor call.
Also, we don't need a new random for each

src/main/java/com/minecolonies/core/client/gui/AbstractWindowSkeleton.java
```diff
@@ -108,6 +119,81 @@ public final void registerButton(final String id, final Consumer<Button> action)
...
+        this.capabilities.add(capability);
+        return capability;
+    }
+
+    /**
+     * Add a module to this window. Extending the original logic of the window.
+     *
+     * @param capabilityBuilder the new capability.
+     */
+    public final <T extends IWindowWithLayoutModule> T registerLayoutModule(final Function<AbstractWindowSkeleton, T> capabilityBuilder, int xPos, int yPos)
```
Raycoms: Still called capability, why do we need so many registration methods?

src/main/java/com/minecolonies/core/client/gui/modules/IWindowWithLayoutModule.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.ldtteam.blockui.Pane;
+import net.minecraft.resources.ResourceLocation;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Extensible logic for windows, without having to use new base classes.
+ * Additionally, this class requires you to implement a layout, which will be automatically constructed and injected into the window.
+ */
+public interface IWindowWithLayoutModule extends IWindowModule
```
Raycoms, line 11: what kind of layout?

src/main/java/com/minecolonies/core/client/gui/citizen/MainWindowCitizen.java
```diff
@@ -103,14 +103,14 @@ public void onButtonClicked(@NotNull final Button button)
             final String label = button.getID().replace(PLUS_PREFIX, "");
             final Skill skill = Skill.valueOf(StringUtils.capitalize(label));
 
-            Network.getNetwork().sendToServer(new AdjustSkillCitizenMessage(colony, citizen, 1, skill));
+            Network.getNetwork().sendToServer(new AdjustSkillCitizenMessage(citizen.getColony(), citizen, 1, skill));
```
Raycoms: instead of all of these changes and requiring the citizen view to have the colony view method, you could've just had the parent WindowCitizen building have the cached colony instead

src/main/java/com/minecolonies/core/client/gui/modules/IWindowWithLayoutModule.java
Thodor12: The layout modules automatically insert the required XML part into the parent view, so it requires less duplication within the XML as well.
So every module with a request tree, automatically inserts the required XML bit for setting up the list, then the module can execute the code, knowing that the correct XML is present.

The only thing you have to do is that during registration you must give an X and Y coordinate where you want to position the layout at.

src/main/java/com/minecolonies/core/client/gui/WindowPostBoxMain.java
```diff
@@ -183,12 +190,46 @@ private void deliverPartialClicked(@NotNull final Button button)
         }
     }
 
+    /**
+     * Reusable method for the postbox to render the tabs.
+     */
+    public static void registerPostboxTabs(final AbstractWindowSkeleton window, final IBuildingView buildingView)
+    {
+        final TabsWindowModule tabsWindowCapability = window.registerModule(TabsWindowModule::new, new Random(buildingView.getID().hashCode()));
```
Raycoms: They are still being called capability in almost every class

Raycoms: Seems a bit weird to register them with each call I think. It's either "loadModule" done every time or "registerModule" done once imo.

Raycoms: @Thodor12

someaddons (dismissed):

PR #11339, opened 2025-11-05
Fix building names in research unlocks

### LINKED ISSUE #11328: Brickyard is still refered to as stone smeltery in research

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

minecolonies-1.1.1175-1.21.1-snapshot

### Structurize Version

structurize-1.0.790-1.21.1-snapshot

_No response_

### Current Behavior

Noticed the new names for the huts haven't been changed in the research tree yet.

Image

### Expected Behavior

Updated name

### Reproduction Steps

Look at stone smeltery research

### Logs

none

### Anything else?

_No response_

### Footer

---
 ## Viewers

Thodor12:
Closes #11328

- The building names in the research unlocks no longer use separated translation keys, instead opting to grab them directly through the ResearchEffect instance.
- To achieve this I've started utilizing the ModResearchEffects appropriately, by introducing new registry entries, and dynamically constructing an `IResearchEffect` based on the registry type.
  - I've consolidated the `ModResearchEffect` and `ModResearchEffectInitializer`, simliar to other registries.
- Repurposed the `"effect": true` field to define the effect type, with fallback code that if the value is still true, it will refer the "global" effect type (for datapacks)
- Building effect types simply refer to the building ID, which will then be used in the `getName()` method.
- Modified the `IResearchEffect` to return `Components` in favor of `TranslatableContents`, so each effect can independently choose how they go about constructing the name.
- Modified the `CITIZENCAP` research effect to get it's own `IResearchEffect` instance, to cope with the special condition of being based on the citizen limit configuration. This way we can remove the technical debt in the window code with that explicit condition over there, this way it's a little more dynamic.
- Datagenned the research files anew, to update them to the new JSON structure

## Testing

Review please


Raycoms (commented):
This breaks all research datapacks right?

Thodor12:
> This breaks all research datapacks right?

No, there's fallback logic, if `"effect": true` is still detected it always generates a GlobalResearchEffect

src/main/java/com/minecolonies/api/research/AbstractResearchProvider.java
```diff
@@ -615,22 +615,35 @@ public Research addEffect(final ResourceLocation effect, final int level)
...
+         * Add an unlock building effect to the research.  Research Effects are applied on completion,
+         * and remain unless the colony is destroyed or the research is undone.
+         * See ModBuildings for a list of supported buildings.  Whenever possible, use the public static String BUILDINGNAME_ID constants from ModBuildings
+         * Buildings with no applicable research effects loaded default to unlocked.
+         * Multiple Effects are supported. Defaults to unlocking the highest possible building level.
+         *
+         * @param buildingBlock the building block to lock behind this research.
+         * @return this
+         */
+        public Research addEffect(final AbstractColonyBlock<?> buildingBlock) {
```
Raycoms: formatting

Raycoms (commented):
I feel like a new effect type for this is overkill. Can't we just in the data define an optional "translationkey"? That we then load for "Unlocks %s" ? If not present, we go for the default key atm.

Thodor12:
> I feel like a new effect type for this is overkill. Can't we just in the data define an optional "translationkey"? That we then load for "Unlocks %s" ? If not present, we go for the default key atm.

The problem is with GlobalResearchEffect there's no possible way to add other translation fields into the TranslatableContents, it only ever takes the original 4 arguments. The entire logic behind the class is different.

At least this thing was already a registry and made it so it could support multiple implementations, we just never made use of it until now.

armele:
Will this still support third-party buildings from other mods?

Thodor12:
> Will this still support third-party buildings from other mods?

Yes, the translation string can't be modified, so it'll always say "Unlocks %s", but it'll always cope with whatever the current building name is.

PR #11340, opened 2025-11-06
Add more features to server managers to configure visibility of admins

Thodor12:
- Entering/leaving messages can be disabled through a mod event. By default this happens in spectator and addons can define custom checks for this as well.
- Colony permission checks can be bypassed by admins, by always allowing any action. The minimum permission level to allow this can be configured through the server configuration, but defaults to permission level 2 (gamemaster).
- The above also inhibits any permission events from being made in the townhall, because those only get made if a person doesn't have permission to do an actions (which admins in this case override)

(Will port first and test on 1.21 because it's easier to do runServer with Neo)

## Testing

Review please


someaddons:
Why do we need mod events?

Thodor12:
> Why do we need mod events?

Reason is because I wanted to tie this into VanishMod

someaddons:
wouldnt it be easier to just never report for spectator mode?

Thodor12:
> wouldnt it be easier to just never report for spectator mode?

The event handles this condition by default but allows external additional conditions (such as vanish)

someaddons (dismissed):

src/main/java/com/minecolonies/core/colony/permissions/Permissions.java
```diff
@@ -669,6 +671,12 @@ public Set<ColonyPlayer> getFilteredPlayers(@NotNull final Predicate<Rank> predi
     @Override
     public boolean hasPermission(@NotNull final Player player, @NotNull final Action action)
     {
+        if (player.hasPermissions(IMinecoloniesAPI.getInstance().getConfig().getServer().permissionEventMinBypassPermLevel.get()))
```
Raycoms: I'd say this should be: if (hasPermission(getRank(player), action)
{
true
}
else if player.hasPerm
{
log
true
}
false
}

someaddons (approved):

PR #11361, opened 2025-11-11
Crafting-based milk bottles

armele:
Closes None

-  Cowherder milking is somewhat unreliable, as they milk occasionally, don't request their inputs through the crafting system and don't deliver their product when complete.  As a result, dairy can be a bottleneck to the food delivery.  With cooldown on milk bottles now enforced, players are looking for a more reliable (and automated) way to address that food bottleneck.
- This PR introduced the "Dairyworker" - a crafter that works out of the Cowboy building and crafts milk bottles through the request system.  This still requires a) empty milk bottles and b) cows.
- The crafting of dairy products (only those which require **only** milk or soy milk) has also been moved to the Dairyworker, taking a bit of load off of the Chef, who struggles to keep up with large level 5 colonies.
- The dairyworker will give an Interaction icon when there are no adult cows available to milk (and will not craft).
- This PR does not remove the existing milking capability from the Cowboy.  (I wanted to gauge the reaction to this step before investing time in that one.)

## Testing

Review please


armele:
Closed in favor of rebalancing herder milk production, per discord comments.

PR #11368, opened 2025-11-13
Update schematics to minecraft 1.21.1

someaddons:
- Update schematics to minecraft 1.21.1

## Testing

Review please


MotionlessTrain:
I assume that from now on, I need to do the same when making schematic PRs to 1.21?

someaddons:
ye would be good, that avoids datafixers requiring to run and those are very slow.
The new command is /structurize updateSchematicPack <pack>
So gotta put the files into a pack inside the blueprints folder and then run the command

MotionlessTrain:
I’ll see whether I will spin up dev mode for this, or whether I can find a way to run the command from the command line 😄

PR #11371, opened 2025-11-15
Refactor sort message to be building-agnostic

armele:
Closes None

- Small refactoring of SortWarehouseMessage to SortBuildingMessage so the same path could be used to create sorting modules of other buildings, if desired.

## Testing

Review please


Thodor12:
Whilst I'm not opposed to the idea, do you intend to use this for something? Addons can theoretically register their own message that does the same

src/main/java/com/minecolonies/core/network/messages/server/colony/building/warehouse/SortBuildingMessage.java
```diff
@@ -0,0 +1,70 @@
...
+     * 
+     * @param ctxIn The context of the payload.
+     * @param player The player that sent the message.
+     * @param colony The colony that the building is in.
+     * @param building The building that needs to be sorted.
+     */
+    @Override
+    protected void onExecute(final IPayloadContext ctxIn, final ServerPlayer player, final IColony colony, final IBuilding building)
+    {
+        if (building.getBuildingLevel() >= requiredSortLevel)
```
Thodor12: Maybe if we're touching this anyway, could we maybe have an interface on `IBuilding` for `IBuildingSortableInventory` that is applied to the warehouse. That interface then requires you to define the level at which you are allowed to sort, so we don't have to pass this value along to the server.

(Technically someone could submit the message to the server from a modified client passing a different minimum sort level, which would void this restriction)

armele: I was looking at putting sorting into my Recycling building, and wanted to limit the amount of code that got needlessly reproduced.  I'm happy to implement as an IBuildingSortableInventory approach or simply implement my own message.  (The latter is trivial, of course - I just try to avoid duplication of code where a small tweak to core Minecolonies could avoid it.)

Thodor12: The change to make it available for every building is ok, if you could just make that interface with 2 methods actually. 1 to check the required level to unlock sorting, the other is to actually do the sorting.
So that you could also choose for a different sorting strategy as well, (method would default to SortingUtils)

armele: Updated with the ISortableBuilding approach, and placed that interface on the Warehouse.  Retested warehouse sorting.

src/main/java/com/minecolonies/core/network/messages/server/colony/building/warehouse/SortBuildingMessage.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param ctxIn The context of the payload.
+     * @param player The player that sent the message.
+     * @param colony The colony that the building is in.
+     * @param building The building that needs to be sorted.
+     */
+    @Override
+    protected void onExecute(final IPayloadContext ctxIn, final ServerPlayer player, final IColony colony, final IBuilding building)
+    {
+        @SuppressWarnings("unchecked")
+        T sortableBuilding = (T) building;
```
Thodor12: No generics, just use instanceof

src/main/java/com/minecolonies/api/colony/buildings/ISortableBuilding.java
```diff
@@ -0,0 +1,35 @@
+package com.minecolonies.api.colony.buildings;
+
+import javax.annotation.Nonnull;
+
+import com.minecolonies.api.inventory.api.CombinedItemHandler;
+import com.minecolonies.core.util.SortingUtils;
+
+import net.minecraft.core.HolderLookup;
+
+public interface ISortableBuilding
```
Thodor12: It's better if you make ISortableBuilding extend from IBuilding, that way you can pass the argument of ISortableBuilding into the message, and don't need the weird generic stuff

armele: Agreed - done.

Thodor12 (dismissed):

src/main/java/com/minecolonies/api/colony/buildings/ISortableBuilding.java
```diff
@@ -0,0 +1,35 @@
...
+        return DEFAULT_REQUIRED_SORT_LEVEL;
+    }
+
+    /**
+     * Sort the inventory of this building using the given provider.
+     * The implementation of this method is usually a call to {@link SortingUtils#sort(HolderLookup.Provider, CombinedItemHandler)}.
+     * @param provider the provider to use for sorting.
+     * @param inventoryHandler the inventory handler to sort.
+     */
+    default public void sort(@Nonnull final HolderLookup.Provider provider, final CombinedItemHandler inventoryHandler)
```
Raycoms: I'd be fine to have this even as part of IBuilding, what do you think @someaddons?

someaddons: ye should be fine

armele: Done! Also in light of discord discussion, changed level-specific method to a "canSort" boolean.

src/main/java/com/minecolonies/api/colony/buildings/IBuilding.java
```diff
@@ -533,4 +539,28 @@ default void writeToItemStack(final ItemStack stack)
      */
     @NotNull
     List<BlockPos> getLocationsFromTag(@NotNull final String tagName);
+
+
+
+    /**
+     * Checks if the building can be sorted.
+     * Be default, this will return true if the building level is greater than or equal to {@link #DEFAULT_REQUIRED_SORT_LEVEL},
```
Thodor12: Oh yeah? But, but, but, it clearly says `return false`

armele: Doc'd, then refactored.  🤦  Fixed.

Thodor12 (approved):

someaddons (approved):

Raycoms:
Please port

PR #11386, opened 2025-11-21
Small field rework

Raycoms:
Closes #
Closes #

- Adjust importance of some of the notify messages
- Allow fields to persist the size in the tileEntity (for style designers) (Also allow setting the field size outside of a colony)
- Allow fields to be more dynamic in size (e.g. not only a radius of 5, but they could also be long and thin).
- Adjust some of the farmer logic a little bit

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/blocks/BlockScarecrow.java
+++ b/src/main/java/com/minecolonies/core/blocks/BlockScarecrow.java
@@ -178,7 +178,7 @@ public class BlockScarecrow extends AbstractBlockMinecoloniesDefault<BlockScarec
         final IColony colony = IColonyManager.getInstance().getColonyByPosFromWorld(worldIn, pos);
         if (colony != null)
         {
-            colony.getBuildingManager().addBuildingExtension(FarmField.create(pos));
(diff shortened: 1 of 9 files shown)


Thodor12:
Could this be done via NBT tags perhaps (or components I guess). Otherwise I'm going to have to rework it again in my code because I cannot support fixed stuff locked in the TE, that's not dynamic

Raycoms:
> Could this be done via NBT tags perhaps (or components I guess). Otherwise I'm going to have to rework it again in my code because I cannot support fixed stuff locked in the TE, that's not dynamic

We need it in the tileEntity otherwise it won't be saved in the schem properly.

Thodor12:
But nbt/components are saved too, are they not? I specifically took everything out of the TE upon the first rework to the field system.

I will have to remove this again on my own update later if this is kept.

MotionlessTrain:
Components are on items, are they not?
The schematics only have blocks in them, not the field item (unless someone put it in an inventory, which they shouldn’t)

someaddons (approved):

PR #11389, opened 2025-11-22
finish crafter cleanup

Raycoms:
Closes #
Closes #

- Crafter rework, go through the same main crafter loop to simplify secondary crafting types.
- Adjust Smelter crafter and concrete mixer to this.
- Don't place construction tape for pasted buildings

## Testing

Review please


src/main/java/com/minecolonies/core/colony/buildings/modules/AbstractCraftingBuildingModule.java
```diff
@@ -1074,21 +1080,6 @@ public String getId()
...
-                    hasFuel = true;
-                    break;
-                }
-            }
-
-            if (!hasFuel)
-            {
-                return null;
-            }
             return super.getFirstFulfillableRecipe(stackPredicate, count, considerReservation);
```
someaddons: how about we remove this override when all it does is calling super? :D

someaddons (approved):

PR #11392, opened 2025-11-24
Rework environmental damage adaption of raiders

someaddons:
Closes #
Closes #

- Raiders no longer are fully environmental damage immune, instead we use a hybrid approach allowing the player to weaken them with traps but not kill them. Depending on difficulty the raiders will survive with 20-60% health left.
- Invulnerability frame for environmental damage is now 1.5sec and no longer scales

## Testing

Review please


Raycoms:
Please port

PR #11396, opened 2025-11-25
Introduce proper solution for operator bypass permissions

Thodor12:
- Allows each action to define whether operator bypassed is allowed for the given action
- Disables allowance on guard_attack

## Testing

Review please


someaddons (changes requested):
The op rank is much cleaner/more elegant of a solution, it is not only guard attack which has issues with allowing all

Thodor12:
I disagree, that's very dirty code and this can achieve the same, whilst being directly configurable on the action itself.

Thodor12:
Ray asked me for a fix, I had no time yesterday, so Ray made a temporary solution. This is my solution.

someaddons:
we have permissions for receiving messages, op will get all messages of all colonies. We have permissions for keeping a colony loaded, OP keeps all colonies loaded etc.....

Having a dedicated rank allows us to nicely select the permissions needed, it is by no means "dirty code"

Thodor12:
And now you define it in the action itself which can be operator bypassed, also a nice list. I checked the difference in permissions:

```
    GUARDS_ATTACK(1),
    EDIT_PERMISSIONS(4),
    RECEIVE_MESSAGES(6),
    EXPLODE(22),
    RECEIVE_MESSAGES_FAR_AWAY(23),
    CAN_KEEP_COLONY_ACTIVE_WHILE_AWAY(24),
    HURT_CITIZEN(26),
    HURT_VISITOR(27),
    MAP_DEATHS(29),
```

Thodor12:
The message things, sure.

But the other ones make no sense to have them disabled. I can hurt citizens if I want, I can hurt visitors if I want. I can edit the permissions if I want, don't even know what map deaths is for

Thodor12:
Also sidenote, what the hell do we even have an EXPLODE permission for?

MotionlessTrain:
Afaik whether you are allowed to set off tnt using flint and steel or a flame arrow

uecasm:
Aren't both of these going to make permissions entirely useless in an SP creative test world?

Thodor12:
> Aren't both of these going to make permissions entirely useless in an SP creative test world?

What permissions do you have to worry about in an SP world to begin with, where you're the only player acting as a colony owner?

> Afaik whether you are allowed to set off tnt using flint and steel or a flame arrow

That is not bound to the player, the explosion action determines whether an explosion event can be set off. The entity in question is not a player, it's the tnt block entity causing the explosion.

Thodor12:
Also this solution will stay as I pointed out to Ray in DM. If you were to add new permissions you must immediately think when making the action whether it should be OP or not.

If you have a separate rank list, in another file, you will not be prompted to think about whether this is OP overridable, so it's going to end up being forgotten, and then you end up with more permissions that are suddenly not OP overridable.

Better to think of it the first time around when you implement it, rather than having to come back later to a bug report.

uecasm:
> What permissions do you have to worry about in an SP world to begin with, where you're the only player acting as a colony owner?

Previously you were able to abandon a colony and demote yourself from officer or downgrade officer perms, and they applied, so could be tested.

> you will not be prompted to think about whether this is OP overridable

You're not exactly prompted with an optional arg 😛

Raycoms:
I thought long and hard about it and I think the current version is the best option. The operator bypass is more of a niche thing compared to the full system and especially now with the cleanup we got rid of all the constructor arguments, so I don't want to introduce a new one for this niche application. Esp considering the current one works well

PR #11401, opened 2025-11-26
flatten saturation and only decrease during day

Raycoms:
Closes #
Closes #

- Don't decrease at night
- Flatten curve a bit

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/citizen/EntityCitizen.java
+++ b/src/main/java/com/minecolonies/core/entity/citizen/EntityCitizen.java
@@ -1908,7 +1908,7 @@ public class EntityCitizen extends AbstractEntityCitizen implements IThreatTable
      */
     private boolean decreaseIdleSaturation()
     {
-        if (citizenData != null && level() != null && !level().isNight())
+        if (citizenData != null && level() != null && !level().isNight() && !citizenSleepHandler.isAsleep())
         {
             final int buildingLevel = citizenData.getHomeBuilding() == null ? 0 :  citizenData.getHomeBuilding().getBuildingLevelEquivalent();
             double decrease = switch (buildingLevel)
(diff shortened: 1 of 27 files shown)


someaddons (dismissed):

src/main/java/com/minecolonies/core/entity/citizen/EntityCitizen.java
```diff
@@ -1908,12 +1908,12 @@ public int getTeamId()
      */
     private boolean decreaseIdleSaturation()
     {
-        if (citizenData != null)
+        if (citizenData != null && level() != null && !level().isNight())
```
Thodor12: Maybe better to switch this to check if the citizen is sleeping? (To cope with the research that extends working day)

MotionlessTrain: What if you don’t have a tavern or residences yet (at the start of the game). Not sure whether citizens would then be sleeping

Raycoms: Intentionally not making it citizen sleeping because we don't want to "over incentivize players sleeping the nights away".

Thodor12: Should you at least respect the working hours that the research respects then, rather just asking Minecraft if it's night?

Raycoms: If I'm not mistaken, citizens tend to finish work early to have enough time to get home before it gets night. I think they don't extend work into the night, but just finish a bit later.

Also, the idea is that the time we ignore is the "unproductive time the player can skip by sleeping". This is the time that should matter for this.

someaddons (approved):

PR #11402, opened 2025-11-26
Cavalry 2 of 4 (CavalryHorseEntity)

armele:
Closes None

# Context
PR 1 of 4: Tags & associated utility methods. (Complete)
**PR 2 of 4: Cavalry Horse Entity (This PR)**
PR 3 of 4: Stable, Stablemaster and Horse Goals
PR 4 of 4: Cavalry Job

Second: Cavalry Horse Entity 
- These are required for both the stable and cavalry job
- It will have no impact on gameplay (though someone could /summon a horse entity, if desired)

## Testing

Review please


src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
```diff
@@ -0,0 +1,876 @@
...
+import net.minecraft.world.entity.ai.goal.BreedGoal;
+import net.minecraft.world.entity.ai.goal.FloatGoal;
+import net.minecraft.world.entity.ai.goal.FollowParentGoal;
+import net.minecraft.world.entity.ai.goal.LookAtPlayerGoal;
+import net.minecraft.world.entity.ai.goal.RandomLookAroundGoal;
+import net.minecraft.world.entity.ai.goal.RandomStandGoal;
+import net.minecraft.world.entity.ai.goal.WaterAvoidingRandomStrollGoal;
+import net.minecraft.world.entity.ai.navigation.PathNavigation;
+
+public class CavalryHorseEntity extends Horse
```
someaddons: missing javadoc on the class and its fields

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
```diff
@@ -0,0 +1,876 @@
...
+
+    /**
+     * A slimmed version of the tick function that only runs on the server side.
+     * This function is responsible for updating the horse's slim state based on 
+     * whether it is being ridden by a guard or trainer. If the horse is being ridden, 
+     * it will be slimmed. Otherwise, it will not be slimmed unless it is unable to 
+     * widen safely, in which case it will be given a certain number of ticks to try 
+     * and move to a safe location.
+     */
+    protected void slimTick()
```
someaddons: think I'd just not do this for performance reasons, we can just keep the slim hit/collision box and make it look bigger visually

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
```diff
@@ -0,0 +1,876 @@
...
+     * A portion of that damage is also applied to the combat cooldown counter.
+     *
+     * @param damageSource the source of the damage
+     * @param damageAmount the amount of damage to apply
+     * @return true if the damage was applied, false otherwise
+     */
+    @Override
+    public boolean hurt(@Nonnull DamageSource damageSource, float damageAmount)
+    {
+        // TODO: Create research that provides combat cooldown mitigation
```
someaddons: we also need to decide whether horse and rider are one unit togher and die together, or if the horse can die and the guard dismounts(horse does take damage first then?)

src/main/java/com/minecolonies/core/client/render/CavalryOverlayLayer.java
```diff
@@ -0,0 +1,60 @@
...
+import net.minecraft.client.renderer.entity.RenderLayerParent;
+import net.minecraft.client.model.HorseModel;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.animal.horse.Horse;
+
+import javax.annotation.Nonnull;
+import com.minecolonies.api.util.constant.Constants;
+import com.minecolonies.core.entity.other.cavalry.CavalryHorseEntity;
+
+public class CavalryOverlayLayer extends RenderLayer<Horse, HorseModel<Horse>>
```
Raycoms, line 17: javadoc on class and constructor

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
```diff
@@ -0,0 +1,876 @@
...
+    private static final String NBT_COMBAT_COOLDOWN = "combatCooldown";
+
+    private static final int RESERVATION_EXPIRATION_LIMIT = 200;
+    private int reservationExpiration = 0;
+
+    /*
+     * A measure of how unprepared for combat a mount is, representing damaged equipment, morale,
+     * fatigue or other non-HP factors.
+     */
+    private static final EntityDataAccessor<Float> DATA_COMBAT_CD =
```
Raycoms: If possible, I'd avoid these. We should probably have a representation of the horse in the colony somewhere with a server/client side version that we sync ourselves manually.

We still have those in the citizens but I want to get rid of them for a while as they're quite slow to access and result in inconsistencies sometimes.

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
```diff
@@ -0,0 +1,876 @@
...
+    {
+        super.onSyncedDataUpdated(key);
+
+        if (key == DATA_IS_SLIM) 
+        {
+            this.refreshDimensions();
+        }
+    }
+
+    public void setStable(BlockPos pos, ResourceKey<Level> dim)
```
Raycoms: Yeah, all this kind of data like stable pos should just be in the colony representation of the horse.

Raycoms: This way, we can similar to citizens, respawn/recall horses for example. We always know where they are even if the chunk is unloaded

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
Raycoms: Yeah, it's probably fine to always have the small collision box, but have a larger hit box.

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
```diff
@@ -0,0 +1,876 @@
...
+        super.tick();
+        // logActiveGoals();
+
+        if (!level().isClientSide)
+        {
+            slimTick();
+
+            if (hasReservation())
+            {
+                if (reservationExpiration > RESERVATION_EXPIRATION_LIMIT)
```
Raycoms, line 485: these reservations I'd also have on the colony level. And we can "slow tick" them down there too just fine.

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
Raycoms: Good point. Maybe we "sum" their health together and I think it's fine for them to die together then too.

armele:
Thanks for all the feedback. When you talk about having things "on the colony level", and syncing them ourselves - could you point me to a similar pattern that already exists?  I'll follow that for a model.

Raycoms:
> Thanks for all the feedback. When you talk about having things "on the colony level", and syncing them ourselves - could you point me to a similar pattern that already exists? I'll follow that for a model.

CitizenData/CitizenDataView

src/main/java/com/minecolonies/core/entity/other/cavalry/CavalryHorseEntity.java
armele: For this scenario when the rider dismounts (to go and eat, for example, or go to sleep) I intend to implement a proportional split of the "combined entity" health based on each separate entities max health.  Let me know if you recommend a different approach.

armele:
Just to be sure I'm not going down the wrong path here, moving the synced data elements, stable position, reservations, etc off the entity to the colony level will result in something like this:

IAnimalManager -> Model on IEntityManager
AnimalManager implements IAnimalManager

IAnimalData -> Model on ICitizenData
AnimalData implements IAnimalData

IAnimalDataView -> Model on ICitizenDataView
AnimalDataView implements IAnimalDataView

ColonyViewAnimalViewDataMessage -> Model on ColonyViewCitizenViewMessage

-- Integrate manager into Colony
-- Integrate manager into ColonyPackageManager
-- Integrate manager into ColonyView
-- Update cavalry logic to use AnimalManager

**Edit: I want to confirm that I am going down this path, by the way.  It may be a few days before the refactoring and retesting is complete.**

armele:
Significantly refactored things to introduce the AnimalData/AnimalDataView pattern suggested. Code changes with this PR are now more intrusive, but will support a much nicer division of labor.  Introduced the IManagedAnimal interface, allowing animals other than cavalry horses to be supported with this structure.

someaddons (approved):

Raycoms:
Please port

PR #11410, opened 2025-11-29
Fix postbox minimum stock not working [1.20]

Thodor12:
See #11409


armele:
Worth adding a log statement for the null condition, in case the module check gets flipped to the other notation again?

Thodor12:
> Worth adding a log statement for the null condition, in case the module check gets flipped to the other notation again?

Nah, it's not that desperate enough that it needs logging. I'm more so confused why this got changed to begin with.

PR #11411, opened 2025-11-30
Simplify permissions

Raycoms:
Closes #
Closes #

- Remove unused isSubscriber option
- Move hostile options like attack guards to hostile type instead of being a permission that confuses player
- Move loading colony & messages to colony manager instead of separate perms
- Remove free block perms and make it !hostile

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/colony/permissions/Action.java
+++ b/src/main/java/com/minecolonies/api/colony/permissions/Action.java
@@ -8,7 +8,7 @@ public enum Action
     //counts for citizen and huts.
     ACCESS_HUTS(0),
     //If guards can attack, player can attack back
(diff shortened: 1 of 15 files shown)


someaddons (approved):

uecasm:
A useful thing with the previous implementation is that you could give someone perms equivalent to officer without making them a colony manager, which still limits the kinds of things they can do (most notably, commands).

Raycoms:
> A useful thing with the previous implementation is that you could give someone perms equivalent to officer without making them a colony manager, which still limits the kinds of things they can do (most notably, commands).

That was part of the intention. People would set "a lot of people" into "all on" and would load a lot more colonies, potentially get them raided etc.

PR #11412, opened 2025-11-30
Pathfinding stuckhandler improvement

someaddons:
- Improved stuckhandler, it now tries to advance the path node index as intended and reduced initial delay to 5 seconds

## Testing
- [x ] Yes I tested this before submitting it.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
+++ b/src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
@@ -939,6 +939,14 @@ public class MinecoloniesAdvancedPathNavigate extends AbstractAdvancedPathNaviga
     @Override
     protected void followThePath()
     {
+        if (path.getNextNodeIndex() == 0)
+        {
+            if (BlockPosUtil.dist(path.getNextNodePos(), ourEntity.blockPosition()) > 2)
+            {
+                Log.getLogger().warn("far distance for start");
+            }
+        }
+
         // TODO: Rework pathfollow
         getSpeedFactor();
         final int curNode = path.getNextNodeIndex();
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
```diff
@@ -939,6 +939,14 @@ private boolean handleEntityInWater(int oldIndex, final PathPointExtended pEx)
     @Override
     protected void followThePath()
     {
+        if (path.getNextNodeIndex() == 0)
+        {
+            if (BlockPosUtil.dist(path.getNextNodePos(), ourEntity.blockPosition()) > 2)
+            {
+                Log.getLogger().warn("far distance for start");
```
Raycoms: keep this in?

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/PathingStuckHandler.java
```diff
@@ -341,6 +334,23 @@ private void tryUnstuck(final NAV navigator)
         }
         delayToNextUnstuckAction = 50;
 
+        if (stuckLevel < 0)
```
Raycoms, line 337: Does this mean for the first 25 seconds it tries to go index by index ahead?

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
someaddons: uh no :D

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/PathingStuckHandler.java
someaddons: yes and no, it goes index by index 5times and waits 1.5sec after to see if that caused progression on the path. All those values are in ticks, so 30 = 1.5sec

PR #11413, opened 2025-12-01
Refactor settings

armele:
Closes None

- Refactor SettingsModuleWindow to allow it to be extended with a different UI layout, if desired.
- Use case: I intend to add a button to the settings window for a custom building in the Trade Post mod.

## Testing

Review please


src/main/java/com/minecolonies/core/client/gui/modules/SettingsModuleWindow.java
```diff
@@ -112,4 +112,13 @@ public void updateElement(final int index, @NotNull final Pane rowPane)
             }
         });
     }
+
+    /**
+     * Gets the resource location for the window's layout.
+     * @return the resource location
+     */
+    protected static ResourceLocation getWindowResourceLocation()
```
someaddons, line 120: static?

armele: Doesn't change per instance - but more importantly can't call a non static function from the super() call in the constructor.

Thodor12: Protected static is a bad idea, give an alternative constructor in that case. However realistically, is this needed? The setting components can already provide their own view, this specific view is just a simple list that aggregates all the settings.

someaddons: when this call is a static util this PR simply does nothing. Compiler will inline the call and its literally the same as before? :D

armele: Ah, silly me... I thought I could still override it in a child.  Ok, I will close this PR and rethink the approach.

Thodor12:
```java
    public SettingsModuleWindow(final SettingsModuleView moduleView)
    {
        this(moduleView, new ResourceLocation(Constants.MOD_ID, "gui/layouthuts/layoutsettings.xml"));
    }

    protected SettingsModuleWindow(final SettingsModuleView moduleView, final ResourceLocation windowId)
    {
        super(moduleView, windowId);
        settingsList = window.findPaneOfTypeByID(LIST_SETTINGS, ScrollingList.class);
    }
```

Thodor12: That being said, I do not think this is a good idea. These windows enforce the existence of certain GUI objects. I do not want to leave that up to chance to an addon to fill correctly. That's why you should be making your own modules for this stuff.

armele: That's fair. My approach so far has been to introduce new modules for things significantly different (trade connections, recycler status, pet assignment, etc) and to adapt modules where things are very similar (item lists).  I'll bias more towards new modules unless the functionality is identical.

Thodor12: I'd recommend you actually create a custom setting class that actually incorporates your button you wanted. That way you do not have to stray from the settings module default implementation at all. You can design individual setting options exactly how you feel like.

armele: Excellent, thanks for the tip... I'll explore that.

PR #11419, opened 2025-12-03
Adjust research cost source

### LINKED ISSUE #10954: The "Builder Modes" research option should not consume enchanted diamond axes

- [ ] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.939-1.21.1alpha

### Structurize Version

1.0.778-1.21.1

_No response_

### Current Behavior

When starting the "Builder Modes" research option in the university, diamond axes with enchantments can be consumed. 

### Expected Behavior

Enchanted items, which are often items a player has spend a lot of time and experience crafting, should not be consumed by the university. 

### Reproduction Steps

1. do thing
2. bad thing happen
3. cry

### Logs

.

### Anything else?

_No response_

### Footer

---
## Viewers

armele:
Allows costs to be paid from items in the university hut block. Prioritizes the university hut block before the player inventory. Ignores enchanted items or items with custom name tags.

Closes #10954 
Closes Discussion at [Plea for an Accessibility Feature](https://github.com/ldtteam/minecolonies/issues/10954)

- Allows research costs to be paid from the university hut block inventory.
- Prioritizes items in the university hut block over those in a player inventory.
- Ignores items with enchantments or custom name tags.
- If a "garbage" enchanted item is desired for use in research, it can be placed in a grindstone to strip it down to the enchanted version.

## Testing
-- Test cost paid only from hut block.
-- Test cost paid only from player inventory.
-- Test cost paid split across inventories.
-- Test skipping enchanted item.

Review please


Shadizar:
Garbage items are often cursed or otherwise unable to be "normalized"...
I'd request that portion not be done please (at least for the hutblock content portion)

rainweaver76:
Aye, I'd skip that ignore and have it take whatever is in the hutblock. Would be a good way to offload some junk items if they have weird enchantments,. Really the only change here would be to require the research to take from the hut inventory instead of player inventory. You still have to go queue the research, just not have to carry things back and forth and risk an oops.

armele:
I'm totally open to that - I'd love to have one core dev opinion on that before I make the change, though, since I know it was received with great "meh" when I suggested it in chat. ;)

*edit* - for clarity, loose match in the university hut block, strict match on player inventory.  Chance of loss of special item then requires a user error, explicitly...  Low risk.

someaddons:
I think we should add a Chest icon to the building's research UI (where the 3 branches are listed below is space) which has a hover tooltip explaining that you can put items in there to start research

src/main/java/com/minecolonies/core/research/GlobalResearch.java
```diff
@@ -164,21 +173,47 @@ public boolean canDisplay(final int uni_level)
...
         if (costList.isEmpty())
         {
             return true;
         }
 
+        final IItemHandler playerInventory = new InvWrapper(Minecraft.getInstance().player.getInventory());
+        IItemHandler univInventory = null;
+
+        BlockEntity be = player.level().getBlockEntity(universityPos);
+        if (be != null)
```
Raycoms: The itemhandler here is very inefficient. Use the "item in building, or item count in building" methods in the InventoryUtils here. Those are a lot faster. Player Inventory can't be null, as nonNull, and we use the Intellij "NotNull" annot

src/main/java/com/minecolonies/core/research/LocalResearchTree.java
```diff
@@ -158,8 +162,10 @@ public void attemptBeginResearch(final Player player, final IColony colony, fina
                 colony.getResearchManager().markDirty();
                 return;
             }
-            final InvWrapper playerInv = new InvWrapper(player.getInventory());
-            if (!research.hasEnoughResources(playerInv))
+            
+            BlockPos universityPos = colony.getBuildingManager().getBestBuilding(player.getOnPos(), BuildingUniversity.class);
```
Raycoms: Can't we send the building through here into this method instead of going through all buildings?

src/main/java/com/minecolonies/core/research/LocalResearchTree.java
```diff
@@ -177,15 +183,44 @@ public void attemptBeginResearch(final Player player, final IColony colony, fina
                     }
                 }
             }
+
+            IItemHandler univInventory = null;
+
+            BlockEntity be = player.level().getBlockEntity(universityPos);
+            if (be != null)
```
Raycoms: I think we have still more efficient building methods than going through the joint item handler.

Raycoms:
> I think we should add a Chest icon to the building's research UI (where the 3 branches are listed below is space) which has a hover tooltip explaining that you can put items in there to start research

Yes, that's probably a good idea.

src/main/java/com/minecolonies/core/research/LocalResearchTree.java
armele: This gets called from both the client (WindowResearchTree) and the server. Server side is easy to switch to InventoryUtils building helpers. Client side we have the BuildingView (not the building). I'm unsure of  a client-side approach to leverage the InventoryUtils helpers. Any advice you may have would be appreciated. I don't love the idea of splitting it into sided logic, but I suppose that could be one approach...

Raycoms: Shouldn't be hard to translate some of those to be sideless (Accept some sideless interface).
All it needs is the positions of the containers and then properly handle the racks for which the info is present on both sides

armele:
> > I think we should add a Chest icon to the building's research UI (where the 3 branches are listed below is space) which has a hover tooltip explaining that you can put items in there to start research
> 
> Yes, that's probably a good idea.

image

I've also made the chest clickable to open the inventory itself, for convenience.

someaddons:
I'd center it and make it much bigger^^

armele:
> I'd center it and make it much bigger^^

image
?

src/main/java/com/minecolonies/core/research/LocalResearchTree.java
armele: While refactoring for this I believe I have discovered that the hut block itself is not present in the container list (which includes building racks) of the building view. Is it ok to fix this, or is this by design?

Raycoms: We should be handling the hut block manually additionally so that would be a bigger refactor then.

armele:
Ok, I incorporated all the code review feedback. It grew a bit in terms of code touched. I re-tested all the client paths that touched the BuildingView.getContainerList (now BuildingView.getContainers) as well as the scenarios described in the PR, the fact that enchanted items will be taken from the building inventory to pay research costs but NOT the player inventory.

src/main/java/com/minecolonies/core/tileentities/TileEntityColonyBuilding.java
```diff
@@ -759,4 +759,39 @@ private void processBlueprint(final Blueprint blueprint)
...
+     * Get the BlockPos of the containers.
+     *
+     * @return a list of BlockPos representing the containers.
+     * If the building is null, an empty list is returned.
+     * If the building view is null, an empty list is returned.
+     * If the level is client side, the containers are retrieved from the building view.
+     * If the level is server side, the containers are retrieved from the building.
+     */
+    @Override
+    public List<BlockPos> getContainers()
```
Raycoms: Can't we call this from the building, instead of through the TE?

src/main/java/com/minecolonies/core/research/LocalResearchTree.java
```diff
@@ -177,15 +184,47 @@ public void attemptBeginResearch(final Player player, final IColony colony, fina
...
             {
                 int toRemoveLeft = cost.getCount();
 
                 for (Item item : cost.getItems())
                 {
-                    final List<Integer> slotsWithMaterial = InventoryUtils.findAllSlotsInItemHandlerWith(playerInv, stack -> stack.getItem().equals(item));
-                    for (Integer slotNum : slotsWithMaterial)
+                    if (building != null)
+                    {
+                        final Map<IItemHandler,List<Integer>> univSlotsWithMaterial = InventoryUtils.findAllSlotsInProviderWith(building, stack -> IGlobalResearch.isUniversityResearchMatch(stack, item));
```
Raycoms: This should probably be a util as well

src/main/java/com/minecolonies/core/tileentities/TileEntityColonyBuilding.java
armele: It could, but unless I'm thinking about it wrong, it will be an extra step.  To go through the building (or building view) we'd introduce a sided check to grab either the building or building view from the block position.

src/main/java/com/minecolonies/core/research/LocalResearchTree.java
armele: Refactored as a util for sharing item reduction across a building and player inventory (with separate predicates to support this use case).

src/main/java/com/minecolonies/core/tileentities/TileEntityColonyBuilding.java
Raycoms: While for Colonies we have a nice generic IColony, for buildings we currently don't have it, right. Which makes it a bit weird handing around buildings generically. I'm currently doing some refactoring for that. We should just already have an IBuildingInventory everywhere as input where we query this. Either in the research Ui we got the buildingview, or in the actual serverside logic where we got the building

uecasm: Most interactions with inventories need to be side-aware anyway, since client-side is read-only.  This shouldn't really be hard since the client-side check would only be in the tooltip and the server-side action would only be in the message handler, both of which are strongly sided.

armele: Refactored to make a sided helper that only needs the BlockPos, and took the IBuildingInventory out of the TE such that it only belongs to IBuilding and IBuildingView.  Retested all-in-hut, all-in-player and split scenarios.

someaddons (dismissed):

armele:
Please remove the "Untested" tag on this PR.  I have re-tested the PR after updating with latest changes in version/main.

armele:
I have re-tested this one as well - please remove the "Untested" label.

armele:
Please remove the "Untested" label - I have re-merged from main and re-tested (as well as removing some unused imports).

armele:
I have converted this PR to the refactored API.  I got rid of the previously-introduced IBuildingInventory in favor of moving the necessary methods to the new ICommonBuilding.  I re-tested these scenarios:
- All items in player inventory
- All items in building inventory
- Split items across both inventories
- Enchanted item NOT taken from player inventory
- Enchanted item IS taken from building inventory.

Please remove the "Untested" label.  Thanks!

src/main/java/com/minecolonies/core/util/BuildingUtils.java
```diff
@@ -185,4 +189,36 @@ else if (blockState.getBlock() instanceof DirectionalBlock)
...
+     * Get the ICommonBuilding for a given BlockPos on a specific side (client or server).
+     * If the building doesn't exist, it returns null.
+     * @param level the level.
+     * @param pos the position of the building.
+     * @return the ICommonBuilding of the building, or null.
+     */
+    public static ICommonBuilding buildingInventoryForSide(Level level, BlockPos pos)
+    {
+        ICommonBuilding buildingInventory = null;
+
```
Raycoms: extra spaces

src/main/java/com/minecolonies/core/util/BuildingUtils.java
```diff
@@ -185,4 +189,36 @@ else if (blockState.getBlock() instanceof DirectionalBlock)
...
+     * @param level the level.
+     * @param pos the position of the building.
+     * @return the ICommonBuilding of the building, or null.
+     */
+    public static ICommonBuilding buildingInventoryForSide(Level level, BlockPos pos)
+    {
+        ICommonBuilding buildingInventory = null;
+        
+
+        if (level.isClientSide)
```
Raycoms: As this returns ICommonBuilding you can now do colony.getCommonBuildingManager().getBuilding to simplify the below logic

armele: I might need a bit of guidance here. Obtaining the colony was a side-specific method in my original code. Is there now a side-agnostic way to retrieve the colony?  I tried this, but from the client it returned "null" for the colony: 

```        
        ICommonBuilding buildingInventory = null;
        
        IColony colony = IColonyManager.getInstance().getColonyByPosFromWorld(level, pos);
        if (colony != null)
        {
            Log.getLogger().info("BuildingUtils.commonBuildingFromPosition() retrieved Colony {} at position {}.", colony.getID(), pos);
            buildingInventory = colony.getCommonBuildingManager().getBuilding(pos);
        }

        return buildingInventory;```

Raycoms: public IColony getIColony(@NotNull final Level w, @NotNull final BlockPos pos)

armele: Never mind - I found IColony colony = IColonyManager.getInstance().getIColony(level, pos);

armele:
I have updated the PR to reflect the code review comments and re-tested these scenarios. Please remove the "Untested" label.  Thanks!

src/main/java/com/minecolonies/core/util/BuildingUtils.java
Raycoms: okay, I'm sorry as this is nitpicky. But this method can go still smaller.

If colony != null 
return .... getBuilding

return null;

Don't need to save buildingInventory at the top level and we can yeet the white spaces at the top too.
Can you also @NotNull final the two params?

armele:
Addressed code review feedback and re-tested. Please remove the "Untested" label.  Thanks!

Raycoms (approved):

someaddons (approved):

PR #11443, opened 2025-12-16
fix player tag getting checked instead of entity it interact with

cylian914:
the tag minecolonies:allowinteract was being check against the player instead of the entity making it not work

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/permissions/ColonyPermissionEventHandler.java
+++ b/src/main/java/com/minecolonies/core/colony/permissions/ColonyPermissionEventHandler.java
@@ -435,7 +435,7 @@ public class ColonyPermissionEventHandler
             return;
         }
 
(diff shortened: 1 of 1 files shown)


Thodor12:
Thanks for the PR, but the tag is deprecated given that people can configure free to interact blocks in the townhall. So the tag and the things loaded from config can actually just be removed in the first place

Raycoms:
Incorrect, this is still relevant.

MotionlessTrain:
^ This is used for things like corpses, which should be able to be interacted with anywhere in a colony, if someone died there, without first waiting for the owner to give access to that one block to freely interact with the corpse

Raycoms (approved):

someaddons (approved):

someaddons:
The tag is a setting for all colonies, while the colonies individual settings need to be configured on a per colony basis

PR #11451, opened 2025-12-21
Panel & Dining Hall adjustment

Raycoms:
Closes #
Closes #

- fix panel pathing issue 
- make request stack based at dining hall again

## Testing

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/util/BlockPosUtil.java
+++ b/src/main/java/com/minecolonies/api/util/BlockPosUtil.java
@@ -792,7 +792,7 @@ public final class BlockPosUtil
      * @param pos1Z start-z
      * @param pos2X end-x
      * @param pos2Z end-z
-     * @return the direction.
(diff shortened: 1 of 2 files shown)


MotionlessTrain:
Would this not cause issues when there are only enough ingredients in the colony to fulfil a request of 16 items?
I thought I noticed that crafters usually like to make the entire amount of items in case of a “deliver what’s available” request, so if there items available for 16 items only, that seemed to fail

Raycoms:
> Would this not cause issues when there are only enough ingredients in the colony to fulfil a request of 16 items? I thought I noticed that crafters usually like to make the entire amount of items in case of a “deliver what’s available” request, so if there items available for 16 items only, that seemed to fail

It's at min count 1, and recently adjusted crafting to "under deliver" as well.
So that should be fine

MotionlessTrain:
Ah, ok. I think I missed the update to crafters

someaddons (approved):

PR #11458, opened 2025-12-23
Select res UI port

someaddons:
Closes #
Closes #

Migrate UI's to use the new unified selectresource window from structurize
Farm field UI now displays biome and its climate for planting crops

Review please


Raycoms:
please port

PR #11466, opened 2025-12-27
Caching

someaddons:
Closes #
Closes #

- Caches login for a while

Review please(in discord)

### DIFF
--- a/src/main/java/com/minecolonies/core/client/auth/UnlockToken.java
+++ b/src/main/java/com/minecolonies/core/client/auth/UnlockToken.java
@@ -1,0 +1,284 @@
+package com.minecolonies.core.client.auth;
+
+import com.google.gson.JsonObject;
+
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.security.MessageDigest;
+import java.util.Base64;
+import java.util.UUID;
+
+import static com.minecolonies.core.generation.DataGeneratorConstants.GSON;
+
+public final class UnlockToken
+{
+    private static final byte[] STATIC_SECRET =
+        "9ccvhWUtG4FYXd14vKtZkntUC1DrqOWD".getBytes(StandardCharsets.UTF_8);
+
+    private static final String HMAC_ALGO = "HmacSHA256";
+    public static final  String VER       = "ver";
+    public static final  String TIME_NOW  = "iat";
+    public static final  String TIME_EXP  = "exp";
+    public static final  String UUID      = "uuid";
+    public static final  String UNLOCK    = "unl";
+
+    /**
+     * Unlockable features
+     */
+    public static final String SKIN           = "skins";
+    public static final int    UNLOCK_MINUTES = 60 * 24 * 7;
+
+    /**
+     * Path to the unlock token file, saved in a OS temp data folder
+     */
+    private static final Path tokenFile = getBaseDir().resolve("unlock.cache");
+
+    private UnlockToken() {}
+
+    /**
+     * Saves the token to disk
+     *
+     * @param token token string
+     */
+    public static void save(String token)
+    {
+        try
+        {
+            Files.createDirectories(tokenFile.getParent());
+            Files.writeString(
+                tokenFile,
+                token,
+                StandardCharsets.UTF_8,
+                StandardOpenOption.CREATE,
+                StandardOpenOption.TRUNCATE_EXISTING,
+                StandardOpenOption.WRITE
+            );
+        }
+        catch (IOException e)
+        {
+            // Ignore: failing to cache doesnt matter
+        }
+    }
+
+    /**
+     * Create a new String token for the unlocked feature
+     *
+     * @param playerUuid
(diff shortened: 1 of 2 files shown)


Raycoms (dismissed):

marchermans (changes requested):
This is a very bad idea.
You store the secret key in a publicly available source.

It should suffice to cache the key for the session and nothing more, you don't need to write that to disk.

Thodor12:
I actually made a similar class in another PR I was working on which I think gets the same idea across in a nicer solution:

https://gist.github.com/Thodor12/e66f68dbb22f1487c9d6c32e44c9d9d5

It uses a LoadingCache which is a persistant map which keeps the feature unlocks available for an hour before recycling the information out of the map

src/main/java/com/minecolonies/core/client/auth/UnlockToken.java
```diff
@@ -0,0 +1,284 @@
...
+        {
+        }
+    }
+
+    public static Path getBaseDir()
+    {
+        String os = System.getProperty("os.name").toLowerCase();
+        String home = System.getProperty("user.home");
+
+        if (os.contains("win"))
```
Thodor12, line 243: This feels like a lot of extra conditions that aren't necessary. Just dump it in the user.home folder always, who cares where the hell we place this file?
Users don't really need to touch this file so it doesn't matter at all whether it lives in appdata or wherever

src/main/java/com/minecolonies/core/client/auth/UnlockToken.java
```diff
@@ -0,0 +1,284 @@
...
+        String xdg = System.getenv("XDG_CONFIG_HOME");
+        if (xdg != null)
+        {
+            return Paths.get(xdg, "minecolonies");
+        }
+
+        return Paths.get(home, ".config", "minecolonies");
+    }
+
+    private static String base64(byte[] data)
```
Thodor12, line 267: Can't we inline the few usages of these methods, it's not enough to justify an entirely new method in my eyes

src/main/java/com/minecolonies/core/client/auth/UnlockToken.java
```diff
@@ -0,0 +1,284 @@
...
+        catch (IOException e)
+        {
+            return null;
+        }
+    }
+
+    /**
+     * Delete the token
+     */
+    public static void delete()
```
Thodor12, line 227: Unused

src/main/java/com/minecolonies/core/client/gui/townhall/WindowMainPage.java
```diff
@@ -339,10 +344,20 @@ public void checkFeatureUnlock()
...
+                }
+
+                // Checks the token file incase the api returned false, to buffer auth issues until the token runs out
+                isFeatureUnlocked.set(isFeatureUnlocked.get() || UnlockToken.isFeatureEnabledFor(playerUuid, SKIN));
             }
             catch (IOException e)
             {
-                e.printStackTrace();
+                // Checks the token file when the api has issues
+                isFeatureUnlocked.set(UnlockToken.isFeatureEnabledFor(playerUuid, SKIN));
```
Thodor12, line 359: Wait, this isn't even caching, caching implies you read the cache before you make the external call, which is not even what happens. This needs to be actual caching by not reading the API, or change the wording of the description of the PR that this is a local fallback.

src/main/java/com/minecolonies/core/client/auth/UnlockToken.java
```diff
@@ -0,0 +1,284 @@
...
+    public static final  String VER       = "ver";
+    public static final  String TIME_NOW  = "iat";
+    public static final  String TIME_EXP  = "exp";
+    public static final  String UUID      = "uuid";
+    public static final  String UNLOCK    = "unl";
+
+    /**
+     * Unlockable features
+     */
+    public static final String SKIN           = "skins";
```
Thodor12, line 29: Something that hasn't been considered at all, even though it's not an issue per sé right now. If we have multiple feature types, there can only be 1 file, and this data isn't aggregated into the file.
So the implementation with being able to define multiple feature types right now is practically useless, as the files will continually overwrite one another and invalidate one another. (Because the signature contains the unlock type which was requested, when you save the file this signature (thus unlock type) is stored in the file, if another feature type comes along and wants to save the file, it's going to overwrite the local cache file with a different signature referring to another feature type)

src/main/java/com/minecolonies/core/client/gui/townhall/WindowMainPage.java
```diff
@@ -339,10 +344,20 @@ public void checkFeatureUnlock()
...
                 reader.close();
                 isFeatureUnlocked.set(Boolean.parseBoolean(response.toString()));
+                if (isFeatureUnlocked.get())
+                {
+                    final String token = UnlockToken.create(playerUuid, SKIN, UNLOCK_MINUTES);
+                    UnlockToken.save(token);
+                }
+
+                // Checks the token file incase the api returned false, to buffer auth issues until the token runs out
+                isFeatureUnlocked.set(isFeatureUnlocked.get() || UnlockToken.isFeatureEnabledFor(playerUuid, SKIN));
```
Thodor12, line 354: Shouldn't we treat the API as a higher priority no matter what? If we get a valid response from the API we should treat that as the truth, and not allow fallbacks to be used in that case.

Thodor12 (changes requested):

PR #11475, opened 2026-01-01
Better Builder State Checking

### LINKED ISSUE #11414: Beehives being replaced on Apiary upgrade, and likely repairs.

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.21

### MineColonies Version

minecolonies-1.1.1218-1.21.1-snapshot

### Structurize Version

structurize-1.0.796-1.21.1-snapshot

_No response_

### Current Behavior

When currently upgrading, or likely repairing the Apiary, if a beehive is in a state where it contains honey and/or bees, the builder will request a new empty beehive and replace it, destroying anything that was contained inside. 

### Expected Behavior

The builder should ignore the state of any beehive when upgrading or repairing, unless it's missing, and leave it as it.

### Reproduction Steps

1. Make an Apiary, make sure there are no worker to avoid harvesting
2. Make sure the beehive(s) are filled with honey and bees
3. Upgrade or repair the apiary

### Logs

None required

### Anything else?

I'm not sure if it's just the honey in the hive, or if bees inside are also affecting it, but when it happened to me, the few bees I had vanished right after the upgrade, which might mean they were inside the beehive when the builder replaced it, destroying them.

### Footer

---
 ## Viewers

Raycoms:
Closes #11414
Closes #

I want this to be a bit of a discussion before merging.
a) What other blocks do we want this for. 
b) How about fence-gates & doors. Domum doors we need to do a tile entity comparison, so checking a generic door tag is a bit messy. Do we want a separate tag for this, or do we just add generic door tag here and check for domum manually?
b) The flower pot check is weird. It exists there for a while but it just leaves an empty flower pot somewhere even if in the next upgrade there is a flower in it. I'd suggest removing it, but we should test if we run into any bugs with it.

- Added tag of blocks where state changes should be ignored
- Added vanilla/minecolonies compost barrels & beehives

Review please


someaddons:
I think it'd be easier to handle it on the placement handler level, the general placement handler could ignore blocks in the tag, the other handlers could do it individually like our DO door handler etc

Raycoms:
> I think it'd be easier to handle it on the placement handler level, the general placement handler could ignore blocks in the tag, the other handlers could do it individually like our DO door handler etc

I had wondered about that as well, but I felt like this might be not great performance wise, as we'd iterate all placement handlers for each position.

someaddons:
not sure why we'd need to iterate all placement handlers? I'd do it similarly to existing handlers, they accept handling a block type, and if the world state equals the to be placed state in the general block placement handler and the block is in the ignorediff tag then do nothing on placement. Looking at it yesterday some of them already have similar checks

MotionlessTrain:
The entire list gets iterated until one specifies that it can handle the block
The catch all is at the end of the list, so having too much placement handlers means that all of them are examined very often (I noticed they are called quite often, by both the build options screen and the builder at several steps)

someaddons:
ye I know how the placement handlers work, ray made it sound like he'd add an additional iteration somewhere. 
Iterating the placement handlers isnt that expensive and the blocks we'd skip are a marginal percentage so performance wise this isnt important

someaddons (approved):

PR #11489, opened 2026-01-09
Level 0 Home Safeguard

armele:
Closes: Eliminates this exception if a citizen is in a house that is level 0.

```
Condition check for state ACTIVE threw an exception:
java.lang.IllegalArgumentException: bound must be positive
	at java.base/java.util.Random.nextInt(Random.java:557) ~[?:?]
	at TRANSFORMER/minecolonies@1.1.1241-1.21.1-snapshot/com.minecolonies.core.colony.CitizenData.update(CitizenData.java:1649) ~[minecolonies-1.1.1241-1.21.1-snapshot.jar%23202!/:1.1.1241-1.21.1-snapshot]
```

- Safeguard Level 0 homeBuildings as already done for Null homeBuilding.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/CitizenData.java
+++ b/src/main/java/com/minecolonies/core/colony/CitizenData.java
@@ -1641,7 +1641,7 @@ public class CitizenData implements ICitizenData
             return;
         }
 
-        final int homeBuildingLevel = homeBuilding == null ? 1 : homeBuilding.getBuildingLevel();
+        final int homeBuildingLevel = homeBuilding == null ? 1 : Math.max(homeBuilding.getBuildingLevel(), 1);
         if (leisureTime > 0)
         {
             leisureTime -= tickRate;


armele:
I have tested this change. Please remove the "Untested" label. Thanks!

someaddons:
How does a citizen end up with a level 0 home building? level 0's should not assign citizens

armele:
I have an addon mod use case where a work building can double as a home, and assignment is intended after placement (as opposed to after construction - similar to hiring a builder in a level 0 builder hut).  I also contemplated (but did not test) that perhaps this condition would be hit if you were to deconstruct a residence?

someaddons:
Deconstructed buildings keep their levels, dont think we should allow assinging citizens that early except for the builder where we have to

Raycoms:
Allowing a level 0 building to be a home building causes other issues too like where do they go to sleep, there is no bed, etc.

PR #11492, opened 2026-01-12
Fix format specifiers in translations [1.20]

### LINKED ISSUE #11491: Some sentences in the Japanese language file display elements like "%1$d" as-is.

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1135-1.21.1-snapshot

### Structurize Version

1.0.790-1.21.1-snapshot

1. NeoForge 21.1.218
2. 1.0.208-1.21.1-snapshot
3. 1.0.221-snapshot

### Current Behavior

I play Minecolonies in Japanese, but for a while now, I've been bothered by how some villager comments display things like "%1$d" as-is.
Image
(I modified the language file to verify if I could resolve this personally, so it may differ from the actual text (but the issue is real).)

### Expected Behavior

I believe "%1$d" is being incorrectly converted to "%1%d".

### Reproduction Steps

1. In Minecolonies, progress to the point where a certain building is under construction with the villagers' assistance.
2. Talk to the villager who gave you the request.

### Logs

There are no logs.

### Anything else?

I'm not very good at English, so I used DeepL Translate to help me. Sorry for the awkward English.

### Footer

---
 ## Viewers

Thodor12:
Closes #11491

- Regex replaced all %d and %[1-9]$d into %s or number equipvalents

Review please


src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -267,8 +267,8 @@
   "minecolonies.config.canplayeruseresetcommand.comment": "Should players be allowed to use the /mc colony requestsystem-reset command?",
 
   "minecolonies.config.default.boolean": "[Default: %b]",
```
uecasm, line 269: %b is also not a thing

src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -267,8 +267,8 @@
...
-  "minecolonies.config.default.int": "[Default: %d, min: %d, max: %d]",
-  "minecolonies.config.default.long": "[Default: %d, min: %d, max: %d]",
+  "minecolonies.config.default.int": "[Default: %s, min: %s, max: %s]",
+  "minecolonies.config.default.long": "[Default: %s, min: %s, max: %s]",
   "minecolonies.config.default.double": "[Default: %f, min: %f, max: %f]",
```
uecasm, line 272: And neither is %f

uecasm: Actually wait, this is the reverse issue.  At least currently, the "config" translation keys don't go through the normal Minecraft translation code (which requires %s regardless of type) but directly uses `String.format` and friends (which do require type-specific placeholders).  So keep %f and revert back to %d for int/long here; this is a rare case where that's still needed.

Though having said that, %s should still work for all, it just might be less than ideal formatting.

someaddons (approved):

PR #11494, opened 2026-01-16
Full Dark Oak update

Jinotad:
Closes #
Closes #

Updated all of Dark Oak style: 

Updated tags and removed waypoints for all huts schematics
Broke and replaced all racks.
Fixed the Netherminer catching on fire
Added Gatehouse and plantation field

-

Review please


Jinotad:
I've tested the schematics, which should be the only changes.

PR #11498, opened 2026-01-18
Animal defense

armele:
Closes: Discord report https://discord.com/channels/139070364159311872/1125094832143077456/1462177040538009885

- Slight enhancement to the exception catching around RecipeAnalyzer.createAnimal()
- Note that the root cause is *not* Minecolonies, but this change allows that problem to be handled and reported more gracefully.  (As reported, the exception prevents world loading. With this change the underlying problem is logged and deferred and the game can start normally.)

Review please


armele:
Please remove the "Untested" tag. This change was tested.  (Specifically, I introduced a NoClassDefFoundError within the createAnimals loop, verified the logging and proper world start. This test code was removed prior to submitting the PR.)

src/main/java/com/minecolonies/core/colony/crafting/RecipeAnalyzer.java
```diff
@@ -123,9 +127,9 @@ public static List<Animal> createAnimals(@NotNull final Level level)
                     animals.add(animal);
                 }
             }
-            catch (final Exception ex)
+            catch (final LinkageError | Exception ex)
```
uecasm, line 130: You could probably just use `Throwable` here instead, in case of other weirdness.

armele: I did consider that but I worried it might be too broad (sucking up memory errors and other truly critical stuff)...

uecasm (dismissed):

Raycoms (dismissed):

someaddons:
Why do we need to add a try catch for a simple bug in another mod? :D

Raycoms:
> Why do we need to add a try catch for a simple bug in another mod? :D

There are a couple of reasons things can break there when mods do weird things. It's better to not crash on our side in an analysis just.

armele:
That was kinda my thought, too - not a minecolonies problem, but it can LOOK like one, and it just helps make the ecosystem better.  And easy! :)

someaddons:
> > Why do we need to add a try catch for a simple bug in another mod? :D
> 
> There are a couple of reasons things can break there when mods do weird things. It's better to not crash on our side in an analysis just.

by that logic we'll need to try-catch any entitytype.create call :D

src/main/java/com/minecolonies/core/colony/crafting/RecipeAnalyzer.java
```diff
@@ -123,9 +127,9 @@ public static List<Animal> createAnimals(@NotNull final Level level)
                     animals.add(animal);
                 }
             }
-            catch (final Exception ex)
+            catch (final LinkageError | Exception ex) 
             {
-                Log.getLogger().error("Couldnt analyze animal", ex);
+                Log.getLogger().error("Couldnt analyze animal: {}", id, ex);
```
Raycoms: Could you add here some additional text with:
Please report this to the mod author of: {}

someaddons (dismissed):

someaddons:
We should make sure to also open a bug report on the mod triggering this crash

src/main/java/com/minecolonies/core/colony/crafting/RecipeAnalyzer.java
armele: Done!

armele:
I have updated the code per the PR comment, and re-tested. (Please remove the "Untested" label.)  Thanks.

Raycoms:
Please port

PR #11502, opened 2026-01-21
Increase license and add usage description

Thodor12:
# Changes proposed in this pull request
- Add a license in the blueprints folder, no longer permitting reuse of blueprints
- Add a description of what our license implies for usage in forking

Review please

PR #11503, opened 2026-01-21
Increase license and add usage description [1.21]

Thodor12:
See #11502


marchermans (changes requested):
See other mainline pr

PR #11508, opened 2026-01-25
Fix/stuck craft req

Raycoms:
Closes #
Closes #

- Fix crafting requests getting stuck between assignment and resolving
-

Review please


src/main/java/com/minecolonies/api/colony/requestsystem/resolver/IRequestResolver.java
```diff
@@ -85,15 +85,15 @@ default void onRequestAssigned(@NotNull final IRequestManager manager, @NotNull
     }
 
     /**
-     * Indicates that a assigned request has been cancelled. Is called before graph is updated.
+     * Indicates that an assigned request has been cancelled. Is called before graph is updated.
      *
      * @param manager The manager that indicates the cancelling
      * @param request The request that has been cancelled.
      */
     void onAssignedRequestBeingCancelled(@NotNull IRequestManager manager, @NotNull IRequest<? extends R> request);
```
someaddons, line 93: two slightly differently named callbacks with the same desc? Why do we have two, what is the difference :D

Raycoms: Says in the top. One if before graph is updated, one is after

someaddons: ah missed that small diff :D

src/main/java/com/minecolonies/core/colony/jobs/AbstractJobCrafter.java
```diff
@@ -388,6 +388,11 @@ private void cancelAssignedRequests()
         {
             getColony().getRequestManager().updateRequestState(t, RequestState.FAILED);
         }
+
+        for (final IToken<?> t : getAssignedTasks())
```
someaddons, line 392: hm shouldnt the task queue and assigned tasks match? Aka contain the same requests?

Raycoms: No, they are completely separate queues.
task queue = tasks that are ready to craft
assigned queue = tasks missing materials

marchermans (approved):

src/main/java/com/minecolonies/core/colony/jobs/AbstractJobCrafter.java
someaddons: well going from the names it would be:
assigned = all requests for this crafter, no matter which state
taskqueue = order of operation on the assigned requests

someaddons (approved):

PR #11511, opened 2026-01-25
Animal defense port

armele:
Port of #11498 

Review please


armele:
Tested as with the 1.20 version - added a fake NoClassDefFoundError throw and verified logging worked as expected.  (Removed test code after test.)  Please remove the "Untested" label.  Thanks!

PR #11520, opened 2026-01-30
Smeltery Warehouse Min

armele:
Closes: [Discord Suggestion](https://discord.com/channels/472875599422291968/1158182298596954152)
Closes: Unreported bug with warehouse minimum preservation.

- Add a setting to the smeltery to indicate how many items should be left in the warehouse (similar to composter).
- During testing, identified that the logic to preserve items had an error.  For example, if two qualifying (different) stacks of 10 were in the warehouse, and the min-to-keep number was 8, what got kept was 2 of the 10 instead of 8 of the 10.  This PR includes a fix for that.

# Testing:
- Tested the default "all available items" path.
- Tested the "specific configured items" path.
- For both paths, observed warehouse depletion down to (but not below) configured amount.

Review please


armele:
I have tested this code.  Please remove the "Untested" label.  Thanks!

src/main/java/com/minecolonies/core/colony/requestsystem/resolvers/core/AbstractWarehouseRequestResolver.java
```diff
@@ -172,16 +175,36 @@ public List<IToken<?>> attemptResolveRequest(@NotNull final IRequestManager mana
...
+                        {
+                            storages.put(new ItemStorage(stack.getA()), storages.getOrDefault(new ItemStorage(stack.getA()), 0) + stack.getA().getCount());
+                            continue;
+                        }
+                        int toKeepThisStack = toKeep - alreadyKept;
+                        storages.put(new ItemStorage(stack.getA()), storages.getOrDefault(new ItemStorage(stack.getA()), 0) + toKeepThisStack);
+                        totalAvailable += stack.getA().getCount() - toKeepThisStack;
+                        continue;
+                    }
+                    totalAvailable -= toKeep;
```
Raycoms: This seems wrong. On each iteration we decrease total available?

armele:
Fixed the double-subtraction when computing warehouse availability and minor performance tuning.  Re-tested scenarios demonstrating min warehouse preservation with the Smeltery. Please remove the "Untested" flag.

Raycoms (approved):

someaddons (approved):

PR #11534, opened 2026-02-07
Announce closest colony by name

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/1469479157846249663)

- Changes the `/mc whereami` command to report the name and ID of the closest colony when you're outside any colony, not just the distance.
    - Perhaps we might want to restrict this on PVP servers or if mapping permission is denied?
    - The translation key was changed to avoid weird behaviour with outdated translations due to new arguments, and also to be more consistent with other keys.
- Fixes formatting (`%.2f` is not valid for Component).
- Switches to direct (as bee flies) distance rather than Manhattan distance (otherwise the float formatting made no sense anyway).


Raycoms (approved):

someaddons (approved):

Raycoms:
please backport

PR #965, opened 2017-05-10
Bug/1.11 remove some warnings

xavierh:
Review please

PR #11571, opened 2026-03-02
fix(miner): solving #965

muellerjp:
- BuildingConstants.java: renaming NBT Tag
- BuildingMiner.java: renaming BlockPos var, removing hardcoded requests for cobblestone, renaming var when (de)serializing NBT, renaming getCobbleLocation function
- EntityAIStructureMiner.java: renamed REQUEST_BATCH, more renaming and adapting to the new naming, removed getLadderBackFillBlock function hardcoding cobble and netherrack, functions now use getMainFillBlock
- MineColonies.java: Renaming Tag
- MinerLevelManagementModule.java: renaming function
- SchematicTagConstants.java: Renaming cobbletag to tag ladderback

Closes #
Closes #

-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/util/constant/SchematicTagConstants.java
+++ b/src/main/java/com/minecolonies/api/util/constant/SchematicTagConstants.java
@@ -39,7 +39,7 @@ public class SchematicTagConstants
     /**
      * Miner tags
      */
-    public static final String TAG_LADDER_BACK = "ladder_back";
+    public static final String TAG_LADDER_BACK = "cobble";
     public static final String TAG_LADDER = "ladder";
 
     /**
--- a/src/main/java/com/minecolonies/api/util/constant/SchematicTagConstants.java
+++ b/src/main/java/com/minecolonies/api/util/constant/SchematicTagConstants.java
@@ -39,7 +39,7 @@ public class SchematicTagConstants
     /**
      * Miner tags
      */
-    public static final String TAG_COBBLE = "cobble";
+    public static final String TAG_LADDER_BACK = "ladder_back";
     public static final String TAG_LADDER = "ladder";
 
     /**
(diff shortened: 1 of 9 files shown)


muellerjp:
i accidentially linked the wrong ticket...
This one is the correct one:
https://github.com/ldtteam/minecolonies-features/issues/965

src/main/java/com/minecolonies/api/util/constant/SchematicTagConstants.java
```diff
@@ -39,7 +39,7 @@ public class SchematicTagConstants
     /**
      * Miner tags
      */
-    public static final String TAG_COBBLE = "cobble";
+    public static final String TAG_LADDER_BACK = "ladder_back";
```
muellerjp: Renaming this tag does require changes in all existing styles.
Is that acceptable? Otherwise the tag stayes cobble for backwards compat, but it then no longer has to be cobblestone.

uecasm: It would be better to retain the existing tag regardless of being slightly less accurate.  Updating style packs is difficult, especially since there are a lot of external ones now.  And even nether styles used "cobble" as the tag despite placing netherrack instead, so there's precedent.

muellerjp: Ok, than i change this back.

muellerjp:
image
Tested in with the `gradew runClient`. Redstone block ladders are now possible :D

-> Wiki should recommend to use as fillerblock what ever the style uses as ladderback block. Otherwise when building the central shaft the ladder will be deconstructed and rebuild in the process.

Raycoms:
Isn't this the same as: https://github.com/ldtteam/minecolonies/pull/11573?

What's the dif?

MotionlessTrain:
Target branch

muellerjp:
> Isn't this the same as: #11573?
> 
> What's the dif?

yeah, the target branch, i got told that if i wanna have it in 1.21.1 i should do a separate PR targeting that version :)

PR #11576, opened 2026-03-09
Show helpful tooltips for dyed items

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/473575384445878273/1479807305930506302) (mostly)

- Modifies the tooltip for leather armor so that it shows the actual dye color required for this particular item.
    - Intended to help understand requests from blueprints that use dyed armor on stands.
    - Only works on single-dyed armor, not multi-colored armor.
    - Can be disabled in client config if you have another mod that does this.
image

Review please (should port)

### DIFF
--- a/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
+++ b/src/main/java/com/minecolonies/api/compatibility/CompatibilityManager.java
@@ -18,6 +18,7 @@ import com.minecolonies.api.util.*;
 import com.minecolonies.core.colony.crafting.CustomRecipeManager;
 import com.minecolonies.core.colony.crafting.LootTableAnalyzer;
 import com.minecolonies.core.generation.ItemNbtCalculator;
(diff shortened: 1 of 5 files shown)


Raycoms:
Can't we just add this without a config setting to the request window tooltips?

uecasm:
> Can't we just add this without a config setting to the request window tooltips?

Not really, because if they have another mod that adds dye colour tooltips then it will probably still show there too.  The point is to let the player (or more likely the modpack author) decide which is more useful to keep, if the duplication annoys them.  This is enabled by default, though.

On a separate note, I've been wondering whether we should also support double-dyes; it would be fairly trivial to add (the same is not true for N-dyes).  Currently I've limited it to just single dyes because that's all that builders are supposed to be using, since it was otherwise too hard to figure out; but if we have a nice tooltip for it then perhaps we could relax that.

someaddons (approved):

Raycoms:
please port

PR #11578, opened 2026-03-09
Fix: Pathfinding, Quest System, and AI bug fixes

ZeOs360:
Closes #
Closes #

- AbstractPathJob.java - pos.offset(BLOCKPOS_DOWN) //BlockPos is immutable
- QuestManager.java - isDirty = false; //to prevent constant re-serialization every tick.
- QuestDialogueInteraction.java - index++ //to fix the infinite logic loop in participant dialogues
- EntityAIEatTask.java - eatPos != null  //checked eatPos != null instead of restaurant != null

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/interactionhandling/QuestDialogueInteraction.java
+++ b/src/main/java/com/minecolonies/core/colony/interactionhandling/QuestDialogueInteraction.java
@@ -196,11 +196,20 @@ public class QuestDialogueInteraction extends StandardInteraction
         String localText = text.getString();
         if (localText.contains("$") && colonyQuest != null)
         {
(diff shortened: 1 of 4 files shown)


Raycoms (commented):
This smells a bit like an AI PR, a lot of the changes don't make a lot of sense to me.

ZeOs360:
> This smells a bit like an AI PR, a lot of the changes don't make a lot of sense to me.

Of course, I used AI assistance, but not entirely. I got carried away and PR turned to a kitchen sink. I think there are a few genuine fixes here, so I removed the rest and left only these 4.

someaddons:
thanks for your contributions, I've added them to my PR here: https://github.com/ldtteam/minecolonies/pull/11587
(had to adapt them slightly) which got merged so they're in now :) Closing this

PR #11614, opened 2026-04-04
Fix: Too many torches being requested in Colonial medium quarry in 1.20

MotionlessTrain:
Closes #
Closes #

- The prevProgressPos of the iterator was broken for the LayeredBlueprintIterator for quarries, as it was not properly delegated from the backing iterator to the layer iterator.
  It always returned `NULL_POS` instead. In case of a Colonial medium quarry, as it reached its limits whilst halfway through a layer,
  that meant it started anew on that layer in the next colony tick, so it was endlessly getting the same torches as requirements, and never progressing 
  This PR fixing the delegation, such that the layerBlueprintIterator can pick up where it left off

Review please


MotionlessTrain:
I tested this before making the PR, so the label can be removed

PR #11616, opened 2026-04-04
Use derived spawn point.

armele:
Closes: None

- Visitor manager was calculating a "good" spawn point based on a proximity reference point, but then ignoring the calculated spawn point.  This adjusts the entity position set to use the calculated spawn point.

Review please


armele:
I tested by building out a test world, plopping down a tavern and witnessing that tavern visitors spawned successfully.  Please remove the "Untested" label.

someaddons (approved):

PR #11626, opened 2026-04-19
confirm window for residence with missing infra & tp

Raycoms:
Closes #
Closes #

- Residence build/upgrade button has a useful tooltip to guide player
- Needs extra confirmation on build if req are not met
- UI confirm for tp not chat confirm

image

(Note, button says Yes/Cancel now)

Review please


Talyda:
~~Is this just to go from level 1 to level 2 housing? Cos it might be helpful to say *why* that matters (ie level 1 housing they get hungry slower) Otherwise it's just an extra step without any real explanation?~~

Ok, reading the lang file showed the changes and now I understand!

someaddons (approved):

PR #11629, opened 2026-04-22
Allow more carpets to be valid for schools

Thodor12:
See #11628


Raycoms (dismissed):

someaddons:
if we do this we should probably allow it by the wool carpet tag(and add the DO blocks to it?).
We got a conceptual problem though with this:
For example our deser oasis style school has floating carpet as roofing(as it was meant to be used) and normal carpet beneath
image

Thodor12:
> if we do this we should probably allow it by the wool carpet tag(and add the DO blocks to it?). We got a conceptual problem though with this: For example our deser oasis style school has floating carpet as roofing(as it was meant to be used) and normal carpet beneath image

That is a valid problem, but how would we distinguish between that?

I have one potential solution that keeps working backwards compatible. We still push this change through, but we add a tag to the schematics that allows you to define a subset of carpets to use. If you do not provide a single block with said tag in the schematic, all carpets are considered valid.

What do you think @someaddons?

Raycoms:
I do think we should move towards a tag based solution, yes.

uecasm:
~I'm pretty sure Ray meant the other kind of tag (tag tool tag, not block tag).  Like how the archery will use **only** the "work" tag to indicate standing positions, but if there are no such tags specified then it will look for glowstone blocks.  Similarly, this should probably look for a "carpet" or "chair" tag etc, and if none found, then fall back to carpet blocks.~
Nvm, I read further and see you did do that one too.

Having a carpet block tag is not a terrible idea as well (though in general inheritance checks are better, so it *should* be pointless), but it does nothing to solve someaddons' desire to exclude certain blocks for specific schematics.

Raycoms:
Yes, mirality is right

someaddons:
ye I think the proper solution is using a tag tool tag, e.g. "sitting" similar to the restaurant, and fallback to registered blocks for missing tags, so its backwards compatible

uecasm:
To be clear, since it looks like you might have missed the edited comment, the latest code does use the sitting tag tool tag (which I had initially missed).

It just also uses a block tag, which feels a little unnecessary, but not actually bad.

Raycoms:
I wouldn't use a block tag alongside, otherwise we can get backwards problems again when packmakers do derp

someaddons:
not to mention that old schematics using floating carpets will break still, until they get an update to include the sitting tag. I'd keep the old logic as is, normal carpet only and add the new sitting tag ontop

kezmodius:
Wasn't the original discord request for this specifically for adding floating carpets?

MotionlessTrain:
With a supported sit tag, that is solved in a different way, without breaking floating carpets which have been placed as roof in existing schools

Thodor12:
So then what, we just remove all the carpet conditions I added, keep the original condition and just require people to use sit tags going forward?

If so we might as well just remove the carpet positions in the next version and just make it mandatory to use the sit tags. That way designers have full control over it.

someaddons:
ye just keep the old code as fallback for a schematic missing sit tags(could also log the schematic in question), no need to change the old stuff since it works and is only used as fallback after

someaddons (dismissed):

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingSchool.java
```diff
@@ -124,21 +119,38 @@ public CompoundTag serializeNBT()
...
      * Get a random place to sit from the school.
      *
+     * @param random the random obj.
      * @return the place to sit.
      */
     @Nullable
-    public BlockPos getRandomPlaceToSit()
+    public BlockPos getRandomPlaceToSit(final RandomSource random)
     {
+        final List<BlockPos> validSitLocations = getLocationsFromTag(TAG_SITTING);
```
Raycoms, line 127: use our utils random instead

Thodor12: This is similar to how it's done in the archery, we really ought to make a choice and stick with it at one point, because we're using God knows what all randomly

Raycoms: Yeah, our random utils is the best choice for most cases where we pipeline entity randoms through atm

someaddons: > This is similar to how it's done in the archery, we really ought to make a choice and stick with it at one point, because we're using God knows what all randomly

ColonyConstants.rand is that choice

src/main/java/com/minecolonies/apiimp/initializer/InteractionValidatorInitializer.java
```diff
@@ -37,9 +37,10 @@
 import static com.minecolonies.api.util.constant.translation.RequestSystemTranslationConstants.REQUEST_SYSTEM_BUILDING_LEVEL_TOO_LOW;
 import static com.minecolonies.core.colony.buildings.modules.BuildingModules.RESTAURANT_MENU;
 import static com.minecolonies.core.entity.ai.workers.crafting.EntityAIWorkSmelter.ORE_LIST;
+import static com.minecolonies.core.entity.ai.workers.production.agriculture.EntityAIWorkFisherman.SUBOPTIMAL_POND_COMPLAINT_DISTANCE;
 import static com.minecolonies.core.util.WorkerUtil.getLastLadder;
 import static com.minecolonies.core.util.WorkerUtil.isThereCompostedLand;
-import static com.minecolonies.core.entity.ai.workers.production.agriculture.EntityAIWorkFisherman.SUBOPTIMAL_POND_COMPLAINT_DISTANCE;
+
 /**
```
Raycoms, line 43: reset change on this one so this file isn't touched please

src/main/java/com/minecolonies/core/entity/ai/workers/education/EntityAIWorkPupil.java
```diff
@@ -30,9 +30,9 @@
 import static com.minecolonies.api.research.util.ResearchConstants.TEACHING;
 import static com.minecolonies.api.util.constant.Constants.DEFAULT_SPEED;
 import static com.minecolonies.api.util.constant.Constants.TICKS_SECOND;
-import static com.minecolonies.api.util.constant.TranslationConstants.PUPIL_NO_CARPET;
-import static com.minecolonies.api.util.constant.StatisticsConstants.LEVELS_GAINED;
 import static com.minecolonies.api.util.constant.StatisticsConstants.ITEM_USED;
+import static com.minecolonies.api.util.constant.StatisticsConstants.LEVELS_GAINED;
+import static com.minecolonies.api.util.constant.TranslationConstants.PUPIL_NO_CARPET;
```
Raycoms, line 33: same here

someaddons (approved):

Issue #11622, opened 2026-04-11
Death Cordinate in Townhall can't display full cordinates if a number is too big


Jinotad:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Are you using the latest MineColonies Version?

- [x] I am running the latest **beta/release** version of MineColonies for my Minecraft version.
I am also running the latest versions of other mods that are part of my problem.

### Did you check on the Wiki? or ask on Discord?

- [x] I checked the MineColonies Wiki and made sure my issue is not covered there. Or I was sent from discord to open an issue here.

### What were you playing at the time? Were you able to reproduce it in both settings?

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

minecolonies-1.1.1299-1.21.1-snapshot

### Structurize Version

structurize-1.0.816-1.21.1

### Current Behavior

Ray said to hub this:

When a civ dies, its death is recorded in the town hall.  If the X number is too large, it can no longer display the complete coordinates.

<img width="819" height="538" alt="Image" src="https://github.com/user-attachments/assets/84dc83ef-15a4-41e1-9943-5c8e0ae427f4" />

<img width="826" height="509" alt="Image" src="https://github.com/user-attachments/assets/92473414-3812-4c5b-9f3b-c2f5ecf99814" />

### Expected Behavior

Properly display coordinate on larger numbers

### Reproduction Steps

Start a colony at 11,000x spawn a civ and kill it.  Look at townhall logs

### Logs

none

### Footer

---
 ## Viewers
  *  Add a 👍 reaction to the bug report if you are also affected. This helps the bug report become more visible without cluttering the comments.
  *  ⚠️ Only add a comment if you have new insights or background information not already mentioned. Off-topic or "+1" comments will be deleted to keep the discussion focused.

PR #11631, opened 2026-04-24
Fix layout changes in Townhall infopage

Multiplication-golf:
See #11622 

-Enlarges the coordinates values box in the Townhall infopage

Review please

Dynamic box sizing might be a future update that could be considered as a addon of the BlockUI project


Raycoms (approved):

someaddons (approved):

Multiplication-golf:
Please remove the untested label

Thodor12:
Wouldn't it almost be better if we just showed a word saying "coordinates" and then just render the actual coordinates in a tooltip?

uecasm:
> Wouldn't it almost be better if we just showed a word saying "coordinates" and then just render the actual coordinates in a tooltip?

If it was inline text, maybe, but here there's a dedicated row for the coords so it's better to just show them.  It was just not wide enough before (which originally wouldn't have mattered, but somewhere around 1.18 text started wrapping & clipping rather than overflowing if the box was too small).

Raycoms:
> > Wouldn't it almost be better if we just showed a word saying "coordinates" and then just render the actual coordinates in a tooltip?
> 
> If it was inline text, maybe, but here there's a dedicated row for the coords so it's better to just show them. It was just not wide enough before (which originally wouldn't have mattered, but somewhere around 1.18 text started wrapping & clipping rather than overflowing if the box was too small).

tbf I see this as a blockui bug that it doesn't overflow

PR #11636, opened 2026-04-27
Fix: starvation via courier

### LINKED ISSUE #11421: [Bug]: Couriers continuously move Restaurant food to Warehouse resulting in starvation

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1218-121.1-snapshot

### Structurize Version

1.0.794-121.1-snapshot

Domum Ornamentum 1.0.223-snapshot

### Current Behavior

Couriers are slowly removing cooked food from the Restaurant inventory and transporting it to the Warehouse. This happens even though the food items are actively set on the Restaurant menu. As a result, the Restaurant inventory keeps emptying, and colonists are starving. The Warehouse accumulates stacks of cooked food, but it is not brought back to the Restaurant even when colonists are hungry.

I have confirmed that I do not have a "Minimum Stock" set for these items in the Warehouse. The waiter appears to request the warehouse to store the items, initiating a loop where food is cooked, taken to the restaurant then taken to be stored in the warehouse, and never eaten unless I actively take them back to the restaurants myself.

### Expected Behavior

Food listed on the Restaurant menu should remain in the Restaurant inventory to be served to colonists. If the inventory is overflowing and there is need for items to be stored away, the system should keep at least one stack of each item in the restaurant menu in the storage. And when the restaurant needs food, they should be able to first check the warehouse for stored food first, before the request going to the chef.

### Reproduction Steps

Reproduction Steps:
First it is important to mention this issue doesn't happen quickly. It takes time, but it always happens.
- Have a functional Restaurant and Warehouse.
- Ensure the Restaurant has cooked food on the menu (e.g., minecolonies food).
- Over time, Couriers will start taking the food from the Restaurant to the Warehouse.
- Watch as the Restaurant empties out and colonists begin to starve while food sits in the Warehouse.

### Logs

https://pastebin.com/PbV5hJEV

### Anything else?

Image

Image

Image

Image

### Footer

---
 ## Viewers

Multiplication-golf:
Closes #11421

Sorry wrong issue

- Blocks couriers from removing food from the restaurant's inventory
- other items can still be removed

If there is a better way to get the building type from the targetBuilding, please let me know.

- I am having trouble locating the file that contains the system that creates the PICKUP requests for the delivery man
- I plan on moving the logic to that file and refactoring it to fit their

I cannot test this till Friday, but it successfully compiled, so it should work fine.

Review please


Thodor12:
I keep clicking the wrong button, excuse me

src/main/java/com/minecolonies/core/entity/ai/workers/service/EntityAIWorkDeliveryman.java
```diff
@@ -252,8 +252,19 @@ private boolean pickupFromBuilding(@NotNull final IBuilding targetBuilding)
         {
             return false;
         }
+        final ItemStack activeStack;
+
+        if (targetBuilding.getBuildingDisplayName().equalsIgnoreCase("restaurant")) {
```
Thodor12: This kind of filtering logic does not go directly into the courier AI, this is controlled by buildings.

Thodor12: Using the building its display name is HIGHLY volatile. What if we rename the building to something that is no longer called a restaurant, then this logic breaks.

Thodor12 (changes requested):
Ok, few questions, what caused you to make this change, because food is already kept by the restaurant?

The logic introduced is also fairly flawed and inhibits existing food keeping rules already set on the restaurant, so all in all I think this is a bad idea and bad way of going about it.

Thodor12:
I think this definitely isn't the solution, looking at the linked issue it's not even clear yet where the bottleneck in the system is. So we cannot provide a fix for this until we know what the bottleneck is in the first place.

This way at least is definitely not the right way, the courier should not be blocking pickups like this.

PR #11640, opened 2026-04-30
Add optional bottom-up clearing for builders

GeneraBlack:
Closes #
Closes #

-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingModules.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingModules.java
@@ -414,6 +414,7 @@ public class BuildingModules
         .with(BuildingBuilder.MODE, new StringSetting(BuildingBuilder.AUTO_SETTING, BuildingBuilder.MANUAL_SETTING))
         .with(AbstractCraftingBuildingModule.RECIPE_MODE, new CrafterRecipeSetting())
         .with(BuildingBuilder.BUILDING_MODE, new BuilderModeSetting())
+        .with(BuildingBuilder.CLEAR_MODE, new StringSetting(BuildingBuilder.CLEAR_TOP_DOWN_SETTING, BuildingBuilder.CLEAR_BOTTOM_UP_SETTING))
         .with(AbstractBuilding.USE_SHEARS, new BoolSetting(false))
         .with(BuildingMiner.FILL_BLOCK, new BlockSetting((BlockItem) Items.DIRT)), () -> SettingsModuleView::new);
     public static final BuildingEntry.ModuleProducer<SimpleCraftingModule,CraftingModuleView> BUILDER_CRAFT                   =
--- a/src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingResourcesModule.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/modules/BuildingResourcesModule.java
@@ -176,7 +176,8 @@ public class BuildingResourcesModule extends AbstractBuildingModule implements I
     {
         return (buckets.isEmpty()
             || ((AbstractBuildingStructureBuilder) building).getProgress() == null
-            || ((AbstractBuildingStructureBuilder) building).getProgress().getB() == BuildingProgressStage.CLEAR) ? null : buckets.getFirst();
+            || ((AbstractBuildingStructureBuilder) building).getProgress().getB() == BuildingProgressStage.CLEAR
+            || ((AbstractBuildingStructureBuilder) building).getProgress().getB() == BuildingProgressStage.CLEAR_BOTTOM_UP) ? null : buckets.getFirst();
     }
 
     /**
--- a/src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingBuilder.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingBuilder.java
@@ -4,6 +4,7 @@ import com.ldtteam.blockui.views.BOWindow;
 import com.minecolonies.api.colony.ICitizenData;
 import com.minecolonies.api.colony.IColony;
 import com.minecolonies.api.colony.IColonyView;
+import com.minecolonies.api.colony.buildings.IBuilding;
 import com.minecolonies.api.colony.buildings.modules.settings.ISettingKey;
 import com.minecolonies.api.colony.workorders.IBuilderWorkOrder;
 import com.minecolonies.api.colony.workorders.IServerWorkOrder;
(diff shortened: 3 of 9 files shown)


GeneraBlack:
Title:
Add an optional bottom-up clearing pass for builders

Description:
## Summary

This change adds a second clearing mode for builders: an optional bottom-up clearing pass before the normal build flow starts.

When enabled, the builder will:
- clear the area from bottom to top first
- then run the normal top-down clearing pass
- continue with the existing build stages afterwards

This preserves the original behavior for existing colonies because the default mode remains top-to-bottom.

## Why

The current builder clearing flow only works from top to bottom.

That works for many cases, but it is not ideal when preparing areas that contain blocks affected by gravity or similar world physics. In those cases, clearing from the bottom up first can be useful, but a final top-down pass is still required to remove blocks that may fall or update after the initial pass.

## What changed

- Added a new builder setting for clearing strategy:
  - Top to Bottom
  - Bottom to Top
- Added a dedicated `CLEAR_BOTTOM_UP` progress stage for new builder work orders
- Wired the builder AI to optionally prepend a bottom-up clearing pass before the existing clear/build stages
- Kept the existing top-down clear pass after the bottom-up pass
- Updated builder-specific clearing logic so blocks that already match the blueprint are skipped instead of being removed unnecessarily
- Updated related builder flow checks so the new clearing stage is treated consistently by:
  - resource/bucket access
  - assistant hammer placement gating
  - structure replacement rules

## Behavior notes

- Default behavior is unchanged
- The new mode only affects new builder clearing work
- Already correct blocks are now preserved during builder clearing instead of being removed just because the builder is in a clear phase

## Validation

- Compiles successfully with `./gradlew compileJava`
- No automated gameplay test was added
- Recommended manual verification:
  - normal top-down clearing still works as before
  - bottom-up mode clears from below first, then completes a top-down cleanup pass
  - correctly matching blueprint blocks are preserved during clearing
  - gravity-affected blocks such as sand or gravel are fully cleaned up after the second pass

Talyda:
Can I ask (and I am not a coder, so sorry if this is obvious) how does it make sense to clear from bottom up? Don't you risk the builder clearing blocks that are holding up gravity blocks, and then getting buried by them? I thought that was why it is top down?

Shadizar:
also excitement like "I just opened the bottom of a lake or lava pool" would be a condideration

GeneraBlack:
> Can I ask (and I am not a coder, so sorry if this is obvious) how does it make sense to clear from bottom up? Don't you risk the builder clearing blocks that are holding up gravity blocks, and then getting buried by them? I thought that was why it is top down?



> also excitement like "I just opened the bottom of a lake or lava pool" would be a condideration

to those concerns, that already does make a mess with the current top down system, the builders do not remove water or lava during the clearing phase, also gravel falls down if its a underground sytle.

The Bottom Up methode does 2 things, it clears first from bottom to top and than again from top to bottom to remove blocks with gravity like sand or gravel.

Its mainly designed to solve the teleportation issue if players build under ground, where the builder creates a cavity thats not accessible and than teleports into it (most likely filled with hostile mobs)

It makes it easyer for the builder to do underground styles.

Also its totaly the players choice. The Builder hut got 1 more option now, default is still top to bottom.

Talyda:
> to those concerns, that already does make a mess with the current top down system, the builders do not remove water or lava during the clearing phase, also gravel falls down if its a underground sytle.

So this still doesn't make sense to me. Yes, gravel falls down, but it doesn't fall down into a big empty space they are standing in. It very rarely, if ever, falls on them. If you have them clearing out the space from underneath, then they are standing in that space underneath, and anything they make fall down (gravel, sand, lava whatever) falls ON them.

It doesn't actually stop them making a cavity that's not accessible at all, it just moves the cavity from the top of the build area to the bottom of the build area  (unless you have made an opening right next to where they are going to be digging out, and know for sure that they are going to dig closest to that area first)

I'm probably not understanding something, but no matter how I try and see this in my head, it just doesn't seem like it would actually improve the builder. All I see is them getting buried in gravel LOL!

GeneraBlack:
> > to those concerns, that already does make a mess with the current top down system, the builders do not remove water or lava during the clearing phase, also gravel falls down if its a underground sytle.
> 
> So this still doesn't make sense to me. Yes, gravel falls down, but it doesn't fall down into a big empty space they are standing in. It very rarely, if ever, falls on them. If you have them clearing out the space from underneath, then they are standing in that space underneath, and anything they make fall down (gravel, sand, lava whatever) falls ON them.
> 
> It doesn't actually stop them making a cavity that's not accessible at all, it just moves the cavity from the top of the build area to the bottom of the build area (unless you have made an opening right next to where they are going to be digging out, and know for sure that they are going to dig closest to that area first)
> 
> I'm probably not understanding something, but no matter how I try and see this in my head, it just doesn't seem like it would actually improve the builder. All I see is them getting buried in gravel LOL!

I am building a lot underground, so there is no perfect solution.

What i repeditly see with the top down digging is.
Builder teleports into the void he digged and dies there.
Lava/Water spreading excessivly bause the spread in the cavity worsens.
No way to light up the digged out space cause builder removes the torches layer by layer (cant get it mobsafe)

yes gravel and sand are a problem but they continue to be in boath styles of clearing out the area without double checking. 

In Styles like https://ladyshatana.wixsite.com/mcguides/cavern you do have the opening from which they can start.

GeneraBlack:
for example i lost like 3 builders in this area due to hostile spawning (the path at the bottom was precleared by me)

Talyda:
Yeah, I've used cavern a lot =) You need to babysit the builders until they have finished. Price of playing a more difficult style!
What I do personally is put on builders goggles, and go in and place the lights in the top area for them. I don't think I ever had them die from gravel falling on their heads though in top down digging - cos they simply are not standing anywhere that gravel can fall on them, cos they already cleared out above wherever they are standing!

GeneraBlack:
come on we are discussing about an alternative non default mode for the builder.
The MR is absolutely minimal.

Thodor12:
Have you actually tested this works, how can they navigate to the top of the area if there's no path to go up? At least with top down they can emergency teleport to the top to clear out the rest. But bottom top then there's nowhere to teleport

Raycoms:
This probably doesn't work well in a lot of cases as a builder mode. The modes have to be the same y direction because certain blocks like torches, lanterns, etc rely on this order to not break due to a lack of support.
We do full blocks in one dir, and deco blocks in the other direction for example. Similar with liquid placement

https://proton.me/mail/home

GeneraBlack:
> Have you actually tested this works, how can they navigate to the top of the area if there's no path to go up? At least with top down they can emergency teleport to the top to clear out the rest. But bottom top then there's nowhere to teleport



> Have you actually tested this works, how can they navigate to the top of the area if there's no path to go up? At least with top down they can emergency teleport to the top to clear out the rest. But bottom top then there's nowhere to teleport

Thats one of the Drawbacks if the Structure is to tall you have to give him support, after the bottom up pass he will do a normal pass again to remove supports or grafity blocks.

Changes when to babysit the builder

> This probably doesn't work well in a lot of cases as a builder mode. The modes have to be the same y direction because certain blocks like torches, lanterns, etc rely on this order to not break due to a lack of support. We do full blocks in one dir, and deco blocks in the other direction for example. Similar with liquid placement https://proton.me/mail/home


Building itself didnt change.

In the code you can see that bottom up just adds  itself as a precurour to the normal buildchain.


final BuildingProgressStage[] stages = useBottomUpClearing
                                    ? new BuildingProgressStage[] {CLEAR_BOTTOM_UP, CLEAR, BUILD_SOLID, WEAK_SOLID, CLEAR_WATER, CLEAR_NON_SOLIDS, DECORATE, SPAWN}
                                    : new BuildingProgressStage[] {CLEAR, BUILD_SOLID, WEAK_SOLID, CLEAR_WATER, CLEAR_NON_SOLIDS, DECORATE, SPAWN};

someaddons:
one drawback this ofc does have is that it'll increase total build time for the extra iteration

GeneraBlack:
> one drawback this ofc does have is that it'll increase total build time for the extra iteration

yes definitly, boath methodes have there benefits and drawbacks, its totaly in the control of the player what they prefere.

Gues the key argument is (that i should have leaded with) give the player an other way of freedom and let them choose what they prefere.
And i gues the freedom of choice and how to play your colony and in what stlye is the biggest feature in minecolonies.

someaddons:
the important part would be that the player understands which to use in which situation/style

PR #11641, opened 2026-04-30
fix: citizen AI and combat improvements

ZeOs360:
getCapability - LazyOptional.of() was being called on every invocation. It has been cached as a field and is now invalidated in setRemoved().
callForHelp - this.blockPosition() was being called on every comparison in the sort comparator. It is now pre-fetched as selfPos.
checkIfValidDamageSource - Although addPlayer() is server-side, it was also running on the client side. !level.isClientSide has been added.
BasicStateMachine - DateTimeFormatter was being created separately for each instance. It has been made static final.

Closes #
Closes #

-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/entity/ai/combat/threat/ThreatTable.java
+++ b/src/main/java/com/minecolonies/api/entity/ai/combat/threat/ThreatTable.java
@@ -65,6 +65,12 @@ public class ThreatTable<T extends LivingEntity & IThreatTableEntity>
         {
             return;
         }
+
+        final ThreatTableEntry currentTarget = (currentTargetIndex >= 0 && currentTargetIndex < threatList.size())
+            ? threatList.get(currentTargetIndex) : null;
+        threatList.removeIf(entry -> !entry.getEntity().isAlive());
+        currentTargetIndex = currentTarget != null ? Math.max(0, threatList.indexOf(currentTarget)) : 0;
+
         ThreatTableEntry threatTableEntry = null;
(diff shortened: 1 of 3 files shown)


Raycoms:
Did you use AI to aid you in this PR?

src/main/java/com/minecolonies/api/entity/ai/combat/threat/ThreatTable.java
```diff
@@ -65,6 +65,12 @@ public void addThreat(final LivingEntity attacker, final int additionalThreat)
         {
             return;
         }
+
+        final ThreatTableEntry currentTarget = (currentTargetIndex >= 0 && currentTargetIndex < threatList.size())
+            ? threatList.get(currentTargetIndex) : null;
+        threatList.removeIf(entry -> !entry.getEntity().isAlive());
+        currentTargetIndex = currentTarget != null ? Math.max(0, threatList.indexOf(currentTarget)) : 0;
```
someaddons: it is intentionally not checking the full list on threat being added as threat added occurs very frequently, instead sanity checks like is alive are done when retrieving a new target.

ZeOs360: thanks for the context I'll revert the changes.

ZeOs360:
> Did you use AI to aid you in this PR?

I used AI for analysis. The ThreatTable change was a mistake I didn't verify it against the existing design intent.

PR #11645, opened 2026-05-01
Add Enchanter scroll quests (5 quests) (1.21.1)

rainweaver76:
Ray request that I finally had the brain power to complete:

Quest 1: "Things That Go Bump in the Night" (bumpinthenight.json)
Scroll: Spatial Guard Reinforcements Scroll | Requires: Enchanter's Tower level 2+ + Enchanter citizen
Flow: Opening dialogue → 4 deliveries (scroll_tp, lapis x5, ender pearl, paper) → "I'll work tonight, come back tomorrow!" → chained squirrels-reveal dialogue + rewards
Rewards: 3x scroll_guard_help, 5x lapis_lazuli, 1x ender_pearl, 1x paper, enchanter happiness boost

Quest 2: "A Promise to Keep" (apromisetokeep.json)
Scroll: Teleport Scroll | Requires: Enchanter L1 + Knight citizen
Flow: Enchanter explains the guard's dilemma → deliver 3x paper + 1x compass → enchanter sends scroll ahead to guard → find the guard (target 1, nexttask icon) → guard reveals another guard covered their shift, mom sent a pie!
Reward: 3x scroll_tp + 1x pumpkin_pie + guard happiness

Quest 3: "A Day in the Field" (adayinthefield.json)
Scroll: Group Teleport Scroll | Requires: Enchanter L2
Flow: Enchanter wants to take scholars herb-gathering but needs a Group Scroll → deliver 3x scroll_tp → "come back tomorrow" → enchanter: "amazing haul, everyone home safe, here's the scroll AND your materials back!"
Reward: 1x scroll_area_tp + 3x scroll_tp (all yours back) + happiness

Quest 4: "Where's the Builder?" (wheresthebuilder.json)
Scroll: Worker-Where-Are-You Scroll | Requires: Enchanter L3 + Builder citizen
The return mechanic is used here for a genuine wait: objective 4 has two options:
• "Take your time, I'll check back!" → return (citizen keeps nexttask icon, player leaves)
• "I'm back - is the scroll ready?" → advances (only visible on the return visit)
Flow: Enchanter overheard tavernkeeper gossip → deliver 3x scroll_tp + 6x glowstone_dust + 2x paper → come back → enchanter dramatically "uses" scroll → find builder (nexttask icon) → builder is napping in the warehouse basement, begs for secrecy → back to enchanter → enchanter laughs
Reward: 5x scroll_highlight + all ingredients back + both enchanter & builder happiness

Quest 5: "Ancient Magic" (ancientmagic.json)
Scroll: Sacred Scroll of Regeneration | Requires: Hospital L1 + Enchanter + Healer citizen
The return option on the opening dialogue means if the player doesn't have the scroll yet, they can come back later without canceling the quest. The enchanter keeps their icon.
Flow: Enchanter senses ancient magic → player says "I think I have it" → deliver 1x scroll_buff directly to the doctor (target 1) → doctor breaks down with gratitude
Reward: 3x Luck potions (given as 3 separate entries since potions are unstackable) + doctor happiness for 14 days (doubled, because they were overwhelmed)


kezmodius:
I just want to let you know that some modpacks (ATM for example) remove the regen scroll from loot tables, so the final quest might be undoable in those.

rainweaver76:
Bugger, I had added a check for that, I must have nipped it when I was fixing typos. Let me look at that again and make sure I get it back in. That quest should only trigger as available once the player has found the scroll.

rainweaver76:
Ah crud no, I remember now why I removed it. Like I said, need coffee. There's no advancement trigger in MCol for finding a scroll in the wild, so there was no way to cleanly add that as a requirement without adding a new class. It would also only count per *colony* and not per player, so if anyone who was an officer in any way found one it would trigger for everyone else, and then only if they were online at the time. And I could just imagine the confusion that would cause. I also wanted to avoid adding any new classes and keep it strictly to the quest jsons themselves so I scrapped the trigger idea. This may just be a case of "I have not come across anything like that yet" having to be the player's final answer :(

rainweaver76:
I realize now that the modern baker skins were smooshed into this PR so I am going to close this one and open two separate ones since that makes more sense.

PR #11648, opened 2026-05-02
Fix duplication glitch with Sable (Create : Aeronautics)

Daudeuf:
Implement Clearable interface to some blocks to made it work with Sable

### DIFF
--- a/src/main/java/com/minecolonies/core/tileentities/TileEntityGrave.java
+++ b/src/main/java/com/minecolonies/core/tileentities/TileEntityGrave.java
@@ -20,6 +20,7 @@ import net.minecraft.network.FriendlyByteBuf;
 import net.minecraft.network.RegistryFriendlyByteBuf;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.protocol.game.ClientboundBlockEntityDataPacket;
+import net.minecraft.world.Clearable;
 import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.inventory.AbstractContainerMenu;
@@ -40,7 +41,7 @@ import static com.minecolonies.api.util.constant.NbtTagConstants.TAG_DECAY_TIMER
 /**
  * Tile entity for the graves.
  */
-public class TileEntityGrave extends AbstractTileEntityGrave
+public class TileEntityGrave extends AbstractTileEntityGrave implements Clearable
 {
     /**
      * The content of the chest.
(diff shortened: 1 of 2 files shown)


someaddons:
its pretty unrealistic that each and every mod does implement this tbh, sable needs to handle it better

Thodor12:
Under which instance is this actually needed? Are you really going to turn buildings into planes? That feels like a recipe for disaster to begin with

Daudeuf:
I know it seems unrealistic and should be handled differently, but currently putting a rack and then transforming it into a vehicle allows objects to be duplicated, which is problematic in the context of a project I am designing where I don't trust everyone not to use these bugs.

Thodor12:
But why would you put a rack into one of these sublevels? Racks are intended solely to be used in schematics. In fact if you put them into one of these sublevels, they no longer function at all. Because a broken rack gets de-registered from a building, so if you were to reinstantiate the sublevel back into the world, the rack already is nonfunctional at that point

MotionlessTrain:
Wouldn’t a rack just work as a chest if not used within a building?

Thodor12:
Yeah but then just use a chest, racks don't need this logic merely in support of a strange physics addon. I see no use-case here at all

MotionlessTrain:
Racks should not have a duplication glitch
If this is publicly known, that can be enough reason for modpack makers to stop including minecolonies in modpacks. Simply because people can use racks instead of chests, just because of the dupe bug

Thodor12:
No, I agree with Sam that sable should fix this. This is why the IItemHandler capability was invented. We expose item handlers so third party mods can have a common way of interacting with blocks, they should be using that to clear the block.

Not to mention that Sable has a godforsaken amount of incompatibilities already

Thodor12:
Honestly I think if Sable isn't improved quickly then Create Aeronautics is going to die really quickly because of the enormous washing list of incompatible mods it has

Raycoms:
I think it's fine to implement those methods, doesn't hurt?

MotionlessTrain:
Same here

Raycoms (approved):
Sable should definitely add support for ItemHandlers properly.

Daudeuf:
I think you're right, Sable does implement more than one method to clear blocks, however the Clearable interface is a basic Minecraft interface, all mods should implement it correctly, it's not something proprietary.
But I think leaving a duplication bug, even temporarily, is not an option.

Fernyam:
Sable developers are already working on a solution to this problem.

Daudeuf:
Hello, I was wondering if you planned to fix this duplication bug (even temporarily) or not?

Raycoms:
yes, we plan to merge this

someaddons (approved):

Raycoms:
Can this be ported to 1.20.1?

Daudeuf:
Not usefull for Create Aeronautics, but maybe for Valkyrien Skies

PR #11652, opened 2026-05-04
Improve builder lava safety and pathfinding on 1.21

GeneraBlack:
## Summary
This PR ports and stabilizes our builder and pathing lava-safety changes for the 1.21 branch.

## Changes
- add worker and structure AI debug output for in-game inspection
- improve builder retreat and safe-stand handling around lava hazards
- avoid repeated retries on the same blocked hazardous target
- fix central pathfinding so flowing lava is treated as unsafe
- add a one-block lava safety margin in central pathfinding

## Validation
- 1.21 build passes
- 1.21 test task passes
- in-game behavior verified around flowing lava and builder work positions

## Review note
The main intended behavior change is stricter rejection of paths directly adjacent to lava.

### DIFF
--- a/src/main/java/com/minecolonies/core/debug/messages/QueryCitizenAIHistoryMessage.java
+++ b/src/main/java/com/minecolonies/core/debug/messages/QueryCitizenAIHistoryMessage.java
@@ -6,6 +6,7 @@ import com.minecolonies.api.colony.ICitizenDataView;
 import com.minecolonies.api.colony.IColony;
 import com.minecolonies.api.util.constant.Constants;
 import com.minecolonies.core.debug.DebugPlayerManager;
+import com.minecolonies.core.entity.ai.workers.AbstractEntityAIBasic;
 import com.minecolonies.core.entity.citizen.EntityCitizen;
 import com.minecolonies.core.network.messages.server.AbstractColonyServerMessage;
 import net.minecraft.network.RegistryFriendlyByteBuf;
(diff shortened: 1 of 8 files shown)


GeneraBlack:
This PR also adds builder-side safeguards for unsafe mining positions. Builders now avoid mining setups that could trigger falling blocks above the target, and they will not mine the block beneath their own position when doing so would create an unsafe fall. When possible, they first try to reposition to a safer adjacent stand.

someaddons:
I dont think we should be doing this as the lava scanning during pathfinding adds additional performance overhead and we do not want citizens to be always fully safe. They're less intended like robots and more like humans who do make mistakes, a builder running into a fire from lava occasionally is calculated in, which is why we also have AI specifically for environmental damage, where they walk away from the danger

Raycoms:
I think some of the debug things are nice, and if there is any bug where the builder could mine a block and fall that should not happen (there should be checks in place for this, if not we have to fix this).

For the general pathing it is a large overhead for the lava edge case at the moment. If anything I think we could mark on detecting lava in the pathfinder the nearby nodes (if exist) as "lava nearby" and give them a higher cost.

GeneraBlack:
> I think some of the debug things are nice, and if there is any bug where the builder could mine a block and fall that should not happen (there should be checks in place for this, if not we have to fix this).
> 
> For the general pathing it is a large overhead for the lava edge case at the moment. If anything I think we could mark on detecting lava in the pathfinder the nearby nodes (if exist) as "lava nearby" and give them a higher cost.

yea the builder doesnt check if he would fall if he removes a block seen it multiple times today XD quiet funny to watch how he runs up the stairs just to fall again XD

Raycoms:
> > I think some of the debug things are nice, and if there is any bug where the builder could mine a block and fall that should not happen (there should be checks in place for this, if not we have to fix this).
> > For the general pathing it is a large overhead for the lava edge case at the moment. If anything I think we could mark on detecting lava in the pathfinder the nearby nodes (if exist) as "lava nearby" and give them a higher cost.
> 
> yea the builder doesnt check if he would fall if he removes a block seen it multiple times today XD quiet funny to watch how he runs up the stairs just to fall again XD

then we definitely need to make the mining steps more sensible. Do you mean mining during the clearing step, or do you mean during deco? (Placing air in the deco step)

GeneraBlack:
> > > I think some of the debug things are nice, and if there is any bug where the builder could mine a block and fall that should not happen (there should be checks in place for this, if not we have to fix this).
> > > For the general pathing it is a large overhead for the lava edge case at the moment. If anything I think we could mark on detecting lava in the pathfinder the nearby nodes (if exist) as "lava nearby" and give them a higher cost.
> > 
> > 
> > yea the builder doesnt check if he would fall if he removes a block seen it multiple times today XD quiet funny to watch how he runs up the stairs just to fall again XD
> 
> then we definitely need to make the mining steps more sensible. Do you mean mining during the clearing step, or do you mean during deco? (Placing air in the deco step)

during the clearing step the miner doesnt check if he is standing on the block he is mining, that can be prevented by checking his current position and the next block position to clear if x and z of the block match with the x z  positioon of the miner you can simply return invalid working position.

That basicly would prevent the miner from dropping or getting burried by gravel. My solution was a bit "overenginered"
(Forbids standing on top or under a block to mine)

the Y cordinate is basicly irrelevant in that calculation.

Raycoms:
I pushed a recent fix for the save standing, thank you for this PR. I will close it in favor of those.

PR #11675, opened 2026-05-19
Create better workflow for managing tested status

Thodor12:
# Checklist

- Improve the untested label workflow by creating a status check instead. Labels will not be used anymore and instead a required status check will block the PR from being merged until the author comments the exact phrase: `I confirm my changes are tested` (the content is trimmed in case they make a fault).
- The status check is automatically set upon opening and upon every push, and this way it no longer requires contributors to remove the labels
- And lastly there's an actual verifiable check at the end of the PR that blocks from merging (provided Ray doesn't bypass it)

Review please


Raycoms:

uecasm:
Why all the tech shenanigans when we could just use a simple checkbox again?  It's still up to Ray whether to merge it either way.  And requiring additional comments etc just causes email spam.

PR #11690, opened 2026-05-29
Allow an option to light up a scarecrow with a lantern

Thodor12:
# Checklist

# Related issues
Closes [discord request](https://discord.com/channels/472875599422291968/1508099338650521803)

- Blockstate property to control whether the lantern is active or not, rather than a whole container implementation
- Due to the fact this is a blockstate property schematic designers can choose to turn this option on by default
- Proper neighbouring handling like doors to make sure the prop is synced through both blocks
- Breaking the block drops the lantern out
- Lantern is visually rendered when equipped with some extra render code in the TileEntityScarecrowRenderer
- Placement handler adds extra requirement to provide a lantern if the blockstate property for lantern is true

It is not possible to remove the lantern from the field unless you break the block, because I couldn't think of a nice way to right click the lantern out considering the block already has a GUI on it.

See it in action here:
https://gyazo.com/3e3abdf1a088e4dc05a79763835fdfac

Review please


someaddons (approved):
nice addition

PR #11697, opened 2026-06-01
Abandoned colony improvements

Thodor12:
# Checklist

# Related issues
Closes [discord issue](https://discord.com/channels/139070364159311872/653530684748595211/1510411735004348486) (other styles cannot be added to worldgen)

- Makes it possible for addons to register worldgen styles into the primary minecolonies empty_colony structure set, to share the same placement pool.
  - No code needed for addons, besides 1 singular method invocation:
```java
@Mod(MOD_ID)
public class StyleColonies
{
    public static final String MOD_ID = "stylecolonies";

    public StyleColonies()
    {
        ColonyStructureRegistry.register(new ResourceLocation(MOD_ID, "steampunk_colony"));
    }
}
```
  - For the rest you just define the structure files as normal within the mod, and the generation of the structure will automatically be picked up by Minecolonies
- Rewrite the empty_colony structure set into a dynamic pack resource, acting almost as an in memory datapack that resolves the structure set, whilst also allowing modification by addons to inject other worldgen styles.
- With this change it was suddenly also possible to control abandoned colony spawning via a common config (not server config, loaded too late), so I added this too so we no longer have to rely on external datapacks for this.

Review please


src/main/java/com/minecolonies/core/structures/ColonyStructureRegistry.java
```diff
@@ -0,0 +1,223 @@
...
+     */
+    public static void register(final ResourceLocation structure)
+    {
+        ADDON_STRUCTURES.add(structure);
+    }
+
+    /**
+     * Registers a synthetic in-memory datapack that serves the merged {@code minecolonies:empty_colony}
+     * structure set JSON, combining all built-in and addon structures into a single placement grid.
+     * The pack is placed at {@link Pack.Position#TOP} to override any other pack that may provide the same file.
```
someaddons: can you still override it with a normal datapack then?

Thodor12: I don't strictly know the loading order what it'd do now, but worst case you can turn it off via config, which means this internal one won't be activated

MotionlessTrain: That might be important, before we get a lot of questions about people wondering why the datapack suddenly stopped working

someaddons: ye we want to keep the datapacks working, as they're not only used for altering the content of the structure set, but also for adjusting the spacing/seperation between the structures

Thodor12: In that case I feel like it's better I add more config options to control those placement patterns, no?

Thodor12:
```java
            placement.addProperty("salt", 1225566777);
            placement.addProperty("spacing", 95);
            placement.addProperty("separation", 45);
            placement.addProperty("type", "minecraft:random_spread");
```
These 4, but primarily just the middle 2, right

someaddons: idk, removing the datapack back to a config def feels like a step backward

Thodor12: Why, config means we have a first class way of turning it off or on. That way we don't have to rely on an external datapack to manage it.
Right now we're effectively putting downloads into someone else their hands because we have no way of turning this feature off. With this we can

MotionlessTrain: This is completely different from other mods and vanilla, however
That is all datapack based, since 1.18.2. If we were to put it in configs, we would confuse modpack makers, as they can’t datapack their world gen anymore in a uniform way

Thodor12: Of course they can datapack it, no clue what you're on about. For all intents and purposes this IS a datapack installed in the datapacks folder. Which is coded to serve a singular file in a dynamic fashion.

If they want to bring their own worldgen they can still register structure sets themselves, nothing is stopping them from doing so. This file however just always comes last, but may be turned off by the config. It's a slightly different flow than what some people may be used to.

And on top of that, unless they are EXPLICITLY trying to override this one singular file, anything else they modify will work just fine.

uecasm: Modpack authors may indeed want to override the empty_colony.json, to make themed worlds where some styles cannot generate or have different spread parameters.  It feels like fighting the Minecraft design to replicate these settings in a config file instead, which is what it sounds like you're doing.

And it becomes harder to override the file in the normal way because the default file is invisible (generated in code rather than existing in github).

Not sure if worldgen supports tags, but something like that (where it merges by default but can be explicitly replaced) might be a better way to go.

A single config "disable all worldgen" would be good, though, so people don't have to hunt down a special datapack for that.  IIRC we used to have that but it was removed because the config didn't load soon enough.

Thodor12: I did put the config in common, not server, server couldn't be loaded quick enough, common could be loaded quick enough on the other hand

MotionlessTrain: > It's a slightly different flow than what some people may be used to.

That is exactly the issue. We are now undermining Minecraft’s design to have worldgen fully datapackable, and so this will lead to questions on discord, and/or frustration, as people don’t expect needing to disable something in configs for datapacks to work

> A single config "disable all worldgen" would be good, though, so people don't have to hunt down a special datapack for that. IIRC we used to have that but it was removed because the config didn't load soon enough.

Indeed, with the move to datapacks in 1.18.2, the config couldn’t work anymore, so it was removed

Thodor12: > Not sure if worldgen supports tags, but something like that (where it merges by default but can be explicitly replaced) might be a better way to go.

I did consider allowing the API to be done via different datapack esque files where you could choose which structures to include or not. But frankly if I turn the config off and then supply my own empty_colony file you'll have the same effect.

So in my eyes, is it worth it creating an entirely different JSON structure elsewhere and merging those together, no

Thodor12: @MotionlessTrain @someaddons I moved it to position BOTTOM, so it should appear first, but given its unique filename it should still work.
This still allows it to be overridden by datapacks if so desired, but aside from limiting which structures you want yourself I no longer see a point of doing this.
I am leaving the config option in naturally since that's a more straightforward way of disabling it, and a less annoying one imho.

Raycoms:
So I thought long and hard about this. I think the best approach is to minimize the amount of code addons need. It's okay if they need code, but we should just minimize it. (E.g. allow them to extend our classes and call utility methods in our class, etc). So that they just need the Mod.java + a few method calls with a string parameter in it. Yes, it would be a completely new type of structure (not part of our set) but that's okay enough I'd say.

Thodor12:
> So I thought long and hard about this. I think the best approach is to minimize the amount of code addons need. It's okay if they need code, but we should just minimize it. (E.g. allow them to extend our classes and call utility methods in our class, etc). So that they just need the Mod.java + a few method calls with a string parameter in it. Yes, it would be a completely new type of structure (not part of our set) but that's okay enough I'd say.

So that means this change is ok, given it's only a singular method call?
(Plus it doesn't impede on vanilla datapacks regardless of what)

Issue #11699, opened 2026-06-03
builders not breaking blocks


TreeDurr:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Are you using the latest MineColonies Version?

- [x] I am running the latest **beta/release** version of MineColonies for my Minecraft version.
I am also running the latest versions of other mods that are part of my problem.

### Did you check on the Wiki? or ask on Discord?

- [x] I checked the MineColonies Wiki and made sure my issue is not covered there. Or I was sent from discord to open an issue here.

### What were you playing at the time? Were you able to reproduce it in both settings?

- [x] Single Player
- [ ] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1327-1.21.1-snapshot

### Structurize Version

1.0.830-1.21.1

### Current Behavior

the builder is currently unable to build the builders hut because it cant mine blocks

### Expected Behavior

i expected the builder to start building the builders hut

### Reproduction Steps

1. make a new town (im using pagoda)
2. put down a builders hut
3. assign building to builders hut, wait until a civilian takes up the job
4. itll throw the error

### Logs

https://gist.github.com/TreeDurr/5627c49060e5d0c5aa5d1a00f92e7a5e

### Footer

---
 ## Viewers
  *  Add a 👍 reaction to the bug report if you are also affected. This helps the bug report become more visible without cluttering the comments.
  *  ⚠️ Only add a comment if you have new insights or background information not already mentioned. Off-topic or "+1" comments will be deleted to keep the discussion focused.

Talyda:
Need the whole latest.log please!

And also just a note to anyone reading, it happened in minecolonies ending in 319 as well

muellerjp:
short question, which types of blocks does the builder not mine?

Talyda:
> short question, which types of blocks does the builder not mine?

Was dirt - they broke it themselves, then the builder errored out again on dirt 
(sorry to butt in, but figured getting you the info quicker was better LOL!)

muellerjp:
freshly tested right now with the following setup:

<img width="3039" height="1185" alt="Image" src="https://github.com/user-attachments/assets/41f5ae5e-55fb-4da1-a9fb-7a8571b46614" />

the builder, using pagoda style, successfully broke every block with a wooden tool. Only observation i could make was that the mining animation did not always show, but the blocks were broken non the less.

short fotomontage of the screenshots

[...]

PR #11700, opened 2026-06-03
 Porting 1.21 Item Compat API to main

muellerjp:
feat(compat): porting changes from  version/1.21 to main. see #11647 for details

# Checklist

i have used AI to summarize the differences between 1.21 and main, so that i did not have to read 777 (at the time) commits.

# Related issues
Closes #
Closes #

-
-

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/compatibility/Compatibility.java
+++ b/src/main/java/com/minecolonies/api/compatibility/Compatibility.java
@@ -4,6 +4,7 @@ import com.minecolonies.api.compatibility.dynamictrees.DynamicTreeProxy;
 import com.minecolonies.api.compatibility.resourcefulbees.IBeehiveCompat;
 import com.minecolonies.api.compatibility.tinkers.SlimeTreeProxy;
 import com.minecolonies.api.compatibility.tinkers.TinkersToolProxy;
+import com.minecolonies.api.crafting.ItemStorage;
 import com.minecolonies.api.equipment.registry.EquipmentTypeEntry;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.NonNullList;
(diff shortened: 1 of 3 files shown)


Raycoms:
Could this https://github.com/ldtteam/minecolonies/issues/11699 be related to the changes btw?

muellerjp:
i'll take a look at it right now

Talyda:
> Could this #11699 be related to the changes btw?

I actually initially thought that, but they had the same error in the version ending in 319 as well (unless some parts of changes happened earlier than I thought they did!)

muellerjp:
Short update here from the bug ticket raycoms linked:

Tested the claimes in the bug ticket and investigated the partial log that was available.
Could neither reproduce on a fresh install. The log points to a function that these changes here never touched.
In addition based on the type of error chances are that another mod or mixin is at work in the background.
Clarifying requires a full log, as requested by talyda in the bug ticket.

Raycoms:
This one might be related to missing some saveguards: https://github.com/ldtteam/minecolonies/issues/11703

Thodor12:
This isn't really this PR its fault Ray. Those more vanilla shields are implemented in SUCH a dumb way. They rely on server configuration to calculate the max damage on their shields.

That is honestly such a stupid idea, because server configuration is calculated post registry instantiation, so they should be doing this in common config.

muellerjp:
Short scope question, where do you guys draw the line between "minecolonies" issue and "other mod" issue with modpacks?
Based on my log analytics i agree with Thodor that the fix should be made on the side of More Vanilla Shields.

Regarding the aspect of missing saveguards: Do you prefer a silently failing minecolonies or should it visibly crash upon incompatibilities?
I personally lean to the second because then everyone can see what works and what doesnt.

someaddons:
its a bug in the other mod yes, but those bugs are frequent which is why when scanning item registries we always wrap it in try catches and skip the erroring items, would suggest to do the same here

Raycoms:
You can send an error log with "item X error with X" https://proton.me/mail/home

muellerjp:
I'll add the catches and log an error as ray suggested.
Will do the changes in a PR to 1.21 first.

Raycoms:
Awaiting safeguards pr then =)

PR #11714, opened 2026-06-16
Bottle safety

armele:
# Checklist

AI was used to review the changed code. 

# Related issues
Closes various discord conversations / discussion

- Halves the chance of breaking glass bottles by the baker.

image

Review please


Thodor12:
Did you actually discuss if this is allowed? Iirc this was done intentionally by Ray

Raycoms:
As the skill is flat you probably need to make it skill/2 somewhere

Thodor12:
> As the skill is flat you probably need to make it skill/2 somewhere

Pretty sure that's usually what the quality is used for

Raycoms:
On a general note, weight = 100 making it 50% seems weird @uecasm is that the case?

Thodor12:
> On a general note, weight = 100 making it 50% seems weird @uecasm is that the case?

That's normal, 2 loot pool entries both with weight 100, that makes it 50/50

Raycoms:
Ah, now I saw it, it's 100 weight empty item and 100 weight glass bottle now. and the 100 weight glass bottle is further influenced by the skill, although I'm not entirely sure how that affects the overall outcome atm. From the screenshot of JEI it seems like flat skill makes it 50% + skill% which would end up with 150% =P (or capped after skill level 50) so it should be skill/2 then we pass into it.

Thodor12:
Assuming that the skill level (SL) is being fed into the luck value of the loot table you'd get this:

Old:
SL   0: 100 -   0 = 100% to lose bottle
SL  25:  75 -  25 =  75% to lose bottle
SL  50:  50 -  50 =  50% to lose bottle
SL  75:  25 -  75 =  25% to lose bottle
SL 100:   0 - 100 =   0% to lose bottle

New:
SL   0: 100 - 100 =  50% to lose bottle
SL  25:  75 - 125 =  37% to lose bottle
SL  50:  50 - 150 =  25% to lose bottle
SL  75:  25 - 175 =  12% to lose bottle
SL 100:   0 - 200 =   0% to lose bottle

Remember that the weight + the quality is cumulative first, then a percentage is created from it.
The number in the quality gets increased by X times the luck value. So a quality of 1 means every skill level nets you 1 extra point. A value of -1 means you're subtracting a point of the weight.

So the loot table determines the final weights based after the qualities are applied, then sums the values and does the weighted table random.

Also it's totally possible that the JEI integration just assumes something and isn't actually doing the correct calculation.
And `+ skill modifier` is a bit vague, but that's a different issue.

Raycoms (approved):

uecasm (approved):

someaddons (approved):

Raycoms:
please port

armele:
Is it possible this needs to be ported to 1.21 still?

armele:
> Is it possible this needs to be ported to 1.21 still?

Oh yes, and I see you asked me to do that... so this is on me. :)

PR #11730, opened 2026-06-27
Fix/water stuff

uecasm:
# Checklist

# Related issues
Closes [discord issue](https://discord.com/channels/472875599422291968/1520258678899478698/1520258678899478698)
Closes [discord issue](https://discord.com/channels/139070364159311872/724810019928801290/1520260849636216943)

- Fixes an unrelated bug where fences and walls didn't get placed waterlogged even when the blueprint said they should.
- Adds a new intermediate build stage to let all water and waterlogged blocks be placed strictly after normal blocks but before decorative blocks, regardless of location.
- Moves farmland and sugarcane to decorative placement since they're water-dependent, and also flowers and crops just because that makes sense.

WARNING: since this adds a build stage it might break work orders already in progress at the DECORATE stage or higher.  They should just need to be restarted or repaired to recover.

Review please (do not port... yet)

### DIFF
--- a/src/main/java/com/minecolonies/core/placementhandlers/GeneralBlockPlacementHandler.java
+++ b/src/main/java/com/minecolonies/core/placementhandlers/GeneralBlockPlacementHandler.java
@@ -7,6 +7,7 @@ import com.ldtteam.structurize.placement.IPlacementContext;
 import com.ldtteam.structurize.placement.handlers.placement.IPlacementHandler;
 import com.ldtteam.structurize.util.BlockUtils;
 import com.minecolonies.api.compatibility.candb.ChiselAndBitsCheck;
+import com.minecolonies.api.util.BlockStateUtils;
 import com.minecolonies.api.util.Log;
 import com.minecolonies.api.util.WorldUtil;
 import net.minecraft.core.BlockPos;
@@ -17,12 +18,10 @@ import net.minecraft.world.InteractionHand;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.context.BlockPlaceContext;
 import net.minecraft.world.level.Level;
-import net.minecraft.world.level.block.FenceBlock;
-import net.minecraft.world.level.block.FenceGateBlock;
-import net.minecraft.world.level.block.IronBarsBlock;
-import net.minecraft.world.level.block.WallBlock;
+import net.minecraft.world.level.block.*;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.block.state.properties.BlockStateProperties;
 import net.minecraft.world.phys.BlockHitResult;
 import net.minecraft.world.phys.Vec3;
 import org.jetbrains.annotations.NotNull;
@@ -59,7 +58,17 @@ public class GeneralBlockPlacementHandler implements IPlacementHandler
                     new BlockHitResult(new Vec3(0, 0, 0), Direction.DOWN, pos, true)));
                 if (tempState != null)
                 {
-                    placementState = tempState;
+                    placementState = BlockStateUtils.copyProperties(placementState, tempState,
+                        BlockStateProperties.UP,
+                        BlockStateProperties.NORTH,
+                        BlockStateProperties.NORTH_WALL,
+                        BlockStateProperties.EAST,
+                        BlockStateProperties.EAST_WALL,
+                        BlockStateProperties.SOUTH,
+                        BlockStateProperties.SOUTH_WALL,
+                        BlockStateProperties.WEST,
+                        BlockStateProperties.WEST_WALL,
+                        PillarBlock.COLUMN);
                 }
             }
             catch (final Exception ex)
--- a/src/datagen/generated/minecolonies/data/minecolonies/tags/block/decoblocks.json
+++ b/src/datagen/generated/minecolonies/data/minecolonies/tags/block/decoblocks.json
@@ -12,10 +12,14 @@
     "minecraft:tube_coral_block",
     "minecraft:bell",
     "minecraft:lantern",
+    "minecraft:farmland",
+    "minecraft:sugar_cane",
+    "minecraft:cactus",
     "minecolonies:gate_wood",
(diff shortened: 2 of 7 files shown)


src/main/java/com/minecolonies/core/placementhandlers/GeneralBlockPlacementHandler.java
```diff
@@ -59,7 +58,17 @@ public ActionProcessingResult handle(
...
+                        BlockStateProperties.UP,
+                        BlockStateProperties.NORTH,
+                        BlockStateProperties.NORTH_WALL,
+                        BlockStateProperties.EAST,
+                        BlockStateProperties.EAST_WALL,
+                        BlockStateProperties.SOUTH,
+                        BlockStateProperties.SOUTH_WALL,
+                        BlockStateProperties.WEST,
+                        BlockStateProperties.WEST_WALL,
+                        PillarBlock.COLUMN);
```
uecasm, line 71: I'm assuming that the purpose of the prior code was to preserve these placement properties based on current adjacency at placement time.  This was what was losing other properties like WATERLOGGED; for good measure this will now preserve any properties _other_ than those listed here.

I'm quite dubious whether `PillarBlock.COLUMN` should be in this list, but then since that's its only property I can only assume that it was intended if it was listed at all.

Raycoms:
Is there some way we can avoid adding an additional stage, merging some existing stages or whatever? I'd really like to avoid adding additional stages as each stage, esp for larger blueprints take up a lot of space and compute

uecasm:
We could make one stage iterate twice, but that just seems strictly worse than having another stage.  In most blueprints it should flick through the water stage pretty fast since there won't be much water.

Raycoms:
Yeah I mean twice doesn't help, it's the amount of times I'm concerned about. And even with little water we still have to check each position. Unless we do some kind of smart indexing

uecasm:
We could iterate the blocks just once at the very start to build a stage -> pos map, then iterate that later.  It'd take a tiny bit more memory and take a bit more time getting started before they actually build (and would need re-doing each restart, since we shouldn't persist it).

I don't think it should be worth it, though, iterating things is something computers do really fast anyway.

someaddons:
didnt this use to work before?

uecasm:
> didnt this use to work before?

#11231 did move waterlogged block placement later (to decoration phase) since it was inconsistent with normal water blocks and caused spillage.  This has prompted discovery of some additional issues and feature requests.

someaddons:
it might be better to just revert to the old behaviour then, some spillage shouldnt be as big of a problem

uecasm:
This is better though.

someaddons:
Why?

uecasm:
Because this fixes spillage while still ensuring correct placement timing.  And more importantly, ensuring that waterlogged blocks are always placed as waterlogged, whereas previously they could sometimes be placed as dry for no readily apparent reason.

Raycoms:
> We could iterate the blocks just once at the very start to build a stage -> pos map, then iterate that later. It'd take a tiny bit more memory and take a bit more time getting started before they actually build (and would need re-doing each restart, since we shouldn't persist it).
> 
> I don't think it should be worth it, though, iterating things is something computers do really fast anyway.

Iterating is okay (it's not just iterating it's our iterating mode on top of it), problem is x*y*z waterlogged checks. It piles up. Builders because of all the iterating and checks already are the #2 worker to show up on profiles (after guards)

uecasm:
FYI:
image
On the left is the correct test blueprint placed using this PR.  All three water blocks (might be hard to see but one is pure water, one is waterlogged coral, one is waterlogged fence) are placed before the farmland and sugarcane.

On the right is the incorrect blueprint placed in version 1326 (before #11231).  The farmland is placed first and might dry out, and the fence is placed not waterlogged.  So that not working does precede the recent change.

Raycoms:
We should place fence during solid stage and water during the normal water stage which should be pre deco, no?

https://proton.me/mail/home

uecasm:
> We should place fence during solid stage and water during the normal water stage which should be pre deco, no?

No, in 1326 and earlier it does place the farmland and fence (not waterlogged) during solid stage, and water is placed during the deco stage mixed with other decos, which can cause issues.  The coral is also placed during deco since it's explicitly flagged as a deco item.

With the earlier change and this PR, it places the fence/water/coral together, strictly before the farmland/sugarcane.

I did wonder if it might be an interesting alternative to place the fence (unwaterlogged) during the solid phase, then waterlog it during the water/deco phase, but this seemed more complicated and still didn't really solve the problem that Shadizar had.

Raycoms:
Water clearing is done after solid, we could do water placement there too

uecasm:
Can't modify water clearing without changing Structurize, it's handled in there instead.  Could add water placement to the CLEAR_NON_SOLIDS phase instead of a separate phase just after it, but CLEAR_NON_SOLIDS is run top-down rather than bottom-up, which seems like the wrong way to place water.

uecasm:
> Iterating is okay (it's not just iterating it's our iterating mode on top of it), problem is xyz waterlogged checks. It piles up. Builders because of all the iterating and checks already are the https://github.com/ldtteam/minecolonies/pull/2 worker to show up on profiles (after guards)

Maybe this is worthwhile then but it ought to be a separate PR.

> We could iterate the blocks just once at the very start to build a stage -> pos map, then iterate that later. It'd take a tiny bit more memory and take a bit more time getting started before they actually build (and would need re-doing each restart, since we shouldn't persist it).

src/main/java/com/minecolonies/core/placementhandlers/GeneralBlockPlacementHandler.java
someaddons: Better description is probably that this will now drop any property of the placement state except for those listed, compared to before

Raycoms:
> Can't modify water clearing without changing Structurize, it's handled in there instead. Could add water placement to the CLEAR_NON_SOLIDS phase instead of a separate phase just after it, but CLEAR_NON_SOLIDS is run top-down rather than bottom-up, which seems like the wrong way to place water.

Could we then adjust structurize to do also apply waterlogging during this step instead?

PR #11748, opened 2026-07-19
Pathfinding/Pathfollow

someaddons:
# Checklist
- [x ] I have read and accept the Contributing Guidelines
- [ x] I have tested and confirmed my changes are working on the most recent commit of this pull request

# Related issues
Closes #
Closes #

Pathfinding improvements:
- Search now aborts on higher cost nodes instead of score, which includes the heuristic already
- Added node badDirectionCost for traversing directional blocks in the perpendicular direction, causing citizens to use stairs,ladders etc from a more favorable direction
- Corner nodes no longer are passed on to pathfollow

Pathfollow:
- Fix slow speed while turning not working as intended, causing citizens to jump too far when going up a block while changing direction
- Citizens now handle ladder entrances better and no longer get stuck on the ladder when trying to enter from the side

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
+++ b/src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
@@ -422,7 +422,7 @@ public abstract class AbstractPathJob implements Callable<Path>, IPathJob
             }
 
             // Don't keep searching more costly nodes when there is a destination
-            if (reachesDestination && node.getScore() > bestNode.getScore())
+            if (reachesDestination && node.getCost() > bestNode.getCost())
             {
                 if (reevaluteHeuristic(bestNode, reachesDestination))
                 {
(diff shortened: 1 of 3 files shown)


src/main/java/com/minecolonies/core/entity/pathfinding/pathjobs/AbstractPathJob.java
```diff
@@ -1176,6 +1212,12 @@ private Path finalizePath(final MNode targetNode)
                 addPathNodeToDebug(node);
             }
 
+            if (node.isCornerNode())
```
Raycoms, line 1215: Could the path end on a corner node and throw an exception here because no parent?

Raycoms: Ah nvm, we do a node.parent != null check a tthe otp

Raycoms (approved):

PR #11753, opened 2026-07-22
Configure Herding Via Datapack

armele:
# Checklist

-- AI was used to analyze and refine my proposed implementation
-- AI was used to code review my changes

# Related issues
Closes None - feature request

- Herders can now be configured via datapack at **minecolonies\tags\entity_types\herders** using tags matching their job ID:
-- beekeeper.json
-- chickenherder.json
-- cowboy.json
-- rabbitherder.json
-- shepherd.json
-- stablemaster.json
-- swineherder.json
- By default, all previously supported animals were configured as built-in datapacks (and these were added to datagen)
- Specialized operations
-- Cowboy milking now uses the **cowboy_milkable_animals** entity tag instead of hardcoded Cow and Goat eligibility. The operation retains the existing MineColonies behavior of exchanging the configured empty container for the configured milk product.
-- Stewing now uses the **cowboy_stewable_animals** entity tag. The cowboy performs the animal’s normal bowl interaction. This allows compatible modded animals to return their own bowl-based products.
-- Shearing now uses the **shepherd_shearable_animals** entity tag. Successful operations collect the item stacks returned by the animal’s own onSheared() implementation, allowing modded animals to provide their normal shearing products.
-- Sheep dyeing remains sheep-specific because Forge does not provide a generic dyeable-entity interface.
- Statistics were updated to specify the type of animal for butchering, breeding, milking and shearing.
- A few prior bugs were fixed where herders were recording stats as "Used" items instead of "Got" items when they picked things up.
- A prior bug with cowboys not requesting milk containers for goats was fixed.
- The minimum herd size now applies to each type of animal for which the herder is responsible. So for example if a herder is maintaining pigs and boars they will retain 2 of each.
- JEI guards were added to animal-specific drops (like wool from sheep)
- Modpack authors should use common sense when deciding what animals a herder might be configured to handle; the code does not guard against bad modpack decisions.

# Deliberate "non-changes" 
...And the thought process behind them
- Introducing custom breeding items introduces a lot of complexity for low value, so herders use their normal prior breeding items rather than allowing specification of custom breeding items for datapacked animals.
- There is no reliable "dyeable" interface, so the shepherd only dyes sheep.

# Testing
- All base minecolonies herders were tested with their default animals
- A custom testing datapack was created with the Naturalist mod, with ducks added to the chicken herder, boars to the swine herder, deer to the cowboy, snails to the shepherd.
- Stats and logs were monitored to make sure herders maintained all the animals they were responsible for, and that special cases like shearing didn't cause problems for animals that don't drop wool (hence the snails to the shepherd).
- Sampled JEI guards to ensure they worked as expected

Review please


src/main/java/com/minecolonies/core/colony/buildings/modules/AnimalHerdingModule.java
```diff
@@ -38,9 +40,30 @@ public class AnimalHerdingModule extends AbstractBuildingModule implements IHasR
...
+
+    /**
+     * Creates a herding module with a programmatic animal predicate.
+     *
+     * @param jobEntry the job associated with the module.
+     * @param animalPredicate predicate selecting animals handled by the module.
+     * @param breedingItem the item used by the herder to breed compatible animals.
+     */
+    public AnimalHerdingModule(@NotNull final JobEntry jobEntry,
                                @NotNull final Predicate<Animal> animalPredicate,
```
Thodor12, line 67: If we have a tag, is there any point in keeping this overload?

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingBeekeeper.java
```diff
@@ -263,18 +264,21 @@ public List<IGenericRecipe> getRecipesForDisplayPurposesOnly(@NotNull Animal ani
...
-                    .withRequiredEntity(animal.getType())
-                    .build());
-
-            recipes.add(GenericRecipe.builder()
-                    .withOutput(Items.HONEY_BOTTLE)
-                    .withInputs(List.of(List.of(Items.GLASS_BOTTLE.getDefaultInstance())))
-                    .withRequiredEntity(animal.getType())
-                    .build());
-
+            if (animal instanceof Bee)
```
Thodor12, line 267: This statement actually makes me wonder if this is the right approach to begin with. If we have to go about making conditional statements like this, what is the point of the tag to begin with then.

I get why you'd do this, but I think some kinds of things really just need code implementations, that's the point of the module system, create more modules to create more supported functionality. Not everything has to be solved with a tag because you can't further configure the system.

src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/EntityAIWorkCowboy.java
```diff
@@ -124,7 +124,7 @@ else if (building != null && building.getFirstModuleOccurance(BuildingCowboy.Her
     {
         final List<ItemStorage> list = super.getExtraItemsNeeded();
         if (building != null && building.getFirstModuleOccurance(BuildingCowboy.HerdingModule.class).canTryToMilk() &&
-              !searchForAnimals(a -> a instanceof Cow && !(a instanceof MushroomCow)).isEmpty())
+              !searchForAnimals(a -> (a instanceof Cow || a instanceof Goat) && !(a instanceof MushroomCow)).isEmpty())
```
Thodor12: Also again, this kind of logic kinda breaks the idea behind tags in the first place.

Thodor12 (dismissed):

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingBeekeeper.java
armele: I think what might be a reasonable compromise is to say that the Beekeeper specifically is not available for customization by datapack because it is too specialized. I could get behind that. I don't think this example is a good reason to throw out the whole concept.

src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/EntityAIWorkCowboy.java
armele: This one is fixable with datapack if desired; we could easily add a tag for "milkable creatures".  Complexity is not high.  (I was skeptical of the value, which is why I didn't include it originally.)

src/main/java/com/minecolonies/core/colony/buildings/modules/AnimalHerdingModule.java
armele: My thought process was in case unknown add-on mods used the existing constructor... this allows backwards compatibility.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingBeekeeper.java
armele: Also I think it's worth mentioning that changes that allow modpack authors to do things with datapacks are far more accessible than code-based changes that require add-on mods or compat mods to be created.  My personal opinion is that a solution that is all datapackable and gives 90% of the "perfect implementation" is a better choice than a solution that is 100% of the "perfect implementation" but requires a compat mod be written.

src/main/java/com/minecolonies/core/entity/ai/workers/production/herders/EntityAIWorkCowboy.java
armele: Milking and "stewing" are now tag-driven as well.  Sheep coloration remains a special case that applies specifically only to sheep.

armele: Shearing also is tag driven.

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingShepherd.java
```diff
@@ -104,11 +105,14 @@ public List<IGenericRecipe> getRecipesForDisplayPurposesOnly(@NotNull Animal ani
...
             final List<IGenericRecipe> recipes = new ArrayList<>(super.getRecipesForDisplayPurposesOnly(animal));
 
-            recipes.add(GenericRecipe.builder()
-                    .withOutputs(ForgeRegistries.ITEMS.tags().getTag(ItemTags.WOOL).stream().map(ItemStack::new).toList())
-                    .withRequiredTool(ModEquipmentTypes.shears.get())
-                    .withRequiredEntity(animal.getType())
-                    .build());
+            if (animal instanceof Sheep)
+            {
+                recipes.add(GenericRecipe.builder()
```
Raycoms, line 110: Shepherd output should always be wool, right?

Raycoms (commented):
I still need to properly review the AI code. I think this is a good idea overall.
As you're already touching this, how hard is it to allow female/male animal separation?
(That could be just the presence of two tags)

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingShepherd.java
armele: For animals that support the "sheareable" interface it might yeild something else depending on their implementation. I left out code agnostic.

armele:
> I still need to properly review the AI code. I think this is a good idea overall.
> As you're already touching this, how hard is it to allow female/male animal separation?
> (That could be just the presence of two tags)

What would be an example of this? Like "bulls" for male cattle and "cows" for female, and ensuring mixed gender for successful breeding? I don't think it would be hard, but am not aware of gendered animals in vanilla, so it seems pretty niche?

someaddons:
I see that you're trying to achieve more widespread animal compatibility, but using Entity Tags probably misses the point.
Entity tags are suitable for selecting candidate entities, but not for defining behavioral compatibility. It makes the animals configurable while leaving breeding, population grouping, products, interactions, item requirements and display logic job-specific. 
As a result, invalid configurations silently produce synthetic or partial behavior, and several tags do not correspond to actual runtime capabilities.
Either we need to vastly extend the data driven approach to include a ton of behaviours, or we do it in-code with additional modules as Thodor suggested. I think the module approach is better overall

Raycoms:
> > I still need to properly review the AI code. I think this is a good idea overall.
> > As you're already touching this, how hard is it to allow female/male animal separation?
> > (That could be just the presence of two tags)
> 
> What would be an example of this? Like "bulls" for male cattle and "cows" for female, and ensuring mixed gender for successful breeding? I don't think it would be hard, but am not aware of gendered animals in vanilla, so it seems pretty niche?

As the tags are mostly useful for the animal mods, a lot of animal mods do have some with genders as well. I feel like implementing this without any genders makes this probably a lot more niche. But maybe I didn't get the main way people would use this.

Like what I would expect is that some modpacks have additional/different animals. E.g. replace vanilla cow with something that doesn't extend cow etc.

armele:
Responding to a mix of the comments - my thinking is the main use of this is simple addition of "vanilla-like" animal handling. So where my chicken farmer currently just does chickens, now they can do ducks, turkeys, etc.  where my swingeeser can currently do only pigs, this would allow boar, moose, etc. 

Products of butchering are a non-issue as that is loot table driven.

Interactions cover the current options ("I can get things with shears", "I can get milk with buckets", "I can get things with bowls") I agree that a new interaction would require new code to support it and a new tag with which to classify animals.

In my mind the one drawback of this implementation is that it allows herders to bemreed animals with items that players cannot use. To me this is acceptible given the extra complexity and limited benefit supporting that would involve.

I see this as a nice low risk compatibility improvement.

That said, if you (core team) prefer the module-based approach I can understand that, and will close this PR.

armele:
Maybe my last ditch sales pitch ;) - this seems like a simple and quick way to make the game a little more fun and immersive, with no down side. The additional code complexity is negligible.

someaddons:
I mean the idea is not bad just that tags arent a great tool to do it. Think we may need to go back to the source and check which actual animals you want support for which are like cow but not instanceof cow.

Could also have a nice little middleway and show in the UI which entities are actually eligible in the pack

PR #11759, opened 2026-07-25
Add waiter to JEI

uecasm:
# Checklist

# Related issues
Closes [discord request](https://discord.com/channels/472875599422291968/1530069945319166112)

- Add waiter to JEI

Review please (should port)


someaddons:
not sure if we want to displayer the waiter, as we want people to use the cook and having recipe partially show for the waiter could confuse the users. Could e.g. think its bugged because its just showing some recipes, not discover the dedicated cooking job etc

uecasm:
The waiter is the cook 😆 but I know you meant chef.

If you check the linked discord conversation, the idea was to show both the waiter and the chef as being able to cook basic food, so that they know the restaurant will work ok by itself before they've unlocked the chef, and that they don't have to teach the chef the cooking recipes (just the crafting ones) since the waiter can handle it, but also that they still can if they wanted to.  Which sounded reasonable to me.

PR #11760, opened 2026-07-25
Show JourneyMap overlay for force-loaded chunks

uecasm:
# Checklist

# Related issues
Closes [discord request](https://discord.com/channels/472875599422291968/473575384445878273/1518649172285194383)

- Adds an optional (disabled by default) overlay for the fullscreen map only, which shows which chunks are force-loaded by the colony.
    - It still uses the colony team colour but with a different intensity, so you can have both overlays visible at the same time if you like and they can still be distinguished, though it might be simpler to turn on only one at a time or set only one to Filled and the other to Framed.
- Adds a general radius-around-player update for chunk status (matching the build tool border renderer) to avoid cases where clients miss an update and show outdated chunk map data.
    - It still caches to avoid expensive updates when nothing actually changed.
    - You may still need to revisit vaguely near a colony to update its borders in some cases, otherwise they may remain at their last-visited state.

Review please (should port)


someaddons (approved):

Raycoms (approved):

Raycoms:
please port

PR #11771, opened 2026-08-01
Build Order Fixes

### LINKED ISSUE #11388: work order priority doesn't work

- [ ] Single Player
- [x] Multi Player

### Minecraft Version

1.20

### MineColonies Version

1.20.1.1.1.1111

### Structurize Version

1.20.1.1.0.787

_No response_

### Current Behavior

Moving a building to the top priority does not make it the next building that gets built they just go in order of which building was put into queue first.  

### Expected Behavior

expect builder to chose the next building in order of priority rather than the order in which they were added to the list,  this is what used to always happen

### Reproduction Steps

put a bunch of buildings in queue
move one from the bottom to the top 
see what they build next 

### Logs

<script src="https://gist.github.com/anastasia590/1557d83c26010705b130e69cc0991506.js"></script>

### Anything else?

_No response_

### Footer

---
 ## Viewers

armele:
Fix town hall prioritization of build orders and display builder queue in priority order.

# Checklist

AI was used to analyze the current logic for better understanding, and to review my code changes.

# Related issues
Closes #11388

- Work-order priority now controls Builder selection, with claimed orders processed before unclaimed orders and work-order ID breaking "ties". 
- The Builder hut shows work in priority order (with all claimed work ahead of any work available to be assigned when in manual mode).
- The current work order border is highlighted in green for a visual signal of the current job.

# Testing
- Assigned multiple builds to a builder, then reordered the priority in the town hall. Observed that the builder moved on to the reordered "next priority" rather than picking the building in the sequence the builds were requested.
- Verified builder hut ordering matched priority order.
- Verified "available" work orders were present when the builder was in manual mode, and appeared after all assigned/claimed work orders.
- Verified priority persisted through a restart

Review please


src/main/java/com/minecolonies/core/client/gui/modules/building/WorkOrderModuleWindow.java
```diff
@@ -132,7 +140,31 @@ private void updateWorkOrders()
      */
     private void sortWorkOrders()
     {
-        workOrders.sort(Comparator.comparing(IWorkOrderView::getPriority, Comparator.reverseOrder()));
```
Raycoms, line 135: Shouldn't priority ordering for this still be correct?

src/main/java/com/minecolonies/core/colony/workorders/WorkManager.java
```diff
@@ -406,11 +414,22 @@ else if (order.isDirty())
             {
                 order.setClaimedBy(BlockPos.ZERO);
             }
+        }
 
-            tryAssignWorkOrder(order, (b) -> order.getClaimedBy().equals(b.getPosition()));
+        final List<IServerWorkOrder> claimedWorkOrders = workOrders.values().stream()
```
Raycoms: This feels a bit like overkill. Twice a for loop inside the while.

        @NotNull final Iterator<IServerWorkOrder> iter = workOrders.values().iterator();
shouldn't we just here take the priority ordered list instead?

src/main/java/com/minecolonies/core/colony/workorders/WorkManager.java
```diff
@@ -47,6 +47,14 @@ public class WorkManager implements IWorkManager
     private static final String TAG_WORK_ORDERS = "workOrders";
     private static final String TAG_NEW_SYSTEM  = "newsystem";
 
+    /**
+     * Orders work orders by their user-selected priority, with a stable tie-breaker.
+     */
+    private static final Comparator<IServerWorkOrder> WORK_ORDER_COMPARATOR = Comparator
```
Raycoms: We should make this static and use it in the other place too I feel

Raycoms (commented):
Some comments

src/main/java/com/minecolonies/core/client/gui/modules/building/WorkOrderModuleWindow.java
armele: The "group" concept is what was behind this change.  So all "claimed" work orders are grouped before "unclaimed" work orders, then priority within those two groups. That way if you have multiple builders in "Manual" mode and multiple unclaimed work orders to be assigned, the queue at the builder hut will show actual assignments in priority order prior to as-yet unassigned build requests.

src/main/java/com/minecolonies/core/colony/workorders/WorkManager.java
armele: Maybe I'm seeing this wrong, but the for loops are outside the while?  The while loop is unchanged - following it is the first for loop on claimed work orders - to see if any builders are ready to take the build order as their active job. Then follows the unclaimed work order - to see if any builders are ready to claim an as-yet unreserved work order.

Raycoms: ah, I didn't see above the closing }.

But we iterate a ton of times over the list here. In the while, + for each filter and sort and then over the filered and sorted once another time. Why can't we add the claimed/unclaimed ones together as before? Esp if we do the while in a sorted iteration (and keep that same already once ordered list for the 2nd loop)

src/main/java/com/minecolonies/core/colony/workorders/WorkManager.java
armele: Do you mean the equivalent priority/ID comparator in WindowInfoPage, and the secondary ordering in WorkOrderModuleWindow? Good idea... Since those operate on IWorkOrderView rather than IServerWorkOrder, I will move the comparator to the common IWorkOrder API type and reuse it from both server and client code.  (And re-test, of course!)

src/main/java/com/minecolonies/core/colony/workorders/WorkManager.java
armele: This is valid... let me rethink the approach here. I think we could do this in two loops - one for cleanup and one for assignment.

armele: I have reworked this method, and also got some performance back out of the tryAssignWorkOrder, too, as it was continuing to iterate buildings without purpose after the assignment was made.

Raycoms (approved):

someaddons (approved):

PR #11794, opened 2026-08-25
Graveyard Quality of Life

armele:
# Checklist

AI was used for code review.

# Related issues
Closes [Discord Feedback](https://discord.com/channels/472875599422291968/1540574537500987422)

- Add a "X of Y Plots Full" label on the graveyard module.

image

- Add min stock module to graveyard (for shovels, really)
- Add a burial event when a citizen is successfully buried.  (Anticipation here is of future add-on mod use of this event.)
- During testing noticed that a body recovered at the end of the day would not be buried if a sleep cycle interrupted the burial. Fixed this.

# Testing
- Verified undertaker buried dead citizens successfully.
- Verified X of Y Plots Full label was updated correctly on burial.
- Verified min stock module worked as expected. 

Review please


src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingGraveyard.java
```diff
@@ -172,6 +172,21 @@ public Set<Tuple<BlockPos, Direction>> getGravePositions()
         return visualGravePositions;
     }
 
+    /**
+     * Get the positions of all burial plots registered from the graveyard schematic.
+     *
+     * @return an unmodifiable set of burial plot positions.
+     */
+    public Set<BlockPos> getGravePlotPositions()
```
Raycoms: I think we can do without this, as we don't do any checks in here, we can do the loop on the visualGravePositons just

src/main/java/com/minecolonies/core/colony/buildings/modules/GraveyardManagementModule.java
```diff
@@ -126,6 +129,22 @@ public void serializeToView(@NotNull final FriendlyByteBuf buf)
         {
             buf.writeUtf(citizenName);
         }
+
+        // Graveyard burial plots.
+        if (building instanceof BuildingGraveyard graveyard)
```
Raycoms: This is always BuildingGRaveyard, no?

armele: It *should* be, agreed. I have been getting in this habit to protect against the case where some add-on author puts the module on another building for some reason...

Raycoms: tbf, the grave positions should probably live inside the module just, then we don't care at all, right?

armele: That is a good idea, but I wonder if the risk of making that change for an existing colony introduces too much complexity or risk for migrating the data upon the upgrade.

Raycoms: I think that should be fine. It's long term more stable instead.

armele: Refactored and retested - including a one-time migration from the building-owned graveyard positions.

src/main/java/com/minecolonies/core/colony/buildings/modules/GraveyardManagementModule.java
```diff
@@ -45,11 +52,21 @@ public class GraveyardManagementModule extends AbstractBuildingModule implements
...
+    private static final String TAG_VISUAL_GRAVES_BLOCKPOS = "visualgravesblockpos";
+
+    private static final String TAG_VISUAL_GRAVES_FACING = "visualgravesfacing";
+
     /**
      * The list of resting citizen in this graveyard.
      */
     private final List<String> restingCitizen = new ArrayList<>();
 
+    private final Set<Tuple<BlockPos, Direction>> gravePositions = new HashSet<>();
```
Raycoms, line 78: javadoc

src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingGraveyard.java
```diff
@@ -130,15 +103,8 @@ public void deserializeNBT(final CompoundTag compound)
...
-        final ListTag visualGraveTagList = compound.getList(TAG_VISUAL_GRAVES, Tag.TAG_COMPOUND);
-        for (int i = 0; i < visualGraveTagList.size(); ++i)
-        {
-            final CompoundTag graveCompound = visualGraveTagList.getCompound(i);
-            final BlockPos graveLocation = BlockPosUtil.read(graveCompound, TAG_VISUAL_GRAVES_BLOCKPOS);
-            final Direction graveFacing = Direction.byName(graveCompound.getString(TAG_VISUAL_GRAVES_FACING));
-            visualGravePositions.add(new Tuple<>(graveLocation, graveFacing));
-        }
+        // Safely migrate grave positions from the building to the module
+        getModule(GraveyardManagementModule.class).migrateLegacyGravePositions(compound);
```
Raycoms, line 107: I think we can simplify the migration logic, we just call from here the module "loadfromnbt" method, and we give it the compound where it can load it from, so it doesn't need a special method

armele: I don't think we want to call the whole loadfromNBT - that's why I did the targeted migration of just the plot positions, because we're merging the plot positions from the building into the module data that already exists.  If you are open to retaining the targeted migration I'd prefer that as safer...

src/main/java/com/minecolonies/core/colony/buildings/modules/GraveyardManagementModule.java
armele: Added

src/main/java/com/minecolonies/core/colony/buildings/modules/GraveyardManagementModule.java
```diff
@@ -41,15 +48,40 @@ public class GraveyardManagementModule extends AbstractBuildingModule implements
...
 
+    /**
+     * Indicates whether the grave positions were loaded from the module data (as opposed to the legacy building data).
+     */
+    private boolean gravePositionsLoadedFromModuleNbt;
+
+    /**
+     * The set of registered grave plot positions and their facing.
+     */
+    private final Set<Tuple<BlockPos, Direction>> gravePositions = new HashSet<>();
```
someaddons, line 78: I know its from the previous code already, but why do we even need to store the facing? It feels odd to me to store this here when the block already has the facing value

armele: I will admit I did not think about the "why" for the current state... From looking at the code now, though, I see it is used to place the block in the world (in buryCitizensHere).  So effectively we're proactively getting the information from the schematic instead of deferring it until placement time, when we'd need to check the schematic facing anyway in order to set the block in the world with the desired facing of the schematic designer (since the block in the world doesn't have the facing on it - it is not yet a grave).

src/main/java/com/minecolonies/core/colony/buildings/modules/GraveyardManagementModule.java
```diff
@@ -41,15 +48,40 @@ public class GraveyardManagementModule extends AbstractBuildingModule implements
...
+
     /**
      * The list of resting citizen in this graveyard.
      */
     private final List<String> restingCitizen = new ArrayList<>();
 
+    /**
+     * Indicates whether the grave positions were loaded from the module data (as opposed to the legacy building data).
+     */
+    private boolean gravePositionsLoadedFromModuleNbt;
```
Raycoms: this should not be necessary, it can only load it from one source.

armele: Fair point.  Removed and retested.

src/main/java/com/minecolonies/core/colony/buildings/modules/GraveyardManagementModule.java
someaddons: Are we not just replacing unnamed graves with named ones? both have a facing attribute

Raycoms: No the unnamed graves are in the schematic only afair, not in the world.

someaddons:
```
    @Override
    public void registerBlockPosition(@NotNull final BlockState state, @NotNull final BlockPos pos, @NotNull final Level world)
    {
        super.registerBlockPosition(state, pos, world);
        if (state.getBlock() == ModBlocks.blockNamedGrave)
        {
            visualGravePositions.add(new Tuple<>(pos, state.getValue(AbstractBlockMinecoloniesNamedGrave.FACING)));
        }
    }
```

 So we fill the information when the builder places the grave store it and reuse the information later instead of just using the information of the block in the world, where we have the information from in the first place.
    Seems quite odd :D

MotionlessTrain: The named grave actually doesn’t place. The [placement handler](https://github.com/ldtteam/minecolonies/blob/1b03b4fcaea3a31fe520b1fe21ebdcfd7c3cc30a/src/main/java/com/minecolonies/core/placementhandlers/NamedGravePlacementHandler.java#L29-L43) has a noop for placement (it is seen as placeholder, basically)

armele: Right, so the thing in the world (an AIR block, I suppose it would be every time) doesn't have facing.

someaddons: so the block does not get placed but still gets registered, thats an odd hack for this :D

Raycoms (approved):

someaddons (approved):

armele:
Do you need me to port this one?

Raycoms:
yes please

PR #11801, opened 2026-08-31
Make upgradeContainers agree with registerBlockPosition

genericJE:
# Checklist

AI (Claude) was used for all of this change: finding the inconsistency, writing the patch to `BuildingWareHouse.upgradeContainers`, and drafting this description.

I have left the second box unticked deliberately. The change was verified against a JUnit test that is not included here, because the test source set is not enabled in this repository and turning it on would mean editing `gradle.properties`. `./gradlew build` passes. It has not been tested in a running game, so I cannot honestly claim the second box yet.

# Related issues
No existing issue. I looked and did not find one covering this.

- `BuildingWareHouse.upgradeContainers` now catches each rack up to the warehouse storage upgrade level, using the same loop that `registerBlockPosition` already uses, instead of stepping each rack once.
- The storage upgrade counter is incremented before the loop, so it is the level racks are brought up to.

The two methods enforce the same invariant, that every rack sits at the current storage upgrade level, but they did not agree. `registerBlockPosition` loops until a rack reaches the level. `upgradeContainers` stepped each rack once and advanced the counter regardless of whether a given rack was reached, so if `world.getBlockEntity(pos)` did not resolve to a rack for one position, that rack was left short of slots with nothing to correct it later.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingWareHouse.java
+++ b/src/main/java/com/minecolonies/core/colony/buildings/workerbuildings/BuildingWareHouse.java
@@ -164,17 +164,23 @@ public class BuildingWareHouse extends AbstractBuilding implements IWareHouse
     @Override
     public void upgradeContainers(final Level world)
     {
-        if (getFirstModuleOccurance(WarehouseModule.class).getStorageUpgrade() < MAX_STORAGE_UPGRADE)
+        final WarehouseModule module = getFirstModuleOccurance(WarehouseModule.class);
+        if (module.getStorageUpgrade() < MAX_STORAGE_UPGRADE)
         {
+            module.incrementStorageUpgrade();
             for (final BlockPos pos : getContainers())
             {
                 final BlockEntity entity = world.getBlockEntity(pos);
                 if (entity instanceof TileEntityRack && !(entity instanceof TileEntityColonyBuilding))
                 {
-                    ((AbstractTileEntityRack) entity).upgradeRackSize();
+                    // Catch racks up to the storage upgrade level; one that missed an earlier
+                    // upgrade would otherwise stay short of slots until the hut is rebuilt.
+                    while (((AbstractTileEntityRack) entity).getUpgradeSize() < module.getStorageUpgrade())
+                    {
+                        ((AbstractTileEntityRack) entity).upgradeRackSize();
+                    }
                 }
             }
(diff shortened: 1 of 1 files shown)


someaddons:
we are relying on the world and container lists matching at many places, which is why we add/remove containers on world changes. The world not having the container at that place means it does not exist, and later adding is now allowed either, so this change makes little sense

Raycoms:
> we are relying on the world and container lists matching at many places, which is why we add/remove containers on world changes. The world not having the container at that place means it does not exist, and later adding is now allowed either, so this change makes little sense

I think it's in case a rack gets broken and repaired a rack will not get upgraded properly. But that might need another repair way than this.

someaddons:
but when a rack is broken its no longer on the containers list

someaddons:
ah you mean the intention is to upgrade racks added after a repair or building upgrade? We should already have that logic I recall, and ye this one would not cut it as the upgrade is called once

Raycoms:
> ah you mean the intention is to upgrade racks added after a repair or building upgrade? We should already have that logic I recall, and ye this one would not cut it as the upgrade is called once

Like, a creeper explodes some racks, the racks are replaced at repair but not at their upgraded version. We might not have that.

someaddons:
```
    @Override
    public void registerBlockPosition(@NotNull final Block block, @NotNull final BlockPos pos, @NotNull final Level world)
    {
        if (block instanceof BlockMinecoloniesRack)
        {
            final BlockEntity entity = world.getBlockEntity(pos);
            if (entity instanceof TileEntityRack)
            {
                ((AbstractTileEntityRack) entity).setInWarehouse(true);
                while (((TileEntityRack) entity).getUpgradeSize() < getFirstModuleOccurance(WarehouseModule.class).getStorageUpgrade())
                {
                    ((TileEntityRack) entity).upgradeRackSize();
                }
            }
        }
        super.registerBlockPosition(block, pos, world);
    }
```

Seems we actually do have that, and the call this PR modifies is only used on the warehouse upgrade button push

MotionlessTrain:
Can’t you only upgrade the rack storage at level 5?
So it is in fact never used in the upgrade function? (More future proofing, if it would be made possible at level 4 or lower)

someaddons:
I'm talking about the rack upgrade button, not the building upgrade. And register is called when the builder places the rack or on repair

PR #11802, opened 2026-08-31
Persist rack size when a warehouse storage upgrade grows it

genericJE:
# Checklist

To be upfront: AI (Claude) was used for all of this change, at my prompting. That covers finding the cause, the one added line in `TileEntityRack.upgradeRackSize()`, and this description. Please review it critically.

# Related issues
No existing issue. I looked and did not find one covering this.

This does not overlap with my other open pull request, which changes `upgradeContainers`. This one is only about persistence, and the two can be taken independently.

- Call `setChanged()` at the end of `TileEntityRack.upgradeRackSize()`, so a rack that grows during a warehouse storage upgrade is actually written back.

`upgradeRackSize()` replaced the inventory and sent a block update, but never marked the block entity changed:

```java
        inventory = tempInventory;
        final BlockState state = level.getBlockState(worldPosition);
        level.sendBlockUpdated(worldPosition, state, state, 0x03);
        invalidateCap();
    }
```

Nothing flagged the chunk unsaved, and chunk saving skips chunks that are not marked unsaved, so the larger size existed only in memory. On reload the racks reverted to their previous size, while the storage upgrade counter in the colony data persisted normally. The warehouse then believes it is upgraded while its racks are back to the base slot count, so `searchMostEmptyRack` runs out of completely empty slots earlier than it should and reports the warehouse as full.

I think this may be behind the reports of a warehouse announcing that it is full while its racks still have room, where setting Repair Building restores it. A repair rebuilds the racks through `registerBlockPosition` and dirties those chunks as it places blocks, so that time the sizes survive.

Testing, on 1.21.1 with a level 5 warehouse holding 184 racks across six chunks. Before the change, storage upgrades persisted only in the single chunk holding the hut block, 18 racks, and the remaining 166 were back to the base size after a reload, including after a clean quit to title. With the change, all 184 racks in all six chunks kept their new size, and they were written during an ordinary autosave rather than needing a shutdown. I read the rack sizes back out of the region files directly rather than counting slots in the GUI, because a double rack shows both halves combined and is easy to misread.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/tileentities/TileEntityRack.java
+++ b/src/main/java/com/minecolonies/core/tileentities/TileEntityRack.java
@@ -281,6 +281,9 @@ public class TileEntityRack extends AbstractTileEntityRack implements IMateriall
         final BlockState state = level.getBlockState(worldPosition);
         level.sendBlockUpdated(worldPosition, state, state, 0x03);
         invalidateCap();
+        // Without this the enlarged rack is never written back, so the extra slots are
+        // lost on reload while the warehouse still counts the storage upgrade.
+        setChanged();
     }
 
     @Override


Thodor12:
There is one thing about letting AI an do an audit. But if you do not verify these issues yourself in-game where possible, then what's the point? Maybe the AI doesn't see something in our enormous codebase.

This isn't broken at all, maybe it's an improvement but unless there's a physical report of this being broken in-game, there is no need for this change at all. And frankly speaking, I don't think this is broken or a bug report would've surfaced LONG ago on this topic.

MotionlessTrain:
> This isn't broken at all, maybe it's an improvement but unless there's a physical report of this being broken in-game, there is no need for this change at all.

The last part of the PR description, which tells you how to test it, does seem to have found a bug
And I can remember there being some wonkiness with it, but nobody ever made a bug report. And this is a feature which is not used very often, I believe

genericJE:
@Thodor12 I have tested this in a running instance. I verified the bug, that the racks weren't upgrading / not saving properly. Then, after compiling a new jar with this 1 line it upgraded the racks properly and saved them to disk properly.

I launched this investigation because I am experiencing a bug on my server that I host and play on with my friends. And the last few times after I launched the server after full shutdown the courier was complaining about not having enough space in the warehouse even though there were tons of empty racks. I have not tried it on that server, because i don't want to distribute this fix to my friends if it's not in a release. I'm hoping this will also resolve that problem. 

I have some in-game screenshots that show the racks at different upgrade levels after reloading the world.

MotionlessTrain:
Yeah, I did have the feeling that this was a shortcut of a bug report and an immediate fix for it

Raycoms (approved):

genericJE:
I'm running this on my server now and it seems to have fix my original problem as well. I believe the original cause was porting the 1.19 version over to 1.21.

someaddons (approved):

PR #11807, opened 2026-09-01
Request related fixes

Raycoms:
# Checklist

# Related issues
Closes #
Closes #

- Fix buggy remote fulfillment
- More gracefully handle requests for raw things in the dining hall

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/client/gui/modules/RequestTreeWindowModule.java
+++ b/src/main/java/com/minecolonies/core/client/gui/modules/RequestTreeWindowModule.java
@@ -5,6 +5,7 @@ import com.ldtteam.blockui.PaneBuilders;
 import com.ldtteam.blockui.controls.*;
 import com.ldtteam.blockui.views.ScrollingList;
 import com.minecolonies.api.colony.IColonyView;
+import com.minecolonies.api.colony.requestsystem.location.ILocation;
 import com.minecolonies.api.colony.requestsystem.request.IRequest;
 import com.minecolonies.api.colony.requestsystem.request.RequestState;
 import com.minecolonies.api.colony.requestsystem.requestable.IDeliverable;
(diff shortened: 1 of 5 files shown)


src/main/java/com/minecolonies/core/client/gui/modules/RequestTreeWindowModule.java
```diff
@@ -274,7 +275,12 @@ public final boolean isCancellable(final IRequest<?> request)
      */
     public final boolean isFulfillable(final IRequest<?> request)
     {
-        if (!(this instanceof IRequestTreeSupportsFulfill) || !(request.getRequest() instanceof IDeliverable deliverable))
+        if (!(this instanceof IRequestTreeSupportsFulfill requestTreeSupportsFulfill) || !(request.getRequest() instanceof IDeliverable deliverable))
```
someaddons: uhm why are we using this weird mixin style interface check cast here for a child class of this class? That should be just normal calls you can override?

src/main/java/com/minecolonies/core/colony/buildings/modules/RestaurantMenuModule.java
```diff
@@ -133,42 +131,56 @@ public void onColonyTick(@NotNull final IColony colony)
...
                     {
-                        final int qty = Math.min(STACKSIZE, Math.min(requestStack.getMaxStackSize(), delta));
                         final MinimumStack stack = new MinimumStack(requestStack, false, true, ItemStackUtils.EMPTY, qty, 1);
 
+                        stack.setCanBeResolvedByBuilding(false);
+                        building.createRequest(stack, true);
+                    }
+                    else if (!rawStack.isEmpty()
+                        && rawRequest == null
+                        && colony.getRequestManager().getRequestForToken(request).getState().ordinal() < RequestState.IN_PROGRESS.ordinal())
```
someaddons, line 166: This seems a bit odd, when the rawstacks of a menu item in the buildings inventory, e.g. noncooked potatoes do exists(are not empty) we request more raw potatoes directly?

Raycoms: The rawStack.isEmpty check just checks if there is a raw item version of that.

someaddons: so if we have a raw version and there is an existing request for the nonraw version we now additionally request the raw version manually?

Raycoms: before we were requesting the raw version additionally "randomly". Now, we only request the raw version, if we tried requesting the cooked version unsuccessfully.

someaddons (approved):

PR #11813, opened 2026-09-04
Add-On Mod Happiness Support

armele:
# Checklist
- [X] I have read and accept the [Contributing guidelines](https://github.com/ldtteam/.github/blob/master/CONTRIBUTING.md)

AI was used for analyzing the failure path of why my 3rd party happiness modifiers weren't working, and to review the corrections.

# Related issues
Closes None

Happiness modifier deserialization was restricted by a hard-coded list of native modifier instance IDs. This changes the logic such that modifier reconstruction is now validated against the registered modifier type stored in the serialized data instead of the modifier’s instance ID.

## Before
When loading saved citizen data or synchronizing citizen data to a client, CitizenHappinessHandler only reconstructed a modifier when its instance ID appeared in VALID_HAPPINESS_MODIFIERS.

As a result, add-on mods could register custom happiness modifier types and functions, but could not successfully persist or synchronize modifier instances with add-on-defined IDs. These modifiers were discarded during deserialization and therefore could not reliably survive world reloads, reach clients, or appear in the existing happiness UI.

Unknown modifier types could also cause unsafe registry lookups during loading.

## After
Modifier reconstruction is now validated against the registered modifier type stored in the serialized data instead of the modifier’s instance ID.

This allows add-on mods to:
- Register custom happiness modifier types and happiness functions.
- Use their own modifier instance IDs.
- Persist custom modifiers across save and reload.
- Synchronize custom modifiers to clients.
- Display custom modifiers in the existing citizen happiness UI.
- Register against stable, public registry ID constants rather than duplicating MineColonies’ registry-name strings.

Missing, malformed, or broken modifier types are logged and safely skipped without interrupting the rest of citizen-data deserialization.

Preconstructed native modifiers continue to be updated in place, preserving runtime-only configuration such as their rollover predicates. The legacy VALID_HAPPINESS_MODIFIERS constant remains available for API compatibility but is deprecated and no longer controls reconstruction.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/api/entity/citizen/happiness/HappinessRegistry.java
+++ b/src/main/java/com/minecolonies/api/entity/citizen/happiness/HappinessRegistry.java
@@ -10,6 +10,7 @@ import net.minecraft.resources.ResourceLocation;
 import net.minecraftforge.registries.IForgeRegistry;
 import net.minecraftforge.registries.RegistryObject;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -69,33 +70,35 @@ public class HappinessRegistry
      * @param persist  whether we're reading from persisted data or from networking.
      * @return the modifier instance.
      */
+    @Nullable
     public static IHappinessModifier loadFrom(@NotNull final CompoundTag compound, final boolean persist)
     {
-        final ResourceLocation modifierType = compound.contains(NbtTagConstants.TAG_MODIFIER_TYPE)
-                                                ? new ResourceLocation(compound.getString(NbtTagConstants.TAG_MODIFIER_TYPE))
-                                                : new ResourceLocation(Constants.MOD_ID, "null");
-        final IHappinessModifier modifier = getHappinessTypeRegistry().getValue(modifierType).create();
-
-        if (modifier != null)
+        final ResourceLocation modifierType = ResourceLocation.tryParse(compound.getString(NbtTagConstants.TAG_MODIFIER_TYPE));
+        if (modifierType == null || !getHappinessTypeRegistry().containsKey(modifierType))
         {
-            try
+            Log.getLogger().warn("Unknown Happiness Modifier type '{}', its state cannot be restored.", modifierType);
+            return null;
+        }
+
+        try
+        {
+            final HappinessFactorTypeEntry entry = getHappinessTypeRegistry().getValue(modifierType);
+            final IHappinessModifier modifier = entry == null ? null : entry.create();
+            if (modifier == null)
             {
-                modifier.read(compound, persist);
-            }
-            catch (final RuntimeException ex)
-            {
-                Log.getLogger()
-                  .error(String.format("A Happiness Modifier %s has thrown an exception during loading, its state cannot be restored. Report this to the mod author",
-                    modifierType), ex);
+                Log.getLogger().warn("Happiness Modifier type '{}' has no usable factory, its state cannot be restored.", modifierType);
                 return null;
             }
-        }
-        else
-        {
-            Log.getLogger().warn(String.format("Unknown Happiness Modifier type '%s' or missing constructor of proper format.", modifierType));
-        }
 
-        return modifier;
+            modifier.read(compound, persist);
+            return modifier;
+        }
+        catch (final RuntimeException ex)
+        {
+            Log.getLogger().error("A Happiness Modifier of type '{}' has thrown an exception during loading, its state cannot be restored. Report this to the mod author.",
+              modifierType, ex);
+            return null;
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/minecolonies/api/util/constant/HappinessConstants.java
```diff
@@ -60,6 +60,12 @@ public final class HappinessConstants
     public static final String FOOD             = "food";
     public static final String HADGREATFOOD     = "greatfood";
 
+    /**
+     * Legacy list of MineColonies happiness modifier instance IDs.
+     *
+     * @deprecated Modifier reconstruction is validated through the happiness factor type registry instead.
+     */
+    @Deprecated
     public static final Set<String> VALID_HAPPINESS_MODIFIERS = ImmutableSet.of(HOMELESSNESS,
```
Thodor12: Probably fine to remove this too. Minecolonies doesn't need to always bin compatible as we do not directly need to justify ourselves against addons. (And realistically this is one of those things that addons should never directly talk to)

src/main/java/com/minecolonies/apiimp/CommonMinecoloniesAPIImpl.java
```diff
@@ -313,13 +313,13 @@ public void onRegistryNewRegistry(final NewRegistryEvent event)
                        .setIDRange(0, Integer.MAX_VALUE - 1), (b) -> questDialogueAnswerRegistry = b);
 
         event.create(new RegistryBuilder<HappinessRegistry.HappinessFactorTypeEntry>()
-                       .setName(new ResourceLocation(Constants.MOD_ID, "happinessfactortypes"))
+                       .setName(HappinessRegistry.HAPPINESS_FACTOR_TYPE_REGISTRY_ID)
```
Thodor12, line 316: Probably better to keep the consistency with the old code here and not reuse constants for this. (This has changed in 1.21)

src/main/java/com/minecolonies/apiimp/initializer/ModHappinessFactorTypeInitializer.java
```diff
@@ -19,8 +18,8 @@
  */
 public final class ModHappinessFactorTypeInitializer
 {
-    public final static DeferredRegister<HappinessFactorTypeEntry> DEFERRED_REGISTER_HAPPINESS_FACTOR = DeferredRegister.create(new ResourceLocation(Constants.MOD_ID, "happinessfactortypes"), Constants.MOD_ID);
-    public final static DeferredRegister<HappinessFunctionEntry> DEFERRED_REGISTER_HAPPINESS_FUNCTION = DeferredRegister.create(new ResourceLocation(Constants.MOD_ID, "happinessfunction"), Constants.MOD_ID);
+    public final static DeferredRegister<HappinessFactorTypeEntry> DEFERRED_REGISTER_HAPPINESS_FACTOR = DeferredRegister.create(HAPPINESS_FACTOR_TYPE_REGISTRY_ID, Constants.MOD_ID);
```
Thodor12, line 21: Ditto here for the constants, do not use constants.

src/main/java/com/minecolonies/api/entity/citizen/happiness/HappinessRegistry.java
```diff
@@ -69,33 +70,35 @@ public IHappinessModifier create()
...
-        if (modifier != null)
+        try
         {
-            try
-            {
-                modifier.read(compound, persist);
-            }
-            catch (final RuntimeException ex)
+            final HappinessFactorTypeEntry entry = getHappinessTypeRegistry().getValue(modifierType);
+            final IHappinessModifier modifier = entry == null ? null : entry.create();
```
Thodor12, line 86: Instead of doing this, isn't it better to put the early return here and then calling .create() on a known non null instance. Given this method should not be nullable.

Thodor12, line 86: Also this is realistically dead code given that you already have a containsKey check ahead of this.

src/main/java/com/minecolonies/api/entity/citizen/happiness/HappinessRegistry.java
```diff
@@ -69,33 +70,35 @@ public IHappinessModifier create()
      * @param persist  whether we're reading from persisted data or from networking.
      * @return the modifier instance.
      */
+    @Nullable
```
Thodor12, line 73: You didn't change the caller to account for the fact this can return null now?

Thodor12 (changes requested):

Thodor12:
Also, side question, but the citizen UI is currently constrained to the fact a certain amount of happiness values are prepared. Is the UI even ready to dynamically receive other happiness modifiers and/or is it even a scroll container yet?

src/main/java/com/minecolonies/api/util/constant/HappinessConstants.java
armele: Removed!

armele:
> Also, side question, but the citizen UI is currently constrained to the fact a certain amount of happiness values are prepared. Is the UI even ready to dynamically receive other happiness modifiers and/or is it even a scroll container yet?

Because my attempt to add modifiers failed upstream from this, I haven't gotten that far in evaluating the impact of adding the new ones.  My plan was to get past this persistence block, go again at adding custom happiness modifiers (out of Salvation, fwiw) and then evaluate UI impact.

src/main/java/com/minecolonies/apiimp/initializer/ModHappinessFactorTypeInitializer.java
armele: So for the port, would I change this to a constant in 1.21.1 with the port?

Thodor12: In 1.21 they are already constants

src/main/java/com/minecolonies/api/entity/citizen/happiness/HappinessRegistry.java
armele: That's a fair point...  Will refactor.

src/main/java/com/minecolonies/api/entity/citizen/happiness/HappinessRegistry.java
armele: Caller was null safe already, actually.  Just added the annotation for clarity.
```
                    final IHappinessModifier modifier = HappinessRegistry.loadFrom(compoundTag, persist);
                    if (modifier != null)
                    {
                        happinessFactors.put(modifier.getId(), modifier);
                    }
```

PR #11814, opened 2026-09-05
Fix nearby building previews rotating differently from the goggles

genericJE:
# Checklist
- [ ] I have tested and confirmed my changes are working on the most recent commit of this pull request

# Related issues
Same symptom as ldtteam/Structurize#609, which was closed without a fix.

- With the build tool out, the preview of a nearby building could show a different rotation than the goggles show for the same building. The neighbour preview used the building view's rotation while the goggles use the work order's rotation, and the two can differ. The neighbour preview now uses the work order rotation when the building has one.
- `calcRotation` compared the world facing of the hut block against the unmirrored blueprint facing. The mirror is applied before the rotation, so mirrored huts whose hut block faces east or west ended up with a rotation that was 180 degrees off. It now mirrors the blueprint facing before comparing.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/client/render/worldevent/ColonyBlueprintRenderer.java
+++ b/src/main/java/com/minecolonies/core/client/render/worldevent/ColonyBlueprintRenderer.java
@@ -415,6 +415,15 @@ public class ColonyBlueprintRenderer
             final AABB blueprintAABB = AABB.encapsulatingFullBlocks(zeroPos, zeroPos.offset(blueprint.getSizeX() - 1, blueprint.getSizeY() - 1, blueprint.getSizeZ() - 1))
                     .inflate(2 + MinecoloniesAPIProxy.getInstance().getConfig().getClient().neighborbuildingrange.get());
 
+            final Map<BlockPos, RotationMirror> workOrderRotations = new HashMap<>();
+            for (final IWorkOrderView workOrder : ctx.nearestColony.getWorkOrders())
+            {
+                if (workOrder instanceof WorkOrderBuildingView)
+                {
+                    workOrderRotations.put(workOrder.getLocation(), workOrder.getRotationMirror());
+                }
+            }
+
             for (final IBuildingView buildingView : ctx.nearestColony.getClientBuildingManager().getBuildings().values())
             {
                 final BlockPos currentPosition = buildingView.getPosition();
@@ -435,7 +444,8 @@ public class ColonyBlueprintRenderer
                         schemPath = schemPath.substring(0, schemPath.length() - 1) + buildingView.getBuildingMaxLevel() + ".blueprint";
 
                         final String structurePack = buildingView.getStructurePack();
-                        final BlueprintCacheKey key = new BlueprintCacheKey(structurePack, schemPath, buildingView.getRotationMirror());
+                        final RotationMirror rotationMirror = workOrderRotations.getOrDefault(currentPosition, buildingView.getRotationMirror());
+                        final BlueprintCacheKey key = new BlueprintCacheKey(structurePack, schemPath, rotationMirror);
 
                         desired.put(currentPosition,
                             new PendingRenderData(key, currentPosition, 0,
--- a/src/main/java/com/minecolonies/core/tileentities/TileEntityColonyBuilding.java
+++ b/src/main/java/com/minecolonies/core/tileentities/TileEntityColonyBuilding.java
@@ -749,7 +749,8 @@ public class TileEntityColonyBuilding extends AbstractTileEntityColonyBuilding i
             return;
         }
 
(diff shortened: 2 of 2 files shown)


src/main/java/com/minecolonies/core/client/render/worldevent/ColonyBlueprintRenderer.java
```diff
@@ -415,6 +415,16 @@ public Map<BlockPos, PendingRenderData> getDesiredBlueprints(final WorldEventCon
             final AABB blueprintAABB = AABB.encapsulatingFullBlocks(zeroPos, zeroPos.offset(blueprint.getSizeX() - 1, blueprint.getSizeY() - 1, blueprint.getSizeZ() - 1))
                     .inflate(2 + MinecoloniesAPIProxy.getInstance().getConfig().getClient().neighborbuildingrange.get());
 
+            // Prefer the work order rotation so this matches the goggles.
```
Thodor12: Please remove self documenting comments. AI always adds unnecessary commentary that goes against the "self documenting code" way of working.

src/main/java/com/minecolonies/core/tileentities/TileEntityColonyBuilding.java
```diff
@@ -749,7 +749,9 @@ private void calcRotation(final Blueprint blueprint)
             return;
         }
 
-        final int structureRotation = structureState.getValue(AbstractBlockHut.FACING).get2DDataValue();
+        // Mirror is applied before rotation, so compare against the mirrored facing.
```
Thodor12: Remove unnecessary comments

Thodor12 (changes requested):

src/main/java/com/minecolonies/core/client/render/worldevent/ColonyBlueprintRenderer.java
```diff
@@ -435,7 +444,8 @@ public Map<BlockPos, PendingRenderData> getDesiredBlueprints(final WorldEventCon
                         schemPath = schemPath.substring(0, schemPath.length() - 1) + buildingView.getBuildingMaxLevel() + ".blueprint";
 
                         final String structurePack = buildingView.getStructurePack();
-                        final BlueprintCacheKey key = new BlueprintCacheKey(structurePack, schemPath, buildingView.getRotationMirror());
+                        final RotationMirror rotationMirror = workOrderRotations.getOrDefault(currentPosition, buildingView.getRotationMirror());
```
Raycoms, line 447: I am a bit confused how the buildingView can get a different rotationMirror compared to the workorder.

genericJE: Another migration issue (1.19 -> 1.21) that arrose. Here was what the AI told me:


> They are two copies. The building keeps its own `rotationMirror` in `AbstractSchematicProvider`, written at placement and sent to the view. The work order copies the tile entity's value when it is created (`WorkOrderBuilding.java:73`), and the tile entity recomputes its value from the hut block facing when it has none, or after `reactivate()` in `processBlueprint`. Nothing writes that back to the building.
>
> My world came from 1.19, which stored `rotation` and `mirror`. 1.21 only reads the `rotMir` byte, unguarded at `AbstractSchematicProvider.java:188`, and the fallback was removed in deffdd6f3b, so every building here read as `NONE` while the tile entity derived the real rotation. The view says `NONE`, the work order says the real one, and the neighbour preview flips whenever the build tool is out. Since 1.21 has saved the world since, `rotMir` is now stored as 0, so a `contains` guard would not repair it.
>
> For mirrored huts the derivation itself was also 180 degrees off for east/west facings, which is the second hunk.
>
> Using the work order rotation here makes the neighbour preview agree with the goggles and with what the builder will build. If you would rather fix it at the source, the building would need to re-derive its value from the tile entity when one is loaded, which is roughly what deffdd6f3b removed. Happy to do that instead if you prefer.

Talyda: > Another migration issue (1.19 -> 1.21) that arrose. Here was what the AI told me:

So you've upgraded your world from 1.19 to 1.21? And forge to neoforge?

genericJE: Yes, I created an issue here and have mentioned some migration artefacts. https://github.com/ldtteam/minecolonies/issues/11797

MotionlessTrain: Yeah, that is not supported. We can’t help with any issues that arise, as there were breaking changes between versions by intention, as I said before

PR #11817, opened 2026-09-10
Render citizen armour through the vanilla armour layer

### LINKED ISSUE #11816: [1.21.1] Citizens do not render modded enchantment glints because CitizenArmorLayer bypasses the vanilla armour layer

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1381-1.21.1

### Structurize Version

1.0.832-1.21.1

1. NeoForge 21.1.250
2. BlockUI 1.0.211-1.21.1
3. Domum Ornamentum 1.0.231-main
4. Quark 4.1-484
5. Glint & Glamour 1.7.0-1.21.1
6. Glint Customizer 1.0FIX-1.21.1
7. DarkGlint 1.1.6

### Current Behavior

Mods that customise the enchantment glint have no effect on citizens. The armour renders, but always with the default purple glint.

With Quark, applying a colour rune to an enchanted armour piece works normally when a player wears it. Give the identical item to a guard and the glint is the vanilla purple instead of the rune colour. Both are visible in the same frame, side by side.

DarkGlint is unaffected, because it only replaces textures/misc/enchanted_glint_entity.png and needs no code hook. So the glint pass itself works on citizens; what does not work is anything that hooks how the render type is chosen.

Only armour on the display-armour path is affected. When getDisplayArmor(slot) is empty and CitizenArmorLayer falls through to super.renderArmorPiece, the glint is correct.

### Expected Behavior

Citizens should render armour the way every other humanoid entity does, going through HumanoidArmorLayer so that NeoForge's hooks and any mod built on them apply to citizens as well as players.

### Reproduction Steps

1. Install MineColonies and Quark on 1.21.1 / NeoForge.
2. Enchant an armour piece and apply a Quark colour rune to it, for example a white rune on a diamond chestplate.
3. Wear it. The glint is the rune colour.
4. Put an identical piece into a guard's armour slot.
5. The guard's glint is the default purple instead of the rune colour.

You can do the same with Glint & Glamour

### Logs

No log needed, it's a rendering issue.

### Anything else?

The cause looks to be that CitizenArmorLayer#renderArmorPiece reimplements HumanoidArmorLayer#renderArmorPiece rather than calling it, so no mod hooking the vanilla armour layer ever affects citizens. 

Reading that copy against the NeoForge-patched original, it is also missing setupModelAnimations, getArmorLayerTintColor, getDefaultDyeColor and the empty-tint layer skip, so animated armour and per-layer tinting look like they are silently dropped on citizens too.

That would make this the same root cause as #11260, which was fixed for the model hook only.

Not version specific: seen on 1.1.1319 and 1.1.1368, and CitizenArmorLayer.java is byte identical at 1319, 1368 and 1381.

### Footer

---
 ## Viewers

leniver:
# Checklist

   - Diagnosis of the issue
   - Code analysis and research on Neoforge code

# Related issues
Closes #11816

- CitizenArmorLayer now uses the vanilla super.renderArmorPiece for display armor instead of its own copy of it
- Removed the private methods that are no longer needed (renderModel, renderTrim, renderGlint)
- Removed the duplicate getDisplayArmor call, which queried the citizen inventory twice

I went from a small change to makes the glint work with quark to a full review the way Citizen Armor are rendered. Citizen have their armor rendered a little bit differently than other entities that the reason why `renderArmorPiece` is overridden. 
The armor is gotten from `getDisplayArmor` instead of `getItemBySlot` and it does an additional check to make sure the `armorItem` is not in the blacklist.
The issue is that some mods are mixin this method to add effects on the armor (glint, etc..) and the overriden method will not catch those.

Because this method is called at each rendering, I tried to have the smallest possible impact on performance. It must also not change any behavior on the server, which is why I chose `citizen.armorItems.set` instead of `citizen.setItemSlot`, which could have some side effects.

Review please

### DIFF
--- a/src/main/java/com/minecolonies/core/client/render/CitizenArmorLayer.java
+++ b/src/main/java/com/minecolonies/core/client/render/CitizenArmorLayer.java
@@ -138,47 +138,30 @@ public class CitizenArmorLayer<T extends AbstractEntityCitizen, M extends Humano
 
     private void renderArmorPiece(PoseStack poseStack, MultiBufferSource bufferSource, T citizen, EquipmentSlot equipmentSlot, int light, A armor, final ICitizenDataView citizenDataView)
     {
-        ItemStack itemstack = citizenDataView.getDisplayArmor(equipmentSlot);
-        if (itemstack.isEmpty())
-        {
-            itemstack = citizen.getItemBySlot(equipmentSlot);
-        }
-        Item armorItem = itemstack.getItem();
+        final ItemStack displayArmor = citizenDataView.getDisplayArmor(equipmentSlot);
+        final ItemStack itemstack = displayArmor.isEmpty() ? citizen.getItemBySlot(equipmentSlot) : displayArmor;
+        final Item armorItem = itemstack.getItem();
 
-        if (armorItem instanceof ArmorItem armoritem && !disabledFromRendering.contains(armorItem))
+        if (armorItem instanceof ArmorItem && !disabledFromRendering.contains(armorItem))
         {
             try
             {
-                if (citizenDataView.getDisplayArmor(equipmentSlot).isEmpty())
+                if (displayArmor.isEmpty())
                 {
                     super.renderArmorPiece(poseStack, bufferSource, citizen, equipmentSlot, light, armor);
                 }
-                else if (armoritem.getEquipmentSlot() == equipmentSlot)
+                else
                 {
-                    this.getParentModel().copyPropertiesTo(armor);
-                    this.setPartVisibility(armor, equipmentSlot);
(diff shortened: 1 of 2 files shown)


leniver:
btw, I saw also that the renderArmorPiece is marked as deprecated with only 6 arguments we should use the 12 arguments, but I have not done it because they should be 0.0F anyway.
ref: https://github.com/neoforged/NeoForge/blob/b92c510d64ed9506140d74538f4699277e8193bd/patches/net/minecraft/client/renderer/entity/layers/HumanoidArmorLayer.java.patch

someaddons (approved):

PR #11821, opened 2026-09-13
Improve node reference and debug tracking

someaddons:
# Checklist

# Related issues
Closes #
Closes #

- Simplified navigator node tracking, it now always retains up to date references to the previous, current and next node.
- Debug tracking no longer needs to be manually done at different places, it is now part of advancing a node on the path

Review please


src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
```diff
@@ -745,38 +758,26 @@ private boolean handleLadders()
...
+            final double nextZ = (double) currentNode().z + (double) ((int) (this.mob.getBbWidth() + 1.0F)) * 0.5D;
 
             final double diffX = Math.abs(this.mob.getX() - nextX);
             final double diffY = Math.abs(this.mob.getY() - nextY);
             final double diffZ = Math.abs(this.mob.getZ() - nextZ);
 
-            final PathPointExtended previous = getPreviousNode();
             // Ladder entry needs more exact position tracking, we want to center the citizen before doing movement in another axis
-            if (previous == null || !previous.isOnLadder() || previous.y == getNextNode().y)
+            if (previousNode() == null || !previousNode().isOnLadder() || previousNode().y == currentNode().y)
```
Raycoms, line 770: Why the func instead of field access? Then we don't need the functions below?

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
```diff
@@ -806,8 +802,7 @@ private boolean handleLadders()
                 }
 
                 //  Ladder Workaround
-                final PathPointExtended afterNext = getNextNextNode();
```
Raycoms, line 809: Not needed?

src/main/java/com/minecolonies/core/network/messages/client/SyncPathReachedMessage.java
```diff
@@ -75,12 +64,9 @@ public void onExecute(final NetworkEvent.Context ctxIn, final boolean isLogicalS
     {
         for (final MNode node : PathfindingDebugRenderer.lastDebugNodesPath)
         {
-            for (final BlockPos reachedPos : reached)
+            if (reached.getX() == node.x && reached.getY() == node.y && reached.getZ() == node.z)
```
Raycoms, line 67: So we only ever show the most recent?

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
someaddons: just a small precaution to people dont accidently modify the value

src/main/java/com/minecolonies/core/entity/pathfinding/navigation/MinecoloniesAdvancedPathNavigate.java
someaddons: just replaced it with the cached var

src/main/java/com/minecolonies/core/network/messages/client/SyncPathReachedMessage.java
someaddons: no its the same logic just for a single position, as we are only sending single position updates for reached. The only exception was when pathfollow skipped two nodes, but that now just sends two additional messages

Raycoms (approved):
Please port

PR #11824, opened 2026-09-13
Tavern music tweaks

armele:
# Checklist
- [X] I have read and accept the [Contributing guidelines](https://github.com/ldtteam/.github/blob/master/CONTRIBUTING.md)

AI was used to analyze the current logic, and code review the changes.

# Related issues
Closes [Discord Conversation](https://discord.com/channels/472875599422291968/1548109169557508187)

- Changes Tavern music to use the RECORDS slider instead of the AMBIENT slider.
- Adds a setting in the tavern that allows music to be turned off.
- When changed, setting applies to next play attempt (will not interrupt music already playing).

Review please


src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -2155,6 +2155,13 @@
   "block.minecolonies.blockhutstable.name": "Stable",
   "com.minecolonies.coremod.gui.maintab": "Main",
   "com.minecolonies.coremod.setting.minecolonies:producedirt": "Produce Dirt:",
+  "com.minecolonies.coremod.setting.minecolonies:tavern_music_volume": "Tavern Music Volume:",
+  "com.minecolonies.coremod.setting.tooltip.minecolonies:tavern_music_volume": "Sets this Tavern's base music volume. Your personal Music volume slider also applies.",
+  "com.minecolonies.coremod.setting.tavern_music_volume.off": "Off",
+  "com.minecolonies.coremod.setting.tavern_music_volume.25": "25%",
+  "com.minecolonies.coremod.setting.tavern_music_volume.50": "50%",
```
Raycoms: This feels like a waste of lang keys. Can just be 25/50/75/100%. Also "Off" should be pre-existing.

someaddons:
> AI was used to analyze the current logic, and code review the changes.

I'm pretty sure AI generated like half of the changes xD
agree with ray, group the keys together, also just group the String constants together instead of adding javadoc bloat for all

uecasm:
I was thinking there could perhaps be a client setting to configure whether to use MUSIC or AMBIENT.  Some people might turn off (or way down) MUSIC because the default music gets annoying, yet still want to hear the tavern music.

Thodor12:
I also feel like this is not a good idea to reuse an existing selection setting component for this. A slider makes more sense for this, that being said we have no slider components, but a set of buttons in a horizontal row with basic numbers in them is probably more logical.

That also being said, should we be responsible for controlling volume sliders in the first place? An on/off is probably warranted, but a full slider system feels like that is something that Mojang should be solely in control of.

Also, with that I think the setting key should state "Tavern plays music", rather than controlling the volume.

armele:
> I also feel like this is not a good idea to reuse an existing selection setting component for this. A slider makes more sense for this, that being said we have no slider components, but a set of buttons in a horizontal row with basic numbers in them is probably more logical.
> 
> That also being said, should we be responsible for controlling volume sliders in the first place? An on/off is probably warranted, but a full slider system feels like that is something that Mojang should be solely in control of.
> 
> Also, with that I think the setting key should state "Tavern plays music", rather than controlling the volume.

I think especially with @uecasm 's suggestion of allowing the player to pick between ambient and music we could then just have an on/off and let them use their music settings as desired to control the volume...  Unless someone feels strongly to the contrary I'll drop volume, just have on/off and create a toggle for Music/Ambient channel.

Thodor12:
> > I also feel like this is not a good idea to reuse an existing selection setting component for this. A slider makes more sense for this, that being said we have no slider components, but a set of buttons in a horizontal row with basic numbers in them is probably more logical.
> > That also being said, should we be responsible for controlling volume sliders in the first place? An on/off is probably warranted, but a full slider system feels like that is something that Mojang should be solely in control of.
> > Also, with that I think the setting key should state "Tavern plays music", rather than controlling the volume.
> 
> I think especially with @uecasm 's suggestion of allowing the player to pick between ambient and music we could then just have an on/off and let them use their music settings as desired to control the volume... Unless someone feels strongly to the contrary I'll drop volume, just have on/off and create a toggle for Music/Ambient channel.

I'm not sure if that really solves the problem. Also I'm not sure if MUSIC is actually the correct channel either. Realistically it should be RECORDS.
The intention behind this music is that it's background music played by a band performing inside the tavern. So the most logical is probably that this is closest to something being played in a jukebox.

But this is just A or B at this point. Just give them an option to turn it on and off and let's not dynamically mess at which source it's played with.

armele:
Simplified significantly with this feedback in mind. Tavern setting is simply on/off.  Changing the setting will not interrupt music already playing (applies to next play check). No tavern-level volume control. Moved Tavern music to "RECORDS" sound source.

src/main/java/com/minecolonies/core/colony/buildings/modules/TavernBuildingModule.java
```diff
@@ -119,15 +128,18 @@ public void onPlayerEnterBuilding(final Player player)
             }
 
             if (count < 2)
+
```
Thodor12, line 131: You have some stray whitelines here

src/main/resources/assets/minecolonies/lang/manual_en_us.json
```diff
@@ -2154,6 +2154,7 @@
   "block.minecolonies.blockhutstable": "Stable",
   "block.minecolonies.blockhutstable.name": "Stable",
   "com.minecolonies.coremod.gui.maintab": "Main",
+  "com.minecolonies.coremod.setting.minecolonies:playmusic": "Play Tavern Music",
```
Thodor12, line 2157: We should probably actually call it "Play Music at Night", Tavern is already implied cause its the settings of the tavern of course. And then it's also clear the music is only playing at night and not during the day, so the user can expect what the feature otherwise would do.

someaddons (approved):

Raycoms:
Please port

PR #11822, opened 2026-09-13
fix 11808

Raycoms:
# Checklist
- [ ] I have read and accept the [Contributing guidelines](https://github.com/ldtteam/.github/blob/master/CONTRIBUTING.md)
- [ ] I have tested and confirmed my changes are working on the most recent commit of this pull request

- [ ] I have used AI in the creation of this pull request

# Related issues
Closes #11808
Closes #

# Changes proposed in this pull request
-
-

Review please

PR #11827, opened 2026-09-15
Fix/food history double record

### LINKED ISSUE #11826: [1.21.1] Food history still wrong after #11726: dining hall meals are recorded twice, and an empty history never fills

- [x] Single Player
- [x] Multi Player

### Minecraft Version

1.21

### MineColonies Version

1.1.1387-1.21.1-snapshot

### Structurize Version

1.0.832-1.21.1

NeoForge 21.1.250
BlockUI 1.0.209-1.21.1
Domum Ornamentum 1.0.231

### Current Behavior

When a citizen eat a meal at a dinner hall, it's recorded wrongly in the citizen's food history. 

The #11726 fix changed getLastEaten() from peek() to poll(), that didn't remove the duplicate and added a second problem.
`poll()` has a side effect that is, it remove the first element from the history.

1. History with at least one entry: each dining hall meal adds the food twice and deletes the oldest entry, even when the history is not full.

   Citizen with 6 entries, one meal of Stuffed Pita:
   before: **Fried Rice**, Fried Rice, Stew Trencher, Stew Trencher, Lamb Stew, Lamb Stew
   after:  Fried Rice, Stew Trencher, Stew Trencher, Lamb Stew, Lamb Stew, **Stuffed Pita, Stuffed Pita**

   Citizen with a full history, one meal of Lamb Stew:
   before: **Stew Trencher, Schnitzel**, Schnitzel, Stuffed Pepper, Sushi Roll, Sushi Roll, Stuffed Pita, Stuffed Pita, Ramen, Ramen
   after:  Schnitzel, Stuffed Pepper, Sushi Roll, Sushi Roll, Stuffed Pita, Stuffed Pita, Ramen, Ramen, **Lamb Stew, Lamb Stew**

   A citizen fed only by the dining hall ends up with every meal in pairs, so 10 entries hold at most 5 different foods. Their diversity can never go above 5, however varied the menu is.

2. Empty history: a dining hall meal is recorded at the first bite and removed again when the meal ends, so the history stays empty, however much is eaten.

   Newly recruited citizen, one meal of two Hand Pies:
   first bite: Hand Pie (saturation 32.5)
   meal over:  (empty) (saturation 60)

   As long as they only eat at the dining hall, such a citizen gets no food complaints, no food happiness factor and no diversity effect on disease chance.

There is a side bug that I have found while debuging.
/mc citizens info throws StringIndexOutOfBoundsException when the food history is empty, so it cannot show it.

### Expected Behavior

Each meal adds each food eaten once, and no other entry is removed except by the 10 entry limit. A citizen fed from a dining hall with enough different foods can reach the diversity their house level requires. 

### Reproduction Steps

1. MineColonies 1.1.1387, a staffed dining hall with several MineColonies foods in stock and on the menu.
2. Pick a citizen with a few food history entries who carries no food. Run `/mc citizens info <colony> <citizen>` and note the history.
3. In Creative, run `/mc citizens modify <colony> <citizen> saturation = 2` and let them eat at the dining hall.
4. Run `/mc citizens info` again: the food appears twice at the end and the oldest entry is gone.
5. Repeat with a newly recruited citizen (info throws while the history is empty): after the meal the history is still empty.

### Logs

https://gist.github.com/leniver/c888a572754acc484db0abe0fddddef2 and https://gist.github.com/leniver/aeb2d74c9bda3f46948bcdcb58ea8318

### Anything else?

It tooks me a while to understand exactly what was going wrong because with the patch citizens are stoping to complain but for the wrong reason.
I coded with the help of AI a small tool to follow different citizen of a colony on a external panel.
https://github.com/leniver/minecolonies-citizen-watch

It helped me to make them hungry, check their activity etc...

Here what I found:

`EntityAIEatTask.eat()` records the food on the first bite of a dining hall meal (`restaurant != null`), then at the end of the meal loops over the foods eaten and records each one again unless `foodHandler.getLastEaten()` returns it:
https://github.com/ldtteam/minecolonies/blob/0484da4a9d/src/main/java/com/minecolonies/core/entity/ai/minimal/EntityAIEatTask.java#L210-L231

`getLastEaten()` was meant to return the most recent meal (its Javadoc, and the "eaten this very recently" check it had in FoodUtils until #11263). On the `EvictingQueue`, `peek()` returns the oldest entry instead, so the skip never matched and the food was recorded twice since #10368. `poll()` still returns the oldest entry and also removes it:
https://github.com/ldtteam/minecolonies/blob/0484da4a9d/src/main/java/com/minecolonies/core/entity/citizen/citizenhandlers/CitizenFoodHandler.java#L84-L87

So at the end of a meal:
- history not empty: `poll()` removes the oldest entry, a different food, so the food is added a second time (case 1)
- history empty: the only entry is the food just added, `poll()` removes it, it matches, nothing is added back (case 2)
- meal with no restaurant set: no first bite record, the food is recorded once, correctly

Recording on the first bite at a dining hall is useful: `eat()` calls `hasFood()` before every bite, so the food just recorded scores worst in `getBestFoodForCitizen()` and a citizen carrying another menu food switches to it. Away from a dining hall it would break the meal (with `menu == null`, a full history and low diversity, `getBestFoodForCitizen()` `returns -1` for food already in the history, `hasFood()` fails and the citizen leaves to find a restaurant), which is probably why #11263 made only dining hall meals record early.

Some function have side effect which make the debuging harder to do, for example `hasFood()` which should tell us if a citizen has food, also set the inventory slot where the food is located.

### Footer

---
 ## Viewers

leniver:
# Checklist

   - Created a debuging [tool](https://github.com/leniver/minecolonies-citizen-watch) to watch food history
   - Used to help locating some of the code

# Related issues
Closes #11826

- It fixes the way food is recorded in history when eating from the dining hall.
- It fixes a side bug that was throwing an exception when the food history of a citizen was empty and an operator was asking for info about that citizen.
- It adds a new command for citizens, allowing an operator to clear the food history.

Like explained in issue #11826, the change to `getLastEaten()` is introducing 2 new bugs. So what I did is store the first bite into `recordedFood` and compare with it instead of relying on the history that gets modified.
I have not touched `getLastEaten()`. By inspecting the code, it seems to not be used anywhere anymore. Maybe we could remove it or fix it to return the last eaten food without modifying the history.

While trying to debug, I was missing a functionality to clear the food history, so I implemented it to help me fully test the changes. I was doing it by recruiting new citizens, but it's more reliable to have a function to do so.

Review please


leniver:
a 1.20-era call was carried over during a port, and it doesn't compile against 1.21.
It has been introduced with this PR: #11822

src/main/java/com/minecolonies/core/entity/ai/minimal/EntityAIEatTask.java
```diff
@@ -224,12 +230,13 @@ private IState eat()
 
         for (final Item foodItem : eatenFood)
         {
-            if (foodHandler.getLastEaten() != foodItem)
```
Raycoms, line 227: What is the problem with this?

leniver: `lastEatenFoods` is a `EvictingQueue`. When we `addLastEaten`, it appends to the tail.
`getLastEaten()` is `lastEatenFoods.poll()`, which remove and return the head (oldest entry) not the last eaten one.

So in the loop, for each `eatenFood`:
- It deletes the history: each call drop one entry from the front of the queue. Eating a meal of three bites removes the three oldest entry.
- It compares agains the wrong end: Our goal is to not record twice the same food, by comparing with the wrong entry it makes the check meaningless.
- It is only reached in the final pass after eating, which is also where the second `addLastEaten` for the same meal came from (the first happens at the top when `eatenFood` is empty)

The effect is it will record twice the meal if the citizen have eaten from the dining hall, because a citizen go to the dining hall at low saturation and need at least 2 bites of a food to get full.
Second effect, if the history is empty it will first put the first bite in the history because `eatenFood` is empty and when comparing it's removing it from history and do not add because it's the same.

leniver: We could have an extrem situation if we have a food that required 10+ bites to get max saturation and the history is full of other food it will first time replace the full history with this eaten food (because different) and the next meal will delete the full history.

Raycoms:
Pushed the real fix.

leniver:
Clean fix.
But after testing it, I saw that it was not resolving all the issues. 

If a citizen eat at the dining hall, everything work fine the history get filled and if full the food registration works.
If a citizen get feed by a player, it's working fine. History get updated correctly.

The issue is when the citizen eat from his inventory/hut the same food over-and-over with a non-full history. Because he doesn't eat this in a restaurant, and the last history item is the same it will never update the history.
This is specially important for nether mine, but could be applied to any hut.

To test this:
- set a minimum stock to a hut (to make sure it has food) with only 1 type of food
image

- make sure to hire a citizen that has a history of max 8 items (new citizen or use my command to clear the food history)
- edit the saturation of that citizen a few time and check each time his history.

The history will stop growing, breaking the food logic.

Raycoms:
> Clean fix. But after testing it, I saw that it was not resolving all the issues.
> 
> If a citizen eat at the dining hall, everything work fine the history get filled and if full the food registration works. If a citizen get feed by a player, it's working fine. History get updated correctly.
> 
> The issue is when the citizen eat from his inventory/hut the same food over-and-over with a non-full history. Because he doesn't eat this in a restaurant, and the last history item is the same it will never update the history. This is specially important for nether mine, but could be applied to any hut.
> 
> To test this:
> 
> * set a minimum stock to a hut (to make sure it has food) with only 1 type of food
> 
> image
> * make sure to hire a citizen that has a history of max 8 items (new citizen or use my command to clear the food history)
> * edit the saturation of that citizen a few time and check each time his history.
> 
> The history will stop growing, breaking the food logic.

Yes, we should probably have a random "trigger every 10" requirement for this. So they don't overeat the same

src/main/java/com/minecolonies/core/entity/ai/minimal/EntityAIEatTask.java
```diff
@@ -210,6 +215,7 @@ private IState eat()
         if (eatenFood.isEmpty() && restaurant != null)
```
leniver, line 215: Why do we have the `&& restaurant != null` check?
It has been introduced in october 2025, making eating food from restaurant behave differently that from the inventory/hut. 
Not sure what it was fixing

leniver:
I found an other weird behavour.

When the history is full and the citizen try to eat from his inventory.
1) If diversity is OK, then everything works fine he will eat the food he has in his inventory
2) If diversity is not-OK then he will go to the restaurant (which is a good behavour) but will still eat the food he has in his inventory instead of getting served a better food by the cook.

But we need to be carefull, because for hut that have their own menu (like the nether mine) it should never go to restaurant.
(I have not checked if that was the case)

leniver:
> Yes, we should probably have a random "trigger every 10" requirement for this. So they don't overeat the same

This happens if the player is configuring the hut to have only 1 meal, if the history fill then the citizen after some time will complain that he is eating always the same food.

Talyda:
I thought they weren't supposed to eat from huts that weren't the dining hall?  Just from player feeding them/their own inventory and dining hall? Or was it they eat from their inventory if there is no dining hall? I have super vague recollection of something along those lines a long while back

MotionlessTrain:
I thought they put some extra food in their building, and could start eating that later. Though I don’t believe it is intended to have a max stock for food there, to have them always do so. That may be an unintended side effect
Besides, the nether miner does have a menu, for food to bring with them to the nether. Not necessarily for them to eat whilst in the overworld, that is what the dining hall is for

leniver:
I will run some tests, but that sound like a good way of handling "non-restaurant" meal.

What about the 2 others commit?
The command is simply a tool, usefull but very limited to such issues.

Issue #11780, opened 2026-08-10
Client crash in AbstractEntityCitizen.getTexture when rendering Tavern visitors after colony load


brian-stearne:
## Is there an existing issue for this?

- [x] I have searched the existing issues.

Searches covered open and closed reports using the exact exception, `AbstractEntityCitizen.getTexture`, `DATA_STYLE`, `DATA_TEXTURE_SUFFIX`, visitor/Tavern texture crashes, colony loading, and dimension transitions. The closest reports were #6039, #6152, #9769, #9912, and #10753, but each has a different failure mode or code path.

## Are you using the latest MineColonies Version?

- [x] I am running the latest beta/release version of MineColonies for Minecraft 1.21.1.

The crash reproduces with MineColonies `1.1.1362-1.21.1-snapshot`, currently listed as the latest 1.21.1 beta on CurseForge. It also reproduced in a clean test with MineColonies `1.1.1368-1.21.1-snapshot` as the only enabled MineColonies JAR.

## Did you check on the Wiki? or ask on Discord?

- [x] I checked the MineColonies Wiki and did not find this crash documented.

## What were you playing at the time?

- [x] Single Player
- [ ] Multi Player

## Minecraft Version

1.21

## MineColonies Version

`minecolonies-1.1.1362-1.21.1-snapshot.jar`

Also reproduced with `minecolonies-1.1.1368-1.21.1-snapshot.jar` as the only enabled MineColonies JAR.

## Structurize Version

`structurize-1.0.832-1.21.1-snapshot.jar`

## Related Mods and their Versions

1. NeoForge `21.1.238`
2. BlockUI `1.0.209-1.21.1`
3. Domum Ornamentum `1.0.234-snapshot-main`
4. Multi-Piston `1.2.58-1.21.1`
5. MineColonies Tweaks `1.21.1-3.32`
6. MineColonies Compatibility `1.21.1-3.42`
7. StyleColonies `1.15.56-1.21.1`
8. TownTalk `1.2.0`
9. MC Talking `1.7.1-neoforge+1.21.1`
10. Disable MineColonies Abandoned Structures `1.21.1-1.0.2`

This is a modpack environment, and the issue has not yet been reproduced in a minimal MineColonies-only instance. No third-party mixin is identified on the failing `AbstractEntityCitizen.getTexture()` frame in the later clean-version crash reports.

## Current Behavior

[...]

PR #11828, opened 2026-09-16
Fix client NPE rendering Tavern visitors: don't copy a null colony texture style into DATA_STYLE

Wesley0444:
### Problem

Clients intermittently crash with `Rendering entity in world` as soon as a `minecolonies:visitor` (Tavern visitor) comes into render range, typically right after arriving next to a colony (teleport, respawn or login):

```
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.toString()" because "s" is null
    at java.lang.String.contains(String.java:2981)
    at com.minecolonies.api.entity.citizen.AbstractEntityCitizen.getTexture(AbstractEntityCitizen.java:281)
    at com.minecolonies.core.client.render.RenderBipedCitizen.getTextureLocation(RenderBipedCitizen.java:146)
```

Line 281 is the `DATA_STYLE` half of the condition (`DATA_TEXTURE_SUFFIX` is line 282), so `DATA_STYLE` was null on the client.

### Cause

The client branch of `VisitorCitizen.aiStep()` copies the colony view's style into synched data without a null check:

```java
this.citizenDataView = colonyView.getVisitor(citizenId);
getEntityData().set(DATA_STYLE, colonyView.getTextureStyleId());
```

`ColonyView.textureStyle` has no initializer and is only assigned in `handleColonyViewMessage` (and `setTextureStyle`), so `getTextureStyleId()` returns null for a view that exists before its first full view message has been applied.

Nothing else can make `DATA_STYLE` null on the client: it is defined as `"default"` in `defineSynchedData`, and a null String cannot be synched from the server. Regular citizens survive the same situation because `EntityCitizen.onSyncedDataUpdated` resets an unknown style to `TEXTURE_PACKS.get(0)`; `VisitorCitizen` extends `AbstractEntityCitizen` directly and has no equivalent guard.

### Fix

Keep the entity's current style when the view does not have one yet. The style is copied again on the next tick that has real data.

### Notes

- This is the crash from #11780, which was closed as caused by the reporter's addon mods. It reproduces on 1.1.1368-1.21.1 in a pack with **no MineColonies addons**: only BlockUI 1.0.211, Domum Ornamentum 1.0.234, Multi-Piston 1.2.58, Structurize 1.0.832 and TownTalk 1.2.0, on NeoForge 21.1.249. No mixin config in any of the 128 client or 135 server jars references MineColonies, and the MineColonies frames in the crash report carry only `{re:classloading}`, so nothing had transformed them.
- `version/main` carries the identical line (`VisitorCitizen.java:521`) if you want the same change there.
- Optional belt-and-braces alternative: initialize `ColonyView.textureStyle = "default"`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Raycoms:
Do you have any multi-threading mod or so installed? ColonyView should be initialized and immedately read, so there should be no moment where the colony view exists but the data is null, unless this is assessed multi-threaded.

Thodor12:
Please fill in the pull request template appropriately

